PR #2, opened 2017-06-12
added attributes to github_user and github_team data sources

mootpt:
- updated go-github
- added gpg and ssh keys to github_user data source and members to github_team
- fixed issues with attributes using pointers instead of functions
- documented changes


josephholsten:
Looks great! Do we need to change things anywhere else for the vendor update?

mootpt:
@josephholsten nope. should be good to merge.

stack72:
Hi @mootpt 

This LGTM! Thanks for this :)

P.

PR #48, opened 2017-09-05
provider: Add insecure mode

georgecodes:
Added a config option to allow use against a GitHub Enterprise installation with self-signed certificates.


majormoses:
needs a rebase to fix conflicts. To be honest I don't find a lot of value doing this as it is super easy to get a trusted cert for free these days and not verifying a cert is basically as bad as not having one. I would think it better to be able to provide a path locally to your certs so it can at least validate that way. Just my `$0.02`.

radeksimko:
Hi @georgecodes 
thank you for your PR and sorry for the long delay in processing it.

Given the age of this PR, I took the liberty and made some changes to:

1. decouple some of the logic and remove unused (commented out) code
2. add a test which exercises this new option for both cases (`insecure` = `true`/`false`)
3. reflect @majormoses's valid comments in docs

I hope you're 👌 with these changes. Your authorship was retained, so your commit will still be credited to you.

With that I'm going to merge this PR, so we can make it part of the next release.

georgecodes:
Thanks very much!

PR #64, opened 2017-12-19
`github_repository` - adding support for `archived` repositories

tombuildsstuff:
- Adds support for archiving repositories
- Updates the SDK to the latest commit
- Fixing the failing deployment tests (by giving them unique names)

After some additional testing with this it appears repositories can only be archived through the API and not unarchived (whereas this is possible through the UI) - so this could do with some 🤔

### DIFF
--- a/website/docs/r/repository.html.markdown
+++ b/website/docs/r/repository.html.markdown
@@ -64,6 +64,8 @@ The following arguments are supported:
 and after a correct reference has been created for the target branch inside the repository. This means a user will have to omit this parameter from the
 initial repository creation and create the target branch inside of the repository prior to setting this attribute.
 
+* `archived` - (Optional) Specifies if the repository should be archived. Defaults to `false`.
+
 ## Attributes Reference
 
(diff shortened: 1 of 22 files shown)


website/docs/r/repository.html.markdown
```diff
@@ -64,6 +64,8 @@ The following arguments are supported:
 and after a correct reference has been created for the target branch inside the repository. This means a user will have to omit this parameter from the
 initial repository creation and create the target branch inside of the repository prior to setting this attribute.
 
+* `archived` - (Optional) Specifies if the repository should be archived. Defaults to `false`.
```
catsby: Need to document the limitation found about unarchiving via the API. If you do choose to add a `CustomizeDiff` func, please document the behavior of going from `true`->`false` (unarchiving)

paultyng:
This should fix #73

majormoses:
Can we update the docs and get this merged?

paultyng:
This was released in v1.1.0

PR #65, opened 2018-01-18
resource/github_repository: Repository rename

### LINKED ISSUE #55: resource/github_repository: Renaming forces new resource
### Terraform Version

v0.10.2

### Affected Resource(s)

github_repository

```hcl
  name = "example-rename" # Change this
}
```

### Expected Behavior

The repository should have been renamed without triggering the "forces new resource" behavior.

### Actual Behavior

The `plan` shows the repository as requiring recreation.

### Output

```
-/+ github_repository.example (new resource required)
      full_name:          "Example/example-rename" => "<computed>"
      git_clone_url:      "git://github.com/Example/example-rename.git" => "<computed>"
      http_clone_url:     "https://github.com/Example/example-rename.git" => "<computed>"
      name:               "example-rename" => "example" (forces new resource)
```

eugenetaranov:
Disables ForceNew for repository name allowing repo rename.
Resolves #55

### DIFF
--- a/github/resource_github_repository.go
+++ b/github/resource_github_repository.go
@@ -24,7 +24,7 @@ func resourceGithubRepository() *schema.Resource {
 			"name": {
 				Type:     schema.TypeString,
 				Required: true,
-				ForceNew: true,
+				ForceNew: false,
 			},
 			"description": {
 				Type:     schema.TypeString,
--- a/github/resource_github_repository_test.go
+++ b/github/resource_github_repository_test.go
@@ -50,9 +50,12 @@ func testSweepRepositories(region string) error {
 
 func TestAccGithubRepository_basic(t *testing.T) {
 	var repo github.Repository
+
 	randString := acctest.RandStringFromCharSet(10, acctest.CharSetAlphaNum)
+
 	name := fmt.Sprintf("tf-acc-test-%s", randString)
 	description := fmt.Sprintf("Terraform acceptance tests %s", randString)
+	updatedDescription := "Updated " + description
 
 	resource.Test(t, resource.TestCase{
 		PreCheck:     func() { testAccPreCheck(t) },
@@ -60,7 +63,7 @@ func TestAccGithubRepository_basic(t *testing.T) {
 		CheckDestroy: testAccCheckGithubRepositoryDestroy,
 		Steps: []resource.TestStep{
 			{
-				Config: testAccGithubRepositoryConfig(randString),
+				Config: testAccGithubRepositoryConfig(name, description),
 				Check: resource.ComposeTestCheckFunc(
 					testAccCheckGithubRepositoryExists("github_repository.foo", &repo),
 					testAccCheckGithubRepositoryAttributes(&repo, &testAccGithubRepositoryExpectedAttributes{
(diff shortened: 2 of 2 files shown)


gabel0287:
What else needs to be done to merge this?

paultyng:
I think in this case it would be great if we can get an acceptance test that actually causes a rename to confirm it works properly.

radeksimko:
Hi folks,
thanks for expressing the interest and thanks @eugenetaranov for the PR.

@majormoses is right that it is safe to make in place update, which we can verify using the passing acceptance test I just added in a form of a commit to this PR.

This works fine as is _only_ if users don't have any other resources depending on that repository, which often may not be the case.

Example:

```hcl
  ~ github_repository.example
      name:       "radek-test-after" => "radek-test-after2"

-/+ github_repository_deploy_key.example_repository_deploy_key (new resource required)
      id:         "radek-test-after:30133846" => <computed> (forces new resource)
      key:        "ssh-rsa AAAAB..." (forces new resource)
      read_only:  "false" => "false"
      repository: "radek-test-after" => "radek-test-after2" (forces new resource)
      title:      "Repository test key" => "Repository test key"

```

As you can see the name is treated as unique ID in other resources, which causes the provider to plan recreation of such dependent resources. Such behaviour is IMO undesired.

We need to figure out how to deal with these references, when a repository is renamed. I'm thinking we could change all these `org/name` based IDs to the unique numeric IDs which (AFAIK) don't change after rename.

I'm not sure if the API supports this though, so I will reach out to the GitHub support with this question.

Thank you all for your patience.

radeksimko:
@docwhat where do you see `repo_id` documented?

majormoses:
> This works fine as is only if users don't have any other resources depending on that repository, which often may not be the case.

Agreed, we have lots of things that depend on the repo (such as team access, branch protections, etc. Though in most of those cases while its undesirable I think would be safe to delete/create most of those resources. Obviously "most cases" scares the hell out of me. I think deploy keys and webhooks would be the exceptions as this could lead to failed automated deploys during that window. The biggest concern I have with this is that say you perform the delete action but then run out of API requests left and are now left in a broken/undesired state for 60 minutes without going to the UI and manually creating.

docwhat:
Finding documentation for this was more of a pain than I thought.

1. I know that the ids are permanent because:

    - I have tested them (via renaming a repository to a new name and organization).
    - It would break `node_id` (example `node_id`: "MDEwOlJlcG9zaXRvcnk5MzQ0NjA5OQ==" which is "010:Repository93446099").

        The Node ID is core to the new /v4/ GraphQL interface. They must be unique for it to work.

2. I know that you can fetch a repository via:

    ```sh
    # Example; my `.netrc` has a GitHub token.
    curl -v --netrc https://api.github.com/repositories/93446099
    ```

3. There references are sparse for the `/repositories/:repository_id` end-points:

    - Sub URLs are mentioned in [_Additional endpoints available for GitHub Apps_](https://developer.github.com/changes/2017-10-20-additional-endpoints-for-github-apps/) as:

        ```
        PUT /repositories/:repository_id/collaborators/:username
        DELETE /repositories/:repository_id/collaborators/:username
        ```

    - There is a mention of similar endpoints in [_Integrations Early Access_](https://developer.github.com/changes/2016-09-14-Integrations-Early-Access/) with regards to the `deployments` sub URL.

        ```
        GET /repositories/:repository_id/deployments/:id
        GET /repositories/:repository_id/deployments/:deployment_id/statuses
        GET /repositories/:repository_id/deployments
        POST /repositories/:repository_id/deployments/:deployment_id/statuses
        POST /repositories/:repository_id/deployments
        ```

4. I couldn't figure out how to search for a repository by ID or Node ID in the v4 (GraphQL) API. :-(

docwhat:
> The biggest concern I have with this is that say you perform the delete action but then run out of API requests left and are now left in a broken/undesired state for 60 minutes without going to the UI and manually creating.

Ugh... I've run across that before.

If we could use the /v4/ GraphQL endpoint then this would be much better since you can send all mutations at once and you won't get "half-a-change".

radeksimko:
I just received a helpful response from GitHub support staff:

> Hi Radek,
> 
> It is not documented, however you can use the numeric repository id already, and you can rely on that continuing to work. The URL format is:
> 
> `https://api.github.com/repositories/:id`
> Also, note that the API will redirect you if the repository has moved or changed names and you fetch it by name:
> 
> https://developer.github.com/v3/#http-redirects
> 
> https://developer.github.com/changes/2015-07-21-automatic-redirects-for-renamed-repositories/
> 
> Please let me know if you have any additional questions.

While it seems it will require some decent refactoring, it may not be as big as switching to GraphQL.

radeksimko:
@cewood That is correct.

I'm going to close this PR for now - for that reason, but anyone should feel free to open a new one which addresses the [concerns I mentioned](https://github.com/terraform-providers/terraform-provider-github/pull/65#issuecomment-411097743) - by changing IDs from slugs to numeric IDs.

Thanks for understanding and thanks @eugenetaranov for the effort here.

bmcstdio:
> (...) anyone should feel free to open a new one which addresses the concerns I mentioned - by changing IDs from slugs to numeric IDs.

@radeksimko I am fairly new to the provider's code, but maybe I could be able to get something going with a little guidance, if that's OK? :)

PR #67, opened 2018-01-25
improved readme

caarlos0:
just small fixes on readme

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,7 +20,7 @@ Clone repository to: `$GOPATH/src/github.com/terraform-providers/terraform-provi
 
 ```sh
 $ mkdir -p $GOPATH/src/github.com/terraform-providers; cd $GOPATH/src/github.com/terraform-providers
-$ git clone git@github.com:hashicorp/terraform-provider-github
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -20,14 +20,16 @@ Clone repository to: `$GOPATH/src/github.com/terraform-providers/terraform-provi
...
+$ git clone git@github.com:terraform-providers/terraform-provider-github.git
 ```
 
 Enter the provider directory and build the provider
 
 ```sh
 $ cd $GOPATH/src/github.com/terraform-providers/terraform-provider-github
 $ make build
+# or if you're on a mac:
+$ gnumake build
```
radeksimko: Where does that come from on Mac? I'd just assume most folks use the GNU tooling from Homebrew which is installed as `make` by default? 🤔

caarlos0: I use gnu tooling from homebrew, but I did not let it override the default tooling, so I have `gfind`, `gnumake`, `gsed`, etc...

maybe it's not common or maybe it is enough to just point out that gnu make should be used...

PR #79, opened 2018-03-13
[WIP] provider: Prevent empty provider config from being used

### LINKED ISSUE #77: github_repository does not have attribute http_clone_url error
I want to create repository and initialize `git` locally after successful creation. Accoding to the doc [providers/github](https://www.terraform.io/docs/providers/github/r/repository.html#http_clone_url) there is the attribute `http_clone_url`.  So my `.tf` file looks like this.

```
variable "github_token" {}

provider "github" {
    token = "${var.github_token}"
    organization = ""
}

resource "github_repository" "sandbox" {
    name = "sandbox-application"
    description = "sandbox application for testing functionality"
    
    auto_init = true
}

# move to github_repository provisioner if possible
resource "null_resource" "init_local" {
    depends_on = ["github_repository.sandbox"]
    
    provisioner "local-exec" {
        command = "git init && git remote add origin ${github_repository.sandbox.http_clone_url}"
    }
}
```
But when I run `apply` I get the error
```
* null_resource.init_local: 1 error(s) occurred:

* Resource 'github_repository.sandbox' does not have attribute 'http_clone_url' for variable 'github_repository.sandbox.
http_clone_url'

Terraform does not automatically rollback in the face of errors.
Instead, your Terraform state file has been partially updated with
any resources that successfully completed. Please address the error
above and apply again to incrementally change your infrastructure.
```

So it seems this attribute does not exist? Or I am doing something wrong?

paultyng:
Fixes #77


website/docs/index.html.markdown
```diff
@@ -34,12 +34,11 @@ resource "github_membership" "membership_for_user_x" {
...
 The following arguments are supported in the `provider` block:
 
-* `token` - (Optional) This is the GitHub personal access token. It must be provided, but
+* `token` - (Required) This is the GitHub personal access token. It must be provided, but
   it can also be sourced from the `GITHUB_TOKEN` environment variable.
 
-* `organization` - (Optional) This is the target GitHub organization to manage. The account
-  corresponding to the token will need "owner" privileges for this organization. It must be provided, but
-  it can also be sourced from the `GITHUB_ORGANIZATION` environment variable.
+* `organization` - (Required) This is the target GitHub organization to manage. The account
```
tombuildsstuff: technically both of these are still optional since they can be sourced from the Environment Variables?

paultyng: Yeah I went to look at how AWS does it, they are inconsistent in their usage of it as well: https://www.terraform.io/docs/providers/aws/#region

IMO it should be marked as `Required` since it is a required piece of data, but the env vars list other ways to pass it, but maybe you feel strongly the other way?

tombuildsstuff: Perhaps it'd be clearer as `(Required unless specified through the environment variable `FOO_BAR`)?

PR #82, opened 2018-03-26
datasource/github_ip_ranges: implement ip datasource via the GH Meta api

### LINKED ISSUE #78: Add data source for GitHub IP ranges
I have a need to restrict a public service to GitHub's various IP ranges, but since GitHub changes these occasionally, I'd love to just use a data source to pull in those values.

I figured a `github_ip_ranges` data source would be the best approach.

This would use the GitHub's [Meta endpoint](https://developer.github.com/v3/meta/). Looking through the go-github source, this API has been mostly implemented, although there are some newer IP ranges in the endpoint (e.g., importer) that are not there.

appilon:
Fix #78 

Implemented data source to pull github's ip addresses. Rigged up the StopContext to pass through to the api call (should probably be done elsewhere in the provider as well).

One of the ip sets (importer) is not yet implemented in the `go-github` library. I'll open a PR for that so we can polish off this data source in a follow up PR.

go-github PR https://github.com/google/go-github/pull/881

### DIFF
--- a/github/data_source_github_ip_ranges.go
+++ b/github/data_source_github_ip_ranges.go
@@ -1,8 +1,6 @@
 package github
 
 import (
-	"context"
-
 	"github.com/hashicorp/terraform/helper/schema"
 )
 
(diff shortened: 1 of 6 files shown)


paultyng:
This would address #78 (just linking it in here).

github/data_source_github_ip_ranges.go
```diff
@@ -0,0 +1,48 @@
...
+				Computed: true,
+				Elem:     &schema.Schema{Type: schema.TypeString},
+			},
+		},
+	}
+}
+
+func dataSourceGithubIpRangesRead(d *schema.ResourceData, meta interface{}) error {
+	client := meta.(*Organization).client
+	ctx := context.Background()
```
paultyng: should this be the `StopContext`?

github/data_source_github_ip_ranges.go
```diff
@@ -0,0 +1,53 @@
...
+func dataSourceGithubIpRangesRead(d *schema.ResourceData, meta interface{}) error {
+	org := meta.(*Organization)
+
+	api, _, err := org.client.APIMeta(org.StopContext)
+	if err != nil {
+		return err
+	}
+
+	if len(api.Hooks)+len(api.Git)+len(api.Pages) > 0 {
+		d.SetId("github-ip-ranges")
```
appilon: I don't know what to set here? I suppose it doesn't matter in this case? Anyone let me know!

paultyng:
These are all always IPv4 I think? Just want to confirm that, and possibly document it.

appilon:
@paultyng the [/meta](https://developer.github.com/v3/meta/) endpoint gives examples that are IPv4 but there doesn't appear to be any docs that guarantee it. Let me know if you want me to change any documentation. Also I opened a PR to go-github to get the last set (importer) we could wait on that PR and then update this one with full coverage!

paultyng:
I'd say just link your PR in a `TODO` comment or something to indicate why it doesn't have that data in the data source, then this is good to go.

paultyng:
This was released in v1.1.0

PR #93, opened 2018-05-09
Add html_url to github_repository

sethvargo:
/cc @radeksimko


paultyng:
This was released in v1.1.0

PR #94, opened 2018-05-09
Update to latest vendor versions

sethvargo:
/cc @radeksimko


paultyng:
Travis is failing on this, so think there must be something weird going on, I'll see if i can diagnose locally.

paultyng:
Sorry that commit rolled it back from the latest commit to v15, let me fix

sethvargo:
LGTM

Issue #41, opened 2017-07-20
Add "topic" to the repo creation 


99:
Hi there,

this is more like a feature request. I don't see in docs option to add repo topics https://github.com/blog/2309-introducing-topics.
Will appreciate if that could be added.

Natalie

Techcadia:
https://developer.github.com/v3/repos/#list-all-topics-for-a-repository

Just adding that this is still in preview. Not sure what the thought process is on implementing preview features into a production level provider.

thomasbrezinski:
@radeksimko if you could confirm whether supporting a preview feature would be approved, I would be happy to take a look at implementing this :)

It's been in preview since January 2017

endorama:
Hello, I would really need this feature, and I would love to see it in terraform. 

Any update on this? I'm willing to help, if needed. Thanks

chandy:
@radeksimko would also love to get confirmation on @thomasbrezinski question about supporting a preview feature

paultyng:
If its in the SDK, I'd be happy to merge a PR for topic support.

Regarding preview features, we generally prefer not to support them until we can be sure an API/SDK is locked down, but seeing as topics seem pretty stable now and are also released in Github Enterprise I think it would be safe to support this feature.

tylersmith34:
I'm interested in this as well and could help with beta/preview testing.

PR #97, opened 2018-05-26
Add repository topics

endorama:
Hello, first contribution here! First of all thanks for the project.

Following #41, this PR should add topics support to `github_repository` terraform resource.

I'm having some trouble compiling due to an error: ( on master branch )
```
./main.go:10:15: cannot use "github.com/terraform-providers/terraform-provider-github/github".Provider (type func() "github.com/terraform-providers/terraform-provider-github/vendor/github.com/hashicorp/terraform/terraform".ResourceProvider) as type "github.com/endorama/terraform-provider-github/vendor/github.com/hashicorp/terraform/plugin".ProviderFunc in field value
```
so I can't run the acceptance tests ( which are included, but untested ).

It's also my first terraform provider contribution and I'm not so skilled in go, so please advice on any error I missed :slightly_smiling_face: 

cc @paultyng ( due to his last comment in #41 )

Thank you!

### DIFF
--- a/github/resource_github_repository.go
+++ b/github/resource_github_repository.go
@@ -92,6 +92,11 @@ func resourceGithubRepository() *schema.Resource {
 				Optional: true,
 				Default:  false,
 			},
+			"topics": {
+				Type:     schema.TypeList,
+				Elem:     &schema.Schema{Type: schema.TypeString},
+				Optional: true,
+			},
 
 			"full_name": {
 				Type:     schema.TypeString,
@@ -137,6 +142,7 @@ func resourceGithubRepositoryObject(d *schema.ResourceData) *github.Repository {
 	licenseTemplate := d.Get("license_template").(string)
 	gitIgnoreTemplate := d.Get("gitignore_template").(string)
 	archived := d.Get("archived").(bool)
+	topics := d.Get("topics").([]string)
 
 	repo := &github.Repository{
 		Name:              &name,
@@ -154,6 +160,7 @@ func resourceGithubRepositoryObject(d *schema.ResourceData) *github.Repository {
 		LicenseTemplate:   &licenseTemplate,
 		GitignoreTemplate: &gitIgnoreTemplate,
 		Archived:          &archived,
+		Topics:            topics,
 	}
 
 	return repo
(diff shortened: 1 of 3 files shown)


github/resource_github_repository.go
```diff
@@ -137,6 +142,7 @@ func resourceGithubRepositoryObject(d *schema.ResourceData) *github.Repository {
 	licenseTemplate := d.Get("license_template").(string)
 	gitIgnoreTemplate := d.Get("gitignore_template").(string)
 	archived := d.Get("archived").(bool)
+	topics := d.Get("topics").([]string)
```
paultyng: This will come out of `ResourceData` as `[]interface{}`, so you will need to write a simple func to "expand" this state to the SDK's representation:

```go
func expandStringList(configured []interface{}) []string {
	if len(configured) == 0 {
		return nil
	}
	vs := make([]string, 0, len(configured))
	for _, v := range configured {
		val, ok := v.(string)
		if ok && val != "" {
			vs = append(vs, val)
		}
	}
	return vs
}
```

github/resource_github_repository_test.go
```diff
@@ -377,6 +405,9 @@ func testAccCheckGithubRepositoryAttributes(repo *github.Repository, want *testA
 		if *repo.HasDownloads != want.HasDownloads {
 			return fmt.Errorf("got has downloads %#v; want %#v", *repo.HasDownloads, want.HasDownloads)
 		}
+		if &repo.Topics != &want.Topics {
```
paultyng: You will probably need to rewrite this comparison somewhat, as the default value for a slice is `nil` but its equivalent to an empty slice, so you'll want to check len and iterate:

```go
if len(want.Topics) != len(repo.Topics) {
			return fmt.Errorf("got has topics %#v; want %#v", repo.Topics, want.Topics)
		}
		for i := range want.Topics {
			if repo.Topics[i] != want.Topics[i] {
				return fmt.Errorf("got has topics %#v; want %#v", repo.Topics, want.Topics)
			}
		}
```

paultyng:
This one was a little weird as topics actually needed to be updated via a different API call, so I went ahead and pushed a new commit with the modifications, I'll need to get someone else to do a review.

chandy:
Thanks all!

endorama:
@paultyng thank you for the merge.

I wasn't able to step in before today, but I really appreciated your review; I'm sorry I didn't updated the code in time.

Thanks!

PR #111, opened 2018-08-04
New Resource: github_organization_project

shihanng:
Hi, this PR adds the ability to manage organization projects to this provider as described in https://developer.github.com/v3/projects/#create-an-organization-project. This is my first time working with Terraform provider, please let me know if I miss anything. Thank you.

### DIFF
--- a/github/resource_github_organization_project.go
+++ b/github/resource_github_organization_project.go
@@ -1,0 +1,116 @@
+package github
+
+import (
+	"context"
+	"fmt"
+	"strconv"
+
+	"github.com/google/go-github/github"
+	"github.com/hashicorp/terraform/helper/schema"
+)
+
+func resourceGithubOrganizationProject() *schema.Resource {
+	return &schema.Resource{
+		Create: resourceGithubOrganizationProjectCreate,
+		Read:   resourceGithubOrganizationProjectRead,
+		Update: resourceGithubOrganizationProjectUpdate,
+		Delete: resourceGithubOrganizationProjectDelete,
+		Importer: &schema.ResourceImporter{
+			State: schema.ImportStatePassthrough,
+		},
+
+		Schema: map[string]*schema.Schema{
+			"name": {
+				Type:     schema.TypeString,
+				Required: true,
+			},
+			"body": {
+				Type:     schema.TypeString,
+				Optional: true,
+			},
+			"url": {
+				Type:     schema.TypeString,
+				Computed: true,
+			},
+		},
+	}
+}
+
+func resourceGithubOrganizationProjectCreate(d *schema.ResourceData, meta interface{}) error {
+	client := meta.(*Organization).client
+	o := meta.(*Organization).name
+	n := d.Get("name").(string)
+	b := d.Get("body").(string)
+
+	options := github.ProjectOptions{
+		Name: n,
+		Body: b,
+	}
+
+	project, _, err := client.Organizations.CreateProject(context.TODO(), o, &options)
+	if err != nil {
+		return err
+	}
+	d.SetId(strconv.FormatInt(*project.ID, 10))
+
+	return resourceGithubOrganizationProjectRead(d, meta)
+}
+
+func resourceGithubOrganizationProjectRead(d *schema.ResourceData, meta interface{}) error {
+	client := meta.(*Organization).client
+	o := meta.(*Organization).name
+
+	projectID, err := strconv.ParseInt(d.Id(), 10, 64)
+	if err != nil {
+		return err
+	}
+
+	project, resp, err := client.Projects.GetProject(context.TODO(), projectID)
+	if err != nil {
+		if resp != nil && resp.StatusCode == 404 {
+			d.SetId("")
+			return nil
+		}
+		return err
+	}
+
+	d.Set("name", project.GetName())
+	d.Set("body", project.GetBody())
+	d.Set("url", fmt.Sprintf("https://github.com/orgs/%s/projects/%d", o, project.GetNumber()))
+
+	return nil
+}
+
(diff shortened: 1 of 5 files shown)


shihanng:
It seems that there is conflict. Would you like me to rebase?

radeksimko:
@shihanng yes, I merged a few PRs lately and I'm going through others, including this one. I'm going to review it regardless, but rebasing would be incredibly helpful.

Thanks!

github/resource_github_organization_project.go
```diff
@@ -0,0 +1,116 @@
...
+				Required: true,
+			},
+			"body": {
+				Type:     schema.TypeString,
+				Optional: true,
+			},
+			"url": {
+				Type:     schema.TypeString,
+				Computed: true,
+			},
```
radeksimko: How do you feel about coverage of the remaining fields which are available through `PATCH`?
https://developer.github.com/v3/projects/#update-a-project
specifically `state`, `organization_permission` and `public`?

github/resource_github_organization_project.go
```diff
@@ -0,0 +1,116 @@
...
+func resourceGithubOrganizationProjectCreate(d *schema.ResourceData, meta interface{}) error {
+	client := meta.(*Organization).client
+	o := meta.(*Organization).name
+	n := d.Get("name").(string)
+	b := d.Get("body").(string)
+
+	options := github.ProjectOptions{
+		Name: n,
+		Body: b,
+	}
```
radeksimko: Nitpick (non-blocking): I think we could just inline the fields above to the struct to reduce LOC and improve readability a bit.

github/resource_github_organization_project.go
```diff
@@ -0,0 +1,116 @@
...
+
+func resourceGithubOrganizationProjectUpdate(d *schema.ResourceData, meta interface{}) error {
+	client := meta.(*Organization).client
+	n := d.Get("name").(string)
+	b := d.Get("body").(string)
+
+	options := github.ProjectOptions{
+		Name: n,
+		Body: b,
+	}
```
radeksimko: Nitpick - as above.

website/docs/r/organization_project.html.markdown
```diff
@@ -0,0 +1,34 @@
...
+}
+```
+
+## Argument Reference
+
+The following arguments are supported:
+
+* `name` - The name of the project.
+
+* `body` - The body of the project.
```
radeksimko: Can you please document whether each field is `Optional` or `Required` here?

github/resource_github_organization_project_test.go
```diff
@@ -0,0 +1,114 @@
...
+					testAccCheckGithubOrganizationProjectAttributes(&project, &testAccGithubOrganizationProjectExpectedAttributes{
+						Name: "test-project",
+						Body: "this is a test project",
+					}),
+				),
+			},
+		},
+	})
+}
+
```
radeksimko: Do you mind adding a test for import also?

github/resource_github_organization_project.go
```diff
@@ -0,0 +1,116 @@
...
+		if resp != nil && resp.StatusCode == 404 {
+			d.SetId("")
+			return nil
+		}
+		return err
+	}
+
+	d.Set("name", project.GetName())
+	d.Set("body", project.GetBody())
+	d.Set("url", fmt.Sprintf("https://github.com/orgs/%s/projects/%d", o, project.GetNumber()))
```
radeksimko: Is there any reason we should be constructing the URL here instead of reading it from the API? It appears that [`html_url`](https://developer.github.com/v3/projects/#create-an-organization-project) is the attribute we could use?

github/resource_github_organization_project.go
shihanng: :thinking: It seems that the library that we are using does now support these fields yet?
- There is a way to set the `state` via [`ProjectOptions`](https://godoc.org/github.com/google/go-github/github#ProjectOptions) but the value of the `state` is not included in [`Project`](https://godoc.org/github.com/google/go-github/github#Project) so we can't get the current state of the project via the library.
- For `organization_permission`, is seems that there is an issue opened and being worked on currently https://github.com/google/go-github/issues/884
- For `public`, is it not available via the go-github library yet.

github/resource_github_organization_project.go
shihanng: Ah yes, but it is not made available in the go-github library? See [Project](https://godoc.org/github.com/google/go-github/github#Project).

github/resource_github_organization_project_test.go
shihanng: Added test for import in ec08678. Thank you.

website/docs/r/organization_project.html.markdown
shihanng: Added in 2ea7426. Thanks!

github/resource_github_organization_project.go
shihanng: Fixed in f1a04f4, thank you.

github/resource_github_organization_project.go
shihanng: Fixed in f1a04f4, thank you.

shihanng:
@radeksimko, thank you for the reviews. I've addressed some comments in code, and also the rest as comments. Could you please take another look. Thank you.

github/resource_github_organization_project.go
radeksimko: Ah, understood. Thanks for the detailed answer & research. We shall leave it out for the initial implementation then.

shihanng:
Thank you so much 😄! I will work on #115 later on and ping you again when it is ready.

PR #116, opened 2018-08-08
Refactoring & naming cleanup

radeksimko:
Some may argue that removing the `toGithubID` helper function is a step back, but I think it's hiding the intention.

`toGithubID` poses a lot of questions to me - _What ID?_, _Why is it different from the one we're passing in?_ and error (which in theory should never occur, but I'm not a huge fan of hiding errors).

So downsides IMO overweigh the benefits (few saved LOC via abstraction), hence I made the type conversion more explicit.


radeksimko:
Addressed your comment and ran all acceptance tests, all still passing, so I'm going to merge this.

PR #120, opened 2018-08-09
New Resource: github_user_gpg_key

radeksimko:
```
TF_ACC=1 go test ./github -v -run=TestAccGithubUserGpgKey_ -timeout 120m
=== RUN   TestAccGithubUserGpgKey_basic
--- PASS: TestAccGithubUserGpgKey_basic (2.17s)
PASS
ok  	github.com/terraform-providers/terraform-provider-github/github	2.281s
```


github/resource_github_user_gpg_key.go
```diff
@@ -0,0 +1,75 @@
...
+
+	id, err := strconv.ParseInt(d.Id(), 10, 64)
+	if err != nil {
+		return unconvertibleIdErr(d.Id(), err)
+	}
+
+	key, _, err := client.Users.GetGPGKey(context.TODO(), id)
+	if err != nil {
+		d.SetId("")
+		return nil
```
catsby: Should log that the `d.Id()` wasn't found and removed from state

github/resource_github_user_gpg_key.go
```diff
@@ -0,0 +1,75 @@
...
+	client := meta.(*Organization).client
+
+	pubKey := d.Get("armored_public_key").(string)
+
+	key, _, err := client.Users.CreateGPGKey(context.TODO(), pubKey)
+	if err != nil {
+		return err
+	}
+
+	d.SetId(strconv.FormatInt(*key.ID, 10))
```
catsby: Should we use `key.KeyID` here instead? Looking at GitHub docs it looks like `key.ID` is a database index-like sequential key, where as `KeyID` appears to be a unique id.

It also saves us the trouble of using `FormatInt` and error checking

catsby: I guess that depends on what `GetGPGKey` uses... the docs make it look like it uses `key_id` but I can't imagine your tests passed if that's what they expected and you weren't doing that 🤔

catsby: I've answered my own question. How bizarre 🤔

github/resource_github_user_gpg_key.go
radeksimko: Agreed.

PR #121, opened 2018-08-10
Cleanup github_branch_protection

radeksimko:
## Summary

1. get rid of the one-letter variable names
1. reduce LOC a bit by returning early
1. add `WARN` when wiping resource from state, per convention

### DIFF
--- a/github/resource_github_branch_protection.go
+++ b/github/resource_github_branch_protection.go
@@ -4,7 +4,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
-	"net/http"
+	"log"
 
 	"github.com/google/go-github/github"
 	"github.com/hashicorp/terraform/helper/schema"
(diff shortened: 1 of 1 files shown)


github/resource_github_branch_protection.go
```diff
@@ -183,7 +193,7 @@ func resourceGithubBranchProtectionRead(d *schema.ResourceData, meta interface{}
 
 func resourceGithubBranchProtectionUpdate(d *schema.ResourceData, meta interface{}) error {
 	client := meta.(*Organization).client
-	r, b, err := parseTwoPartID(d.Id())
+	repoName, branch, err := parseTwoPartID(d.Id())
```
tombuildsstuff: tbh parseTwoPartId could return a struct so this could become `id.RepoName` / `id.Branch`

github/resource_github_branch_protection.go
```diff
@@ -193,58 +203,71 @@ func resourceGithubBranchProtectionUpdate(d *schema.ResourceData, meta interface
...
-	protectionRequest.Restrictions = res
-
-	protectionRequest.EnforceAdmins = d.Get("enforce_admins").(bool)
+	req.Restrictions = res
 
-	return protectionRequest, nil
+	return req, nil
 }
 
 func flattenRequiredStatusChecks(d *schema.ResourceData, protection *github.Protection) error {
```
tombuildsstuff: shouldn't this be `flattenAndSetRequiredStatusChecks`

github/resource_github_branch_protection.go
```diff
@@ -297,11 +318,9 @@ func flattenRequiredPullRequestReviews(d *schema.ResourceData, protection *githu
...
 		}
-	} else {
-		d.Set("required_pull_request_reviews", []interface{}{})
 	}
 
-	return nil
+	return d.Set("required_pull_request_reviews", []interface{}{})
 }
 
 func flattenRestrictions(d *schema.ResourceData, protection *github.Protection) error {
```
tombuildsstuff: (same here)

github/resource_github_branch_protection.go
```diff
@@ -263,11 +286,9 @@ func flattenRequiredStatusChecks(d *schema.ResourceData, protection *github.Prot
...
 		}
-	} else {
-		d.Set("required_status_checks", []interface{}{})
 	}
 
-	return nil
+	return d.Set("required_status_checks", []interface{}{})
 }
 
 func flattenRequiredPullRequestReviews(d *schema.ResourceData, protection *github.Protection) error {
```
tombuildsstuff: (same here)

github/resource_github_branch_protection.go
radeksimko: Yeah, good point, I'll change that.

github/resource_github_branch_protection.go
radeksimko: Yeah, I'll change it. I kind of wish the function wasn't as greedy and did one thing only, but that's refactoring for another time.

github/resource_github_branch_protection.go
radeksimko: Hm, I'm torn on this... I guess it makes a bit more sense to represent it that way, on the other side it adds 3 characters for each variable occurrence.

I don't have a strong opinion, but I think it's outside of scope of the PRs I opened. 😉

PR #122, opened 2018-08-10
Cleanup github_issue_label

radeksimko:
## Summary

1. get rid of the one-letter variable names
1. add `WARN` when wiping resource from state, per convention
1. take advantage of the SDK helpers and reduce LOC

### DIFF
--- a/github/resource_github_issue_label.go
+++ b/github/resource_github_issue_label.go
@@ -56,50 +56,57 @@ func resourceGithubIssueLabel() *schema.Resource {
 
 func resourceGithubIssueLabelCreateOrUpdate(d *schema.ResourceData, meta interface{}) error {
 	client := meta.(*Organization).client
-	o := meta.(*Organization).name
-	r := d.Get("repository").(string)
-	n := d.Get("name").(string)
-	c := d.Get("color").(string)
+	orgName := meta.(*Organization).name
+	repoName := d.Get("repository").(string)
+	name := d.Get("name").(string)
+	color := d.Get("color").(string)
 
 	label := &github.Label{
-		Name:  &n,
-		Color: &c,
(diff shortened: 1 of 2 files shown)


github/resource_github_issue_label.go
```diff
@@ -56,50 +56,54 @@ func resourceGithubIssueLabel() *schema.Resource {
...
-		Color: &c,
+		Name:  github.String(name),
+		Color: github.String(color),
 	}
 
-	log.Printf("[DEBUG] Querying label existence %s/%s (%s)", o, r, n)
-	existing, _, _ := client.Issues.GetLabel(context.TODO(), o, r, n)
+	log.Printf("[DEBUG] Querying label existence %s/%s (%s)",
+		orgName, repoName, name)
+	existing, _, _ := client.Issues.GetLabel(context.TODO(),
```
tombuildsstuff: I'm presuming one of these is an error, if so, should we handle it?

radeksimko: Actually no, this is intentional, let me find the relevant PR.

radeksimko: https://github.com/terraform-providers/terraform-provider-github/commit/8116f8f656dc770847f18028671a46d35a15f02e

radeksimko: We should probably catch the 422 and 404 response codes here, instead of `existing != nil` though and error out on other error codes.

tombuildsstuff: yep, we should be safe to ignore 422 and 404 and throw for the other errors 👍

radeksimko: 422 is actually error code returned on failed creation, so 404 should suffice.

PR #132, opened 2018-08-13
resource/github_repository_deploy_key: Trim key

### LINKED ISSUE #69: github_repository_deploy_key forces new resource
I'm getting a "forces new resource" every time I run `terraform plan` or `terraform apply` regardless of whether the key already exists.

### Terraform Version

```
Terraform v0.11.2
+ provider.aws v1.6.0
+ provider.github v0.1.1
```

### Affected Resource(s)

```hcl
variable "github_token" {}

provider "github" { 
  organization = "SomeOrg"
  token = "${var.github_token}"
}

resource "github_repository_deploy_key" "webapp_repo_deploy_key" {
    title = "Deploy key"
    repository = "SomeRepo"
    key = "redacted"
    read_only = "true"
}
```

### Expected Behavior
Once the deploy key exists, don't recreate it.

### Actual Behavior
Terraform recreates the resource:

```
-/+ github_repository_deploy_key.webapp_repo_deploy_key (new resource required)
      id:         "SomeRepo:26817612" => <computed> (forces new resource)
      key:        "redacted" (forces new resource)
      read_only:  "true" => "true"
      repository: "SomeRepo" => "SomeRepo"
      title:      "Deploy key" => "Deploy key"


### Steps to Reproduce
1. `terraform plan`  (`terraform apply` does recreate the resource)

### PR DESCRIPTION
**Pull request by radeksimko**
Fixes #69

```
TF_ACC=1 go test ./github -v -run=TestAccGithubRepositoryDeployKey_ -timeout 120m
=== RUN   TestAccGithubRepositoryDeployKey_basic
--- PASS: TestAccGithubRepositoryDeployKey_basic (4.29s)
=== RUN   TestAccGithubRepositoryDeployKey_importBasic
--- PASS: TestAccGithubRepositoryDeployKey_importBasic (3.69s)
PASS
ok  	github.com/terraform-providers/terraform-provider-github/github	8.108s
```

### CODE REVIEW

File: github/resource_github_repository_deploy_key_test.go [hunk 1]
```diff
@@ -11,21 +13,43 @@ import (
 	"github.com/hashicorp/terraform/terraform"
 )
 
+func TestSuppressDeployKeyDiff(t *testing.T) {
+	oldV := "ssh-rsa AAAABB...cd+== terraform-acctest@hashicorp.com\n"
+	newV := "ssh-rsa AAAABB...cd+=="
+	if suppressDeployKeyDiff("test", oldV, newV, nil) {
+		t.Fatalf("Expected %q and %q to be suppressed", oldV, newV)
```
appilon: did you mean NOT suppressed in the error message?

github/resource_github_repository_deploy_key_test.go
```diff
@@ -11,21 +13,43 @@ import (
 	"github.com/hashicorp/terraform/terraform"
 )
 
+func TestSuppressDeployKeyDiff(t *testing.T) {
```
appilon: I'm not really strongly opinionated on this, but typically I do 1 validation per test and typically the function name is quite verbose, almost describing the test circumstance. What is more important though is our unit test style is consistent, so if this is consistent with everything it's all good with me

github/resource_github_repository_deploy_key_test.go
appilon: This is the only real blocker for me on the PR, either the message is missing `NOT` or the conditional should be negated.

radeksimko: You're right, the reason the test was passing is because I accidentally swapped `oldV` and `newV`. 🙈 
Fixed now!

github/resource_github_repository_deploy_key_test.go
radeksimko: That's fair point, I'll break it apart.

radeksimko: Actually we do have many unit tests (set aside acceptance tests which can have multiple steps) which have multiple test cases in them and I think it's fine? e.g. https://github.com/terraform-providers/terraform-provider-aws/blob/master/aws/validators_test.go#L12-L54

The thing which I agree needs changing though is the error message, so it's more obvious where each failure is coming from. 👍 I will fix that.

github/resource_github_repository_deploy_key_test.go
```diff
@@ -11,21 +13,60 @@ import (
...
+	testCases := []struct {
```
appilon: nice table driven tests!

PR #133, opened 2018-08-13
resource/github_repository_webhook: Avoid spurious diff

### LINKED ISSUE #81: github_repository_webhook secret always generates an "update in-place"
### Terraform Version
Terraform v0.11.5
+ provider.github v1.0.0

### Affected Resource(s)
- github_repository_webhook

```hcl

provider "github" {
    token        = "${var.github_token}"
    organization = "${var.github_organization}"
}

resource "github_repository_webhook" "example" {
    repository = "example"
    name       = "web"
    active     = true
    events     = ["push"]
  
    configuration {
        url          = "https://example.com"
        content_type = "json"
        secret       = "This is a secret"
        insecure_ssl = 0
    }
}
```

### Debug Output
This gist shows three commands.  `plan`, `apply`, and `plan` again.  The first two create a new webhook with a secret.  The last `plan` shows that it wants to overwrite the secret.

https://gist.github.com/enobrev/0d574a7e396e4a0b81ec1851b21e812c

### Expected Behavior
```bash
No changes. Infrastructure is up-to-date.
```

I'm not sure how this is generally handled, where we cannot confirm the remote value on the provider with what was set in the template, but generating a new "update in-place" every time terraform is run isn't ideal.

### Actual Behavior
```bash
Resource actions are indicated with the following symbols:
  ~ update in-place

  ~ github_repository_webhook.example
      configuration.secret: "********" => "This is a secret"

```

### Steps to Reproduce
1. Create a `github_repository_webhook` with `configuration.secret` set to any string (besides "********")
2. `terraform plan`
3. `terraform apply`
4. `terraform plan`

### Assumptions
Since the github api will return "********" instead of the actual password, terraform will compare what's returned by the api and what's set in the template, and since there's a difference, try to write what's in the template.

### References
https://developer.github.com/v3/repos/hooks/

### Workarounds
For the time being, I've been creating webhooks with the correct `configuration.secret`, and then after it's been "applied", I change the value in of `configuration.secret` my template to "********".

radeksimko:
Fixes #81

```
TF_ACC=1 go test ./github -v -run=TestAccGithubRepositoryWebhook_ -timeout 120m
=== RUN   TestAccGithubRepositoryWebhook_basic
--- PASS: TestAccGithubRepositoryWebhook_basic (6.25s)
=== RUN   TestAccGithubRepositoryWebhook_secret
--- PASS: TestAccGithubRepositoryWebhook_secret (3.80s)
=== RUN   TestAccGithubRepositoryWebhook_importBasic
--- PASS: TestAccGithubRepositoryWebhook_importBasic (4.94s)
=== RUN   TestAccGithubRepositoryWebhook_importSecret
--- PASS: TestAccGithubRepositoryWebhook_importSecret (5.05s)
PASS
ok  	github.com/terraform-providers/terraform-provider-github/github	20.069s
```

### DIFF
--- a/github/migrate_github_repository_webhook.go
+++ b/github/migrate_github_repository_webhook.go
@@ -1,0 +1,55 @@
+package github
+
+import (
+	"fmt"
+	"log"
+	"strings"
+
+	"github.com/hashicorp/terraform/terraform"
+)
+
+func resourceGithubRepositoryWebhookMigrateState(v int, is *terraform.InstanceState, meta interface{}) (*terraform.InstanceState, error) {
+	switch v {
+	case 0:
+		log.Println("[INFO] Found GitHub Repository Webhook State v0; migrating to v1")
+		return migrateGithubRepositoryWebhookStateV0toV1(is)
+	default:
+		return is, fmt.Errorf("Unexpected schema version: %d", v)
+	}
+}
+
+func migrateGithubRepositoryWebhookStateV0toV1(is *terraform.InstanceState) (*terraform.InstanceState, error) {
+	if is.Empty() {
+		log.Println("[DEBUG] Empty InstanceState; nothing to migrate.")
+		return is, nil
+	}
+
+	log.Printf("[DEBUG] GitHub Repository Webhook Attributes before migration: %#v", is.Attributes)
+
+	prefix := "configuration."
+
+	delete(is.Attributes, prefix+"%")
+
+	// Read & delete old keys
+	oldKeys := make(map[string]string, 0)
+	for k, v := range is.Attributes {
+		if strings.HasPrefix(k, prefix) {
+			oldKeys[k] = v
+
+			// Delete old keys
+			delete(is.Attributes, k)
+		}
+	}
+
+	// Write new keys
+	for k, v := range oldKeys {
+		newKey := "configuration.0." + strings.TrimPrefix(k, prefix)
+		is.Attributes[newKey] = v
+	}
+
+	is.Attributes[prefix+"#"] = "1"
+
+	log.Printf("[DEBUG] GitHub Repository Webhook Attributes after State Migration: %#v", is.Attributes)
+
+	return is, nil
+}
--- a/github/resource_github_repository_webhook.go
+++ b/github/resource_github_repository_webhook.go
@@ -28,6 +28,9 @@ func resourceGithubRepositoryWebhook() *schema.Resource {
 			},
 		},
 
+		SchemaVersion: 1,
+		MigrateState:  resourceGithubRepositoryWebhookMigrateState,
+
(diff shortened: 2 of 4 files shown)


github/migrate_github_repository_webhook.go
```diff
@@ -0,0 +1,55 @@
...
+		if strings.HasPrefix(k, prefix) {
+			oldKeys[k] = v
+
+			// Delete old keys
+			delete(is.Attributes, k)
+		}
+	}
+
+	// Write new keys
+	for k, v := range oldKeys {
```
appilon: any chance this could be done all in the previous loop?

github/resource_github_repository_webhook.go
```diff
@@ -46,8 +49,38 @@ func resourceGithubRepositoryWebhook() *schema.Resource {
...
+						"content_type": {
+							Type:     schema.TypeString,
+							Optional: true,
+						},
+						"secret": {
+							Type:      schema.TypeString,
+							Optional:  true,
+							Sensitive: true,
+							DiffSuppressFunc: func(k, oldV, newV string, d *schema.ResourceData) bool {
+								maskedSecret := "********"
```
appilon: I haven't really worked with sensitive attributes, but it feels strange to me for a maskedSecret to ALWAYS be 8 asterisks, is this code robust?

github/migrate_github_repository_webhook.go
radeksimko: That was actually the very first version of this PR 😃until I saw the test intermittently failing as the for loop kept ranging over some new elements which were just added.

Why? https://golang.org/ref/spec#For_range

> The iteration order over maps is not specified and is not guaranteed to be the same from one iteration to the next. If a map entry that has not yet been reached is removed during iteration, the corresponding iteration value will not be produced. **If a map entry is created during iteration, that entry may be produced during the iteration or may be skipped. The choice may vary for each entry created and from one iteration to the next.**

Short answer: No.

Feel free to have a play yourself: https://play.golang.org/p/Tm1FPyjwd0m 😉

github/resource_github_repository_webhook.go
radeksimko: GitHub unfortunately doesn't document this, but I assumed it's a static placeholder they just return. Why would/should the API ever return different number of asterisks?

btw. this is unrelated to how we treat sensitive data in the schema via `Sensitive`, it's more related to how the backend API (GitHub in this case) implements it.

appilon: okay, I think a code comment might be appropriate in this case mentioning it appears to be an undocumented GH api response.

github/migrate_github_repository_webhook.go
appilon: ahhhh yes I see now. 👍

PR #134, opened 2018-08-15
resource/github_organization_webhook: Avoid spurious diff

radeksimko:
Similar to #133 the same bug appears in org webhook, which has pretty much the same schema, so I refactored some of the functionality to reuse it.

Here's a test failing prior to the patch:

```
TF_ACC=1 go test ./github -v -run=TestAccGithubOrganizationWebhook_ -timeout 120m
=== RUN   TestAccGithubOrganizationWebhook_basic
--- PASS: TestAccGithubOrganizationWebhook_basic (3.22s)
=== RUN   TestAccGithubOrganizationWebhook_secret
--- FAIL: TestAccGithubOrganizationWebhook_secret (1.38s)
	testing.go:518: Step 0 error: After applying this step, the plan was not empty:

		DIFF:

		UPDATE: github_organization_webhook.foo
		  configuration.secret: "********" => "VerySecret"

		STATE:

		github_organization_webhook.foo:
		  ID = 44243925
		  provider = provider.github
		  active = true
		  configuration.% = 4
		  configuration.content_type = json
		  configuration.insecure_ssl = 0
		  configuration.secret = ********
		  configuration.url = https://www.terraform.io/webhook
		  events.# = 1
		  events.2342231791 = pull_request
		  name = web
		  url = https://api.github.com/orgs/terraformtesting/hooks/44243925
FAIL
FAIL	github.com/terraform-providers/terraform-provider-github/github	4.633s
```

## After patch

```
TF_ACC=1 go test ./github -v -run=TestAccGithubOrganizationWebhook_ -timeout 120m
=== RUN   TestAccGithubOrganizationWebhook_basic
--- PASS: TestAccGithubOrganizationWebhook_basic (2.23s)
=== RUN   TestAccGithubOrganizationWebhook_secret
--- PASS: TestAccGithubOrganizationWebhook_secret (1.21s)
PASS
```

### DIFF
--- a/github/migrate_github_repository_webhook.go
+++ b/github/migrate_github_repository_webhook.go
@@ -8,23 +8,23 @@ import (
 	"github.com/hashicorp/terraform/terraform"
 )
 
-func resourceGithubRepositoryWebhookMigrateState(v int, is *terraform.InstanceState, meta interface{}) (*terraform.InstanceState, error) {
+func resourceGithubWebhookMigrateState(v int, is *terraform.InstanceState, meta interface{}) (*terraform.InstanceState, error) {
 	switch v {
 	case 0:
-		log.Println("[INFO] Found GitHub Repository Webhook State v0; migrating to v1")
-		return migrateGithubRepositoryWebhookStateV0toV1(is)
+		log.Println("[INFO] Found GitHub Webhook State v0; migrating to v1")
+		return migrateGithubWebhookStateV0toV1(is)
 	default:
 		return is, fmt.Errorf("Unexpected schema version: %d", v)
 	}
 }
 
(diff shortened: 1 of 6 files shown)


github/util.go
```diff
@@ -12,6 +12,44 @@ const (
 	maxPerPage = 100
 )
 
+func webhookConfigurationSchema() *schema.Schema {
```
appilon: Don't love where this lives, maybe create a file named something like `webhook_helpers.go`

radeksimko: Ah, yeah, I had mixed feelings about that too, let me clean it up 🙂

radeksimko:
@appilon Moved the schema into its own file, PTAL.

PR #145, opened 2018-09-05
Implement & use RateLimitTransport

### LINKED ISSUE #5: Github Provider: Support API Rate Limiting
_This issue was originally opened by @johnrengelman as hashicorp/terraform#6016. It was migrated here as part of the [provider split](https://www.hashicorp.com/blog/upcoming-provider-changes-in-terraform-0-10/). The original body of the issue is below._

<hr>

The Github API has a rate limit than can be easily overcome when managing an organization.
This ticket is simply a spot to start a conversation about how to properly handle the rate limit.

Details: https://developer.github.com/v3/#rate-limiting

radeksimko:
As described in attached code comments this PR implements best practices GitHub describes [in their official documentation](https://developer.github.com/v3/guides/best-practices-for-integrators/#dealing-with-abuse-rate-limits).

This is second of two PRs addressing #5 
Closes #5

The upside is that with this patch user should never see rate-limit related errors anymore as we'll always retry accordingly and more importantly avoid hitting the limit.

Downsides include slowdown of CUD operations and generally slower UX when managing large number of resources. Admittedly if the previous/current experience was "rate limit errors" and now is slow, but successful apply/destroy I'd still call that a win. 🤷‍♂️ See some figures below.

We could in theory expose a flag to turn off the sleep logic and/or mutex, but I'm not too keen on doing that pre-emptively without further context, esp. because we're _just_ implementing GitHub's own best practices.

### Current implementation

 - creation: 16 seconds
 - refresh: 4 seconds
 - destroy: 9 seconds

### Serializing requests via mutex

 - creation: 147 seconds (`9x` slower)
 - refresh: 25 seconds (`6x` slower)
 - destroy: 48 seconds (`5x` slower)

### 1sec delay between write operations

 - creation: 352 seconds (`22x` and `2.4x` slower)
 - refresh: 25 seconds (`6x` slower and equal)
 - destroy: 150 seconds (`16x` and `3x` slower)

### DIFF
--- a/github/transport.go
+++ b/github/transport.go
@@ -62,11 +62,8 @@ func (rlt *rateLimitTransport) RoundTrip(req *http.Request) (*http.Response, err
 		log.Printf("[DEBUG] Sleeping %s between write operations", writeDelay)
 		time.Sleep(writeDelay)
 	}
-	if rlt.isWriteMethod(req.Method) {
-		rlt.delayNextRequest = true
-	} else {
-		rlt.delayNextRequest = false
-	}
+
+	rlt.delayNextRequest = isWriteMethod(req.Method)
 
 	resp, err := rlt.transport.RoundTrip(req)
 	if err != nil {
@@ -77,15 +74,12 @@ func (rlt *rateLimitTransport) RoundTrip(req *http.Request) (*http.Response, err
 	// Make response body accessible for retries & debugging
 	// (work around bug in GitHub SDK)
 	// See https://github.com/google/go-github/pull/986
-	r1, r2, err := rlt.drainBody(resp.Body)
+	r1, r2, err := drainBody(resp.Body)
 	if err != nil {
 		return nil, err
 	}
 	resp.Body = r1
 	ghErr := github.CheckResponse(resp)
-	if err != nil {
-		return nil, err
-	}
 	resp.Body = r2
 
 	// When you have been limited, use the Retry-After response header to slow down.
@@ -114,9 +108,25 @@ func (rlt *rateLimitTransport) RoundTrip(req *http.Request) (*http.Response, err
 	return resp, nil
 }
 
+func (rlt *rateLimitTransport) lock(req *http.Request) {
+	ctx := req.Context()
+	log.Printf("[TRACE] Aquiring lock for GitHub API request (%q)", ctx.Value(ctxId))
+	rlt.m.Lock()
+}
+
+func (rlt *rateLimitTransport) unlock(req *http.Request) {
+	ctx := req.Context()
+	log.Printf("[TRACE] Releasing lock for GitHub API request (%q)", ctx.Value(ctxId))
+	rlt.m.Unlock()
+}
+
+func NewRateLimitTransport(rt http.RoundTripper) *rateLimitTransport {
+	return &rateLimitTransport{transport: rt}
+}
+
 // drainBody reads all of b to memory and then returns two equivalent
 // ReadClosers yielding the same bytes.
-func (rlt *rateLimitTransport) drainBody(b io.ReadCloser) (r1, r2 io.ReadCloser, err error) {
+func drainBody(b io.ReadCloser) (r1, r2 io.ReadCloser, err error) {
 	if b == http.NoBody {
 		// No copying needed. Preserve the magic sentinel meaning of NoBody.
 		return http.NoBody, http.NoBody, nil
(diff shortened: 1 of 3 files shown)


github/transport.go
```diff
@@ -29,3 +38,119 @@ func (ett *etagTransport) RoundTrip(req *http.Request) (*http.Response, error) {
...
+	}
+
+	rlt.unlock(req)
+
+	return resp, nil
+}
+
+// drainBody reads all of b to memory and then returns two equivalent
+// ReadClosers yielding the same bytes.
+func (rlt *rateLimitTransport) drainBody(b io.ReadCloser) (r1, r2 io.ReadCloser, err error) {
```
appilon: do you forsee this method needing `rlt` in scope? Not a huge preference but I would just define this as a helper `drainBody` as opposed to a method of `*rateLimitTransport`

github/transport.go
```diff
@@ -29,3 +38,119 @@ func (ett *etagTransport) RoundTrip(req *http.Request) (*http.Response, error) {
...
+	rlt.m.Lock()
+}
+
+func (rlt *rateLimitTransport) unlock(req *http.Request) {
+	ctx := req.Context()
+	log.Printf("[TRACE] Releasing lock for GitHub API request (%q)", ctx.Value(ctxId))
+	rlt.m.Unlock()
+}
+
+func (rlt *rateLimitTransport) isWriteMethod(method string) bool {
```
appilon: this method I feel more strongly to having just as a little helper and not a method

github/transport.go
```diff
@@ -29,3 +38,119 @@ func (ett *etagTransport) RoundTrip(req *http.Request) (*http.Response, error) {
...
+	// and restoring bodies between retries below
+	rlt.lock(req)
+
+	// If you're making a large number of POST, PATCH, PUT, or DELETE requests
+	// for a single user or client ID, wait at least one second between each request.
+	if rlt.delayNextRequest {
+		log.Printf("[DEBUG] Sleeping %s between write operations", writeDelay)
+		time.Sleep(writeDelay)
+	}
+	if rlt.isWriteMethod(req.Method) {
```
appilon: nitpick suggestion:
```
rlt.delayNextRequest = rlt.isWriteMethod(req.Method)
```

github/transport.go
```diff
@@ -29,3 +38,119 @@ func (ett *etagTransport) RoundTrip(req *http.Request) (*http.Response, error) {
...
+	// Make response body accessible for retries & debugging
+	// (work around bug in GitHub SDK)
+	// See https://github.com/google/go-github/pull/986
+	r1, r2, err := rlt.drainBody(resp.Body)
+	if err != nil {
+		return nil, err
+	}
+	resp.Body = r1
+	ghErr := github.CheckResponse(resp)
+	if err != nil {
```
appilon: need to check `ghErr` or reuse defined `err`

appilon: looking further down maybe this error check is not wanted at all? seems like `ghErr` is handled below

github/transport.go
appilon: I usually always need a refresher when it comes to golangs `io` package (my mental model always thinks things are backwards). I have a feeling what you've written is exactly what you needed, but would `io.TeeReader` be of any use? passing the first reader to `checkResponse` which will fill the second reader with the content? My apologies if there is some gotcha you already explored.

github/transport.go
radeksimko: Good spot! 🙈

github/transport.go
radeksimko: I don't see how `TeeReader` would be handy without more re-buffering/hacking. 🤔 
It seems to do the same thing, but it requires `io.Writer` as source and `io.ReadCloser` coming from the response body doesn't comply with that interface.

Maybe I'm missing something obvious?

github/transport.go
```diff
@@ -29,3 +38,113 @@ func (ett *etagTransport) RoundTrip(req *http.Request) (*http.Response, error) {
...
+
+	// Make response body accessible for retries & debugging
+	// (work around bug in GitHub SDK)
+	// See https://github.com/google/go-github/pull/986
+	r1, r2, err := drainBody(resp.Body)
+	if err != nil {
+		return nil, err
+	}
+	resp.Body = r1
+	ghErr := github.CheckResponse(resp)
```
appilon: I'm seeing now that `github.CheckResponse` accepts the `response` object, making the use of `TeeReader` impossible. The first time I read it I thought it accepted just a `Reader`, in that case `TeeReader` would have worked nicely
```
var r2 bytes.Buffer
r1 := io.TeeReader(resp.Body, &r2)
ghErr := github.CheckResponse(r1)
resp.Body = &r2
```
Also the interfaces might not have aligned, if `resp.Body` was `ReadCloser` you would have to wrap it just like you did in `drainBody` anyways nevermind me... always looking for an excuse to use `io` builtins 😛

PR #150, opened 2018-09-07
resource/github_repository: Allow setting `default_branch` to `master` on creation

### LINKED ISSUE #146: Cannot set the default branch on a new repository
Right now, an [error is thrown](https://github.com/terraform-providers/terraform-provider-github/blob/master/github/resource_github_repository.go#L161) if a default branch is specified when a respository is being created, even if the `auto_init` value is set to true. We should allow setting the default branch if `auto_init` is true since initializing the repo will necessarily result in the creation of a branch anyway.

ctrom:
Fixes #146.


ctrom:
@radeksimko could I get a review on this?

ctrom:
@radeksimko 

- It isn't possible to set a default branch when creating a repo, but regardless of the value that is passed for the default branch, GitHub will set the default branch to `master`. As a result, I don't think it makes a difference whether we do this on `create` or `update`. Really what I want to achieve here is that since GitHub sets the default branch to `master` when a repo is created, I want the Terraform state to reflect that.
- Perhaps I'm missing it, but I don't see an acceptance test for repo creation. I can attempt to create a new acceptance test for this, but I don't have a lot of experience with golang yet, so please be patient with me while I do.

ewoutp:
This PR would be a big help.
I'm using `github_repository` in a module and I right now cannot set the `default_branch` argument at all.
I like to set it to something like `"${var.default_branch}"`.

This PR would enable me to do so.

Then in the place where I use the module, I could start with `default_branch = "master"` and later change that to `default_branch = "something-else"`

ctrom:
@ewoutp Your situation sounds just like mine. Unfortunately, I don't know enough about golang to finish this quickly and can't take the time right now to learn it. Maybe someone else would be able to take a stab at it?

PR #161, opened 2018-10-18
New Resource: github_user_invitation_accepter

radeksimko:
### Acceptance tests
```
=== RUN   TestAccGithubRepositoryCollaborator_basic
--- PASS: TestAccGithubRepositoryCollaborator_basic (10.16s)
=== RUN   TestAccGithubRepositoryCollaborator_importBasic
--- PASS: TestAccGithubRepositoryCollaborator_importBasic (9.80s)
=== RUN   TestAccGithubUserInvitationAccepter_basic
--- PASS: TestAccGithubUserInvitationAccepter_basic (11.07s)
```

### DIFF
--- a/github/resource_github_repository_collaborator.go
+++ b/github/resource_github_repository_collaborator.go
@@ -2,6 +2,7 @@ package github
 
 import (
 	"context"
+	"fmt"
 	"log"
 
 	"github.com/google/go-github/github"
@@ -35,6 +36,10 @@ func resourceGithubRepositoryCollaborator() *schema.Resource {
 				ForceNew:     true,
 				Default:      "push",
 				ValidateFunc: validateValueFunc([]string{"pull", "push", "admin"}),
+			},
+			"invitation_id": {
+				Type:     schema.TypeString,
+				Computed: true,
 			},
 		},
 	}
@@ -90,6 +95,7 @@ func resourceGithubRepositoryCollaboratorRead(d *schema.ResourceData, meta inter
 		d.Set("repository", repoName)
 		d.Set("username", username)
 		d.Set("permission", permissionName)
+		d.Set("invitation_id", fmt.Sprintf("%d", invitation.GetID()))
 		return nil
 	}
 
--- a/github/resource_github_user_invitation_accepter.go
+++ b/github/resource_github_user_invitation_accepter.go
@@ -1,0 +1,47 @@
+package github
+
+import (
+	"context"
+	"fmt"
+	"log"
+	"strconv"
+
+	"github.com/hashicorp/terraform/helper/schema"
+)
+
+func resourceGithubUserInvitationAccepter() *schema.Resource {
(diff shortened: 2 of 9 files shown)


github/resource_github_user_invitation_accepter_test.go
```diff
@@ -0,0 +1,74 @@
...
+	"github.com/hashicorp/terraform/helper/resource"
+	"github.com/hashicorp/terraform/helper/schema"
+	"github.com/hashicorp/terraform/terraform"
+)
+
+func TestAccGithubUserInvitationAccepter_basic(t *testing.T) {
+	resourceName := "github_repository_collaborator.test"
+	repoName := fmt.Sprintf("tf-acc-test-collab-%s", acctest.RandString(5))
+
+	inviteeToken := os.Getenv("GITHUB_TEST_COLLABORATOR_TOKEN")
```
bflad: Nit: Should we name this `GITHUB_ALTERNATE_ACCOUNT_TOKEN` or something so its reusable in other contexts if necessary?

github/resource_github_user_invitation_accepter.go
```diff
@@ -0,0 +1,57 @@
...
+	"strconv"
+
+	"github.com/hashicorp/terraform/helper/schema"
+)
+
+func resourceGithubUserInvitationAccepter() *schema.Resource {
+	return &schema.Resource{
+		Create: resourceGithubUserInvitationAccepterCreate,
+		Read:   resourceGithubUserInvitationAccepterRead,
+		Delete: resourceGithubUserInvitationAccepterDelete,
```
bflad: Nit: Should we use [`schema.Noop`](https://godoc.org/github.com/hashicorp/terraform/helper/schema#Noop) instead of making an empty, custom function?

github/resource_github_repository_collaborator.go
```diff
@@ -36,6 +37,10 @@ func resourceGithubRepositoryCollaborator() *schema.Resource {
 				Default:      "push",
 				ValidateFunc: validateValueFunc([]string{"pull", "push", "admin"}),
 			},
+			"invitation_id": {
```
bflad: This attribute needs documentation. 😄

github/resource_github_user_invitation_accepter_test.go
radeksimko: Well, the token is always associated with a user and what that user has (or doesn't have) access to, so I feel that the ENV variable name should hint that the token belongs to `GITHUB_TEST_COLLABORATOR`, which we (maybe confusingly?) expect to _not_ to be a member of the org, so we can test collaborator invitations.

github/resource_github_user_invitation_accepter.go
radeksimko: TIL `schema.Noop` is a thing.

PR #163, opened 2018-10-22
Do not error when deleting a repository that does not exist

quantumew:
We had to manually amend state to remove a repository that was manually deleted. That seems wrong. If it 404s just remove it from terraform state. (let me know if you want an ACC test for this scenario)

https://postimg.cc/2LxR6WGF

### DIFF
--- a/github/resource_github_team_repository.go
+++ b/github/resource_github_team_repository.go
@@ -177,7 +177,14 @@ func resourceGithubTeamRepositoryDelete(d *schema.ResourceData, meta interface{}
 
 	log.Printf("[DEBUG] Deleting team repository association: %s (%s/%s)",
 		teamIdString, orgName, repoName)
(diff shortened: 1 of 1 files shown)


radeksimko:
Hi @quantumew 
we already have a guard for this in place on line 108:

https://github.com/terraform-providers/terraform-provider-github/pull/163/files#diff-4a7bb1b9cf90d0441333b134c34b5aa6R108

and we assume (not only in this provider, but in all other official providers) that Terraform is going to refresh prior to deleting, which it does by default (unless you pass `-refresh=false`), hence we prefer to keep this logic in `Read()`.

Do you mind double checking you weren't passing `-refresh=false` to `apply` or `destroy` command?

Thanks.

quantumew:
Ah gotcha, I missed that. I saw the error and made and assumption. I’ll look into it the next time it comes up. Thanks!

PR #174, opened 2018-12-20
Updating label name no longer creates new label.

### LINKED ISSUE #173: Renaming labels creates new labels.
Terraform version: v0.11.8
Provider version: v1.3.0

# Problem

It seems like when changing the name property on a label the provider creates a new label with the new name and leaves the old one intact. 

# Cause

The reason for this is due to the implementation in https://github.com/terraform-providers/terraform-provider-github/blob/master/github/resource_github_issue_label.go.

In `resourceGithubIssueLabelCreateOrUpdate` the following flow is used:

Check if `newName` is a label, if it is update it and if it isn't create a new one. The problem here is that the original name is not the one that is checked. So if a label is changed in terraform from `label1` and `label2`, github is searched for `label2`, which does not exist yet so a new label is created and the old `label1` is left behind.

# Solution

When checking for existence, the original name should be used.

eliasbrange:
Fixes #173

### DIFF
--- a/github/resource_github_issue_label.go
+++ b/github/resource_github_issue_label.go
@@ -75,10 +75,23 @@ func resourceGithubIssueLabelCreateOrUpdate(d *schema.ResourceData, meta interfa
 		ctx = context.WithValue(ctx, ctxId, d.Id())
 	}
 
+	// Pull out the original name. If we already have a resource, this is the
+	// parsed ID. If not, it's the value given to the resource.
+	var originalName string
+	if d.Id() == "" {
+		originalName = name
+	} else {
+		var err error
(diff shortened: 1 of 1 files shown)


paultyng:
This should probably have an acceptance test to simulate the problem its fixing to avoid regression.

eliasbrange:
Probably a good idea yes. Will see if I can find the time someday.

anGie44:
hi @eliasbrange, thanks again for this PR :) Looking through changes in the provider, I see the functionality is available given the changes in #288 - released in v2.4.1. I'm going to close this PR but thanks again for the contribution and if there are any future questions please let us know!

PR #207, opened 2019-04-03
Update go-github to v25

bramwelt:
This updates the go-github module used by the Terraform Github provider to v25.
v22-23 are skipped due to a bug introduced and later fixed related to webhooks.

PR #212, opened 2019-04-23
Provide "require signed commits" on `github_branch_protection`

bltavares:
This commit provides changing the toggle to require signed commits on
    branch protection.

The [API for signed commits](https://developer.github.com/v3/repos/branches/#get-required-signatures-of-protected-branch)
    is separate endpoint from the branch protection url, and does not come on the same payload.
    This means that we need to make an extra request to read and modify it.

This changes requires `go-github` at least v19.

### Needs:
   - [ ] Update `go-github` version: https://github.com/terraform-providers/terraform-provider-github/pull/207
   - [ ] Provide default required reviewers count to avoid GitHub update error: https://github.com/terraform-providers/terraform-provider-github/pull/211

Closes https://github.com/terraform-providers/terraform-provider-github/issues/87


bltavares:
I'll gladly rebase or pick the last commit apart if necessary, but updating the version of `go-github` is important for it. After #207 lands, I'll gladly update the PR.

Note: precompiled plugin to validate the changes [here](https://github.com/bltavares/terraform-provider-github/releases/tag/v1.3.0-wip)

PR #214, opened 2019-04-24
Provide "require signed commits" on `github_branch_protection`

bltavares:
This commit provides changing the toggle to require signed commits on branch protection.

The [API for signed commits](https://developer.github.com/v3/repos/branches/#get-required-signatures-of-protected-branch) is separate endpoint from the branch protection url, and does not come on the same payload. This means that we need to make an extra request to read and modify it.

Closes https://github.com/terraform-providers/terraform-provider-github/issues/87

### DIFF
--- a/github/resource_github_branch_protection.go
+++ b/github/resource_github_branch_protection.go
@@ -130,6 +130,11 @@ func resourceGithubBranchProtection() *schema.Resource {
 				Optional: true,
 				Default:  false,
 			},
+			"require_signed_commits": {
+				Type:     schema.TypeBool,
+				Optional: true,
+				Default:  false,
+			},
 			"etag": {
 				Type:     schema.TypeString,
 				Computed: true,
@@ -165,6 +170,10 @@ func resourceGithubBranchProtectionCreate(d *schema.ResourceData, meta interface
 
 	d.SetId(buildTwoPartID(&repoName, &branch))
 
+	if err = requireSignedCommitsUpdate(d, meta); err != nil {
+		return err
+	}
+
 	return resourceGithubBranchProtectionRead(d, meta)
 }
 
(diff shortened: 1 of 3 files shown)


bltavares:
@paultyng I've updated this PR to use the latest changes recently landed on `master`. It would be really nice to have this landed so we can upgrade to the new release of the provider and drop the fork.

paultyng:
@bltavares would it be possible to add acceptance test coverage and docs for this?

bltavares:
I'm not used to acceptance tests on Terraform. Should I include it on the current branch protection test, or create a new one?

paultyng:
@bltavares adding to the existing one is probably fine if you don't think there is a reason to have an additional test. I'm not sure what the API interactions look like and if it would be useful to have tests turning it on and off, etc.

bltavares:
@paultyng I've changed the test to create the signed commit requirement, then remove it, as well as documentation.

Let me know if they are good enough.

tobypinder:
Hey folks, any update on this? From my perspective, automating the branch protection rules is the most fiddly thing to standardise in the UI and we're very keen to add the signed_commit stuff to our internal state management as well :+1:

tracypholmes:
@bltavares @tobypinder we're reviewing it this week. Thank you for being patient with us!

maelvls:
Weird, as of today the [official docs](https://www.terraform.io/docs/providers/github/r/repository.html) regarding the resource `github_repository ` do not mention `signed_commit` although it has been released 2 weeks ago in v2.3.2.

jcudit:
@maelvls it looks like the documentation updated [landed on a different resource](https://www.terraform.io/docs/providers/github/r/branch_protection.html#require_signed_commits).  Mention me on any PRs that would fix the documentation flow for you in this use case.

PR #181, opened 2019-01-30
Add github_organization_block resource

paultyng:
Fixes #179, #185, #89 
Supersedes #215, #211, #207, #186, #160

PR #215, opened 2019-04-24
[v0.12-beta2] Bump go-github version v24

bltavares:
After bumping to go-github v24, a couple of signatures have changed.

One of them, the `name` on `webhooks` has been removed. Previously, the
only valid name was `web`, and now, it has been removed from the object.

This PR is a breaking change on the provider, as it will not accept
more taking `name` on the webhook resources.


bltavares:
@apparentlymart It would also be nice to land this change before doing the compatibility release. The previous `go get @24` on https://github.com/terraform-providers/terraform-provider-github/commit/525ca4b6708de9ffc00b39414686406d1f57de0a dind't migrate upgrade to `v24`, it only added the dependency as a transitive dependency.

Sorry to keep bothering and thanks a lot for the help recently. Let me know who would be better fit to ping on such issues as well.

paultyng:
@bltavares working on these, as well as #181 and adding 0.12 support. Hopefully should have this ready to go soon, really appreciate your work here!

paultyng:
`2.0.0` has been released, including an SDK bump to v25

PR #221, opened 2019-05-07
Add instructions for getting the acceptance tests passing with any test org

ryndaniels:
The acceptance tests used to be hardcoded to a specific `terraformtesting` organization. This changes that to a variable and adds instructions for people to set up and run the acceptance tests with their own organization.

### DIFF
--- a/README.md
+++ b/README.md
@@ -34,7 +34,8 @@ $ gnumake build
 
 Using the provider
 ----------------------
-## Fill in for each provider
+
+Detailed documentation for the GitHub provider can be found [here](https://www.terraform.io/docs/providers/github/index.html).
 
 Developing the Provider
 ---------------------------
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -63,3 +64,38 @@ In order to run the full suite of Acceptance tests, run `make testacc`.
...
+* delete_repo
+* admin:gpg_key
+
+Once the token has been created, it must be exported in your environment as `GITHUB_TOKEN`.
+
+### GitHub organization
+If you do not have an organization already that you are comfortable running tests again, you will need to [create one](https://help.github.com/en/articles/creating-a-new-organization-from-scratch). The free "Team for Open Source" org type is fine for these tests. The name of the
+organization must then be exported in your environment as `GITHUB_ORGANIZATION`.
+
+### Test repository
```
bflad: Definitely not for this pull request, but we should fix the testing that requires this existing repository to instead automatically create it so this step is not required. 👍

Seems like its related to the `github_repository` data source testing and could be fixed via something like:

```go
func testAccCheckGithubRepositoryDataSourceConfig_fullName(randomName string) string {
	return fmt.Sprintf(`
  name = %[1]q
  # ... appropriate configuration ...
}

data "github_repository" "test" {
  full_name = "${github_repository.test.full_name}"
}
`, randomName)
}

func testAccCheckGithubRepositoryDataSourceConfig_name(randomName string) string {
	return fmt.Sprintf(`
  name = %[1]q
  # ... appropriate configuration ...
}

data "github_repository" "test" {
  name = "${github_repository.test.name}"
}
`, randomName)
}
```

PR #224, opened 2019-05-08
Add base_url clarification for GitHub enterprise

SpamapS:
If /api/v3 is left off of this URL, users will get confusing responses such as 406 Not Acceptable.


megan07:
Hey @SpamapS !  Thanks for opening this PR! It looks good and should make it into the next release!

PR #234, opened 2019-05-31
 data_source/github_repositories: add sort capabilities 

### LINKED ISSUE #206: There is no option to sort repositories in github_repositories data source
Hi there,

### Terraform Version
Terraform v0.11.11

### Affected Resource(s)

- data.github_repositories

```hcl
data "github_repositories" "example" {
  query = "org:hashicorp language:Go"
}
```
### Expected Behavior
I expect the list to be sorted consistently, for example if I add another repository to the organisation.

### Actual Behavior

Possibly the new repos get added to the beginning of the list of repositories.

### Steps to Reproduce

1. `terraform apply`
2. Add a new repository
3. `terraform apply`

### Important Factoids

In the underlying library there are sort options: https://github.com/google/go-github/blob/master/github/search.go#L40
They are not used in the provider: https://github.com/terraform-providers/terraform-provider-github/blob/master/github/data_source_github_repositories.go#L60

tracypholmes:
Fixes #206 and added ability to sort

### DIFF
--- a/github/data_source_github_repositories.go
+++ b/github/data_source_github_repositories.go
@@ -2,6 +2,7 @@ package github
 
 import (
 	"context"
+	"github.com/hashicorp/terraform/helper/validation"
 	"log"
 
 	"github.com/google/go-github/v25/github"
@@ -17,15 +18,21 @@ func dataSourceGithubRepositories() *schema.Resource {
 				Type:     schema.TypeString,
 				Required: true,
 			},
+			"sort": {
+				Type:         schema.TypeString,
+				Default:      "updated",
+				Optional:     true,
+				ValidateFunc: validation.StringInSlice([]string{"stars", "fork", "updated"}, false),
+			},
 			"full_names": {
-				Type: schema.TypeSet,
+				Type: schema.TypeList,
 				Elem: &schema.Schema{
 					Type: schema.TypeString,
 				},
 				Computed: true,
 			},
 			"names": {
-				Type: schema.TypeSet,
+				Type: schema.TypeList,
 				Elem: &schema.Schema{
 					Type: schema.TypeString,
 				},
@@ -39,9 +46,15 @@ func dataSourceGithubRepositoriesRead(d *schema.ResourceData, meta interface{}) 
 	client := meta.(*Organization).client
 
 	query := d.Get("query").(string)
+	opt := &github.SearchOptions{
+		Sort: d.Get("sort").(string),
+		ListOptions: github.ListOptions{
+			PerPage: 100,
+		},
+	}
 
 	log.Printf("[DEBUG] Searching for GitHub repositories: %q", query)
-	fullNames, names, err := searchGithubRepositories(client, query)
+	fullNames, names, err := searchGithubRepositories(client, query, opt)
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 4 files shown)


github/data_source_github_repositories.go
```diff
@@ -2,6 +2,7 @@ package github
 
 import (
 	"context"
+	"github.com/hashicorp/terraform/helper/validation"
```
katbyte: I would run goimports on this file (sorts the imports)

website/docs/d/repositories.html.markdown
```diff
@@ -27,6 +27,8 @@ The following arguments are supported:
 
 * `query` - (Required) Search query. See [documentation for the search syntax](https://help.github.com/articles/understanding-the-search-syntax/).
 
+* `sort` - (Optional) Sorts query. Valid values include `stars`, `fork`, and `updated`. Defaults to `updated`.
```
katbyte: Would this sound better as: 
```suggestion
* `sort` - (Optional) Sorts the repositories returned by the specified attribute. Valid values include `stars`, `fork`, and `updated`. Defaults to `updated`.
```

github/data_source_github_repositories.go
```diff
@@ -39,9 +46,15 @@ func dataSourceGithubRepositoriesRead(d *schema.ResourceData, meta interface{})
 	client := meta.(*Organization).client
 
 	query := d.Get("query").(string)
+	opt := &github.SearchOptions{
```
katbyte: Another option here would have been to pass the sort option into `searchGithubRepositories(client, query, sort)` keeping all the github specific search logic and types together while keeping this function a bit cleaner. WDYT?

github/data_source_github_repositories_test.go
```diff
@@ -50,7 +76,16 @@ func TestAccGithubRepositoriesDataSource_noMatch(t *testing.T) {
...
+	query = %q
 }
 `, query)
 }
+
+func testAccCheckGithubRepositoriesDataSourceConfigWithSort(query, sort string) string {
+	return fmt.Sprintf(`
+data "github_repositories" "test" {
+	query = %q
+	sort = %q
```
katbyte:
```suggestion
	sort  = %q
```

github/data_source_github_repositories_test.go
```diff
@@ -50,7 +76,16 @@ func TestAccGithubRepositoriesDataSource_noMatch(t *testing.T) {
 func testAccCheckGithubRepositoriesDataSourceConfig(query string) string {
 	return fmt.Sprintf(`
 data "github_repositories" "test" {
-	query = "%s"
+	query = %q
```
katbyte: Should we be removing the ""s from around string values? 🤔 While that may work in .12 some people may use the tests as an example and still be running .11. Would it be better to use the syntax that has the most compatibility until we officially retire .11?

katbyte: I just realized %q quoted 😅 It might make make more sense to leave this as "%s" to be explicit that there are quotes there.

website/docs/d/repositories.html.markdown
tracypholmes: Agreed. Was having a time with the wording on that.

github/data_source_github_repositories.go
tracypholmes: That definitely makes sense. Did it this way because it was easier for me to read. I also felt like if someone needed to add additional parameters/values, they could just add it to the function.

github/data_source_github_repositories_test.go
tracypholmes: I originally left it like that. But found over in AWS land, we are trying to be more consistent, so we use `%q` since that's what it does.

Issue #196, opened 2019-02-28
GitHub usernames are case-insensitive and should be treated that way


dgmorales:
GitHub usernames are case-insensitive but case preserving. When searching in the UI and the API, the case does not matter, but the user found is returned with the proper username capitalisation defined by the user. The terraform GitHub provider needs to be aware of that.

When adding a user to an org with the github_membership resource, it will work with whatever casing you specify. BUT, when running again terraform will detect the username difference (dgMorales != dgmorales, for example) and report a change, and try to apply. When applied it will not thrown an error, but it will keep reporting the change again an again ...

### Terraform Version
- Terraform v0.11.11
- provider.github: version = "~> 1.3"

### Affected Resource(s)
- github_membership
- github_team_membership
- both above confirmed, but possibly also any other the accepts an username string as input (github_branch_protection, for example)

### Terraform Configuration Files

See below.

### Debug Output
To make a sensible debug output and configuration example, I need to refactor my code to portrait only this problem. I can do this in a few days, but I'm not sure that is really necessary or helpful in this case. 

Below here I've placed an adapted excerpt from the regular output of my `terraform plan`. My GitHub username is *dgmorales* and suppose I specified **dgMorales** in my terraform code (here I'm wrapping the resources in a user module).

```
-/+ module.dgmorales_user.github_membership.org-member (new resource required)
      id:       "my-org-name:dgmorales" => <computed> (forces new resource)
      etag:     "W/\"<hex-numbers-string>\"" => <computed>
      role:     "member" => "member"
      username: "dgmorales" => "dgMorales" (forces new resource)

-/+ module.dgmorales_user.github_team_membership.all-org-members (new resource required)
      id:       "<team-id-number>:dgMorales" => <computed> (forces new resource)
      etag:     "W/\"<hex-numbers-string>\"" => <computed>
      role:     "member" => "member"
      team_id:  "<team-id-number>" => "<same-team-id-number>"
      username: "dgmorales" => "dgMorales" (forces new resource)
```

That shows both with -refresh=true or -refresh=false (the state is saved with the correct username case). It keeps happening after running apply.

### Expected Behavior

I believe the provider should either:

1. Check the username is in a different case, and throw an error (despite having found the correct user)
2. Ignore differences in case only and not report a change for that.

I would vote on alternative 2.

### Actual Behavior

Plan keeps reporting a change every time. 

### Steps to Reproduce

[...]

PR #241, opened 2019-06-14
Bug Fix: GitHub Usernames are Case Insensitive

megan07:
Addressing #196 where GitHub usernames were being treated as case-sensitive in the diff.
Before fix:
```
Step 1 error: After applying this step, the plan was not empty:

        DIFF:

        DESTROY/CREATE: github_team_membership.test_team_membership
          etag:     "W/\"52d84b121c000efbcf03e7e877c4fbf7\"" => "<computed>"
          id:       "3289152:Mbang-github2" => "<computed>"
          role:     "member" => "member"
          team_id:  "3289152" => "3289152"
          username: "mbang-github2" => "Mbang-github2" (forces new resource)
```
```
Step 1 error: After applying this step, the plan was not empty:

        DIFF:

        DESTROY/CREATE: github_membership.test_org_membership
          etag:     "W/\"ef0b40d063d184967e965b7d8af83466\"" => "<computed>"
          id:       "mbang-github-acc-test:mbang-github2" => "<computed>"
          role:     "member" => "member"
          username: "mbang-github2" => "Mbang-github2" (forces new resource)
```
```

        When applying changes to
        github_repository_collaborator.test_repo_collaborator, provider "github"
        produced an unexpected new value for was present, but now absent.

        issue tracker.

FAIL
```

After fix:
```
=== RUN   TestAccGithubTeamMembership_basic
--- PASS: TestAccGithubTeamMembership_basic (13.25s)
PASS
```
```
=== RUN   TestAccGithubMembership_basic
--- PASS: TestAccGithubMembership_basic (4.08s)
PASS
```
```
=== RUN   TestAccGithubRepositoryCollaborator_basic
--- PASS: TestAccGithubRepositoryCollaborator_basic (10.31s)
PASS
```

I also tested that if the user DOES update the casing in the GitHub UI, a terraform refresh will pick that up.


github/resource_github_membership_test.go
```diff
@@ -29,6 +30,27 @@ func TestAccGithubMembership_basic(t *testing.T) {
 	})
 }
 
+func TestAccGithubMembership_caseInsensitive(t *testing.T) {
+	if len(testCollaborator) == 0 {
```
nywilken: Nit when checking for empty strings its best to explicitly check for an empty string so that future readers know the variable is a string and not some type that implements the length method. 

```suggestion
	if testCollaborator == "" {
```

github/resource_github_membership_test.go
```diff
@@ -140,3 +162,18 @@ var testAccGithubMembershipConfig string = fmt.Sprintf(`
     role = "member"
   }
 `, testCollaborator)
+
+func testAccGithubMembershipConfig_caseInsensitive() string {
+	otherCase := []rune(testCollaborator)
```
nywilken: If I get the intent right, you are modifying the collaborator username, by toggling the case of the first rune, to test case insensitivity functionality. Is that correct? If so what about changing the function signature to take in an input, and just pass the collaborator name with the expected case. 

```go
func testAccGithubMembershipConfig_caseInsensitive(memberName string) {
return fmt.Sprintf(`
  resource "github_membership" "test_org_membership" {
    username = "%s"
    role = "member"
  }
`, memberName)
}
```

This way in the acceptance test you can run two TestSteps:

```go
Steps: []resource.TestStep{
			{
				Config: testAccGithubMembershipConfig_caseInsensitive(strings.ToUpper(testCollaborator)),
				Check: resource.ComposeTestCheckFunc(
					testAccCheckGithubMembershipExists("github_membership.test_org_membership", &membership),
				),
			},
			{
				Config: testAccGithubMembershipConfig_caseInsensitive(strings.ToLower(testCollaborator)),
				Check: resource.ComposeTestCheckFunc(
					testAccCheckGithubMembershipExists("github_membership.test_org_membership", &membership),
				),
			},
		},
```

github/resource_github_repository_collaborator.go
```diff
@@ -112,14 +114,14 @@ func resourceGithubRepositoryCollaboratorRead(d *schema.ResourceData, meta inter
...
-			if *c.Login == username {
+			if strings.ToLower(*c.Login) == strings.ToLower(username) {
 				permissionName, err := getRepoPermission(c.Permissions)
 				if err != nil {
 					return err
 				}
 
 				d.Set("repository", repoName)
-				d.Set("username", username)
+				d.Set("username", *c.Login)
```
nywilken: You don't need to deference the pointer here, as the `d.Set` will check that it is a pointer and get the value if the pointer is not nil.

nywilken (commented):
@megan07 this looks good so far. I took a quick pass and left a couple of comments. I'll circle back later today.

nywilken (commented):
@megan07 I left a few comments for you, but this is otherwise good. 

```
--- PASS: TestAccGithubMembership_basic (3.30s)
--- PASS: TestAccGithubMembership_caseInsensitive (2.84s)
--- PASS: TestAccGithubMembership_importBasic (2.79s)
--- PASS: TestAccGithubRepositoryCollaborator_basic (9.65s)
--- PASS: TestAccGithubRepositoryCollaborator_importBasic (9.17s)
```

github/resource_github_membership_test.go
```diff
@@ -2,29 +2,52 @@ package github
...
+	"unicode"
 
 	"github.com/google/go-github/v25/github"
 	"github.com/hashicorp/terraform/helper/resource"
 	"github.com/hashicorp/terraform/terraform"
 )
 
 func TestAccGithubMembership_basic(t *testing.T) {
+	if testCollaborator == "" {
+		t.Skip("Skipping because length of `GITHUB_TEST_COLLABORATOR` is not set")
```
paultyng: think this is a copy/paste typo for the message

github/resource_github_membership_test.go
```diff
@@ -134,9 +157,21 @@ func testAccCheckGithubMembershipRoleState(n string, membership *github.Membersh
...
     username = "%s"
     role = "member"
   }
-`, testCollaborator)
+`, username)
+}
+
+func testAccGithubMembershipTheSame(orig, other *github.Membership) resource.TestCheckFunc {
+	return func(s *terraform.State) error {
+		if *orig.URL != *other.URL {
```
paultyng: should this also test to make sure at least one isn't null/empty? or should two empties be equal?

github/resource_github_membership_test.go
```diff
@@ -2,29 +2,52 @@ package github
...
 func TestAccGithubMembership_basic(t *testing.T) {
+	if testCollaborator == "" {
+		t.Skip("Skipping because length of `GITHUB_TEST_COLLABORATOR` is not set")
+	}
+
 	var membership github.Membership
+	var otherMembership github.Membership
+
+	oc := []rune(testCollaborator)
+	if unicode.IsUpper(oc[0]) {
```
paultyng: not sure if its possible, but could `0` ever be a number or symbol in a username?

paultyng (approved):
Some minor comments, but otherwise LGTM

github/resource_github_membership_test.go
megan07: I think, for the intent of the test, two empties would be equal.  I can't think of a time when this would be empty here, but if they were to both be empty, I think it means the case change didn't have an effect.

paultyng (approved):

kjolley-zainar:
I'm still seeing this:

```
% tf --version
Terraform v1.6.6
on darwin_amd64

% tf plan
...
# module.github.github_repository_collaborators.myenv will be updated in-place
  ~ resource "github_repository_collaborators" "myenv" {
        id             = "myenv"
      ~ invitation_ids = {} -> (known after apply)

      - user {
          - permission = "pull" -> null
          - username   = "foo" -> null
        }
      - user {
          - permission = "pull" -> null
          - username   = "bar" -> null
        }
      + user {
          + permission = "pull"
          + username   = "Foo"
        }
      + user {
          + permission = "pull"
          + username   = "Bar"
        }

        # (4 unchanged blocks hidden)
    }
```

esacteksab:
I'm still seeing this as well. Not sure if there is a regression. An example (a bit redacted).

```terraform
  ~ update in-place

  # github_team_members.drupal_fe will be updated in-place
  ~ resource "github_team_members" "drupal_fe" {

      - members {
          - role     = "maintainer" -> null
          - username = "username" -> null
        }
      + members {
          + role     = "maintainer"
          + username = "UserName"
        }

        # (3 unchanged blocks hidden)
    }

  # github_team_members.org_admin will be updated in-place
  ~ resource "github_team_members" "org_admin" {

      - members {
          - role     = "maintainer" -> null
          - username = "username" -> null
        }
      + members {
          + role     = "maintainer"
          + username = "UserName"
        }

        # (4 unchanged blocks hidden)
    }

  # github_team_members.org_devops will be updated in-place
  ~ resource "github_team_members" "org_devops" {

      - members {
          - role     = "maintainer" -> null
          - username = "username" -> null
        }
      + members {
          + role     = "maintainer"
          + username = "UserName"
        }

        # (1 unchanged block hidden)
    }

  # github_team_members.org_eng will be updated in-place
  ~ resource "github_team_members" "org_eng" {

      - members {
          - role     = "maintainer" -> null
          - username = "username" -> null
        }
      + members {
          + role     = "maintainer"
          + username = "UserName"
        }

        # (6 unchanged blocks hidden)
    }

Plan: 0 to add, 4 to change, 0 to destroy.
```

Issue #171, opened 2018-11-29
Webhook secret set via Terraform makes signatures not match


Tenzer:
### Terraform Version
Terraform v0.11.10
+ provider.archive v1.1.0
+ provider.aws v1.49.0
+ provider.github v1.3.0

### Affected Resource(s)
- github_organization_webhook
- github_repository_webhook (assumed, not tested)

### Terraform Configuration Files
```hcl
resource "github_organization_webhook" "webhook" {
  name   = "web"
  events = ["*"]

  configuration {
    url          = "${aws_api_gateway_deployment.github_notifications.invoke_url}"
    content_type = "json"
    secret       = "${local.github_secret}"
    insecure_ssl = false
  }
}
```

### Expected Behavior
I should be able to verify the body of the incoming webhook requests against the signature in the header, using the secret specified in the GitHub webhook.

### Actual Behavior
The signatures never match.

---

I first looked at a bunch of different implementations of GitHub webhook receivers to check if my implementation of the HMAC signature verification was wrong in some way but couldn't find any problems with it. I then tried to set the secret for the webhook directly in GitHub's web UI and then all signatures matched after that.
I tried to make some change to the hook with Terraform and could confirm that the signature then no longer would match again until I set the secret via the GitHub web UI again.

This is a bit tricky to debug further, as I can't see the secret that is being sent to GitHub as part of the TRACE debug output and I can't see the secret in GitHub either, but my only conclusion so far is that Terraform sets the secret to something else than the string I tell it to.

A curiosity is that the secret never is updated in the GitHub hook even though I change it in the configuration - at least from what the output of Terraform show - I don't know if this could be related, that the secret is set to something invalid whenever another change is made to the webhook?

sryabkov:
I think I am experiencing the same issue.  I initially set the repo webhook manually using GitHub UI, and the webhook was firing correctly. I then tried doing it through TF using the same value. `terraform plan` reported that it wanted to update the value. I ran `terraform apply`, after which the webhook started failing. I had to manually reset the secret value using GitHub UI to fix the webhook. 

It does look like the value of `secret` for a repo webhook is not set correctly when using the GitHub TF provider.

niedbalski:
Experiencing the same exact bug. +1

niedbalski:
Well, this is reproducible with 0.11.13 and current master. I think the best approach would be for the config schema to be forcenew = true and remove the diffsupress entirely.

johne:
[...]

PR #251, opened 2019-07-02
Bug Fix: Webhook secret stored incorrectly in state

megan07:
Addressing #171 where the webhook secret set in Terraform makes signatures not match.
Before fix:
```
--- FAIL: TestAccGithubRepositoryWebhook_importSecret (9.38s)
    testing.go:568: Step 1 error: ImportStateVerify attributes not equivalent. Difference is shown below. Top is actual, bottom is expected.

        (map[string]string) (len=1) {
         (string) (len=22) "configuration.0.secret": (string) (len=8) "********"
        }


        (map[string]string) (len=1) {
         (string) (len=22) "configuration.0.secret": (string) (len=18) "RandomSecretString"
        }
```
After fix:
```
=== RUN   TestAccGithubRepositoryWebhook_importSecret
=== PAUSE TestAccGithubRepositoryWebhook_importSecret
=== CONT  TestAccGithubRepositoryWebhook_importSecret
--- PASS: TestAccGithubRepositoryWebhook_importSecret (9.66s)
PASS
```

### DIFF
--- a/github/resource_github_organization_webhook.go
+++ b/github/resource_github_organization_webhook.go
@@ -92,6 +92,11 @@ func resourceGithubOrganizationWebhookCreate(d *schema.ResourceData, meta interf
 	}
 	d.SetId(strconv.FormatInt(*hook.ID, 10))
 
+	if hook.Config["secret"] != nil {
+		hook.Config["secret"] = webhookObj.Config["secret"]
+	}
+	d.Set("configuration", []interface{}{hook.Config})
+
 	return resourceGithubOrganizationWebhookRead(d, meta)
 }
 
@@ -134,6 +139,15 @@ func resourceGithubOrganizationWebhookRead(d *schema.ResourceData, meta interfac
 	d.Set("url", hook.URL)
 	d.Set("active", hook.Active)
 	d.Set("events", hook.Events)
+
+	if len(d.Get("configuration").([]interface{})) > 0 {
+		currentSecret := d.Get("configuration").([]interface{})[0].(map[string]interface{})["secret"]
+
+		if hook.Config["secret"] != nil {
+			hook.Config["secret"] = currentSecret
+		}
+	}
+
 	d.Set("configuration", []interface{}{hook.Config})
 
 	return nil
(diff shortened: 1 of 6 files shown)


github/resource_github_organization_webhook.go
```diff
@@ -92,6 +92,11 @@ func resourceGithubOrganizationWebhookCreate(d *schema.ResourceData, meta interf
 	}
 	d.SetId(strconv.FormatInt(*hook.ID, 10))
 
+	if hook.Config["secret"] != nil {
+		hook.Config["secret"] = webhookObj.Config["secret"]
```
paddycarver: I'm assuming the GitHub API doesn't return the secret, so we have to set it based on what's in the ResourceData? If so, can we add a comment saying as much, just to explain why we have to do this?

paddycarver: Also, what does `hook.Config` get set to, if not `nil`? That looks like the GitHub response, but what does it actually get set to?

github/resource_github_organization_webhook.go
```diff
@@ -134,6 +139,15 @@ func resourceGithubOrganizationWebhookRead(d *schema.ResourceData, meta interfac
 	d.Set("url", hook.URL)
 	d.Set("active", hook.Active)
 	d.Set("events", hook.Events)
+
+	if len(d.Get("configuration").([]interface{})) > 0 {
+		currentSecret := d.Get("configuration").([]interface{})[0].(map[string]interface{})["secret"]
+
+		if hook.Config["secret"] != nil {
+			hook.Config["secret"] = currentSecret
```
paddycarver: Likewise, a comment here just explaining the situation would be 💯

harlowja:
Ah, this explains why terraform screwed up all of our organizations hook secrets! Can we get this merged....

zhenkaiwang:
Hope this get reviewed and merged soon :rocket:

wenbin1989:
Hope this get merged ASAP. we have to apply the patch manually currently, which is not very nice.

Issue #230, opened 2019-05-22
Remove auth requirement for provider


julianvmodesto:
I'd like to consume only the [github_ip_ranges](https://www.terraform.io/docs/providers/github/d/ip_ranges.html) data source, but the provider requires setting an Organization and Token.

Would it be possible to remove this requirement?

majormoses:
duplicate of #190  and proposed as #191

tracypholmes:
Hi, @julianvmodesto! Thank you for opening this issue. This is being closed as a duplicate of #190.

Issue #205, opened 2019-03-27
Allow unauthenticated usage for public API endpoints


StephenKing:
Hi there,

### Terraform Version
Run `terraform -v` to show the version. If you are not running the latest version of Terraform, please upgrade because your issue may have already been fixed.

### Affected Resource(s)
Please list the resources as a list, for example:
- data.github_ip_ranges
- data.github_repositories
- data.github_* actually..

### Terraform Configuration Files
```hcl
provider "github" {
  organization = "" # required setting
  token = "" # required setting
}

data "github_ip_ranges" "this" {}

output "github_ip_ranges_hooks" {
  value = "${data.github_ip_ranges.this.hooks}"
}
```

### Expected Behavior

The API endpoint https://api.github.com/meta can be reached without authentication. It would be easier to use this data source (or the github data sources in general), if the provider wouldn't enforce that a token is given.

I know that rate limiting can kick in a lot earlier then. But for the use case of white-listing the IP addresses that GitHub uses to deliver hooks, our Terraform deployment would be easier and the number of calls sufficiently low.

### Actual Behavior

I'm asked to provide a Github token.

majormoses:
I think this is a duplicate of https://github.com/terraform-providers/terraform-provider-github/issues/190

megan07:
Thanks @StephenKing for opening this issue. It's been closed with #255 and will be released shortly.

PR #255, opened 2019-07-10
Kayrus optional parameters

megan07:
Addressing #230, #205, and #190 where GITHUB_TOKEN should not be required for all data sources.
Before fix:
```
        Error: GET https://api.github.com/users/hashibot: 401 Bad credentials []

          on /var/folders/nj/n5hr70hd2dx45qb__6wl10nc0000gp/T/tf-test353694711/main.tf line 6:
          (source code not available)


FAIL
```
After fix:
```
== RUN   TestProvider_anonymous
--- PASS: TestProvider_anonymous (1.04s)
PASS
```


github/provider.go
```diff
@@ -39,6 +39,12 @@ func Provider() terraform.ResourceProvider {
 				Default:     false,
 				Description: descriptions["individual"],
 			},
+			"anonymous": {
+				Type:        schema.TypeBool,
+				Optional:    true,
+				Default:     false,
```
paddycarver: This will default to falsey anyways, so we can avoid having this Default here.

github/provider.go
```diff
@@ -92,6 +99,8 @@ func init() {
 			"without verifying the TLS certificate.",
 
 		"individual": "Whether to run outside an organization.",
+
+		"anonymous": "Whether to authenticate without a token.",
```
paddycarver: My problem with "whether or not to do X" descriptions of booleans is the cognitive hurdle it takes to determine whether `true` or `false` leads to X actually happening. `Set to true to authenticate without a token` avoids this shortcoming, as an example.

github/provider_test.go
```diff
@@ -90,6 +102,59 @@ func TestProvider_individual(t *testing.T) {
 				Config:      individualProviderConfig + testAccGithubMembershipConfig(username),
 				ExpectError: regexp.MustCompile("This resource requires GitHub organization to be set on the provider."),
 			},
+			{
+				Config:      conflictingProviderConfig + testAccCheckGithubUserDataSourceConfig(username),
+				ExpectError: regexp.MustCompile("If `individual` is true, `organization` cannot be set."),
```
paddycarver: Does this test pass? `conflictingProviderConfig`, from what I can see, only sets `individual` to `true`, it doesn't set an organization.

github/provider_test.go
```diff
@@ -71,6 +71,18 @@ func TestProvider_individual(t *testing.T) {
...
 }
 `
+
+	conflictingProviderConfig := `provider "github" {
+	individual = true
+}
+`
+
+	neitherProviderConfig := `provider "github" {
+	organization = ""
```
paddycarver: I think the more common user scenario, and the one we should probably be testing for, is that they didn't specify organization at all, rather than setting it to an empty string.

megan07: @paddycarver I agree, however, the rest of the tests need GITHUB_ORGANIZATION defined, so, when running the suite it will already be there, this will override that.  Is there a better way to do this?

kayrus:
@megan07 when are you going to make the release with this feature?

Issue #260, opened 2019-07-21
github_user data source should support current authenticated user


mithun:
### Terraform Version

```
$ terraform version
Terraform v0.12.5
+ provider.github v2.2.0
```

### Affected Resource(s)

- `github_user` data source

### Terraform Configuration Files
```hcl
provider "github" {
  individual = true
  version    = "2.2.0"
}

data "github_user" "current" {
  username = ""
}

output "github_current_login" {
  value = data.github_user.current.login
}
```

### Expected Behavior

The current authenticated user's github login is printed as a terraform output.

### Actual Behavior

```
$ terraform apply
data.github_user.current: Refreshing state...

Error: GET https://api.github.com/user/gpg_keys: 404 Not Found []

  on main.tf line 6, in data "github_user" "current":
   6: data "github_user" "current" {
```

### Steps to Reproduce
Please list the steps required to reproduce the issue, for example:
1. `terraform apply`

### Important Factoids

`go-github/v27` allows using an [empty string for the username](https://github.com/google/go-github/blob/v27.0.0/github/users.go#L77) to fetch the current user's details. This API call seems successful. However, the API calls to get GPG and SSH seems to be using the `username` argument which can be an empty string.

kfcampbell:
Closing this as #261 has been merged. Please reopen if this is in error!

PR #261, opened 2019-07-21
Add support to retrieve information about the currently authenticated user

mithun:
Helps address  #260


politician:
Any reason why this hasn't been accepted?

jcudit (approved):
@politician thanks for the bump.

@mithun 🙇 for the contribution.

PR #266, opened 2019-08-06
[WIP] `ImportStateVerify` updates

tracypholmes:


github/resource_github_branch_protection_test.go
```diff
@@ -17,6 +17,7 @@ import (
 func TestAccGithubBranchProtection_basic(t *testing.T) {
 	var protection github.Protection
 
+	resourceName := "github_branch_protection.master"
```
katbyte: Now that we have a variable with the resource name (i might also go with rn over resource name but it doesn't matter) we can replace all instances of that with the variable:
```
					resource.TestCheckResourceAttr(rn, "branch", "master"),
					resource.TestCheckResourceAttr(rn, "enforce_admins", "true"),
					resource.TestCheckResourceAttr(rn, "require_signed_commits", "true"),
					resource.TestCheckResourceAttr(rn, "required_status_checks.0.strict", "true")
```
(applies to all)

github/resource_github_branch_protection_test.go
```diff
@@ -82,23 +89,36 @@ func TestAccGithubBranchProtection_users(t *testing.T) {
...
+				// Possible API bug
+				ImportStateVerifyIgnore: []string{
+					"etag", "restrictions.0.users.#", "restrictions.0.users.35441184",
+				},
+			},
 		},
 	})
 }
 
+/* TODO: Research what's going on with this restrictions bug in the API
```
katbyte: Great comment

github/resource_github_branch_protection_test.go
```diff
@@ -114,57 +134,67 @@ func TestAccGithubBranchProtection_teams(t *testing.T) {
...
 				),
 			},
+			{
+				ResourceName:      rn,
+				ImportState:       true,
+				ImportStateVerify: true,
+				// Possible API bug
+
+				ImportStateVerifyIgnore: []string{
+					"restrictions.0.teams.#",
```
katbyte: We shouldn't ignore bugs, let the test fail so its apparently something needs to be fixed

github/resource_github_branch_protection_test.go
```diff
@@ -82,23 +89,36 @@ func TestAccGithubBranchProtection_users(t *testing.T) {
...
 				Config:      testAccGithubBranchProtectionConfigUser(repoName, "user_with_underscore"),
 				ExpectError: regexp.MustCompile("unable to add users in restrictions: user_with_underscore"),
 			},
+			{
+				ResourceName:      rn,
+				ImportState:       true,
+				ImportStateVerify: true,
+				// Possible API bug
+				ImportStateVerifyIgnore: []string{
+					"etag", "restrictions.0.users.#", "restrictions.0.users.35441184",
```
katbyte: We shouldn't just ignore something to get a test to pass, let it fail, merge anyways so its apparently it needs a fix

github/resource_github_branch_protection_test.go
tracypholmes: Ok. Will change it back.

PR #289, opened 2019-10-22
Feat/add retry

37108:
Adding user and inviting through `terraform apply` sometimes fails due to 422 error.
This is because of Terraform changes "github_membership" before "github_team_membership".

So I added `Retry` in "github_membership" for retrying request if API returns `You must purchase at least one more seat to add this user as a member.`

I did `make test` and I checked passing. But I don't test real environment yet.
This is why I made this as draft. 
I will make an organization and test, but if you help me, I put this in your hands.

## Related Issue
* https://github.com/terraform-providers/terraform-provider-github/issues/177

### DIFF
--- a/github/resource_github_membership.go
+++ b/github/resource_github_membership.go
@@ -7,7 +7,7 @@ import (
 	"time"
 
 	"github.com/google/go-github/v25/github"
-  "github.com/hashicorp/terraform/helper/resource"
+ 	"github.com/hashicorp/terraform/helper/resource"
 	"github.com/hashicorp/terraform/helper/schema"
 )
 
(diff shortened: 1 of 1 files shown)


37108:
I checked this working in my "test Organization".
But I can't test in "Team plan"'s Organization, and I think this test is most important.
Someone can help me?

kpfleming:
Does the GitHub API return a specific, testable, status code in the HTTP response when this situation occurs? It would be much safer to test for that than for the error message itself, which could be changed (or even translated).

37108:
@kpfleming 
In my case, I got 422 status code and error message, "422 You must purchase at least one more seat to add this user as a member."

I already added code that checks error message.

https://github.com/37108/terraform-provider-github/blob/62c2c212805a318eb2bbcfd4552f2c9303016fbf/github/resource_github_membership.go#L73-L80

If this change has a bad influence on other, using only `depends_on` for ordering execution is better and this works well my environment.

PR #300, opened 2019-11-25
go mod tidy

radeksimko:
The update was made [by the renovate bot](https://github.com/terraform-providers/terraform-provider-github/pull/274), but in reality that version was never used, because no import paths were updated.

I turned off the renovate bot for that reason, to prevent similar messy upgrades in the future. We can turn it back on at some point, but someone will need to investigate configuration and whether it's even possible for the bot to update import paths.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,8 +1,8 @@
 module github.com/terraform-providers/terraform-provider-github
 
 require (
+	github.com/google/go-github/v25 v25.1.3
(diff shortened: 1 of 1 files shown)


appilon:
@radeksimko should we run go mod vendor?

radeksimko:
@appilon  `go mod vendor` doesn't really change anything - I tried.

Issue #270, opened 2019-08-21
github_team_repository permission field do not accept "maintain"


Ameausoone:
### Terraform Version

Terraform v0.11.14
provider.github v2.2.0

### Affected Resource(s)
Please list the resources as a list, for example:
- github_team_repository

### Terraform Configuration Files
```hcl
resource "github_team_repository" "team_developers" {
  team_id = "${github_team.team.id}"
  repository = "my-repo"
  permission = "maintain"
}
```

### Debug Output
N/A

### Panic Output
N/A

### Expected Behavior
We expect to assign "maintain" role on repository for specified team.

### Actual Behavior
But this role is still in beta, it doesn't seem to be available through api => https://developer.github.com/v3/repos/collaborators/#add-user-as-a-collaborator.

I got this (expected) error: 

```
Error: github_team_repository.team_developers: maintain is an invalid value for argument permission
```

### Steps to Reproduce
Please list the steps required to reproduce the issue, for example:
1. `terraform apply`

### Important Factoids
I think it will be available once maintain feature will be released, I create this issue to follow when it will be available. 

### References
* api reference : https://developer.github.com/v3/repos/collaborators/#add-user-as-a-collaborator
* Terraform documentation : https://www.terraform.io/docs/providers/github/r/team_repository.html

alexbde:
As stated in #296 this feature is out of beta / generally available: https://github.blog/2019-10-03-delegate-responsibilities-with-expanded-repository-permissions/

Ameausoone:
OK I close this issue, duplicate with #296  thx @alexbde

PR #296, opened 2019-11-15
Support for 'maintain' and 'triage' as team permissions

gposton:
Simple PR to add support for triage and maintain as team roles.

Successfully tested using a terraform apply in our environment.  It works despite Github API documentation not officially showing this is available (as of 11/15/2019).

PR #303, opened 2019-11-26
Add triage and maintain repo permissions

### LINKED ISSUE #279: Extra permissions required
We require the use of the `triage` permission, which is currently still in beta (see https://help.github.com/en/articles/repository-permission-levels-for-an-organization).

What is the view on supporting beta features? Would you accept a PR to add the beta permissions to `github_team_repository`?

liamg:
Closes #279 

This PR makes the `triage` and `maintain` repository permissions available. See https://github.blog/2019-10-03-delegate-responsibilities-with-expanded-repository-permissions/

### DIFF
--- a/github/resource_github_repository_collaborator.go
+++ b/github/resource_github_repository_collaborator.go
@@ -37,7 +37,7 @@ func resourceGithubRepositoryCollaborator() *schema.Resource {
 				Optional:     true,
 				ForceNew:     true,
 				Default:      "push",
-				ValidateFunc: validateValueFunc([]string{"pull", "push", "admin"}),
+				ValidateFunc: validateValueFunc([]string{"pull", "triage", "push", "maintain", "admin"}),
 			},
 			"invitation_id": {
 				Type:     schema.TypeString,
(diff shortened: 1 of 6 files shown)


kpfleming:
Note that this PR contains the same (unrelated) fix as in #310. We really need that one merged so everyone can stop making this same change in their branches :-)

alexbde:
Looks like #296 is another duplicate of this. Please merge one of them :)

alexbde:
Closes #270.

salvianreynaldi:
Hi, 
the code seems to address both `github_team_repository` and `github_repository_collaborator`. However, I get errors when trying to set repository collaborator:
```
unexpected permission value: triage
unexpected permission value: maintain
```
Terraform set the permissions correctly, though. However, I can't then modify the permission (e.g. from maintain/triage to `pull`) in terraform, it gives the same error message, and the permissions are not updated this time.

This doesn't happen with the team repository. Am I missing something?

paultyng:
@salvianreynaldi not that I'm aware of, this is working for me now, but I do occasionally get unexpected diffs with different permission strings, but it doesn't seem to affect the actual runtime permissions available to people. I haven't dug in to the diff/read issue yet to figure out the reason for that, it may be related. 

If you can though, I'd recommend opening a new issue with a repro if you have one, or at least a log.

romaninsh:
The problem is here: https://github.com/terraform-providers/terraform-provider-github/blob/8c9a8c3892278e7a32fda6b8744190994722db33/github/util_permissions.go#L51

anGie44:
Fix to be released in the upcoming `v2.8.0` release.

PR #310, opened 2019-12-30
Correct timestamp for oauth2 module in go.mod

kpfleming:
The timestamp in go.mod does not match the timestamp in the
oauth2 repository, and thus all of the golang tools complain about
it. This patch corrects go.mod and go.sum to match the timestamp.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -5,5 +5,7 @@ require (
 	github.com/hashicorp/terraform v0.12.7
 	github.com/kylelemons/godebug v1.1.0
 	github.com/terraform-providers/terraform-provider-tls v1.2.0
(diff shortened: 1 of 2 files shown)


jcudit:
@kpfleming - see [here](https://github.com/terraform-providers/terraform-provider-github/pull/309#discussion_r362786392) for recent commentary on this subject.  If acceptance testing fails due to this issue, I will push to resolve this.  Otherwise, this is a useful fix to leave documented in case other community members run into the issue.

kpfleming:
Agreed, although the timestamp correction can be handled separately from adding the 1.13 tag to the end of the file. The timestamp is just incorrect :-)

PR #312, opened 2019-12-31
Refactor: cleanup usage of checkOrganization

kpfleming:
Every instance where checkOrganization is called, to ensure that
the provider has been configured with an organization name, the
code also requires the name itself, and obtains it later.

This patch changes the checkOrganization function to return the
name if it is present, and an error if it not.

In addition there were some places where checkOrganization was
called but the organization name was not actually required, or
where it was obtained, included in a log message, but not used
in the resulting API call.

### DIFF
--- a/github/data_source_github_repositories.go
+++ b/github/data_source_github_repositories.go
@@ -43,11 +43,6 @@ func dataSourceGithubRepositories() *schema.Resource {
 }
 
 func dataSourceGithubRepositoriesRead(d *schema.ResourceData, meta interface{}) error {
-	err := checkOrganization(meta)
-	if err != nil {
-		return err
-	}
-
 	client := meta.(*Organization).client
 
 	query := d.Get("query").(string)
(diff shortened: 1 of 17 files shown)


kpfleming:
I've pushed a new commit which also renames the function to `getOrganization` since that's what it does now.

jcudit:
This one triggered two tests to fail:
```
------- Stdout: -------
=== RUN   TestAccGithubRepositoryDataSource_name_existing
=== PAUSE TestAccGithubRepositoryDataSource_name_existing
=== CONT  TestAccGithubRepositoryDataSource_name_existing
--- FAIL: TestAccGithubRepositoryDataSource_name_existing (0.22s)
    testing.go:569: Step 0 error: errors during refresh:
        
        Error: GET https://api.github.com/repos//test-repo: 404 Not Found []
        
          on /opt/teamcity-agent/temp/buildTmp/tf-test087211269/main.tf line 2:
          (source code not available)
        
        
FAIL
```
```
------- Stdout: -------
=== RUN   TestAccGithubBranchProtection_emptyItems
=== PAUSE TestAccGithubBranchProtection_emptyItems
=== CONT  TestAccGithubBranchProtection_emptyItems
--- FAIL: TestAccGithubBranchProtection_emptyItems (5.01s)
    testing.go:569: Step 0 error: errors during apply:
        
        When applying changes to github_repository.test, provider "github" produced an
        unexpected new value for was present, but now absent.
        
        issue tracker.
        
FAIL
```

I've got this queued to investigate further.

PR #315, opened 2020-01-01
Refactor: simplify API result handling

kpfleming:
Every resource `Read` function has to process the results
of its primary GitHub API call, and there are a number of
possible results and actions to be taken.

This patch adds a function to parse the API results and return
an enum indicating the type of result the API returned; this
allows the calling code to use a `switch` statement to handle
each possible case, making the code more readable.

### DIFF
--- a/github/resource_github_branch_protection.go
+++ b/github/resource_github_branch_protection.go
@@ -5,7 +5,6 @@ import (
 	"errors"
 	"fmt"
 	"log"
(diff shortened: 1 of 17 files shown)


jcudit:
This one is having a bad time in CI:
> Process exited with code 1

I've got this queued to investigate but may be a bit more involved than the other related cleanup PRs.

PR #314, opened 2020-01-01
Refactor: clean up buildTwoPartID

kpfleming:
The parameters to buildTwoPartID are not optional,
so they should be passed by-value and not by-pointer.

Signed-off-by: Kevin P. Fleming <kpfleming@bloomberg.net>

PR #313, opened 2019-12-31
Refactor: cleanup usage of parseTwoPartID

kpfleming:
This function is used in many resources to parse colon-separated
two-part identifier strings, and it emits an error message when
the string does not match the expected format. The error message
indicates that the format was expected to be "organization:name"
but this is only true for some resources, not all of them. As a
result users may get an error while attempting to import a resource
which gives them incorrect information about the identitier format.

This patch improves the function to accept a pair of label strings,
and it uses them when constructing the error message, so the message
will be tailored to the call site's desired format for the string.

Signed-off-by: Kevin P. Fleming <kpfleming@bloomberg.net>

PR #311, opened 2019-12-30
Refactor: consolidate resource context creation and etag handling

kpfleming:
This patch refactors resource context handling to use a new helper
function in util.go; this results in consistent context behavior
and etag handling, and simplifies the code.

Signed-off-by: Kevin P. Fleming <kpfleming@bloomberg.net>

Issue #298, opened 2019-11-22
Handling renames of user accounts properly - thoughts?


kpfleming:
### Terraform Version
Terraform v0.12.12
+ provider.github v2.2.1

### Affected Resource(s)
github_team_membership

### Expected Behavior
Terraform should not have planned to make a team membership change because the user was already in the team.

### Actual Behavior
Terraform plans to make a team membership change.

### Steps to Reproduce
1. Add a user to a team, and apply the changes.
2. Ask the user to change their username on github.com.
3. Run 'plan', and see that Terraform can no longer see that the user is in the team.

This is clearly caused by the Terraform configuration (and state) being based on the the GitHub user's account name, and not its internal ID number. The name is changeable by the user at any time, but the internal ID number is not. If the user changes their account name, Terraform will attempt to add the *old* account name to the team, which could be very bad if that account name has been grabbed by a different user.

In my mind I can envision some mechanism to import a github_user as *data* (not a resource) and grab its internal ID number, then throughout the rest of the configuration refer to that github_user resource (as is done for teams already) and use the internal ID number in the API calls (and in the stored state). If the user changes their name, there wouldn't be any effect, other than the local configuration file would have the wrong 'name' for that github_user 'data' block.

kpfleming:
After learning more about the internals of Terraform this clearly can't work, because *data* objects are not persisted in the state.

This would mean that we'd need a `github_user` resource instead: it wouldn't be creatable or destroyable (only importable), but would otherwise behave as proposed above.

I'm going to experiment with this shortly.

jcudit:
As discussed over Zoom, let's check back in after a couple of months while the following tasks are worked on:

- [ ] @kpfleming to perform exploration work around a `github_user` resource
- [ ] @kpfleming to explore using the standard testing framework to express the **Steps to Reproduce**
- [ ] @kpfleming to explore writing an unrelated provider to gain experience that would benefit future improvements for this provider
- [ ] @jcudit to get onboarded as a maintainer and guide the above changes towards a suitable release

kpfleming:
I've completed items 1 and 3 from this list; item 2 requires some documentation of the environment require to be able to run the test suite before I can experiment with it.

PR #316 is a nearly-comprehensive attempt to address this issue; it allows the Terraform user to use numeric user IDs for GitHub users in their configuration, even though the GitHub API does not accept them (they are translated back to usernames just before each API call is issued). There is more work to do, but the current state works for my configuration.

PR #316, opened 2020-01-04
RFC: User identifier stability (possible version 3 of provider)

kpfleming:
This PR is far from being ready to merge but is being opened for discussion and feedback from the community.

As documented in #298, using usernames to refer to GitHub users is somewhat fragile, as users can change their username at any time (and unlike all the other objects in terraform-provider-github, those are not resources being managed by Terraform, but outside of Terraform). At the moment the GitHub API V3 only allows usernames to be used in API calls, but the provider can still offer a more stable interface to Terraform users.

This patch modifies `github_membership`, `github_repository_collaborator`, and `github_team_membership` to use numeric user IDs instead of usernames. The HCL configuration can either hardcode the user IDs (manually entered or provided from an external tool which populates the configuration), or the user can use the `github_user` data source to obtain the user ID at runtime, although doing so means the configuration is still fragile.

All three resources have new schema versions, and StateUpgraders to automatically update existing state files during the first 'apply' run. Of course since the arguments to the resources have been changed, users of the provider will have to edit their configuration files before they can run validate/plan/apply operations.

All three resources also have much friendlier 'import' operations, which don't require the user to provide a numeric user/team/etc. ID (although they can be provided if the user has them); instead the importer translates the username/team slug/etc. to its numeric ID during import.

The documentation has been updated as well, and the unit tests have been updated to ensure they stil compile, but they are *not* updated to actually test the modified resources. That will have to wait until after @jcudit has figured out how people who don't work for Hashicorp can run the test suite :-) I *have* done real-world testing of these changes in a configuration with ~100 users and a couple of dozen repositories/teams.

On a minor note (and a change which could easily be separated), I've enhanced the `github_user` data source to allow the user to specify either a username or a user ID to find the user, and also added a `minimal` flag which suppresses the API calls to obtain SSH and GPG keys, since many users do not need them.

Note that this branch incorporates the changes from #310, #311, #312, #313, #314, and #315. This branch will be updated as those PRs are merged and closed. Also readers paying close attention will see commits which add a `github_user` resource and then removal of it; for now I'm not including it as it would be a non-traditional Terraform resource.


jcudit:
> have to wait until after @jcudit has figured out how people who don't work for Hashicorp can run the test suite :-)

I was pointed in the direction of passing organization and repo values as env variables.  The fix is still pending, but I agree that this is a simple and effective solution.

> Also readers paying close attention will see commits which add a github_user resource and then removal of it; for now I'm not including it as it would be a non-traditional Terraform resource

@kpfleming - I'd like to hear more about what made the `github_user` a non-traditional resource.  When we last spoke it seemed like a good resource to implement.

---

I've got time to read through the PR series today and plan to leave questions and feedback.  I'm still getting comfortable with the release process around this repository, so we should look to involve more reviewers that are familiar with established best practices around merging.

kpfleming:
I implemented `resource_github_user`, with provisions to ensure that Terraform could neither create nor destroy them (they had to be imported). This makes them behave differently from any other Terraform resource, and most importantly they would *not* actually be managed by Terraform even though a casual reader of the HCL would believe that they were.

Originally I implemented them so that user lookups could be done using the numeric ID, and so that the `ETag` value from previous API calls could be reused, but the first can now be done by passing a numeric ID to `data_source_github_user`, and the second is not a significant performance win as best I can tell. If testing proves otherwise, that may be justification for adding the resource, as the data source can't preserve an `ETag`.

The only remaining reason I implemented the resource was so that I could avoid the additional API calls which the data source generates (to obtain GPG and SSH keys), but it was a simple matter to add a boolean argument to the data source to turn off those API calls.

jcudit:
👋 @kpfleming I am working through the acceptance tests this week to get this series merged in.  Thanks for your patience so far.

kpfleming:
I would hold off on that; I'd rather work on the move to the V4 API, which uses stable object IDs by default. Once the V4 API client is available in the provider, I'll rework this branch to use it, as the current mechanism is really ugly.

jcudit:
OK, I will put this series on hold unless requested otherwise.  How about #313?  I see it is partially blocking #318, which is an enhancement I want to start leveraging.

kpfleming:
I was only referring to this specific PR; the refactoring PRs are useful on their own and I'm fine with those being merged.

jcudit:
Thanks for clarifying.  I've begun interacting with this code base's CI workflow and would like to invest in making errors more readily visible to contributors.  The associated cleanup PRs have been manually updated with CI errors obtained from the current system.  I aim to improve this workflow before returning to getting the tests to pass.

PR #326, opened 2020-01-17
Fix getTeamAndUserFromURL function

steakfest:
The github api started returning a different url format, this handles both the old and new string, just in case they change it back.

### DIFF
--- a/github/resource_github_team_membership.go
+++ b/github/resource_github_team_membership.go
@@ -145,7 +145,7 @@ func getTeamAndUserFromURL(url *string) (string, string) {
 
 	urlSlice := strings.Split(*url, "/")
 	for v := range urlSlice {
-		if urlSlice[v] == "teams" {
+		if urlSlice[v] == "teams" || urlSlice[v] == "team" {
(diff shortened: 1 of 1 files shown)


steakfest:
This is a possible alternative to https://github.com/terraform-providers/terraform-provider-github/pull/325 

This seems like a less intrusive way of changing the behavior of the provider.  Keeping the original intent of the code and adjusting for the change, safely.

MichaelMarieJulie:
Thanks you for this fix, it seems working perfectly
May we know when it can be released as it’s really impactful for everybody ?

meabed:
Thanks for the fix! how would you override this directly until the fix is released official ?

TheKevJames:
@meabed see the "quick fix" here: https://github.com/terraform-providers/terraform-provider-github/pull/325#issuecomment-575674622 (note that you'll need to change the github invocation if you'd prefer this fix over the other one). You may also need to roll back your state file: https://github.com/terraform-providers/terraform-provider-github/pull/325#issuecomment-575751214

PR #347, opened 2020-02-05
[fix] Example's template block fixed

Cre-eD:
`repo` is not supported by this module


jcudit:
Thanks for this @Cre-eD

PR #349, opened 2020-02-10
Update repository.html.markdown

rballcmm:
An argument named "repo" is not expected here.


benj-fletch:
Hi @rballcmm, I believe that another contributor also raised a similar PR to fix this issue under https://github.com/terraform-providers/terraform-provider-github/pull/347 - are you happy that this completes your request and this PR can be closed?

Thanks for raising this though!

PR #355, opened 2020-02-13
Fix wrong argument name

eyalzek:


jcudit:
Duplicates https://github.com/terraform-providers/terraform-provider-github/pull/347

PR #361, opened 2020-02-24
Updating docs to include triage and maintain permissions, as per PR #303

MatStace:
I spent a little bit of time last week digging around until I discovered that triage and maintain permissions had been added in #303 so I tweaked the docs to include mention of them.


jcudit:
Thanks for this @MatStace and apologies for your time spent digging in. 🙇

PR #342, opened 2020-02-01
Update go-github to v29.

jakebiesinger-onduo:
Updates to go-github v29. To do this, I ran `go mod vendor` after updating `vendor/modules.txt`.

There were two small interface changes I had to make manually:
 1. `client.Teams.EditTeam` now takes a third option, `removeParent` which I've set to `false`
 2. `client.Repositories.AddCollaborator` now returns a data object representing the invitation. This didn't seem to be used in the surrounding code so I am just ignoring it.



I also ran acceptance tests locally. There are a few failures, but as a new contributor, I'm not sure how to proceed fixing them:
 1.  a different etag than is expected (does this mean an API call that was atomic no longer is?)
```
--- FAIL: TestAccGithubTeam_hierarchical (71.24s)                                                                                                   
    testing.go:569: Step 1 error: ImportStateVerify attributes not equivalent. Difference is shown below. Top is actual, bottom is expected.        
                                                                                                                                                    
        (map[string]string) (len=1) {                                     
         (string) (len=4) "etag": (string) (len=36) "W/\"195adb63bbfd36cd6e36f9baeb47351c\""                                                        
        }                                                                                                                                           
                                                                                                                                                    
                                                                                                                                                    
        (map[string]string) (len=1) {                                     
         (string) (len=4) "etag": (string) (len=36) "W/\"4ba9252760600a41994f7061c0357bdd\""                                                        
        }                                                                 
```


 2. some team changes don't seem to be retained in protections as expected:

```
--- FAIL: TestAccGithubBranchProtection_teams (113.12s)
    testing.go:569: Step 0 error: Check failed: Check 3/15 error: diff "restrictions.teams": (-got +want)
         [
        + "team-1-n6fmt",
        + "team-2-n6fmt",
         ]
--- FAIL: TestAccGithubBranchProtection_teams (154.32s)                                                                                             
    testing.go:569: Step 0 error: Check failed: Check 3/15 error: diff "restrictions.teams": (-got +want)                                           
         [                                                                                                                                          
        + "team-1-m3wpi",                                                                                                                           
        + "team-2-m3wpi",                                                                                                                           
         ]                                                                                                                                          
```

 3. None of the acceptance tests related to a local server are able to run on my machine.

```
        Error: Post https://localhost:58299/orgs/github-test-jakebiesinger/repos: dial tcp 127.0.0.1:58299: connect: connection refused             
```

Issue #271, opened 2019-08-21
Feature request: Support secret creation for github actions


gjambet:
Hi friends,

 it would be super cool to be able to set secrets for actions through terraform while creating a repository. ;)

ty, 

g. (happy action user)

hronix:
Yes please

zzh8829:
github hasn't made api for secrets available yet, we have to wait for now

iniinikoski:
Hi. Any updates on this - is the Secrets API still private...?

gjambet:
❤️

jcudit:
GitHub is aiming to make this endpoint available within the next few sprints.  Mention me here if I don't post an update in a month or so.

ajliv:
Secrets have been added to the GitHub API https://developer.github.com/v3/actions/secrets/

martinssipenko:
We now need to wait till Go GitHub client releases action secrets - https://github.com/google/go-github/pull/1402

benesis002:
+1

btkostner:
It looks like it was merged https://github.com/google/go-github/pull/1402#issuecomment-583939009 and released in the [v29.0.3 tag](https://github.com/google/go-github/blob/v29.0.3/github/actions_secrets.go).

martinssipenko:
I think we should also wait for #342 before we start implementing this one.

martinssipenko:
What do we think about usage? Should action secrets be a new resource - `github_repository_action_secret` and be used as follows?

```hcl
resource "github_repository_action_secret" "foo" {
    repository = "${github_repository.repo.name}"
    name       = "FOO"
    value      = "secret"
}
```

hazcod:
Is there a way to provide ciphertexts in terraform and supply the key via an environment variable?
I can imagine it's problematic to store plaintext github secrets in terraform files ... which might be located on github again.

martinssipenko:
@hazcod I think one option would be that resource accepts two conflicting arguments - `secret` and `encrypted_secret`, The `secret` could be used to pass plain text value which would be encrypted by terraform using GitHubs public key. The `encrypted_secret` would accept already encrypted value and leave the encryption the user.

Off topic: In order to not store secrets in terraform files I use an approach where I store secrets somewhere else, and then use a data provider to get the secret values from there. In that case I only need to reference the value from data source.

martinssipenko:
@radeksimko @jcudit @paultyng what are your thoughts on this?

[...]

PR #362, opened 2020-02-24
Gh 271 GitHub secrets

benj-fletch:
This PR covers functionality described in GH-271

The following functionality has been provided by this new code:

Data source `github_actions_public_key` - retrieving the public key of a repository
Data source `github_actions_secrets` - retrieval of all secrets in a repository
Resource `github_actions_secret` - management of Actions Secrets

See the documentation for details around config for each new item.

Note that this requires #342 to be merged prior to this and so it has been raised as a draft PR

### DIFF
--- a/github/resource_github_actions_secret.go
+++ b/github/resource_github_actions_secret.go
@@ -1,0 +1,169 @@
+package github
+
+import (
+	"context"
+	"encoding/base64"
+	"fmt"
+	"github.com/google/go-github/v29/github"
+	"github.com/hashicorp/terraform-plugin-sdk/helper/schema"
+	"golang.org/x/crypto/nacl/box"
+	"log"
+)
+
+func resourceGithubActionsSecret() *schema.Resource {
+	return &schema.Resource{
+		Create: resourceGithubActionsSecretCreateOrUpdate,
+		Read:   resourceGithubActionsSecretRead,
+		Update: resourceGithubActionsSecretCreateOrUpdate,
+		Delete: resourceGithubActionsSecretDelete,
+
+		Schema: map[string]*schema.Schema{
+			"repository": {
+				Type:     schema.TypeString,
+				Required: true,
+			},
+			"secret_name": {
+				Type:     schema.TypeString,
+				Required: true,
+				ForceNew: true,
+			},
+			"plaintext_value": {
+				Type:      schema.TypeString,
+				Required:  true,
+				Sensitive: true,
+			},
+			"created_at": {
+				Type:     schema.TypeString,
+				Computed: true,
+			},
+			"updated_at": {
+				Type:     schema.TypeString,
+				Computed: true,
+			},
+		},
+	}
+}
+
+func resourceGithubActionsSecretCreateOrUpdate(d *schema.ResourceData, meta interface{}) error {
+	err := checkOrganization(meta)
+	if err != nil {
+		return err
+	}
+
+	client := meta.(*Organization).client
+	owner := meta.(*Organization).name
+	ctx := context.Background()
+
+	repo := d.Get("repository").(string)
+	secretName := d.Get("secret_name").(string)
+	plaintextValue := d.Get("plaintext_value").(string)
+
+	keyId, publicKey, err := getPublicKeyDetails(owner, repo, meta)
+	if err != nil {
+		return err
+	}
+
+	encryptedText, err := encryptPlaintext(plaintextValue, publicKey)
+	if err != nil {
+		return err
+	}
+
+	// Create an EncryptedSecret and encrypt the plaintext value into it
+	eSecret := &github.EncryptedSecret{
+		Name:           secretName,
+		KeyID:          keyId,
+		EncryptedValue: base64.StdEncoding.EncodeToString(encryptedText),
+	}
+
+	_, err = client.Actions.CreateOrUpdateSecret(ctx, owner, repo, eSecret)
+	if err != nil {
+		return err
+	}
+
+	d.SetId(buildTwoPartID(&repo, &secretName))
+	return resourceGithubActionsSecretRead(d, meta)
+}
+
+func resourceGithubActionsSecretRead(d *schema.ResourceData, meta interface{}) error {
+	err := checkOrganization(meta)
+	if err != nil {
+		return err
+	}
+
+	client := meta.(*Organization).client
+	owner := meta.(*Organization).name
+	ctx := context.Background()
+
+	repoName, secretName, err := parseTwoPartID(d.Id(), "repository", "secret_name")
+	if err != nil {
+		return err
+	}
+
+	secret, _, err := client.Actions.GetSecret(ctx, owner, repoName, secretName)
+	if err != nil {
+		d.SetId("")
+		return err
+	}
+
+	d.Set("plaintext_value", d.Get("plaintext_value"))
+	d.Set("updated_at", secret.UpdatedAt.Format("default"))
+	d.Set("created_at", secret.CreatedAt.Format("default"))
+
+	return nil
+}
+
+func resourceGithubActionsSecretDelete(d *schema.ResourceData, meta interface{}) error {
+	err := checkOrganization(meta)
+	if err != nil {
+		return err
+	}
+
+	client := meta.(*Organization).client
+	orgName := meta.(*Organization).name
+	ctx := context.WithValue(context.Background(), ctxId, d.Id())
+
+	repoName, secretName, err := parseTwoPartID(d.Id(), "repository", "secret_name")
+	if err != nil {
+		return err
+	}
+
+	log.Printf("[DEBUG] Deleting secret: %s", d.Id())
+	_, err = client.Actions.DeleteSecret(ctx, orgName, repoName, secretName)
+
+	return err
+}
+
+func getPublicKeyDetails(owner, repository string, meta interface{}) (keyId, pkValue string, err error) {
+	client := meta.(*Organization).client
+	ctx := context.Background()
+
+	publicKey, _, err := client.Actions.GetPublicKey(ctx, owner, repository)
+	if err != nil {
+		return keyId, pkValue, err
+	}
+
+	return publicKey.GetKeyID(), publicKey.GetKey(), err
+}
+
+func encryptPlaintext(plaintext, publicKeyB64 string) ([]byte, error) {
+	publicKeyBytes, err := base64.StdEncoding.DecodeString(publicKeyB64)
+	if err != nil {
+		return nil, err
+	}
+
+	var publicKeyBytes32 [32]byte
+	copiedLen := copy(publicKeyBytes32[:], publicKeyBytes)
+	if copiedLen == 0 {
+		return nil, fmt.Errorf("could not convert publicKey to bytes")
+	}
+
+	plaintextBytes := []byte(plaintext)
+	var encryptedBytes []byte
+
+	cipherText, err := box.SealAnonymous(encryptedBytes, plaintextBytes, &publicKeyBytes32, nil)
+	if err != nil {
+		return nil, err
+	}
+
+	return cipherText, nil
+}
--- a/website/docs/r/actions_secret.html.markdown
+++ b/website/docs/r/actions_secret.html.markdown
@@ -1,0 +1,55 @@
+---
+layout: "github"
+page_title: "GitHub: github_actions_secret"
+description: |-
+  Creates and manages an Action Secret within a GitHub repository
+---
+
+# github_actions_secret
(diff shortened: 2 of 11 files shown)


github/resource_github_actions_secret.go
```diff
@@ -0,0 +1,150 @@
...
+			"secret_name": {
+				Type:     schema.TypeString,
+				Required: true,
+			},
+			"plaintext_value": {
+				Type:      schema.TypeString,
+				Required:  true,
+				Sensitive: true,
+			},
+			"key_id": {
```
martinssipenko: Perhaps it's better to not have these arguments and instead just fetch and use the public key on create?

benj-fletch: It's a good shout. I initially provided these to allow users to determine the key / key id without the need for `github_actions_secrets_public_key` but I think just not including them as you suggest is better.

github/resource_github_actions_secret.go
```diff
@@ -0,0 +1,157 @@
...
+
+func resourceGithubActionsSecret() *schema.Resource {
+	return &schema.Resource{
+		Create: resourceGithubActionsSecretCreateOrUpdate,
+		Read:   resourceGithubActionsSecretRead,
+		Update: resourceGithubActionsSecretCreateOrUpdate,
+		Delete: resourceGithubActionsSecretDelete,
+
+		Schema: map[string]*schema.Schema{
+			"owner": {
```
martinssipenko: Why is the `owner` argument required?  Given you use [`checkOrganization`](https://github.com/terraform-providers/terraform-provider-github/blob/b838aed6391ac450740f023e4a714d6cd449aad3/github/util.go#L16) func, that checks that organization is configured in provider settings, you could get the value from provider config via `meta`. As an example, take a look at [repository webhook](https://github.com/terraform-providers/terraform-provider-github/blob/8dc722aeae3d86b082447e22f34ac69539bc277c/github/resource_github_repository_webhook.go#L102) code.

github/resource_github_actions_secret.go
```diff
@@ -0,0 +1,157 @@
...
+		KeyID:          keyId,
+		EncryptedValue: base64.StdEncoding.EncodeToString(encryptedText),
+	}
+
+	_, err = client.Actions.CreateOrUpdateSecret(ctx, owner, repo, eSecret)
+	if err != nil {
+		return err
+	}
+
+	d.SetId(fmt.Sprintf("%s/%s:%s", owner, repo, secretName))
```
martinssipenko: If we assume that `owner` comes from provider configuration, then you can't have more than a single owner value and the owner part in Id becomes redundant. Have you considered using two part id consisting of repo name and secret name, and combining them using [`buildTwoPartID`](https://github.com/terraform-providers/terraform-provider-github/blob/b838aed6391ac450740f023e4a714d6cd449aad3/github/util.go#L59-L61) utility func, you can find example [here](https://github.com/terraform-providers/terraform-provider-github/blob/8dc722aeae3d86b082447e22f34ac69539bc277c/github/resource_github_repository_deploy_key.go#L84)?

github/resource_github_actions_secret.go
```diff
@@ -0,0 +1,157 @@
...
+	repo := d.Get("repository").(string)
+	owner := d.Get("owner").(string)
+	secretName := d.Get("secret_name").(string)
+
+	secret, _, err := client.Actions.GetSecret(ctx, owner, repo, secretName)
+	if err != nil || secret.Name != secretName {
+		d.SetId("")
+		return err
+	}
+
```
martinssipenko: How come there are no `d.Set()` method calls? I think the secret information could be stored in state. The `UpdatedAt` value from `github.Secret` could be used to determine if the value has changed in GitHub, for example, if `UpdatedAt` stored in state differs from value you got in `GetSecret` response, it means someone has updated the value outside of Terraform, and we probably want to sync it up (reapply value defined in Terraform).

github/resource_github_actions_secret.go
```diff
@@ -0,0 +1,157 @@
...
+func resourceGithubActionsSecretDelete(d *schema.ResourceData, meta interface{}) error {
+	err := checkOrganization(meta)
+	if err != nil {
+		return err
+	}
+
+	client := meta.(*Organization).client
+	ctx := context.WithValue(context.Background(), ctxId, d.Id())
+
+	orgName := d.Get("owner").(string)
```
martinssipenko: Consider getting `orgName` from provider config via `meta` here.

github/resource_github_actions_secret.go
```diff
@@ -0,0 +1,157 @@
...
+	repoName := d.Get("repository").(string)
```
martinssipenko: Perhaps it's better to obtain `repoName` and `secretName` by extracting it from `d.GetId()` using [`parseTwoPartID`](https://github.com/terraform-providers/terraform-provider-github/blob/b838aed6391ac450740f023e4a714d6cd449aad3/github/util.go#L49-L56) utility func?

```go
repoName, secretName, err := parseTwoPartID(d.Id(), "repository", "secret_name")
if err != nil {
    return err
}
```

martinssipenko:
Can't really figure out from the code, are we always updating the secret on GitHub, even if it has not changed locally?

github/resource_github_actions_secret.go
benj-fletch: I was unaware that I could pull the organisation from the `meta` to be honest!
Thinking about this, would there be any case that you would want to manage a secret that wasn't part of the organisation in the provider? I think it might be possible in the current implementation but should not be encouraged - forcing users to have an individual provider per organisation seems more appropriate?

martinssipenko: Yes, same ideology is used when managing other repository resources, webhooks is one example.

github/resource_github_actions_secret.go
benj-fletch: I agree. I saw that function but wanted to use all 3 values. If we remove `owner` as per above I will implement this suggestion

github/resource_github_actions_secret.go
benj-fletch: I wasn't aware that the `d.Set()` functions were required to pass things to state. My understanding was that the existing `d` values passed in would get persisted.
I fully agree with the use of `UpdatedAt` - would you recommend saving the full state of the `github.Secret` ?

benj-fletch:
I didn't think we are updating every time, though reading this review, I was unaware that the secret data wasn't being saved to state so maybe we are always updating. As discussed further up in the review, I think we should add to the state and this will resolve your concern

github/resource_github_actions_secret.go
martinssipenko: I think I would store the `repo` and `name` as par of Id, then store `plaintext_value` as it is and lastly store `updated_at` as computed getting the value from create/update operation.

martinssipenko:
@benj-fletch sweet, if you need any help let me know. I'm happy to review again after you have meade the changes discussed above.

github/resource_github_actions_secret.go
benj-fletch: Sounds like a solid plan - I will also include `created_at` since we have it available and might as well make it available

benj-fletch:
Thanks for the comments @martinssipenko! I have updated the PR to hopefully address all of them. I'll leave it in draft still until the update to go github v29 has occured as I imagine it will create conflicts to resolve

github/resource_github_actions_secret.go
```diff
@@ -0,0 +1,166 @@
...
+	secretName := d.Get("secret_name").(string)
+
+	secret, _, err := client.Actions.GetSecret(ctx, owner, repo, secretName)
+	if err != nil || secret.Name != secretName {
+		d.SetId("")
+		return err
+	}
+
+	d.SetId(buildTwoPartID(&repo, &secretName))
+	d.Set("plaintext_value", d.Get("plaintext_value"))
```
martinssipenko: Not 100% sure but I think is not required, as it already should exist in state.

github/resource_github_actions_secret.go
```diff
@@ -0,0 +1,166 @@
...
+
+	client := meta.(*Organization).client
+	owner := meta.(*Organization).name
+	ctx := context.Background()
+
+	repo := d.Get("repository").(string)
+	secretName := d.Get("secret_name").(string)
+
+	secret, _, err := client.Actions.GetSecret(ctx, owner, repo, secretName)
+	if err != nil || secret.Name != secretName {
```
martinssipenko: Why is `|| secret.Name != secretName` part needed here?

github/resource_github_actions_secret.go
```diff
@@ -0,0 +1,166 @@
...
+		Name:           secretName,
+		KeyID:          keyId,
+		EncryptedValue: base64.StdEncoding.EncodeToString(encryptedText),
+	}
+
+	_, err = client.Actions.CreateOrUpdateSecret(ctx, owner, repo, eSecret)
+	if err != nil {
+		return err
+	}
+
```
martinssipenko: You should set the Id here using `d.SetId(buildTwoPartID(&repo, &secretName))`. Afterwards this func calls `resourceGithubActionsSecretRead` func, which instead should extract repo and secret name from the `d.GetId()` via `parseTwoPartID` func, I've given the example below.

martinssipenko:
Can you rebase from master? #342 was merged, but unfortunatley it includes v29.0.3, but we need v29.0.3 for this to work. The question is should we make update to latest in this or seperate PR?

github/data_source_github_actions_secrets.go
```diff
@@ -0,0 +1,106 @@
...
+
+import (
+	"context"
+	"fmt"
+	"github.com/google/go-github/v29/github"
+	"github.com/hashicorp/terraform/helper/schema"
+	"log"
+)
+
+func dataSourceGithubActionsSecrets() *schema.Resource {
```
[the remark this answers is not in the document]
benj-fletch: I initially added this as I thought it might be useful in future resources / data sources for GitHub actions. Reviewing the available API endpoints I think I agree that this data is useless in the long run. I will remove this data source.

github/resource_github_actions_secret.go
```diff
@@ -0,0 +1,166 @@
...
+	err := checkOrganization(meta)
+	if err != nil {
+		return err
+	}
+
+	client := meta.(*Organization).client
+	owner := meta.(*Organization).name
+	ctx := context.Background()
+
+	repo := d.Get("repository").(string)
```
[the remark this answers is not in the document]
benj-fletch: Yes, I should have caught this in the previous PR comment resolution! Thanks.

github/resource_github_actions_secret.go
benj-fletch: I am also not sure on this point. I left it there such that it was obvious (and definite) it would be set. I don't see any issue with leaving this line in?

github/resource_github_actions_secret.go
benj-fletch: I don't think it is on review!

benj-fletch:
Hi @martinssipenko, thanks a lot for the re-review. I have rebased onto master and hopefully have addressed all your concerns, are you able to have another look?

N.B. This PR now relies on #369 for the go-github v29.0.3 upgrade and will need rebasing again to master following its merge.

btkostner:
@benj-fletch #369 has been merged into master now. This can be rebased and hopefully merged! Thank you for all the work on this!

benj-fletch:
Thanks for the prod @btkostner - I have rebased, fixed a couple of conflicts and squashed the commits. This should be ready for review now! 💯

website/docs/r/actions_secret.html.markdown
```diff
@@ -0,0 +1,55 @@
...
+* `key_id`          - (Required) ID if the key used for encryption
+* `public_key`      - (Required) Public key of the repository to be used in encryption of the `plaintext_value`
+
+## Import
+
+Repositories can be imported using the `name`, e.g.
+
+```
+$ terraform import github_repository.terraform terraform
+```
```
jcudit: This needs to be updated or removed

website/github.erb
```diff
@@ -13,6 +13,9 @@
         <li>
           <a href="#">Data Sources</a>
           <ul class="nav nav-visible">
+            <li>
+              <a href="/docs/providers/github/d/actions_public_key.html">github_actions_public_key</a>
```
jcudit: I always miss this file, thanks.

website/docs/r/actions_secret.html.markdown
benj-fletch: Good catch, classic copy paste error! Will update

PR #382, opened 2020-03-09
Add repository vulnerability alerts

jtsaito:
Provides a GitHub vulnerability alerts resource.

Related to: https://github.com/terraform-providers/terraform-provider-github/issues/238


grobie:
What's the advantage of introducing a new resource for a boolean toggle instead of adding a new attribute to the existing `github_repository` resource?

jtsaito:
@grobie That's a legitimate question.

Sooner or later there should be API endpoints for setting security alerts' recipients (in the web console under `https://github.com/organization-foo/repo-bar/settings/alerts`). These could then be handled as attributes in the resource without bloating the repo resource.  (There are already other resources prefixed with `github_repository_`.)

PR #389, opened 2020-03-16
Typographical errors and consistency

ksatirli:
Hello 👋 

In this PR, I am fixing a handful of typographical errors and inconsistencies* in the documentation.

I've added code highlighting where it was missing, altered cases of words as required and removed or added linebreaks to be more in line with the rest of the documentation.

Due to the nature of this PR, I opted to combine all into one, rather than having separate PRs for each data source and resource.

Let me know if this works?

There are many more (small) inconsistencies, specifically in lists and styling that I'm happy to take on, if there's any interest.


ksatirli:
Thanks for merging this, @jcudit!

PR #398, opened 2020-03-30
Docs: Correct Repository_Collaborator Permission Levels

AjkayAlan:
Looking at https://github.com/terraform-providers/terraform-provider-github/blob/master/github/resource_github_repository_collaborator.go#L41, "triage" and "maintain" are supported. Update the docs to reflect it.


AjkayAlan:
Doing more testing, it looks like this may not be implemented yet, as when I attempt to apply the plan, it fails with an unexpected permission value: maintain.

Closing this PR.

PR #400, opened 2020-03-30
addition of team sync support

anGie44:
Addition of `team_sync_groups` data source and `team_sync_group_mapping` resource to address GitHub's new Team Sync capabilities as discussed in issue: https://github.com/terraform-providers/terraform-provider-github/issues/293


jopedros:
Hi @anGie44  I'm not a contributor in this provider, but should not you create also the documentation pages in /website/docs/r|d ?

Thanks!

anGie44:
ah yes great catch @jopedros, thank you! i'll make the addition.

github/resource_github_team_sync_group_mapping_test.go
```diff
@@ -0,0 +1,97 @@
...
+	updatedDescription := "Terraform acc test group - updated"
+
+	resource.ParallelTest(t, resource.TestCase{
+		PreCheck:     func() { testAccPreCheck(t) },
+		Providers:    testAccProviders,
+		CheckDestroy: testAccCheckGithubTeamSyncGroupMappingDestroy,
+		Steps: []resource.TestStep{
+			{
+				Config: testAccGithubTeamSyncGroupMappingConfig(teamName, groupID, groupName, description),
+				Check: resource.ComposeTestCheckFunc(
```
megan07, line 36: You shouldn't need the `Check` here, if you add a step after this (similar to your last step) that will import, because the import will check the state before and the state after, basically doing the same thing.

github/resource_github_team_sync_group_mapping_test.go
```diff
@@ -0,0 +1,97 @@
...
+	"github.com/hashicorp/terraform-plugin-sdk/terraform"
+)
+
+func TestAccGithubTeamSyncGroupMapping_basic(t *testing.T) {
+	var team github.Team
+
+	rn := "github_team.foo"
+	randStringTeam := acctest.RandStringFromCharSet(10, acctest.CharSetAlphaNum)
+	teamName := fmt.Sprintf("tf-acc-test-%s", randStringTeam)
+	randStringGroup := acctest.RandStringFromCharSet(10, acctest.CharSetAlphaNum)
```
megan07, line 23: You could probably use the same random string for both, if you wanted to clean up the variables here a bit, but totally up to you!

github/resource_github_team_sync_group_mapping_test.go
```diff
@@ -0,0 +1,97 @@
...
+				Check: resource.ComposeTestCheckFunc(
+					testAccCheckGithubTeamExists(rn, &team),
+					resource.TestCheckResourceAttrSet(rn, "groups"),
+					resource.TestCheckResourceAttr(rn, "groups.0.group_id", groupID),
+					resource.TestCheckResourceAttr(rn, "groups.0.group_name", groupName),
+					resource.TestCheckResourceAttr(rn, "groups.0.description", description),
+				),
+			},
+			{
+				Config: testAccGithubTeamSyncGroupMappingConfig(teamName, groupID, groupName, updatedDescription),
```
megan07, line 45: I think we might want to do a few steps here updating the actual groups.  If we can make an array of maybe 2 GitHub groups, then modify it to 1, and/or maybe even 0, just to make sure it's doing that update correctly, I think that might be a good check as well as updating the content of the groups.

github/resource_github_team_sync_group_mapping_test.go
```diff
@@ -0,0 +1,97 @@
...
+func testAccCheckGithubTeamSyncGroupMappingDestroy(s *terraform.State) error {
+	conn := testAccProvider.Meta().(*Organization).client
+	for _, rs := range s.RootModule().Resources {
+		if rs.Type != "github_team_sync_group_mapping" {
+			continue
+		}
+		id := rs.Primary.Attributes["team_id"]
+		groupList, resp, err := conn.Teams.ListIDPGroupsForTeam(context.TODO(), id)
+		if err == nil {
+			if len(groupList.Groups) > 0 {
```
megan07, line 70: We might want to also check groupList.Groups != nil here too.

github/data_source_github_team_sync_groups.go
```diff
@@ -0,0 +1,118 @@
...
+		orgName := meta.(*Organization).name
+		slug := d.Get("team_slug").(string)
+		team, err = getGithubTeamBySlug(ctx, client, orgName, slug)
+	}
+
+	if err != nil {
+		return err
+	}
+
+	idpGroups, _, err := client.Teams.ListIDPGroupsForTeam(ctx, string(team.GetID()))
```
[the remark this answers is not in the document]
anGie44: i think you're right! it felt redundant listing groups in a team for this data source, so organization makes much more sense

github/resource_github_team_sync_group_mapping_test.go
anGie44: will these checks for specific attributes help do the checks instead? rather than check if the actual resource exists

github/resource_github_team_sync_group_mapping_test.go
```diff
@@ -0,0 +1,180 @@
...
+				Config: testAccGithubTeamSyncGroupMappingConfig(teamName),
+				Check: resource.ComposeTestCheckFunc(
+					resource.TestCheckResourceAttrSet(rn, "group.#"),
+					resource.TestCheckResourceAttr(rn, "group.0.name", groupName),
+				),
+			},
+			{
+				Config: testAccGithubTeamSyncGroupMappingEmptyConfig(teamName),
+				Check: resource.ComposeTestCheckFunc(
+					resource.TestCheckResourceAttr(rn, "group.#", "0"),
```
anGie44, line 76: i wasn't sure here if this will check for an empty `group` array

github/resource_github_team_sync_group_mapping_test.go
```diff
@@ -0,0 +1,180 @@
...
+resource "github_team" "test_team" {
+  name        = "%s"
+  description = "team for acc-tests"
+}
+
+resource "github_team_sync_group_mapping" "test_mapping" {
+  team_slug  = github_team.test_team.slug
+  
+  dynamic "group" {
+    for_each = [for g in data.github_organization_team_sync_groups.test_groups.groups : g if g.name == "test_team_group"]
```
megan07, line 130: i think you'll want if `g.group_name`, right?

github/resource_github_team_sync_group_mapping_test.go
```diff
@@ -0,0 +1,180 @@
...
+
+resource "github_team_sync_group_mapping" "test_mapping" {
+  team_slug  = github_team.test_team.slug
+  
+  dynamic "group" {
+    for_each = [for g in data.github_organization_team_sync_groups.test_groups.groups : g if "test_team_group" in g.name]
+    content {
+      group_id          = each.value.group_id
+      group_name        = each.value.group_name
+      group_description = "%s"
```
megan07, line 158: I think maybe better than updating the description here would be to not filter in the for_each, maybe update from just one group to all the groups, or vice versa.

github/resource_github_team_sync_group_mapping_test.go
```diff
@@ -0,0 +1,180 @@
...
+resource "github_team" "test_team" {
+  name        = "%s"
+  description = "team for acc-tests"
+}
+
+resource "github_team_sync_group_mapping" "test_mapping" {
+  team_slug  = github_team.test_team.slug
+  
+  dynamic "group" {
+    for_each = [for g in data.github_organization_team_sync_groups.test_groups.groups : g if "test_team_group" in g.name]
```
megan07, line 154: Same here, I think we want g.group_name

github/resource_github_team_sync_group_mapping_test.go
```diff
@@ -0,0 +1,180 @@
...
+					resource.TestCheckResourceAttrSet(rn, "group.#"),
+					resource.TestCheckResourceAttr(rn, "group.0.name", groupName),
+				),
+			},
+			{
+				Config: testAccGithubTeamSyncGroupMappingAddGroupAndUpdateConfig(teamName, description),
+				Check: resource.ComposeTestCheckFunc(
+					resource.TestCheckResourceAttrSet(rn, "group.#"),
+					resource.TestCheckResourceAttr(rn, "group.0.description", description),
+					resource.TestCheckResourceAttr(rn, "group.1.description", description),
```
megan07, line 40: The test, as written, I think is only adding one group to the team.

github/data_source_github_organization_team_sync_groups.go
```diff
@@ -0,0 +1,90 @@
...
+
+		groups = append(groups, result...)
+
+		if resp.NextPage == 0 {
+			break
+		}
+		options.Page = string(resp.NextPage)
+	}
+
+	d.SetId("github-org-team-sync-groups")
```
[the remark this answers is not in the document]
anGie44: agreed!

github/resource_github_team_sync_group_mapping.go
```diff
@@ -0,0 +1,194 @@
...
+		return err
+	}
+	log.Printf("[DEBUG] Creating team-sync group mapping (Team: %s)", team.GetName())
+	_, resp, err := client.Teams.CreateOrUpdateIDPGroupConnections(ctx, string(team.GetID()), *idpGroupList)
+	if err != nil {
+		return err
+	}
+
+	d.Set("etag", resp.Header.Get("ETag"))
+	d.SetId("github-team-sync-group-mappings")
```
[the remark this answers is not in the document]
anGie44: ah yes, good idea to match the API call

github/resource_github_team_sync_group_mapping_test.go
anGie44: ah so for this check i was imagining the list comprehension `[for g in data.github_organization_team_sync_groups.test_groups.groups : g if "test_team_group" in g.group_name]`  could match more than one group if in the list of possible groups they existed as ["team_team_group_a", "test_team_group_b", ...]

anGie44: but i think it makes more sense to just include all groups like you mentioned in a comment above --which in that case perhaps these attribute checks would need to be in a loop to check the description on all the groups available?

github/data_source_github_organization_team_sync_groups.go
```diff
@@ -0,0 +1,74 @@
...
+}
+
+func dataSourceGithubOrganizationTeamSyncGroupsRead(d *schema.ResourceData, meta interface{}) error {
+	log.Print("[INFO] Refreshing GitHub Organization Team-Sync Groups")
+
+	client := meta.(*Organization).client
+	ctx := context.Background()
+
+	orgName := meta.(*Organization).name
+	options := &github.ListCursorOptions{PerPage: maxPerPage}
```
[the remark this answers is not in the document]
anGie44: ah yes its hiding in `util.go` - https://github.com/terraform-providers/terraform-provider-github/blob/a372e8adfa7d81bf1978298cfaaceaec5fb7db15/github/util.go#L13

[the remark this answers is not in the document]
anGie44: so perhaps in practice we would only lean towards using the cursor if we're expecting the list retrieved from the  `ListIDPGroupsInOrganization` call to be too large to be stored in the `IDPGroupList` object?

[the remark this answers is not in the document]
megan07: Ahh, ok, sorry, I didn't look into that as I was just browsing through.  This looks good, thanks!

jcudit:
I have queued this one to be released after [`v2.7.0`](https://github.com/terraform-providers/terraform-provider-github/milestone/4) which adds a GraphQL client and performs a batch update of how clients are obtained:

```go
meta.(*Organization).client
```

would turn into ...

```go
meta.(*Organization).v3client
```

We can get the conflicts fixed after `v2.7.0` merges, which could happen this week.

anGie44:
@megan07 I was finally able to test the data/resource w/my test github enterprise cloud org configured w/SAML and Azure AD. The changes I pushed up reflect some bugs I found but the API calls are still error prone -- filed the issue/PR here: https://github.com/google/go-github/pull/1484

github/resource_github_team_sync_group_mapping.go
```diff
@@ -0,0 +1,210 @@
...
+				Type:     schema.TypeString,
+				Optional: true,
+			},
+			"team_id": {
+				Type:     schema.TypeString,
+				Optional: true,
+			},
+			"groups": {
+				Type:     schema.TypeList,
+				Optional: true,
```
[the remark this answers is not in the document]
anGie44: I changed this to optional b/c in testing w/ the field as `Required` I found that if I wanted to set this to empty, I wasn't able to use `groups {}` in the TF config as terraform complained the nested attributes needed to be set. or is there another workaround to specify empty block?

gmlewis:
@anGie44 - https://github.com/google/go-github/releases/tag/v31.0.0 has been released, as requested. Thank you for your contributions! They are greatly appreciated!

anGie44:
Note: these changes (and subsequent changes to handle API calls) will be contingent on upgrade to `go-github v31.0.0` with  https://github.com/terraform-providers/terraform-provider-github/pull/424

jlpedrosa:
Hi @megan07 

Is this PR ready to be merged? I don't know if you have any ETA for the merge, we would probably go with a custom build from master when this patch is merged.

anGie44:
Output of acceptance testing (with enterprise account):
```
--- PASS: TestAccGithubTeamSyncGroupMapping_empty (12.61s)
--- PASS: TestAccGithubTeamSyncGroupMapping_basic (17.53s)
--- PASS: TestAccGithubTeamSyncGroupMapping_disappears (18.78s)
--- PASS: TestAccGithubTeamSyncGroupMapping_update (30.68s)
--- PASS: TestAccGithubOrganizationTeamSyncGroupsDataSource_existing (3.39s)
```

anGie44:
Hi @jlpedrosa 👋, please find the changes now in `master`. Let us know if you have any further questions or if we can help in any way. thanks in advance!

jlpedrosa:
Hi @anGie44 

Thanks for the follow up! As soon as it is released we will incorporate it in out project. Will let you know if there's any feedback we can give.

github/resource_github_team_sync_group_mapping.go
```diff
@@ -0,0 +1,210 @@
...
+
+		Schema: map[string]*schema.Schema{
+			"team_slug": {
+				Type:     schema.TypeString,
+				Required: true,
+			},
+			"group": {
+				Type:     schema.TypeSet,
+				Optional: true,
+				ForceNew: true,
```
Jorge-Rodriguez, line 35: @anGie44 I know this has been long part of the provider, but what was the reasoning for this? At first sight, it doesn't seem like an update to the groups should require a new resource. Why wasn't in-place updating possible?

PR #411, opened 2020-04-05
Add getter usage where missing

patrickmarabeas:
Resolves "invalid memory address or nil pointer dereference" errors that
were being encountered - specifically with the branch protection
resource.


patrickmarabeas:
@jcudit a general tidy and fix for one major bug we were encountering with v3 branch protection resource - if this could be slipped into the v2.7.0 release as well.

jcudit:
I've got this one queued.  Ideally this ships end of week.

jcudit:
👍 on the changes but I'm seeing some friction with these tests failing consistently:

```
=== CONT  TestAccGithubMembership_basic
--- FAIL: TestAccGithubMembership_basic (5.41s)
    testing.go:654: Step 1 error: ImportStateVerify attributes not equivalent. Difference is shown below. Top is actual, bottom is expected.

        (map[string]string) {
        }


        (map[string]string) (len=1) {
         (string) (len=4) "role": (string) (len=6) "member"
        }

FAIL
FAIL	github.com/terraform-providers/terraform-provider-github/github	5.673s
```
```
=== RUN   TestAccGithubMembership_caseInsensitive
--- FAIL: TestAccGithubMembership_caseInsensitive (7.23s)
    testing.go:654: Step 2 error: ImportStateVerify attributes not equivalent. Difference is shown below. Top is actual, bottom is expected.

        (map[string]string) {
        }


        (map[string]string) (len=1) {
         (string) (len=4) "role": (string) (len=6) "member"
        }

FAIL
FAIL	github.com/terraform-providers/terraform-provider-github/github	7.502s
```

Still investigating.

github/resource_github_membership.go
```diff
@@ -110,8 +110,8 @@ func resourceGithubMembershipRead(d *schema.ResourceData, meta interface{}) erro
 	}
 
 	d.Set("etag", resp.Header.Get("ETag"))
-	d.Set("username", membership.User.Login)
-	d.Set("role", membership.Role)
+	d.Set("username", membership.GetUser().GetLogin())
+	d.Set("role", membership.GetUser())
```
patrickmarabeas, line 114: Should be: `d.Set("role", membership.GetRole())`

patrickmarabeas:
Unsure if the second fixup commit is needed (as I was playing around with the acceptance test config itself at the same time and was running into state issues 😕) - but seemed unnecessary to get data already on hand.

patrickmarabeas:
@jcudit FYI `fixup!` commits: https://blog.sebastian-daschner.com/entries/git-commit-fixup-autosquash

Allows for ease of visibility on what has changed in the PR while allowing an easy trigger for squashing the fixes into their appropriate commits. The rebase should occur prior to merging.

PR #428, opened 2020-04-16
rebased - provider: Add the possibility to create repositories for non-organization accounts

anGie44:
Continuing the work done by @elislusarczyk in https://github.com/terraform-providers/terraform-provider-github/pull/96 to support individual github accounts for repos


anGie44:
Note: test failures pending https://github.com/terraform-providers/terraform-provider-github/pull/429

Issue #425, opened 2020-04-12
[Feature Request] Option to archive repository on destroy


pmiles01:
[Feature Request] Provide 'Releases' data sourceHi there,

Given the example terraform
```
resource "github_repository" "new_repo" {

  name        = "new_repo"
  description = "My new repo"
  private     = true
  template {
    owner      = "joeblogs"
    repository = "my-template"
  }
}
```

You can terraform apply and it will create the repository as expected, and when you do a terraform destroy, it will delete the repository.

I'd like a new flag, something like 'archive_on_destroy' which would archive but preserve the repository on destroy.

So the new terraform could look like :

```
resource "github_repository" "new_repo" {

  archive_on_destroy = true
  name        = "new_repo"
  description = "My new repo"
  private     = true
  template {
    owner      = "joeblogs"
    repository = "my-template"
  }
}
```

anGie44:
hi @pmiles01! I was looking into this more to see what's available in the current github API and it looks like on destroy, even if you had already updated the repo to set `archived=true`, the repo will not be preserved. would updating your config to set the `archived` param to `true` instead of calling`destroy` work in your use case?

pmiles01:
@anGie44 Thanks for looking at this request.

My particular use case is that I use Terraform to create a bunch of stuff, including git repositories, Google Cloud projects etc etc. The idea is that I use this Terraform every time our product team create a new product idea, and we need a space in which to work.

We use the destroy functionality of Terraform to mainly remove Google Cloud projects, service accounts etc, and ideally wanted to keep, but archive our Terraform created github repositories.

My work around at the moment is to use Terraform resource targeting to only destroy 'Google Cloud Project' related resources, and to specifically exclude my github module from the targeting. Personally I hate this approach, as it feels very wrong to use Terraform targeting.

Any thoughts or advice you have gratefully received.

pmiles01:
Once again, thanks for looking - excited to hear the wider communities thoughts ....

jcudit:
@pmiles01  I am still trying to understand this project's position on adding value.  For example, our underlying library presents the GitHub API to us with no added value and we layer on abstractions using the available primitives to add commonly used functionality.  I think its a useful feature but may only be applicable to users who share this preference.  Embracing targetting or archiving and removing the repository from state are other solutions as well, but I am not sure how common these workflows are used.  My vote goes to deferring the feature until it receives a few more 👍 s.

patrickmarabeas:
[...]

PR #432, opened 2020-04-18
[Feature Request] archive repo on destroy

anGie44:
* Adding optional field of `archive_on_destroy` to the `github_repository` schema per #425

### DIFF
--- a/.github/workflows/dotcom-acceptance-tests.yml
+++ b/.github/workflows/dotcom-acceptance-tests.yml
@@ -1,6 +1,7 @@
 name: Dotcom Acceptance Tests
 
 on:
+  push:
   pull_request:
     types: [opened, synchronize, reopened]
 
--- a/.github/workflows/dotcom-acceptance-tests.yml
+++ b/.github/workflows/dotcom-acceptance-tests.yml
@@ -1,7 +1,6 @@
 name: Dotcom Acceptance Tests
 
 on:
-  push:
   pull_request:
     types: [opened, synchronize, reopened]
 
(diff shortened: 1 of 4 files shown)


github/resource_github_repository.go
```diff
@@ -22,6 +22,7 @@ func resourceGithubRepository() *schema.Resource {
 		Importer: &schema.ResourceImporter{
 			State: func(d *schema.ResourceData, meta interface{}) ([]*schema.ResourceData, error) {
 				d.Set("auto_init", false)
+				d.Set("archive_on_destroy", false)
```
jcudit, line 25: If a repository is imported with configuration that specifies `archive_on_destroy` to be `true`, will `terraform` perform the import and then update the resource's preference in some subsequent step?  That seems like the probable flow, but wanted to ask in case you can confirm this behaviour.

anGie44: ah yes another `plan` and `apply` will be necessary after import as the API doesn't return this info at Read time; i'm now wondering what's a better way to approach this though.. set the `archive_on_destroy` to false here at import time, or don't explicitly set a value here? in either case there will be a non-empty plan after import

anGie44: ah or we could import by `repo_name:archive_on_destroy` instead 🤔

jcudit: @anGie44 I think the value this option provides outweighs the annoyance of a `plan` / `apply` step after an import.

Perhaps we can add a log statement during the import to explain this side effect rather than modifying the `import` ID.

jcudit:
Looks like recent updates to the branch protection plumbing has added conflicts to this PR.  I've got this queued to address this week, but welcome anyone else jumping in to remove conflicts.

.github/workflows/dotcom-acceptance-tests.yml
```diff
@@ -1,6 +1,7 @@
 name: Dotcom Acceptance Tests
 
 on:
+  push:
```
jcudit, line 4:
```suggestion
```

PR #435, opened 2020-04-21
Allow renaming a GitHub repository by removing ForceNew on name

thehunt33r:
This PR removes the `ForceNew` field for the `name` attribute in `resource/github_repository`, which enables renaming repositories via Terraform.

I have tested this change and it behaved as expected.


patrickmarabeas:
See: https://github.com/terraform-providers/terraform-provider-github/pull/65

emmasax4:
This would certainly be helpful! But until it's allowed, the workaround I use when I rename a repo (which is infrequent anyway), is to create a new module, do `terraform state rm` and then import the repository to the new module.

PR #440, opened 2020-04-28
Fix CI Failure For `make tools`

jcudit:
This proposes a fix to CI failures by aligning our method to obtain tools with how the AWS provider has evolved the process in https://github.com/terraform-providers/terraform-provider-aws/pull/8034.  We now `go install` the tooling using vendored go modules.

Fixes https://github.com/terraform-providers/terraform-provider-github/issues/439


jcudit:
@anGie44 thanks for making that PR visible.  I think its worth following the path that it paves eventually.  I'd like to push on merging this one as-is so other efforts can be unblocked.

anGie44:
@jcudit sure thing! agreed :)

PR #424, opened 2020-04-10
upgrade provider to go-github v31.0.0 to support new API methods 

anGie44:
This release of `go-github` will enable provider support of IdP groups for teams ; data and resource to be added in #400 

Changes include:
* import `go-github` references modified from v29 -> v31
* `#GetTeamMembership` -> `#GetTeamMembershipByID`
* `#ListTeamMembers` -> `#ListTeamMembersByID`

PR #441, opened 2020-05-01
Visibility

patrickmarabeas:
Adds the additional visibility parameter allowing for enterprise
accounts to set the repository to be only internally visible.

**NOTE: This is dependant on #424 and should be rebased on master once this has been merged.**

### DIFF
--- a/github/data_source_github_repository.go
+++ b/github/data_source_github_repository.go
@@ -37,6 +37,10 @@ func dataSourceGithubRepository() *schema.Resource {
 				Type:     schema.TypeBool,
 				Computed: true,
 			},
+			"visibility": {
+				Type:     schema.TypeString,
+				Computed: true,
+			},
 			"has_issues": {
 				Type:     schema.TypeBool,
 				Computed: true,
@@ -143,6 +147,7 @@ func dataSourceGithubRepositoryRead(d *schema.ResourceData, meta interface{}) er
 	d.Set("description", repo.GetDescription())
 	d.Set("homepage_url", repo.GetHomepage())
 	d.Set("private", repo.GetPrivate())
+	d.Set("visibility", repo.GetVisibility())
 	d.Set("has_issues", repo.GetHasIssues())
 	d.Set("has_wiki", repo.GetHasWiki())
 	d.Set("allow_merge_commit", repo.GetAllowMergeCommit())
--- a/github/resource_github_repository.go
+++ b/github/resource_github_repository.go
@@ -44,6 +44,12 @@ func resourceGithubRepository() *schema.Resource {
 				Type:     schema.TypeBool,
 				Optional: true,
 			},
+			"visibility": {
+				Type:         schema.TypeString,
+				Optional:     true,
+				Default:      "public",
+				ValidateFunc: validation.StringInSlice([]string{"public", "private", "internal"}, false),
+			},
 			"has_issues": {
 				Type:     schema.TypeBool,
 				Optional: true,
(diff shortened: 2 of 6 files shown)


github/resource_github_repository.go
```diff
@@ -44,6 +44,12 @@ func resourceGithubRepository() *schema.Resource {
 				Type:     schema.TypeBool,
 				Optional: true,
 			},
+			"visibility": {
+				Type:         schema.TypeBool,
+				Optional:     true,
+				Default:      "public",
+				ValidateFunc: validateValueFunc([]string{"public", "private", "internal"}),
```
anGie44, line 51: I think you could also use `validation.StringInSlice()` here though this works :)

github/resource_github_repository.go
```diff
@@ -319,7 +327,6 @@ func resourceGithubRepositoryRead(d *schema.ResourceData, meta interface{}) erro
 	d.Set("svn_url", repo.GetSVNURL())
 	d.Set("git_clone_url", repo.GetGitURL())
 	d.Set("http_clone_url", repo.GetCloneURL())
-	d.Set("archived", repo.GetArchived())
```
anGie44, line 322: I might be missing this in the resource but do we set the `archived` attribute in another place? Ahh or we don't need to re-set because it has a default value?

anGie44:
@patrickmarabeas this looks great! when you get a chance you can also now rebase to use the updated `go-gtihub` lib

github/resource_github_repository.go
patrickmarabeas: Yeah, sorry that looks completely unintentional! Thanks for spotting.

github/resource_github_repository.go
patrickmarabeas: Ah nice. I haven't browsed the validation methods at all - grabbed this one from another resource. This looks much more meaningful.

patrickmarabeas:
Comments resolved, and commits rebased @anGie44.

anGie44:
awesome, thanks for pushing up changes!  do you mind adding test coverage for the attribute as well?  forgot to mention it when I made my initial comments..otherwise changes LGTM 👍

patrickmarabeas:
@anGie44  I've included mention of the property in the tests, but not sure how much more can be done since testing is only done within public orgs... This looks to have the same degree of "testing" as the `private` property now. Any additional suggestions?

github/data_source_github_repository.go
```diff
@@ -37,6 +37,10 @@ func dataSourceGithubRepository() *schema.Resource {
 				Type:     schema.TypeBool,
 				Computed: true,
 			},
+			"visibility": {
+				Type:     schema.TypeBool,
```
anGie44, line 41: ah sorry I missed this! I believe we want `TypeString` here. the test files look fine, just would need to update this in the resource as well

patrickmarabeas: Welp.

patrickmarabeas:
@anGie44 

Added `visibility` to the Check functions.

I don't think testing that edge case should be necessary, as that's just testing the API, not the provider.

anGie44:
Note: rolling this back as the param will override the `private` attribute by default (even if omitted from config); i think we should consider overriding only if a user provides a value for `visibility` @patrickmarabeas -- similar to how the API treats this param

patrickmarabeas:
Ah true. Would removing the default be enough? That should result in a nil setting.

patrickmarabeas:
@anGie44 see #454

PR #444, opened 2020-05-03
Add vulnerability_alerts attribute for repositories

jtsaito:
Add support for new attribute `vulnerability_alerts` for `github_repository`.  GitHub calls this feature as `vulnearability-alerts` in the API, `Security Alerts` in the web console, and "security alerts for vulnerable dependencies" in the documentation.

The PR follows up on https://github.com/terraform-providers/terraform-provider-github/pull/382 and addresses https://github.com/terraform-providers/terraform-provider-github/issues/238.

This would introduce changes for some existing repos with vulnerabilities alerts enabled and show a diff accordingly. To address fix the diff, adding `vulnerability_alerts = true` suffices.

What's slightly weird is that by default GitHub enables this feature on public repos but disables it on private repos.

**References**
- [API for vulnerability-alerts](https://developer.github.com/v3/repos/)
- [Feature documentation](https://help.github.com/en/github/managing-security-vulnerabilities/about-security-alerts-for-vulnerable-dependencies)

**Acc tests**
I ran `make testacc` but only for the new and the basic repo test.

```
TF_ACC=1 go test github.com/terraform-providers/terraform-provider-github/github -v -run=TestAccGithubRepository_vulnerabilityAlerts -timeout 120m
=== RUN   TestAccGithubRepository_vulnerabilityAlerts
=== PAUSE TestAccGithubRepository_vulnerabilityAlerts
=== CONT  TestAccGithubRepository_vulnerabilityAlerts
--- PASS: TestAccGithubRepository_vulnerabilityAlerts (16.84s)
PASS
ok  	github.com/terraform-providers/terraform-provider-github/github	16.870s

TF_ACC=1 go test github.com/terraform-providers/terraform-provider-github/github -v -run=TestAccGithubRepository_basic -timeout 120m
=== RUN   TestAccGithubRepository_basic
=== PAUSE TestAccGithubRepository_basic
=== CONT  TestAccGithubRepository_basic
--- PASS: TestAccGithubRepository_basic (16.53s)
PASS
ok  	github.com/terraform-providers/terraform-provider-github/github	16.555s
```

Also manually tested for private repos and that worked as well.


majormoses:
Yay I am excited to see this, its something I have been looking forward to adding to our internal module for a while and I am currently relying on manually managing these on the core repos in my org. Clearly far from ideal as we know there are new projects being created daily.

I think we should consider having this default with a `null` value and only in the case where the passed value is a boolean we add this to the request. This would effectively maintains Github's current defaults and only looks to update when explicitly told to manage this. In both cases defaulting to `true` or `false` are technically both breaking changes IMO as they realistically require you to take a stance to maintain existing "status quo" updates. Either way the expectation for what should happen will have changed. If we decide we want to always control the default then I think we need to be clear about calling out the impact for users and ensure this is properly versioned as a major release. We can also release with a `null` default now and later take the stance on if we should manage the default for this or not. 

I would prefer we follow the upstream defaults and encourage organizations to have their own module encapsulating their own opinions, standards, etc.

majormoses:
🤔 I just realized we currently have documented that we support terraform `0.10+` and `null` was only added in `0.12` so technically that would be breaking as well. I suppose we could treat an empty string as the same thing as `null` to provide that type of functionality for older terraform? We might also consider dropping `< 0.12` in a major version which appears to have already started happening in other core modules: https://www.hashicorp.com/blog/deprecating-terraform-0-11-support-in-terraform-providers/ I would appreciate a maintainer to weigh in when we should look at doing the same as this may affect what we want to do.

anGie44:
hi @jtsaito, should we add this field to the data source as well?

jtsaito:
@majormoses Thank you for your feedback!

Regarding the `null` value, the [documentation on Terraform expressions](https://www.terraform.io/docs/configuration/expressions.html#null) states:

> If you set an argument of a resource or module to null, Terraform behaves as though you had completely omitted it — it will use the argument's default value if it has one, or raise an error if the argument is mandatory.

As long as the attribute is a boolean the default is either `true` or `false`. I think that a boolean is the correct choice here because it reflects exactly the API.

Forcing users to set the value according to the diff is actually helpful because they can be sure their infrastructure as code reflects the state of the API. This precludes ambiguity about the click-opsed state. Setting the actual value based on the diff requires minimal effort even for hundreds of repos a simple script or editor macro should do the job.

The default `false` seems natural assuming the most common use case is private repositories and disabled alerts.

Please llet me know what you think.

jtsaito:
@anGie44 Sure, if this PR get's accepted I will gladly create another adding the attribute for the datasource!

github/resource_github_repository.go
```diff
@@ -374,6 +405,18 @@ func resourceGithubRepositoryUpdate(d *schema.ResourceData, meta interface{}) er
 		}
 	}
 
+	if d.HasChange("vulnerability_alerts") {
```
anGie44, line 408: my only suggestion here could be to save an API call after creation of a resource by adding the`!d.IsNewResource()` since `Update` immediately follows `Create`

github/resource_github_repository.go
```diff
@@ -118,6 +118,11 @@ func resourceGithubRepository() *schema.Resource {
 					ValidateFunc: validation.StringMatch(regexp.MustCompile(`^[a-z0-9][a-z0-9-]*$`), "must include only lowercase alphanumeric characters or hyphens and cannot start with a hyphen"),
 				},
 			},
+			"vulnerability_alerts": {
+				Type:     schema.TypeBool,
+				Optional: true,
+				Default:  false,
```
anGie44, line 124: my only concern here is, as stated, this default value is only applicable to `private` repos, and a user may unintentionally disable alerts for a `public` repo by omitting this attribute in a config and making it through the `apply` step..could we consider leaving this w/o a default or is the general user experience better with having this value in place?

github/resource_github_repository_test.go
```diff
@@ -557,6 +631,33 @@ func testAccCheckGithubRepositoryTemplateRepoAttribute(n string, repo *github.Re
 	}
 }
 
+func testAccCheckGithubVulnerabilityAlerts(n string, expected bool) resource.TestCheckFunc {
+	return func(s *terraform.State) error {
+		rs, ok := s.RootModule().Resources[n]
```
anGie44, line 636: it may be redundant to check w/in the state here as a call to this function in the acceptance tests is preceded by `testAccCheckGithubRepositoryExists` which checks this for us, so the logic here can be omitted similar to`testAccCheckGithubRepositoryAttributes`

github/resource_github_repository_test.go
```diff
@@ -965,3 +1066,53 @@ resource "github_branch_protection" "repo_name_master" {
...
+  has_issues         = true
+  has_wiki           = true
+  is_template        = false
+  allow_merge_commit = true
+  allow_squash_merge = false
+  allow_rebase_merge = false
+  has_downloads      = true
+  auto_init          = false
+
+	vulnerability_alerts = true
```
anGie44, line 1090: Nit: can be aligned by using `terrafmt`

github/resource_github_repository_test.go
```diff
@@ -965,3 +1066,53 @@ resource "github_branch_protection" "repo_name_master" {
...
+  has_issues         = true
+  has_wiki           = true
+  is_template        = false
+  allow_merge_commit = true
+  allow_squash_merge = false
+  allow_rebase_merge = false
+  has_downloads      = true
+  auto_init          = false
+
+	vulnerability_alerts = false
```
anGie44, line 1115: same nit as above

majormoses:
@jtsaito sorry for the delayed response. AFAIK we can use `null`, set the `type` to `any`, and can raise an exception if its `!nul`, `true`, or `false`. My opinion then if we must take a stance on default we should default to `true` rather than `false` as it will benefit more people with better secure defaults. If you don't trust github/microsoft you should probably move to another platform. There is an inherent trust and a server admin can 100% get your data if they really wanted to regardless of this setting. As both ways are opinions and I didn't want to get into an argument about it (as there is no winner) by keeping the upstream behavior.

github/resource_github_repository.go
jcudit: > could we consider leaving this w/o a default

👍 my vote goes to removing the default.

jtsaito:
@anGie44 Thanks for the review! Sorry, I did not get around to fixing this earlier. I'll have time to fix this by the end of the week and will remove the default.

jtsaito:
@anGie44 Sorry again for the delay. I hope to now have incorporated all your corrections and suggestions. Please check the latest commits and let me know if I should squash all commits into one.

website/docs/r/repository.html.markdown
```diff
@@ -78,7 +78,7 @@ initial repository creation and create the target branch inside of the repositor
 
 * `template` - (Optional) Use a template repository to create this resource. See [Template Repositories](#template-repositories) below for details.
 
-* `vulnerability_alerts` - (Optional) Set to `true` to enable security alerts for vulnerable dependencies. Is `false` by default. Enabling requires alerts to be enabled on the owner level. (Note for importing: GitHub enables the alerts on public repos but disables them on private repos by default.) See [GitHub Documentation](https://help.github.com/en/github/managing-security-vulnerabilities/about-security-alerts-for-vulnerable-dependencies) for details. 
+* `vulnerability_alerts` - Set to `true` to enable security alerts for vulnerable dependencies. Enabling requires alerts to be enabled on the owner level. (Note for importing: GitHub enables the alerts on public repos but disables them on private repos by default.) See [GitHub Documentation](https://help.github.com/en/github/managing-security-vulnerabilities/about-security-alerts-for-vulnerable-dependencies) for details.
```
anGie44, line 81: do you mind adding back the `(Optional)` in the description?

github/resource_github_repository_test.go
```diff
@@ -83,6 +83,7 @@ func TestAccGithubRepository_basic(t *testing.T) {
 						DefaultBranch:       "master",
 						Archived:            false,
 					}),
+					testAccCheckGithubVulnerabilityAlerts(rn, false),
```
anGie44, line 86: in calls to `testAccCheckGithubVulnerabilityAlerts`, we'll want to send the repo name (in this case stored in the `repo` variable) as needed for the github API method; alternatively you could pass the whole `repo` object and extract the name inside the `testAccCheckGithubVulnerabilityAlerts` method

github/resource_github_repository_test.go
```diff
@@ -101,6 +102,7 @@ func TestAccGithubRepository_basic(t *testing.T) {
 						HasProjects:      false,
 						Archived:         false,
 					}),
+					testAccCheckGithubVulnerabilityAlerts(rn, false),
```
anGie44, line 105: similar comment as above

anGie44:
I wanted to also note @jtsaito, this feature request https://github.com/terraform-providers/terraform-provider-github/issues/390 brought to light a side-effect of enabling vulnerability alerts via the API as the dependency_graph is also enabled (https://developer.github.com/v3/repos/#enable-vulnerability-alerts). In this case, I think we should document this in website docs as well.

dkatona:
Hi @jtsaito, we are eagerly waiting for this feature to turn on security checking via terraform - is there any way we can help you to finish this PR and get it merged? :) Thank you!

jcudit:
Looks like recent updates to the branch protection plumbing has added conflicts to this PR.  I've got this queued to address this week, but welcome anyone else jumping in to remove conflicts.

github/resource_github_repository_test.go
```diff
@@ -524,4 +524,4 @@ func init() {
 		Name: "github_repository",
 		F:    testSweepRepositories,
 	})
-}
+}
```
jcudit, line 527:
```suggestion
}

```

website/docs/r/repository.html.markdown
jcudit:
```suggestion
* `vulnerability_alerts` (Optional) - Set to `true` to enable security alerts for vulnerable dependencies. Enabling requires alerts to be enabled on the owner level. (Note for importing: GitHub enables the alerts on public repos but disables them on private repos by default.) See [GitHub Documentation](https://help.github.com/en/github/managing-security-vulnerabilities/about-security-alerts-for-vulnerable-dependencies) for details. 
```

jcudit:
Conflicting tests have been removed and will be re-added in a follow-up commit.

PR #449, opened 2020-05-08
fix git release tag url

smiller171:
While Trying to use the releases functionality I discovered that the release tag url is being built incorrectly, making it impossible to get a release by tag. This PR fixes that mistake

### DIFF
--- a/vendor/github.com/google/go-github/v31/github/repos_releases.go
+++ b/vendor/github.com/google/go-github/v31/github/repos_releases.go
@@ -109,7 +109,7 @@ func (s *RepositoriesService) GetLatestRelease(ctx context.Context, owner, repo 
 //
 // GitHub API docs: https://developer.github.com/v3/repos/releases/#get-a-release-by-tag-name
 func (s *RepositoriesService) GetReleaseByTag(ctx context.Context, owner, repo, tag string) (*RepositoryRelease, *Response, error) {
-	u := fmt.Sprintf("repos/%s/%s/releases/tags/%s", owner, repo, tag)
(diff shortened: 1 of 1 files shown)


anGie44:
hi @smiller171, thanks for submitting this PR! how did you go about using the release functionality? do you mind providing error logs and/or your config to reproduce?

i should note, since the change directly affects a file in the `go-github` library, you would need to submit the PR directly to the `go-github` [project](https://github.com/google/go-github) as within this provider, we directly pull a specific version of the package and a change here will not persist

smiller171:
This is my code:

```
data "github_release" "this" {
  repository  = "redacted"
  owner       = "redacted"
  retrieve_by = "tag"
  release_tag = var.github_release_tag
}

output "github_release_assets_url" {
  value = data.github_release.this.asserts_url
}
```

Error message:
```
Error: GET https://api.github.com/repos/redacted/redacted/releases/tags/master-20200505-updated7: 404 Not Found []

  on lambda.tf line 83, in data "github_release" "this":
  83: data "github_release" "this" {
```

I checked the URL and it is in fact a 404. The correct URL is:
`https://api.github.com/repos/redacted/redacted/releases/tag/master-20200505-updated7`

The only difference being `.../releases/tag/...` works and `.../releases/tags/...` does not

smiller171:
Why is vendor code being committed instead of pulled at build time?

smiller171:
@anGie44 Since this is an upstream issue I'm going to close this PR and open an issue instead. I've opened a PR against the upstream package. https://github.com/google/go-github/pull/1520

PR #452, opened 2020-05-10
Revert "Visibility"

anGie44:
Reverts terraform-providers/terraform-provider-github#441
* introduces a breaking change (default value of false will be applied and override `private` param if  `visibility` is omitted in config)


harlowja:
This is scary :(

harlowja:
Thanks for the revert. eck...

PR #454, opened 2020-05-11
Add visibility parameter to data source and resource

### LINKED ISSUE #304: Support "internal" repository types
Hi,

Currently the provider only supports `private = true`, but there is now a new type called `internal`.
I think this should now be switched to a new variable called `visibility` with three possible values `public`,`private` and `internal`

References:
https://www.terraform.io/docs/providers/github/r/repository.html#private
https://help.github.com/en/github/creating-cloning-and-archiving-repositories/creating-an-internal-repository

patrickmarabeas:
Adds the additional visibility parameter allowing for enterprise
accounts to set the repository to be only internally visible.

Fixes issues within #441 
* Removes default being set
* Removes inclusion of the parameter within unit tests as it will have a nil value

Resolves #304

### DIFF
--- a/github/resource_github_repository.go
+++ b/github/resource_github_repository.go
@@ -41,8 +41,17 @@ func resourceGithubRepository() *schema.Resource {
 				Optional: true,
 			},
 			"private": {
-				Type:     schema.TypeBool,
-				Optional: true,
+				Type:          schema.TypeBool,
+				Computed:      true, // is affected by "visibility"
+				Optional:      true,
+				ConflictsWith: []string{"visibility"},
+				Deprecated:    "use visibility instead",
+			},
+			"visibility": {
+				Type:         schema.TypeString,
+				Optional:     true,
+				Computed:     true, // is affected by "private"
+				ValidateFunc: validation.StringInSlice([]string{"public", "private", "internal"}, false),
 			},
 			"has_issues": {
 				Type:     schema.TypeBool,
@@ -178,6 +187,7 @@ func resourceGithubRepositoryObject(d *schema.ResourceData) *github.Repository {
 		Description:         github.String(d.Get("description").(string)),
 		Homepage:            github.String(d.Get("homepage_url").(string)),
 		Private:             github.Bool(d.Get("private").(bool)),
+		Visibility:          github.String(d.Get("visibility").(string)),
 		HasDownloads:        github.Bool(d.Get("has_downloads").(bool)),
 		HasIssues:           github.Bool(d.Get("has_issues").(bool)),
 		HasProjects:         github.Bool(d.Get("has_projects").(bool)),
@@ -204,6 +214,12 @@ func resourceGithubRepositoryCreate(d *schema.ResourceData, meta interface{}) er
 
 	repoReq := resourceGithubRepositoryObject(d)
 	owner := meta.(*Owner).name
+
+	// Auth issues (403 You need admin access to the organization before adding a repository to it.)
+	// are encountered when the resources is created with the visibility parameter. As
+	// resourceGithubRepositoryUpdate is called immediately after, this is subsequently corrected.
+	repoReq.Visibility = nil
+
 	repoName := repoReq.GetName()
 	ctx := context.Background()
 
@@ -300,6 +316,7 @@ func resourceGithubRepositoryRead(d *schema.ResourceData, meta interface{}) erro
 	d.Set("description", repo.GetDescription())
 	d.Set("homepage_url", repo.GetHomepage())
 	d.Set("private", repo.GetPrivate())
+	d.Set("visibility", repo.GetVisibility())
 	d.Set("has_issues", repo.GetHasIssues())
 	d.Set("has_projects", repo.GetHasProjects())
 	d.Set("has_wiki", repo.GetHasWiki())
@@ -338,6 +355,12 @@ func resourceGithubRepositoryUpdate(d *schema.ResourceData, meta interface{}) er
 	client := meta.(*Owner).v3client
 
 	repoReq := resourceGithubRepositoryObject(d)
+
+	// The endpoint will throw an error if trying to PATCH with a visibility value that is the same
+	if !d.HasChange("visibility") {
+		repoReq.Visibility = nil
+	}
+
 	// Can only set `default_branch` on an already created repository with the target branches ref already in-place
 	if v, ok := d.GetOk("default_branch"); ok {
 		branch := v.(string)
--- a/website/docs/r/repository.html.markdown
+++ b/website/docs/r/repository.html.markdown
@@ -38,6 +38,8 @@ The following arguments are supported:
 
 * `private` - (Optional) Set to `true` to create a private repository.
   Repositories are created as public (e.g. open source) by default.
+  
+* `visibility` - (Optional) Can be `public` or `private`. If your organization is associated with an enterprise account using GitHub Enterprise Cloud or GitHub Enterprise Server 2.20+, visibility can also be `internal`. The `visibility` parameter overrides the `private` parameter.
 
 * `has_issues` - (Optional) Set to `true` to enable the GitHub Issues features
   on the repository.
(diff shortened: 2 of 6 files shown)


patrickmarabeas:
Definitely something funky going on.

github/resource_github_repository.go
```diff
@@ -179,6 +184,7 @@ func resourceGithubRepositoryObject(d *schema.ResourceData) *github.Repository {
 		Description:         github.String(d.Get("description").(string)),
 		Homepage:            github.String(d.Get("homepage_url").(string)),
 		Private:             github.Bool(d.Get("private").(bool)),
+		Visibility:          github.String(d.Get("visibility").(string)),
```
anGie44, line 187: from running the acceptance tests, it seems passing an empty string here causes a strange 403 error about not having admin rights..in this case, it may make more sense to pull out this attribute and only set it if the string's available

github/resource_github_repository.go
```diff
@@ -303,6 +309,7 @@ func resourceGithubRepositoryRead(d *schema.ResourceData, meta interface{}) erro
 	d.Set("description", repo.GetDescription())
 	d.Set("homepage_url", repo.GetHomepage())
 	d.Set("private", repo.GetPrivate())
+	d.Set("visibility", repo.GetVisibility())
```
anGie44, line 312: in cases where `visibility` isn't provided, looks like we'll be left with a non-empty plan as the API returns one of the 3 visibility type values. we can account for this if we make the param also `Computed`

anGie44:
@patrickmarabeas as a sanity check, we should also add a test case that uses a config w/`private=true`

patrickmarabeas:
So there's an issue with the Github API itself. Basically you cannot patch the endpoint with a visibility value that is the same. I don't know if this is a usable parameter for Terraform currently...

`curl -X PATCH https://api.github.com/repos/XXX/test -H 'Content-Type: application/json' -H 'Accept: application/vnd.github.nebula-preview+json' -d '{ "visibility": "private" }'`

```
{
  ...
  "visibility": "private",
  ...
}
```

`curl -X PATCH https://api.github.com/repos/XXX/test -H 'Content-Type: application/json' -H 'Accept: application/vnd.github.nebula-preview+json' -d '{ "visibility": "private" }'`

```
{
  "message": "Visibility is already private.",
  "documentation_url": "https://github.com/pricing"
}
```

`curl -X PATCH https://api.github.com/repos/XXX/test -H 'Content-Type: application/json' -H 'Accept: application/vnd.github.nebula-preview+json' -d '{ "visibility": "internal" }'`

```
{
  ...
  "visibility": "internal",
  ...
}
```

`curl -X PATCH https://api.github.com/repos/XXX/test -H 'Content-Type: application/json' -H 'Accept: application/vnd.github.nebula-preview+json' -d '{ "visibility": "internal" }'`

```
{
  "message": "Visibility is already internal.",
  "documentation_url": "https://github.com/pricing"
}
```

anGie44:
ah interesting, I wonder if for some reason the `go-github` lib handles these requests differently? I ran the `TestAccGithubRepository_basic` but edited the first step's config to set `visibility="public"` as well as in the update step  and the test didn't return an error, but given the results in your above comment, it should've? not sure if this solves the whole problem, but in the `Update` method of the resource, we could guard against this by only adding the `visibility` attribute to the `repoReq` object when there is evidence of change via the `d.HasChange()` method

patrickmarabeas:
Same issues with `go-github`.

Just ran this twice:

```
...
payload := &github.Repository{
  Visibility: github.String("internal"),
}
repo, _, err := client.Repositories.Edit(ctx, "XXX", "test", payload)
...
```

I'll try `d.HasChange`.

anGie44:
> I'll try `d.HasChange`.

you may have to chain this condition with `!d.IsNewResource()` as well since `Update` gets called  right after `Create`

patrickmarabeas:
So there is some naff auth stuff happening with that endpoint. Wish I had taken more notice of your first comment. This occurs when creating a repository with the `visibility` param. I've removed the value for create, with the update subsequently correcting for it.

Looks like it is now behaving as expected:
* Can create without `visibility` set
* Can create with `private` set and no `visibility`
* Creating with `visibility` and `private` will ignore `private` and set `visibility` correctly
* Can update the `visibility` value, or replace with `private`
* Can delete

patrickmarabeas:
Issue logged: https://github.community/t5/GitHub-API-Development-and/v3-repository-visibility-parameter-issues/m-p/57921

patrickmarabeas:
Still needs work, plans are still detecting diffs each run:

```
- visibility             = "private" -> null
```

anGie44:
> It seems once you start using `visibility`, you can't go back to using `private`. Looking at funnelling both params into `visibility`.

😬 yikes, turning out to be more thorny then i expected..i can help investigate on my end as well

patrickmarabeas:
OK - it's just the funkiness of `private` equaling `true` when `visibility` is both `internal` or `private` - so going from `visibility: "internal"` to using `private: true` causes no change. I'm whacking a deprecated sticker on `private`.

patrickmarabeas:
@anGie44 give it a whirl and see what you think.

website/docs/r/repository.html.markdown
```diff
@@ -41,6 +41,10 @@ The following arguments are supported:
 
 * `private` - (Optional) Set to `true` to create a private repository.
   Repositories are created as public (e.g. open source) by default.
+  
+* `visibility` - (Optional) Can be `public` or `private`. If your organization is associated with an enterprise account 
+using GitHub Enterprise Cloud or GitHub Enterprise Server 2.20+, visibility can also be `internal`. The `visibility` 
+parameter overrides the `private` parameter.
```
jcudit, line 47: Can we squash these all into a single line?

```suggestion
* `visibility` - (Optional) Can be `public` or `private`. If your organization is associated with an enterprise account using GitHub Enterprise Cloud or GitHub Enterprise Server 2.20+, visibility can also be `internal`. The `visibility` parameter overrides the `private` parameter.
```

patrickmarabeas: Single line'd

patrickmarabeas:
Ah yes - `visibility` mimics `private` when not set. Latest fixup! should resolve these errors.

@jcudit  Is it possible to run a subset of acceptance tests?

jcudit:
Adding a label to a PR prefixed with `acceptance-test/` and suffixed with the beginning of an acceptance test name will kick off a run.  ☝️ adds the `accceptance-test/TestAccGithubRepository` label which will run the subset of tests starting with the `TestAccGithubRepository` prefix.

patrickmarabeas:
I meant locally - I think Github Actions will have limitations around forks

jcudit:
Ah, locally tests can be invoked via:
```console
TF_ACC=1 go test -v -timeout 30m  ./... -run TestProvider_insecure
```

The [script](https://github.com/terraformtesting/acceptance-tests/blob/master/entrypoint.sh) that runs the automated tests is locally runnable as well.

jcudit:
@patrickmarabeas hare are a pair of commits available to cherry-pick that will get acceptance tests passing:

- https://github.com/terraform-providers/terraform-provider-github/commit/2f2bc1857ede67c100b8476fc6c6bcfd3e7136d6
- https://github.com/terraform-providers/terraform-provider-github/commit/d74cf6ca111c9a1b5c40f0caa8a4ef0925bc5fff

patrickmarabeas:
@jcudit commits cherry-picked and PR rebased

jcudit:
@patrickmarabeas this looks ready to go, however we hit a conflict.  Can you take a last look at this one? This PR is next up to merge and will close out the v2.9.0 milestone.

patrickmarabeas:
@jcudit rebased.

github/resource_github_repository.go
jcudit: Aiming to perform a follow-up pass to address this.  Thanks for raising this.

github/resource_github_repository.go
jcudit: This too is queued to be addressed by a subsequent pass.

PR #459, opened 2020-05-14
Add Overwrite Support For `github_repository_file`

jcudit:
Fixes https://github.com/terraform-providers/terraform-provider-github/issues/438

/cc @techdragon @shoekstra @cornfeedhobo

### DIFF
--- a/github/resource_github_repository_file_test.go
+++ b/github/resource_github_repository_file_test.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"log"
 	"os"
+	"regexp"
 	"strings"
 
 	"encoding/base64"
@@ -302,8 +303,10 @@ func TestAccGithubRepositoryFile_overwrite(t *testing.T) {
 	var commit github.RepositoryCommit
 
 	rn := "github_repository_file.foo"
+	path := ".gitignore"
+	fileContent := "**/.terraform/*"
 	randString := acctest.RandStringFromCharSet(10, acctest.CharSetAlphaNum)
-	path := fmt.Sprintf("tf-acc-test-file-%s", randString)
+	repo := fmt.Sprintf("tf-acc-test-%s", randString)
 
 	resource.Test(t, resource.TestCase{
 		PreCheck:     func() { testAccPreCheck(t) },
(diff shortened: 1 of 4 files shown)


github/resource_github_repository_file_test.go
```diff
@@ -297,7 +298,46 @@ func TestAccGithubRepositoryFile_committer(t *testing.T) {
...
+		},
+	})
+}
+
 func testAccCheckGithubRepositoryFileExists(n, path, branch string, content *github.RepositoryContent, commit *github.RepositoryCommit) resource.TestCheckFunc {
+	hardCodedRepoName := "test-repo"
+	return testAccCheckGithubRepositoryFileExistsWithRepo(n, path, branch, hardCodedRepoName, content, commit)
+}
+
+func testAccCheckGithubRepositoryFileExistsWithRepo(n, path, branch, repo string, content *github.RepositoryContent, commit *github.RepositoryCommit) resource.TestCheckFunc {
```
jcudit, line 340: I went down the path of removing references to "test-repo" and leveraging a per-test generated repository instead.  Those changes turned out to break tests so I've opted to leave the existing code paths alone and expose `testAccCheckGithubRepositoryFileExistsWithRepo` instead.  This allows for creating a repository from a template and overwriting its files during the test suite.

jcudit:
Thanks for raising discussion on the update case.  I had not given it too much thought.

After thinking about this a bit, I recommend we rebrand `overwrite` to `overwrite_on_create`.  
We can then assume that whenever the update code path is executed, we will _always_ be overwriting a file, which matches the existing implementation.

To further safeguard, we can add the `ForceNew` attribute to `overwrite_on_create` so that if configuration changes, we can start fresh with a new file.  What do you think @anGie44 ?

jcudit:
q

jaxxstorm:
Is there any news on when this might be reviewed/merged?

anGie44:
> I recommend we rebrand `overwrite` to `overwrite_on_create`.
I like this idea, suits the intended behavior 👍 my only hesitation would be adding the `ForceNew` attribute since the field is intended to be used only at `create` time and it might be more of a pain point to force recreation for users

jcudit:
Tests are [looking good](https://github.com/terraform-providers/terraform-provider-github/runs/1203794419).  The GHES ones are failing as the infrastructure is down at the moment.

PR #460, opened 2020-05-15
Add Project Card Support

jcudit:
Fixes https://github.com/terraform-providers/terraform-provider-github/issues/433

### DIFF
--- a/github/resource_github_project_card.go
+++ b/github/resource_github_project_card.go
@@ -1,0 +1,160 @@
+package github
+
+import (
+	"context"
+	"log"
+	"net/http"
+	"strconv"
+	"strings"
+
+	"github.com/google/go-github/v32/github"
+	"github.com/hashicorp/terraform-plugin-sdk/helper/schema"
+)
+
+func resourceGithubProjectCard() *schema.Resource {
+	return &schema.Resource{
+		Create: resourceGithubProjectCardCreate,
+		Read:   resourceGithubProjectCardRead,
+		Update: resourceGithubProjectCardUpdate,
+		Delete: resourceGithubProjectCardDelete,
+		Importer: &schema.ResourceImporter{
+			State: resourceGithubProjectCardImport,
+		},
+		Schema: map[string]*schema.Schema{
+			"column_id": {
+				Type:     schema.TypeString,
+				Required: true,
+				ForceNew: true,
+			},
+			"note": {
+				Type:     schema.TypeString,
+				Required: true,
+			},
+			"etag": {
+				Type:     schema.TypeString,
+				Computed: true,
+			},
+			"card_id": {
+				Type:     schema.TypeInt,
+				Computed: true,
+			},
+		},
+	}
+}
+
+func resourceGithubProjectCardCreate(d *schema.ResourceData, meta interface{}) error {
+	err := checkOrganization(meta)
+	if err != nil {
+		return err
+	}
+
+	columnIDStr := d.Get("column_id").(string)
+	columnID, err := strconv.ParseInt(columnIDStr, 10, 64)
+	if err != nil {
+		return unconvertibleIdErr(columnIDStr, err)
+	}
+
+	log.Printf("[DEBUG] Creating project card note in column ID: %d", columnID)
+	client := meta.(*Owner).v3client
+	options := github.ProjectCardOptions{Note: d.Get("note").(string)}
+	ctx := context.Background()
+	card, _, err := client.Projects.CreateProjectCard(ctx, columnID, &options)
+	if err != nil {
+		return err
+	}
+
+	d.Set("card_id", card.GetID())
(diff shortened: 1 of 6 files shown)


github/resource_github_project_card.go
```diff
@@ -0,0 +1,140 @@
...
+)
+
+func resourceGithubProjectCard() *schema.Resource {
+	return &schema.Resource{
+		Create: resourceGithubProjectCardCreate,
+		Read:   resourceGithubProjectCardRead,
+		Update: resourceGithubProjectCardUpdate,
+		Delete: resourceGithubProjectCardDelete,
+		Importer: &schema.ResourceImporter{
+			State: schema.ImportStatePassthrough,
```
anGie44, line 21: we'll need a custom import fn to set the `card_id`

github/resource_github_project_card_test.go
```diff
@@ -0,0 +1,129 @@
...
+			},
+			{
+				Config: testAccGithubProjectCardConfig(updatedNote),
+				Check: resource.ComposeTestCheckFunc(
+					testAccCheckGithubProjectCardExists(rn, &card),
+					testAccCheckGithubProjectCardAttributes(&card, &testAccGithubProjectCardExpectedAttributes{
+						Note: updatedNote,
+					}),
+				),
+			},
```
anGie44, line 43: we'll need an ImportStep here to verify but I believe the resource's Importer func will need an update as well

jcudit:
The following acceptance test failure was introduced:
```
--- FAIL: TestAccGithubProjectCard_basic (13.38s)
    testing.go:654: Step 2 error: ImportStateVerify attributes not equivalent. Difference is shown below. Top is actual, bottom is expected.
        
        (map[string]string) {
        }
        
        
        (map[string]string) (len=1) {
         (string) (len=9) "column_id": (string) (len=7) "9220014"
        }
        
FAIL
FAIL	github.com/terraform-providers/terraform-provider-github/github	13.397s
FAIL
```

Not sure if this is a good candidate for `ImportStateVerifyIgnore` or if I have done something wrong with the import configuration.

jcudit:
@anGie44 this is ready for review again after addressing your feedback.  I ended up using `ImportStateVerifyIgnore` to work around [this error](https://github.com/terraform-providers/terraform-provider-github/pull/460#issuecomment-629659922).  Let me know if you have any other ideas.

github/resource_github_project_card.go
```diff
@@ -0,0 +1,161 @@
...
+	// FIXME: Remove URL parsing if a better option becomes available
+	columnURL := card.GetColumnURL()
+	columnIDStr := strings.TrimPrefix(columnURL, client.BaseURL.String()+`projects/columns/`)
+	columnID, err := strconv.ParseInt(columnIDStr, 10, 64)
+	if err != nil {
+		return unconvertibleIdErr(columnIDStr, err)
+	}
+
+	d.Set("note", card.GetNote())
+	d.Set("column_id", columnID)
```
anGie44, line 103: looks like this assignment is where the import step is failing b/c here `columnID` is an int64 while the schema is expecting a String so the field doesn't get populated.if you pass the string from L96, it should work as expected

github/resource_github_project_card.go
```diff
@@ -0,0 +1,161 @@
...
+	cardID := d.Get("card_id").(int)
+	_, err := client.Projects.DeleteProjectCard(ctx, int64(cardID))
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func resourceGithubProjectCardImport(d *schema.ResourceData, meta interface{}) ([]*schema.ResourceData, error) {
```
anGie44, line 140: this works as expected but i'm wondering if by adding import support to this resource, we could make things a little easier by setting the resource's ID to the card_id (thereby removing the `card_id` field from the schema) and set the `node_id` as it's own field in the schema? that way the Importer func would just be a `ImportStatePassThrough` and the `id` can be an exported attribute

anGie44: ah ignore above comment -- definitely seems useful to store the node_id to help transition between v3 and v4 of the github API

github/resource_github_project_card.go
```diff
@@ -0,0 +1,161 @@
...
+	ctx := context.Background()
+	card, _, err := client.Projects.GetProjectCard(ctx, cardID)
+	if card == nil || err != nil {
+		return []*schema.ResourceData{d}, err
+	}
+
+	d.SetId(card.GetNodeID())
+	d.Set("card_id", cardID)
+
+	return []*schema.ResourceData{d}, resourceGithubProjectCardRead(d, meta)
```
anGie44, line 159: i think usually in these import funcs, the final return just needs the resource data and an explicit `nil` error as the `Read` method will implicitly be called

jcudit:
Rebased this and added passing tests.

PR #96, opened 2018-05-19
provider: Add the possibility to create repositories for non-organization accounts

elislusarczyk:
Fixed the ambiguity regarding the owner parameter name when calling the go-github api and added the possibility for the individual accounts to create repositories.

Here is the 404 error when an individual account tries to create a repository.

Github REST Api resource `/orgs/eliaszs/repos` doesn't exist for user's account.

```
$ TF_LOG=DEBUG terraform apply -target module.github -auto-approve

...

Error: Error applying plan:

1 error(s) occurred:

* module.github.github_repository.hello-stocks: 1 error(s) occurred:

* github_repository.hello-stocks: POST https://api.github.com/orgs/eliaszs/repos: 404 Not Found []
```

The PR is only breaking the compatibility with the env variable name. The `organization` structure has been renamed to `owner` together with the GITHUB_ORGANIZATION -> GITHUB_OWNER. It is now compatible with the go-github api in the most cases. 

According to the spec of this function the `org string` argument is really an owner and not the organization, as it is used in other function across go-github library.

Here is the ambiguous case for the `Create` and below is the correct one for the `Get` function in the go-github library.
```
// Create a new repository. If an organization is specified, the new
// repository will be created under that org. If the empty string is
// specified, it will be created for the authenticated user.
//
// GitHub API docs: https://developer.github.com/v3/repos/#create
func (s *RepositoriesService) Create(ctx context.Context, org string, repo *Repository) (*Repository, *Response, error) {
...
}
```

```
// Get fetches a repository.
//
// GitHub API docs: https://developer.github.com/v3/repos/#get
func (s *RepositoriesService) Get(ctx context.Context, owner, repo string) (*Repository, *Response, error) {
```

Fixes #45

PR #464, opened 2020-05-17
github/provider: introduce owner  

anGie44:
Continuing the work done by @elislusarczyk in #96 to support individual github accounts for repos

Changes here address last review comments in #96 : 
* transition off of `GITHUB_ORGANIZATION` to `GITHUB_OWNER` (across acctests as well)
* check whether the owner is org and if not, return a user-friendly error (#checkOrganization in`util.go`)
* deprecate `organization` in `provider.go`
* skip acctests where organization is needed 
* add `IsOrganization` to `Owner` struct

Output of acceptance tests with `owner` set as Github User:
```
--- PASS: TestEtagTransport (0.00s)
--- PASS: TestRateLimitTransport_abuseLimit_get (0.00s)
--- PASS: TestRateLimitTransport_abuseLimit_post (0.00s)
--- PASS: TestRateLimitTransport_abuseLimit_post_error (0.00s)
--- PASS: TestAccValidateTeamIDFunc (0.00s)
--- PASS: TestAccGithubUtilRole_validation (0.00s)
--- PASS: TestAccGithubUtilTwoPartID (0.00s)
--- PASS: TestAccGithubReleaseDataSource_invalidRetrieveMethodReturnsError (0.03s)
--- PASS: TestAccGithubReleaseDataSource_fetchByTagNoTagReturnsError (1.57s)
--- PASS: TestAccGithubReleaseDataSource_fetchByIdWithNoIdReturnsError (1.62s)
--- PASS: TestAccGithubUserDataSource_noMatchReturnsError (1.72s)
--- PASS: TestAccGithubReleaseDataSource_fetchByLatestNoReleaseReturnsError (2.01s)
--- PASS: TestAccGithubMembershipDataSource_noMatchReturnsError (2.06s)
--- PASS: TestAccGithubIpRangesDataSource_existing (8.22s)
--- PASS: TestAccGithubUserGpgKey_basic (11.04s)
--- PASS: TestAccGithubUserSshKey_basic (11.17s)
--- PASS: TestAccGithubUserDataSource_existing (12.14s)
--- PASS: TestAccProvider_individual (12.56s)

* remaining tests are skipped
```

### DIFF
--- a/github/config.go
+++ b/github/config.go
@@ -69,7 +69,14 @@ func (c *Config) Clients() (interface{}, error) {
 	owner.v4client = v4client
 
 	owner.name = c.Owner
-	if owner.name != "" {
+	if owner.name == "" {
+		// Discover authenticated user
+		user, _, err := owner.v3client.Users.Get(ctx, "")
+		if err != nil {
+			return nil, err
+		}
+		owner.name = user.GetLogin()
+	} else {
 		remoteOrg, _, err := owner.v3client.Organizations.Get(ctx, owner.name)
 		if err == nil {
 			if remoteOrg != nil {
--- a/github/config.go
+++ b/github/config.go
@@ -2,8 +2,6 @@ package github
 
 import (
 	"context"
-	"crypto/tls"
-	"fmt"
 	"net/http"
 	"net/url"
 	"path"
@@ -15,65 +13,33 @@ import (
 )
 
 type Config struct {
-	Token        string
-	Organization string
-	BaseURL      string
-	Insecure     bool
-	Individual   bool
-	Anonymous    bool
+	Token   string
+	Owner   string
+	BaseURL string
 }
 
-type Organization struct {
-	name        string
-	id          int64
-	v3client    *github.Client
-	v4client    *githubv4.Client
-	StopContext context.Context
+type Owner struct {
+	name           string
+	id             int64
+	v3client       *github.Client
+	v4client       *githubv4.Client
+	StopContext    context.Context
+	IsOrganization bool
 }
 
 // Clients configures and returns a fully initialized GithubClient and Githubv4Client
 func (c *Config) Clients() (interface{}, error) {
-	var org Organization
+	var owner Owner
 	var ts oauth2.TokenSource
 	var tc *http.Client
 
 	ctx := context.Background()
 
-	if c.Insecure {
-		insecureClient := insecureHttpClient()
-		ctx = context.WithValue(ctx, oauth2.HTTPClient, insecureClient)
-	}
-
-	// Either Organization needs to be set, or Individual needs to be true
-	if c.Organization != "" && c.Individual {
-		return nil, fmt.Errorf("If `individual` is true, `organization` cannot be set.")
-	}
-	if c.Organization == "" && !c.Individual {
-		return nil, fmt.Errorf("If `individual` is false, `organization` is required.")
-	}
-
-	// Either run as anonymous, or run with a Token
-	if c.Token != "" && c.Anonymous {
-		return nil, fmt.Errorf("If `anonymous` is true, `token` cannot be set.")
-	}
-	if c.Token == "" && !c.Anonymous {
-		return nil, fmt.Errorf("If `anonymous` is false, `token` is required.")
-	}
-
-	if !c.Anonymous {
-		ts = oauth2.StaticTokenSource(
-			&oauth2.Token{AccessToken: c.Token},
-		)
-	}
-
+	ts = oauth2.StaticTokenSource(
+		&oauth2.Token{AccessToken: c.Token},
+	)
 	tc = oauth2.NewClient(ctx, ts)
-
-	if c.Anonymous {
-		tc.Transport = http.DefaultTransport
-	} else {
-		tc.Transport = NewEtagTransport(tc.Transport)
-	}
-
+	tc.Transport = NewEtagTransport(tc.Transport)
 	tc.Transport = NewRateLimitTransport(tc.Transport)
 	tc.Transport = logging.NewTransport("Github", tc.Transport)
 
@@ -99,30 +65,18 @@ func (c *Config) Clients() (interface{}, error) {
 	}
 	v3client.BaseURL = uv3
 
-	org.v3client = v3client
-	org.v4client = v4client
+	owner.v3client = v3client
+	owner.v4client = v4client
 
-	if c.Individual {
-		org.name = ""
-	} else {
-		org.name = c.Organization
-
-		remoteOrg, _, err := org.v3client.Organizations.Get(ctx, org.name)
-		if err != nil {
-			return nil, err
+	owner.name = c.Owner
+	if owner.name != "" {
+		remoteOrg, _, err := owner.v3client.Organizations.Get(ctx, owner.name)
+		if err == nil {
+			if remoteOrg != nil {
+				owner.id = remoteOrg.GetID()
+				owner.IsOrganization = true
+			}
 		}
-		org.id = remoteOrg.GetID()
 	}
-
-	return &org, nil
-}
-
-func insecureHttpClient() *http.Client {
-	return &http.Client{
-		Transport: &http.Transport{
-			TLSClientConfig: &tls.Config{
-				InsecureSkipVerify: true,
-			},
-		},
-	}
+	return &owner, nil
 }
--- a/github/provider.go
+++ b/github/provider.go
@@ -89,7 +89,7 @@ func init() {
 
 		"owner": "The GitHub owner name to manage.",
 
-		"organization": "The GitHub owner name to manage.",
+		"organization": "(Deprecated) The GitHub organization name to manage.",
 
 		"base_url": "The GitHub Base API URL",
 	}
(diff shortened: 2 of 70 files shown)


jcudit:
Gave this one a skim and am 👍 on the approach.  It is a large diff so I plan on reading it over the next few days.  Excited to see this 🚢 !

website/docs/index.html.markdown
```diff
@@ -33,33 +33,17 @@ resource "github_membership" "membership_for_user_x" {
...
-  As the name suggests **this is insecure** and should not be used beyond experiments,
-  accessing local (non-production) GHE instance etc.
-  There is a number of ways to obtain trusted certificate for free, e.g. from [Let's Encrypt](https://letsencrypt.org/).
-  Such trusted certificate *does not require* this option to be enabled.
-  Defaults to `false`.
-
-* `individual`: (Optional) Run outside an organization.  When `individual` is true, the provider will run outside
-  the scope of an organization. Defaults to `false`.
-
-* `anonymous`: (Optional) Authenticate without a token.  When `anonymous` is true, the provider will not be able to
```
jcudit, line 60: We should call out this removal explicitly when we update the CHANGELOG for the release this goes out in.

abitrolly:
Looks like `owner` is not needed at all - only `token` matters https://developer.github.com/v4/guides/forming-calls/#authenticating-with-graphql

For example GitLab doesn't use any kind of username for API access https://github.com/terraform-providers/terraform-provider-gitlab/blob/74c5685c69a9926c39c6beccebc89e29217494b0/gitlab/config.go#L14

jcudit:
@abitrolly this is good news, but may cause us to take a step back and re-evaluate how best to accomplish things.  We are currently trying to leverage the `v3` API and is unclear to me whether relying on the `token` is available outside of `v4`.  My aim is to read more but welcome a more thorough explanation of what could be changed.

jcudit:
@anGie44 - I've gotten this branch into a strange state by fixing conflicts using the web UI 😕.  

Also, I am now realizing that local test runs with the configured user account are incompatible:

```console
--- SKIP: TestAccGithubRepositoryWebhook_secret (0.50s)
    resource_github_repository_webhook_test.go:73: Skipping because GITHUB_OWNER "github-terraform-test-user" is a user, not an organization.
PASS
ok  	github.com/terraform-providers/terraform-provider-github/github	0.867s
```

Should a non-organization user be able to run acceptance tests? My original assumption was that `owner` was an abstraction of `organization` and `personal-account`.  Will give this another read through to better understand things, but please clarify if I've misunderstood.

anGie44:
Yeah ideally we'd want an `owner` set with an individual user to be able to run all tests but that would require that we remove the `checkOrganization` method calls in a lot of the resources/data_sources and also adjust acctests where certain values are only available for organizations vs. users (e.g. branch protection, repo collaborator permissions) ..i remember going about this but held back as that would be a lot to review in one go. instead I was thinking of partitioning out changes like in #465 by resource and/or datasource.

emmasax4:
Is there a new Terraform provider version with these changes yet? I'm running into the same issue:
```
module.example_repo.github_repository.repo: Creating...

Error: This resource requires GitHub organization to be set on the provider.
```

anGie44:
hi @emmasax4 👋 ! Not yet, but will be part of the next provider milestone which is the release of `v2.9.0` 😃

emmasax4:
@anGie44 Awesome! Do we have any idea yet of when that next release will happen?

mariux:
hi @anGie44 .. this causes 'token required' errors in our CI now... so seems to be a breaking change. Is this expected to be breaking as it was released as a minor version change rather than a major one? the token is provided via an environment variable in our case and we are still using GITHUB_ORGANIZATION env to specify the organization. So not expecting this to break in a minor release.

mariux:
we were facing some issues.. will dig deeper before re posting my deleted post... sorry for the noise

anGie44:
@mariux no worries! I believe you should still be able to source your token from an environment variable and initialize the provider as usual however maybe we could reconsider moving this back to `Optional` @jcudit ? at one point the `token` attribute was required and then changed to `Optional` to support the `anonymous` OAuth, though now removed in this version.

abitrolly:
I was going to ask if the next version should be 3.0.0 because of potential compatibility break, but got lazy. :D

jcudit:
😄 was considering a major version bump as well.  Hoping the community is forthcoming with bug reports so we can mop them up during `v2.9.1`.

mariux:
Seems to be only `terraform validate`to be affected.. will get back, once confirmed.

mariux:
I can now confirm that with provider `2.8.0` `terraform validate` runs fine without a token and with `2.9.0` `terraform validate` needs a token to run.

specifying any token (token does not need to be valid) via `GITHUB_TOKEN` environment variable fixes it.

I am not sure if this is worth to be called a full-blown regression as we can easily fix this in our CI by providing the required env variable. (we run automated tests for our open source github modules)

Should I open a bug report anyway?

abitrolly:
@mariux what could you do with anonymous GitHub provider before?

website/docs/index.html.markdown
jamengual: why this was removed and not added to the docs ?, I had to come here to actually find out why

jcudit:
> Should I open a bug report anyway?

@mariux yes, please.  Capturing this is valuable.

github/provider.go
```diff
@@ -97,41 +87,33 @@ func init() {
 		"token": "The OAuth token used to connect to GitHub. " +
 			"If `anonymous` is false, `token` is required.",
```
jamengual, line 88: Anonymous is mention here although it has been completely removed

jamengual: > I can now confirm that with provider `2.8.0` `terraform validate` runs fine without a token and with `2.9.0` `terraform validate` needs a token to run.
> 
> specifying any token (token does not need to be valid) via `GITHUB_TOKEN` environment variable fixes it.
> 
> I am not sure if this is worth to be called a full-blown regression as we can easily fix this in our CI by providing the required env variable. (we run automated tests for our open source github modules)
> 
> Should I open a bug report anyway?

same thing here, we have many automated tests now failing

scottb-isp:
I just started receiving this Warning which seems spurious because I'm not using or declaring `Organization` - I already use `Owner`. Is this Warning always emitted intentionally?

`Warning: "organization": [DEPRECATED] Use owner field (or GITHUB_OWNER ENV variable)`

My codebuild github config looks like this

```
action {
  category         = "Source"
  ~ configuration    = {
    "Branch"               = "***my branch***"
    + "OAuthToken"           = "*******"
    "Owner"                = "***my github company org***"
    "PollForSourceChanges" = "false"
    "Repo"                 = "********removed********"
  }
```

```
Initializing provider plugins...
- Checking for available provider plugins...
- Downloading plugin for provider "github" (hashicorp/github) 2.9.0...
- Downloading plugin for provider "aws" (hashicorp/aws) 2.68.0...
```

mariux:
> @mariux what could you do with anonymous GitHub provider before?

We provide a module and run automated tests for static analysis and unit testing. The static analysis runs tools like `tflint` and `terraform validate`. `terraform validate` started failing with 2.9.0 if no token is provided. As this is a module we do not configure the provider at all but just have a version constraint defined:

```
terraform {
  required_version = "~> 0.12.9"

  required_providers {
    github = "~> 2.4"
  }
}
```

website/docs/index.html.markdown
MarcoDalco: Damn, yes, it took me 2 hours to find out, and now I'm hardcoding the old plugin version. I need anonymous functionality to just pull the GitHub API IPs. That should never require tokens.

jcudit: The CHANGELOG has been retroactively updated to help save time for anyone else affected.  

Thanks for being quick to call this out.

abitrolly: @MarcoDalco open a bug report to make sure such kind of requests are possible? The advantage of token is that it probably raises from GitHub API limits, but I haven't checked. A list of resources that one can use without token would be handy.

MarcoDalco: @abitrolly https://api.github.com/meta?q=hooks "You can add the IP address we use when delivering hooks to your server's allow list. To ensure that you're always checking the right IP address, you can use the /meta endpoint to find the address we use." https://developer.github.com/v3/guides/best-practices-for-integrators/ . Having to periodically update the whitelisted IPs is not great, but it's what seems to be suggested, at the moment. Maybe in the future we could discriminate on some content/token, but right now whitelisting seems to be done that way, based on IPs.

github/config.go
```diff
@@ -99,30 +65,25 @@ func (c *Config) Clients() (interface{}, error) {
...
+		remoteOrg, _, err := owner.v3client.Organizations.Get(ctx, owner.name)
+		if err == nil {
+			if remoteOrg != nil {
+				owner.id = remoteOrg.GetID()
+				owner.IsOrganization = true
+			}
+		}
 	}
+	return &owner, nil
 }
```
skeggse, line 89: this PR removed support for the `insecure` flag, but didn't remove the `insecure` flag from the provider's accepted `insecure` config, can we reinstate the behavior or remove the flag?

Issue #351, opened 2020-02-11
Add github-actions ip addresses on datasource github_ip_ranges


bkamin29:
Hello,

Can you add the publics ip addresses used by github-actions as result of the datrasource `github_ip_ranges` ?

### Terraform Version
```
▶ terraform version
Terraform v0.12.20
+ provider.aiven (unversioned)
+ provider.github v2.3.2
+ provider.google v2.19.0
+ provider.google-beta v3.7.0
+ provider.grafana v1.5.0
+ provider.kubernetes v1.10.0
+ provider.local v1.4.0
+ provider.null v2.1.2
+ provider.random v2.2.1
+ provider.template v2.1.2
+ provider.ultradns v0.1.1
```

### Affected Resource(s)
Please list the resources as a list, for example:
- `github_ip_ranges`

benj-fletch:
Hi @benesis002 I'm not exactly sure what you are after in this request - based on your TF modules I'm guessing you need the IP addresses for some whitelist to allow deployment?

I don't believe that the [Github Meta API](https://developer.github.com/v3/meta/) makes available the IP addresses it uses for hosted runners, though some digging has revealed that they are [hosted on Azure](https://help.github.com/en/actions/reference/virtual-environments-for-github-hosted-runners#ip-addresses-of-github-hosted-runners) and use the public IP addresses of the MS datacentres.

Does this help your use case? If not are you able to provide some more context around what you require the IP addresses for?

benesis002:
Hell

> Hi @benesis002 I'm not exactly sure what you are after in this request - based on your TF modules I'm guessing you need the IP addresses for some whitelist to allow deployment?
> 
> I don't believe that the [Github Meta API](https://developer.github.com/v3/meta/) makes available the IP addresses it uses for hosted runners, though some digging has revealed that they are [hosted on Azure](https://help.github.com/en/actions/reference/virtual-environments-for-github-hosted-runners#ip-addresses-of-github-hosted-runners) and use the public IP addresses of the MS datacentres.
> 
> Does this help your use case? If not are you able to provide some more context around what you require the IP addresses for?

benesis002:
> Hi @benesis002 I'm not exactly sure what you are after in this request - based on your TF modules I'm guessing you need the IP addresses for some whitelist to allow deployment?
> 
[...]

PR #483, opened 2020-06-02
Add "web" and "api" IP blocks published by GitHub

fabiendelpierre:
For some reason the "web" and "api" blocks published at https://api.github.com/meta are missing from the github_ip_ranges data source, this intends to remedy that. I'm not sure why they aren't included in the first place.

Also... I don't code in GoLang and lack the tooling as well as the ability to test this locally, but it seemed easy enough to add the bits of code needed for this PR. Which is to say, I hope this looks OK and I didn't miss anything.

### DIFF
--- a/github/data_source_github_ip_ranges.go
+++ b/github/data_source_github_ip_ranges.go
@@ -14,6 +14,16 @@ func dataSourceGithubIpRanges() *schema.Resource {
 				Computed: true,
 				Elem:     &schema.Schema{Type: schema.TypeString},
 			},
+			"web": {
+				Type:     schema.TypeList,
+				Computed: true,
+				Elem:     &schema.Schema{Type: schema.TypeString},
+			},
+			"api": {
+				Type:     schema.TypeList,
+				Computed: true,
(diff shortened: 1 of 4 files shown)


jcudit:
@fabiendelpierre thanks for getting this published for review.  This one makes sense for an upcoming feature release.  It is now tracked for further testing.

jcudit:
🤔 our underlying [library](https://github.com/google/go-github/blob/v32.0.0/github/misc.go#L142-L165) does not have support for this and the [GraphQL API](https://developer.github.com/v4/explorer/) is also not showing support for the `web` and `api` blocks.

My recommendation is we wait on better support for this data and will ask internally to get clarity on when we can expect these fields to be visible in the GraphQL API.

EricHorst:
It would be great if the "actions" IP ranges could be incorporated into this PR when the underlying hurdles are removed. Like the others it should be similarly simple if the library support is there.

See #351

kfcampbell:
Closing in favor of #1220.

PR #492, opened 2020-06-16
Add missing attributes references for node_id

patrickmarabeas:
Adds missing attribute references for node_id within the:
* repository data source and resource
* team data source and resource
* user data source


patrickmarabeas:
CC @anGie44 @jcudit

patrickmarabeas:
This should probably be assigned to a release so it's not forgotten about

PR #506, opened 2020-07-01
Deprecate `anonymous` Flag For Provider Configuration

jcudit:
Looking for wider feedback on this approach.  Please raise any concerns with this deprecation.  It may be possible to shape `v3.0.0` to include resource-level anonymous access if there are use cases that need better representation.

### DIFF
--- a/github/config_test.go
+++ b/github/config_test.go
@@ -1,0 +1,62 @@
+package github
+
+import (
+	"os"
+	"testing"
+)
+
+func TestConfigClients(t *testing.T) {
+
+	t.Run("returns a client for the v3 REST API", func(t *testing.T) {
+		config := Config{
+			Token:      os.Getenv("GITHUB_TOKEN"),
+			Individual: true,
+		}
+
+		meta, err := config.Clients()
+		if err != nil {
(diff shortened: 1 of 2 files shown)


nwsparks:
Why is this being deprecated? Is there some alternative method of using the provider without providing a token that will be provided?

A common use case for this parameter is to retrieve the github IP range using the `github_ip_ranges` data source as seen below. If this parameter is removed how is it intended that an operation like this be achieved in the future?

```
provider "github" {
  anonymous  = true
  individual = true
}

data "github_ip_ranges" "github_ips" {}
```

jcudit:
@nwsparks see [this comment](https://github.com/terraform-providers/terraform-provider-github/issues/502#issuecomment-652980593) for ongoing discussion around this PR.

jcudit:
The latest commit explores a solution alongside deprecating the flag.  We continue to provide an `anonymous` mode, but we no longer require explicit configuration.  Instead, when the `token` parameter is empty, `anonymous` mode is enabled.

Shipping this will help unblock re-instating the `v2.9.0` features.  Edits here will be preferred when fixing up conflicts when actioning https://github.com/terraform-providers/terraform-provider-github/pull/508.

PR #508, opened 2020-07-02
Re-instate v2.9.0 Diffs For v3.0.0

jcudit:

### DIFF
--- a/github/provider.go
+++ b/github/provider.go
@@ -10,15 +10,22 @@ func Provider() terraform.ResourceProvider {
 		Schema: map[string]*schema.Schema{
 			"token": {
 				Type:        schema.TypeString,
-				Optional:    true,
+				Required:    true,
 				DefaultFunc: schema.EnvDefaultFunc("GITHUB_TOKEN", nil),
 				Description: descriptions["token"],
+			},
+			"owner": {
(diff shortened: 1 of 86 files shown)


github/provider.go
```diff
@@ -10,39 +10,29 @@ func Provider() terraform.ResourceProvider {
 		Schema: map[string]*schema.Schema{
 			"token": {
 				Type:        schema.TypeString,
-				Optional:    true,
+				Required:    true,
```
MarcoDalco, line 13: Please make it optional to support operations suggested by GitHub itself: point 2 of https://developer.github.com/v3/guides/best-practices-for-integrators/ is to whitelist GitHub IPs on for WebHook calls coming into the client infrastructure. That call ( https://api.github.com/meta ) does not require a token. Making the token required in the provider schema will make Terraform users' code unnecessarily complicated in order to provide a bit of information which isn't then used.

jcudit: This is good news. Will look to incorporate this into a follow up commit.

jcudit:
There is some follow up work to do here around testing and finalizing the provider schema ahead of a v3.0.0 release.  Aiming to merge this as-is though in order to contain follow ups in their own PR.

PR #521, opened 2020-07-20
Add v3.0.0 Breaking Changes To Provider Schema

jcudit:
Ahead of our next major release, this PR modifies the provider schema in the following ways:

- `token` becomes optional, with its absence signalling `anonymous` mode
- `organization` is no longer deprecated and instead enables the management of organization-only resources
- `individual` and `anonymous` removed
- `owner` inferred from `organization` when present

This also refactors the provider configuration testing into a new `provider_utils` file alongside the original test file. The aim was to make this suite of tests more readable.

Finally, a trivial `github_organization` data source was added to aid in testing.


jcudit:
```
--- PASS: TestAccConfigMeta (9.11s)
    --- PASS: TestAccConfigMeta/returns_an_anonymous_client_for_the_v3_REST_API (7.30s)
    --- SKIP: TestAccConfigMeta/returns_an_anonymous_client_for_the_v4_GraphQL_API (0.00s)
    --- PASS: TestAccConfigMeta/returns_a_v3_REST_API_client_to_manage_individual_resources (0.25s)
    --- PASS: TestAccConfigMeta/returns_a_v4_GraphQL_API_client_to_manage_individual_resources (0.20s)
    --- PASS: TestAccConfigMeta/returns_a_v3_REST_API_client_to_manage_organization_resources (0.23s)
    --- PASS: TestAccConfigMeta/returns_a_v4_GraphQL_API_client_to_manage_organization_resources (0.19s)
    --- PASS: TestAccConfigMeta/manages_individual_resources_for_a_GHES_deployment (0.94s)
PASS
ok  	github.com/terraform-providers/terraform-provider-github/github	9.789s
```
```
=== RUN   TestAccProviderConfigure
=== RUN   TestAccProviderConfigure/can_be_configured_to_run_insecurely
=== RUN   TestAccProviderConfigure/can_be_configured_to_run_anonymously
=== RUN   TestAccProviderConfigure/can_be_configured_with_an_individual_account
=== RUN   TestAccProviderConfigure/can_be_configured_with_an_organization_account
=== RUN   TestAccProviderConfigure/can_be_configured_with_a_GHES_deployment
--- PASS: TestAccProviderConfigure (21.43s)
    --- PASS: TestAccProviderConfigure/can_be_configured_to_run_insecurely (0.17s)
    --- PASS: TestAccProviderConfigure/can_be_configured_to_run_anonymously (2.67s)
    --- PASS: TestAccProviderConfigure/can_be_configured_with_an_individual_account (5.62s)
    --- PASS: TestAccProviderConfigure/can_be_configured_with_an_organization_account (10.23s)
    --- PASS: TestAccProviderConfigure/can_be_configured_with_a_GHES_deployment (2.74s)
PASS
ok  	github.com/terraform-providers/terraform-provider-github/github	21.758s
```

Added more testing around the GHES use case.  This is functioning well locally and will push to merge this week.  Follow up PRs are expected as iteration continues towards the new acceptance testing workflow.

github/provider_utils.go
```diff
@@ -0,0 +1,189 @@
...
+		}
+	}
+}
+
+func testAccPreCheck(t *testing.T) {
+	if v := os.Getenv("GITHUB_TOKEN"); v == "" {
+		t.Fatal("GITHUB_TOKEN must be set for acceptance tests")
+	}
+	if v := os.Getenv("GITHUB_OWNER"); v == "" {
+		t.Fatal("GITHUB_OWNER must be set for acceptance tests")
```
anGie44, line 33: we should also check for GITHUB_ORGANIZATION here if owner not set

github/config_test.go
```diff
@@ -1,82 +1,163 @@
...
-		// }
-		//
-		// if client := meta.(*Owner).v4client; client == nil {
-		// 	t.Fatalf("failed to return a v4 client")
-		// }
-		//
-		// if client := meta.(*Owner).v3client; client == nil {
-		// 	t.Fatalf("failed to return a v3 client")
-		// }
+	t.Run("manages individual resources for a GHES deployment", func(t *testing.T) {
```
anGie44, line 142: maybe add a skip condition if GHES_BASE_URL is not exported

github/provider_utils.go
```diff
@@ -0,0 +1,189 @@
...
+}
+
+func testAccCheckOrganization() error {
+	baseURL := os.Getenv("GITHUB_BASE_URL")
+	token := os.Getenv("GITHUB_TOKEN")
+
+	config := Config{
+		BaseURL: baseURL,
+		Token:   token,
+		Owner:   testOwner,
```
anGie44, line 56: we'll need to check for either `testOwner` or `testOrganization` before passing one or the other to this config. thinking about this more, I believe this function  can be refactored in the future to be used in a `PreCheck` step and use the `testAccProvider` object that gets populated instead of having to create  a new config. for now, though, I think we can just check the environment variables to not skip tests when the organization is exported

github/provider.go
```diff
@@ -103,13 +91,17 @@ var descriptions map[string]string
...
 
-		"organization": "(Deprecated) The GitHub organization name to manage.",
+		"insecure": "Enable `insecure` mode for testing purposes",
 
-		"base_url": "The GitHub Base API URL",
+		"owner": "The GitHub owner name to manage. " +
+			"Use this field instead of `organization` when managing individual accounts.",
+
+		"organization": "The GitHub organization name to manage. " +
+			"Use this field instead of `owner` when managing organization accounts.",
```
anGie44, line 104: maybe to avoid possible confusion, we can add a `ConflictsWith` attribute to these 2 fields in the schema since they each come with their respective use cases

website/docs/d/organization.html.markdown
```diff
@@ -0,0 +1,20 @@
...
+
+## Example Usage
+
+```hcl
+data "github_organization" "test" { name = "github" }
+```
+
+## Attributes Reference
+
+ * `plan` - The plan name for the organization account
```
anGie44, line 20: Nitpick: though just used for testing, we could still add this data-source to the list of data-sources in `github.erb` as this page is in place.

github/provider_test.go
```diff
@@ -40,301 +34,197 @@ func init() {
...
+	t.Run("can be configured with an individual account", func(t *testing.T) {
+
+		individualConfiguration := fmt.Sprintf(`
+			provider "github" {
+				token = "%s"
+			}
+			data "github_user" "test" { username = "%s" }
+		`,
+			os.Getenv("GITHUB_TOKEN"),
+			os.Getenv("GITHUB_ORGANIZATION"),
```
anGie44, line 131:
```suggestion
			os.Getenv("GITHUB_OWNER"),
```

github/provider_test.go
```diff
@@ -40,301 +34,197 @@ func init() {
...
-}
-`, testOwner, url, url)
+	t.Run("can be configured with an organization account", func(t *testing.T) {
+
+		organizationConfiguration := fmt.Sprintf(`
+			provider "github" {
+				organization = "%s"
+				token = "%s"
+			}
+			data "github_organization" "test" { name = "%s" }
```
anGie44, line 164: similar comment as above

github/provider_test.go
```diff
@@ -40,301 +34,197 @@ func init() {
...
-	"login": "%s",
-	"url" : "%s",
-	"repos_url": "%s/repos"
-}
-`, testOwner, url, url)
+	t.Run("can be configured with an organization account", func(t *testing.T) {
+
+		organizationConfiguration := fmt.Sprintf(`
+			provider "github" {
+				organization = "%s"
```
anGie44, line 161: Nit: could use format with index `"%[1]s"`

github/provider_utils.go
jcudit: pushed a commit to update the check to reflect this

github/provider.go
anGie44: also, we should update `index.html.markdown` with this provider schema change for the no longer deprecated organization field and now optional token field. would be important to note as well that if both owner and organization are not set or set with `null` or empty string, the provider will default to the authenticated user

website/docs/d/organization.html.markdown
jcudit: added to `github.erb` in a recent commit

github/config_test.go
jcudit: added in a recent commit

github/provider_test.go
jcudit: ❤️ did not know this magic existed.  added in a recent commit.

github/provider_utils.go
jcudit: > check for either testOwner or testOrganization before passing one or the other to this config

this has been addressed with a recent commit.

github/provider.go
jcudit: >  add a `ConflictsWith` attribute

Added in a recent commit.

jcudit:
@anGie44 concerns have been resolved.  This is ready for another pass if you have time over the next few days.

github/provider_test.go
```diff
@@ -40,301 +34,196 @@ func init() {
...
 
-func testRespondJson(responseBody string) func(http.ResponseWriter, *http.Request) {
-	return func(w http.ResponseWriter, req *http.Request) {
-		w.Header().Set("Content-Type", "application/json")
-		if _, err := w.Write([]byte(responseBody)); err != nil {
-			return
-		}
-	}
-}
+	t.Run("can be configured with an individual account", func(t *testing.T) {
```
anGie44, line 122: ah i missed this on first review :/ side-effect of adding the `ConflictsWith` property in the provider is that this and the subsequent test can't be run at the same time without 1 or the other failing since only one of the env vars can be set so i think instead of eventually failing at `testAccPreCheckEnvironment` the env var could be checked here in the beginning of the method and call t.Skip()

github/provider_test.go
```diff
@@ -40,301 +34,196 @@ func init() {
...
-func orgResponseBody(port string) string {
-	url := fmt.Sprintf(`https://localhost:%s/v3/orgs/%s`, port, testOwner)
-	return fmt.Sprintf(`
-{
-	"login": "%s",
-	"url" : "%s",
-	"repos_url": "%s/repos"
-}
-`, testOwner, url, url)
+	t.Run("can be configured with an organization account", func(t *testing.T) {
```
anGie44, line 157: similar comment above

github/resource_github_repository_test.go
```diff
@@ -693,11 +693,11 @@ func testAccCreateRepositoryBranch(branch, repository string) error {
 		Owner:   testOwner,
```
anGie44, line 693: hmm looks like it'll be problematic anywhere `Owner:   testOwner` is used in tests ...might make sense to determine the value in provider_utils.go from either of the 2 env vars

anGie44: actually thinking about this more, since this function is only referenced in one PreConfig @L286, it's not too drastic to update this function to use the provider so lines 687-699 can be eliminated and line 700 can be updated to use the provider meta e.g.
```
client := testAccProvider.Meta().(*Owner).v3client
```

and then the Preconfig ref can be simplified to: 
```				
PreConfig: func() { testAccCreateRepositoryBranch("foo", *repo.Name) },
```

radeksimko:
Looks like @anGie44 has more context here, so I'll leave the review to her.

github/resource_github_repository_test.go
jcudit: In the short term as we are building out GHES acceptance testing, my plan is to perform passes over our existing test cases to modernize / minimize where needed.  I'm not fully understanding the change here, but welcome a commit either on this PR or another one that follows.  However in general, I'm hoping that once [tests are expanded](https://github.com/terraform-providers/terraform-provider-github/pull/523), we can align our tests to best fit an Action-oriented test suite.

github/provider_test.go
jcudit: Another approach could be to unset the conflicting variable ahead of each test and `defer` re-instating the variable after the test has run.  This pattern is used in the `insecure` mode tests in the past.  I will experiment with both options and return with a commit.

jcudit: aaa1d86187961252571c3d344acd74d8df746b85 takes a different approach and removes the `ConflictsWith` feature for this use case.  It seems like the cleanest solution for the CI use case, but we likely confuse someone down the line if the docs are not clear.  What do you think?

jcudit:
https://github.com/terraform-providers/terraform-provider-github/pull/544 addresses the [concerns raised](https://github.com/terraform-providers/terraform-provider-github/pull/521#pullrequestreview-458918355) above.  

I will merge this to accommodate progress on refreshing most tests and adding new testing workflows.

colltoaction:
Is it possible that there is an undocumented breaking change here? I had to change from

```tf
provider "github" {
  base_url = "https://my-ghe-host/api/v3"
  organization = "my-org"
}
```

to

```tf
provider "github" {
  base_url = "https://my-ghe-host"
  organization = "my-org"
}
```

Issue #404, opened 2020-04-01
GitHub Enterprise provider compatibility


mengesb:
With the recent provider update to 2.5.0 we experienced an issue with our organization management using this provider.

https://github.com/terraform-providers/terraform-provider-github/compare/v2.4.1...v2.5.0

Included in this, which seems more likely, was an update in google/go-github :
https://github.com/google/go-github/compare/v29.0.2...v29.0.3

With the upgrade in go-github it looks like the APIs surrounding teams got chagned. In at least one case, I see a deprecation message in the documentation. The result is that all github_team resources are thought to not exist, so it wants to re-create them (unsure if creation would even succeed at this point). Additionally github_team_repository resources, which rely on the github_team relationship are also affected and these APIs have changed as well.

While I imagine that this isn't so much as a bug as it is a compatibility issue, I think that GitHub Enterprise users need messaging regarding compatibility with this provider. As it stands, it looks like GitHub Enterprise v2.20.3 is compatible only up to v2.4.1 of this provider.

### Terraform Version
0.12.24

### Affected Resource(s)
- github_team
- github_team_repository

**Note:** When using GitHub Enterprise (tested with v2.20.3)

### Terraform Configuration Files
https://www.terraform.io/docs/providers/github/r/team_repository.html
```hcl
# Add a repository to the team
resource "github_team" "some_team" {
  name        = "SomeTeam"
  description = "Some cool team"
}

resource "github_repository" "some_repo" {
  name = "some-repo"
}

resource "github_team_repository" "some_team_repo" {
  team_id    = "${github_team.some_team.id}"
  repository = "${github_repository.some_repo.name}"
  permission = "pull"
}
```

### Debug Output
https://gist.github.com/mengesb/811cadacde485dba7d36807ff9ae42b5

### Panic Output
N/A

### Expected Behavior
Teams and team repositories shouldn't change after provider update, or there should be some documentation about provider compatibility in the case of GitHub Enterprise vs GitHub

### Actual Behavior
When a plan is ran with provider 2.4.x, and generates resources, an update to provider version 2.5.x results in it re-generating the same team and team_repository resources. If there are members of the team - those trigger regeneration due to vertex changes

### Steps to Reproduce
Please list the steps required to reproduce the issue, for example:
1. `terraform apply` with v2.4.1
2. Update provider to 2.5.0
3. `terraform apply`

### Important Factoids
Github Enterprise v2.20.3

[...]

PR #529, opened 2020-08-12
changelog: add breaking change to 2.5.0 for GHE

docwhat:
This is to document #404 in the changelog so nobody else has to deal with this.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -83,7 +83,7 @@ BUG FIXES:
 
 BREAKING CHANGES:
 
-* `go-github` `v29.03` is incompatible with all currently known versions of GitHub Enterprise Server ([[#404](https://github.com/terraform-providers/terraform-provider-github/issues/404)])
+* `go-github` `v29.03` is incompatible with versions of GitHub Enterprise Server prior to `v2.21.5`. ([[#404](https://github.com/terraform-providers/terraform-provider-github/issues/404)])
 
 ENHANCEMENTS:
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -81,7 +81,7 @@ BUG FIXES:
 
 ## 2.5.0 (March 30, 2020)
 
-BREAKING CHANGES:
+REGRESSION:
 
 * `go-github` `v29.03` is incompatible with versions of GitHub Enterprise Server prior to `v2.21.5`. ([[#404](https://github.com/terraform-providers/terraform-provider-github/issues/404)])
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -81,6 +81,10 @@ BUG FIXES:
 
 ## 2.5.0 (March 30, 2020)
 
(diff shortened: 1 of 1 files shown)


CHANGELOG.md
```diff
@@ -81,6 +81,10 @@ BUG FIXES:
 
 ## 2.5.0 (March 30, 2020)
 
+BREAKING CHANGES:
+
+* `go-github` `v29.03` is incompatible with all currently known versions of GitHub Enterprise Server ([[#404](https://github.com/terraform-providers/terraform-provider-github/issues/404)])
```
majormoses, line 86: looks like this could be updated based on the issue comments https://github.com/terraform-providers/terraform-provider-github/issues/404#issuecomment-694367840

CHANGELOG.md
```diff
@@ -81,6 +81,10 @@ BUG FIXES:
 
 ## 2.5.0 (March 30, 2020)
 
+BREAKING CHANGES:
```
majormoses, line 84: :thinking: if we should use `REGRESSION` as they were not intentional breaking changes and if I understand root cause is that GHE server versions were quite a bit behind a github.com and once they are on new enough versions they should work.

CHANGELOG.md
docwhat: Hopefully, it will be confirmed shortly.

CHANGELOG.md
docwhat: It was broken in a minor version instead of a major number.

majormoses: I hear you and I agree with the sentiment but it's not an intentional change here, which is what `BREAKING CHANGES` typically means. Even in the upstream API this happened because github was not consistent with its api release across products. It's absolutely a bug and I would :+1:  `REGRESSION` or `BUG` as it's more accurate and does not imply intent. I am a maintainer on many projects that use semver very strictly, I weigh such breaking decisions heavily. This implies that the author and reviewer knew about it and decided it was a good idea anyways.

CHANGELOG.md
jcudit: Just getting caught up on this one.  Admittedly, compatible GHES versions are something that our testing is blind to and am hoping this changes going forward.  From the linked issue, it seems like:

> 2.9.2 works fine for us with GHES 2.21.5

> GitHub Enterprise v2.20.3 is compatible only up to v2.4.1 of this provider.

Is there a more accurate statement we can make here? Maybe this?

```suggestion
* `go-github` `v29.03` is incompatible with versions of GitHub Enterprise Server prior to `v2.21.5`. ([[#404](https://github.com/terraform-providers/terraform-provider-github/issues/404)])
```

CHANGELOG.md
jcudit: I like the option to use `REGRESSION`.  The distinction between the two makes sense, thanks for the guidance @majormoses.

```suggestion
REGRESSION:
```

jcudit (approved):

PR #540, opened 2020-08-27
repo webhook url is sensitive

kzw:
This marks URL of repo webhook as sensitive.  A use case is for atlantis server where URL contains basic auth credentials.


majormoses (approved):
LGTM

It might be good to update the docs to indicate why this is sensitive as that may not be obvious if you are not thinking about embedding credentials into the URL and are relying on other mechanisms such as verifying the signature of the payloads from github. I am not disagreeing on the validity of using auth to establish an additional entry point into say an auth proxy, just that we should document it's rationale.

majormoses (approved):
LGTM

jcudit (approved):
🎉

PR #544, opened 2020-09-02
Expand Acceptance Testing (GHES, Anonymous, Organization modes)

jcudit:
Changes here expand our acceptance tests to cover various use cases and deployment targets.  We now have explicit testing of anonymous, individual and organization use cases while also gaining the ability to target `github.com` or `ghes` deployments.  To accomplish this, some tests were rewritten while others were removed entirely.  There has also been duplication added, trading for flexibility in crafting test cases.

Secrets to support the new workflows were aded to the https://github.com/terraformtesting/terraform-provider-github repository.  This allowed for a [full run](https://github.com/terraformtesting/terraform-provider-github/runs/1026206614) against `github.com` to succeed.  Next steps are to migrate ownership of this repository to the [`github` organization](https://github.com/github/terraform-provider-github) and then fill in the secrets there.  Once completed, test runs against a GHES deployment will be available for community use.

/cc https://github.com/terraform-providers/terraform-provider-github/pull/523 (first pass)


.github/workflows/dotcom-acceptance-tests-all.yml
```diff
@@ -0,0 +1,58 @@
...
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+        with:
+          ref: ${{ github.event.pull_request.head.ref }}
+          fetch-depth: 2
+      - name: Acceptance Tests (Anonymous)
+        uses: terraformtesting/acceptance-tests@v2.0.0
+        with:
+          TF_LOG: DEBUG
```
radeksimko, line 26: Am I right that the logs here will then end up somewhere on GitHub Action's "terminal output"? I'm only asking because the log _may_ contain some sensitive data. I know that GHA will mask known secrets automatically, but I just wanted to point this out in case it has any other implications.

Also I'm not sure how useful it really is to have the log mixed with the test output, instead of logging into a file. I guess the downside of logging into a file though is that GHA won't mask the secrets and the only way of exposing that file would be through artifacts?

radeksimko (approved):
I reviewed GitHub Actions code more thoroughly and I admit that I just briefly scanned over the rest (it's a _lot of code_ 😅 ), but I didn't see any obvious blockers. Perhaps @anGie44 will have more detailed feedback.

> Next steps are to migrate ownership of this repository to the github organization and then fill in the secrets there.

I assume you mean `terraform-providers/terraform-provider-github` when you say "this repository". Either way @paultyng might be able to help.

With regards to `terraformtesting` - We tend to use that organization purely for testing and data may get wiped from there from time to time as a result, so we may want need to find a better long-term home for these GitHub Actions.

radeksimko:
btw. that GHES instance's DNS records seem to point to `255.255.255.255`, unsure if you (plan to) use some split-horizon DNS magic to serve correct records just inside GitHub Actions, but CloudFlare's DNS servers respond this:

```
;; ANSWER SECTION:
terraformtesting-ghe.eastus.cloudapp.azure.com.	8 IN A 255.255.255.255
```

.github/workflows/dotcom-acceptance-tests-all.yml
jcudit: Thanks for catching this.  I found this output helpful when debugging while writing tests / fixing bugs.  This should be `INFO` by default and can be toggled while iterating on PRs.

jcudit:
Thanks for the review @radeksimko.  Apologies for lumping together such a the large diff.

I suspect `terraformtesting-ghe.eastus.cloudapp.azure.com` is resolving like that because the VM was turned off to save ⚡.  I'll be turning it back on shortly.

jcudit:
Merged this into the wrong branch 🙃.  I have instead merged to `master` locally and pushed up the [changes](a5fe5d0a57706a316e2e2a7a49e81e1a315989dc..7b4477e64aa0d5eab38cabe032251adef10812a4).

PR #549, opened 2020-09-09
Update github URL

runlevel5:
The git form of the URL is not public-accessible


kpfleming:
It is publicly accessible, but it's not anonymously accessible. Is that what you meant?

PR #553, opened 2020-09-18
change private to visability

Xorima:
private is deprecated.


jcudit (approved):

jcudit (approved):
Thanks for catching this 🙇

PR #554, opened 2020-09-20
Revert "Add vulnerability_alerts attribute for repositories (#444)"

mallyvai:
This reverts commit a4eb4faf494171f1eeef70af5ceb0ed6258b7d02 which was causing the build to fail.

Unsure why the acceptance tests are failing at the git checkout  auth stage...


jcudit:
See [here](https://github.com/terraform-providers/terraform-provider-github/pull/556#issuecomment-711093468).

Issue #566, opened 2020-10-13
3.1.0 has breaking changes which are not identified in changelog


Xorima:
### Affected Resource(s)
- github_branch_protection

### Expected Behavior
Minor bumps should not have breaking changes, https://semver.org/

### Actual Behavior
You have introduced undocumented breaking changes inside the provider without putting anything in the changelog:

- `branch` has become `pattern`
- `repository` seems to have become `repository_id`

**Also there is no update to the documentation for this change**

Please either revert this change or release a 4.0.0 with actual updated documentation.

### References
- https://github.com/terraform-providers/terraform-provider-github/pull/337

mbarany:
I noticed the same. I had to pin the provider to 3.0

mwarkentin:
`repository_id = github_repository.foo.node_id` was the key for me doing the upgrade.

I don't think `node_id` is documented that I could find, but it appears to be related to the upgrade to the v4 (graphql) api.

anthonyangel:
We use a for_each loop to create our repositories, and a separate for_each loop to set up branch protection for 1 or more branches per repository. Being able to use the id (which was just the name) was nice & simple, but having to retrieve the node_id from the repository resource and apply that into the branch_protection resource looks like it's a pain.

kchristensen:
Seems as though the `restrictions` argument to `github_branch_protection` was also renamed, somewhere between 2.9.0 and 3.1.0. We were still on 2.9.0 (mostly due to the lack of actual Github releases so I was unaware new versions were being released).

However the [documentation](https://www.terraform.io/docs/providers/github/r/branch_protection.html#restrictions) doesn't seem to have been updated to reflect any changes.

jcudit:
😞  apologies on this one.

---

> Please either revert this change or release a 4.0.0 with actual updated documentation.

I am inclined to revert the change and cut a new `v3.x` release.  

Can I get some 👍  👎  feedback on this approach?

Krenair:
As long as you release 4.x alongside it without the revert that sounds fine
to me.

nikolay:
@jcudit Not to add to your plate, but we're still waiting for the org-level secrets! The feature was released half a year ago and is still unsupported.

jcudit:
☝️ reverts the changes that removed `repository` and `branch`.  The new fields are now computed so we can keep the same implementation behind the  `v3` configuration interface.  Breaking changes are now queued for the `v4` release, hopefully around end of year.

[...]

PR #337, opened 2020-01-31
Graphql / Uplift branch_protection resource

patrickmarabeas:
Uplifts the branch_protection resource to use v4 API.

API surface has changed slightly.

Migration function only worries about:
- migration of `id` from repository name to the current branch protection's global object ID
- branch -> pattern

**notes:**
- This branch will be rebased as necessary
- Shared consts and functions will be abstracted int util_* files as required, I've simply included them within the one file for simplicity for the time being
- This constitutes a preliminary migration to the v4 API
- Tests will be written / updated once the API surface is nailed down

PR #569, opened 2020-10-15
Fix documentation for `dismissal_restrictions`

woodb:
In the [current documentation for the resource `github_branch_protection`](https://github.com/terraform-providers/terraform-provider-github/blob/52bfa407a0bb9f34f84847f3aaaf57e2c44a8355/website/docs/r/branch_protection.html.markdown), it is noted [under `required_pull_request_reviews`](https://github.com/terraform-providers/terraform-provider-github/blob/52bfa407a0bb9f34f84847f3aaaf57e2c44a8355/website/docs/r/branch_protection.html.markdown) that:

> `dismissal_actors`: (Optional) The list of actor IDs with dismissal access.

This differs from the parameter name `dismissal_restrictions` as defined: https://github.com/terraform-providers/terraform-provider-github/blob/9c532df8193907ff250481be9060290f65bedd66/github/util_v4_consts.go#L15

This PR updates the documentation to reflect the parameter name as defined in `github/util_v4_consts.go`, under the assumption that the intended parameter name was `dismissal_restrictions` rather than `dismissal_actors`.

Relates to #337 and comments in #566

### DIFF
--- a/website/docs/r/branch_protection.html.markdown
+++ b/website/docs/r/branch_protection.html.markdown
@@ -80,7 +80,7 @@ The following arguments are supported:
 `required_pull_request_reviews` supports the following arguments:
 
 * `dismiss_stale_reviews`: (Optional) Dismiss approved reviews automatically when a new commit is pushed. Defaults to `false`.
-* `dismissal_actors`: (Optional) The list of actor IDs with dismissal access.
+* `dismissal_restrictions`: (Optional) The list of actor IDs with dismissal access.
 * `require_code_owner_reviews`: (Optional) Require an approved review in pull requests including files with a designated code owner. Defaults to `false`.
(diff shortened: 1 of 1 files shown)


woodb:
I couldn't find contributing documentation, so I'm not sure why so many of these checks are failing. Please let me know how I can help resolve those.

jcudit:
Thanks for this.  All these failing checks are definitely confusing.  Contributing docs are a great idea and will aim to follow up.  Lets get some more feedback in https://github.com/terraform-providers/terraform-provider-github/issues/566 before proceeding here.  Will keep an 👀  out for suggested next steps.

woodb:
> Thanks for this. All these failing checks are definitely confusing. Contributing docs are a great idea and will aim to follow up. Lets get some more feedback in #566 before proceeding here. Will keep an 👀 out for suggested next steps.

 Sounds good, thanks @jcudit!

jcudit (approved):

PR #570, opened 2020-10-19
Revert Config Changes To `github_branch_protection`

jcudit:
`repository` was previously removed in favour of `repository_id`.  We 
revert to `repository` here and transition `repository_id` to a computed 
value that is derived from a GraphQL lookup against the provided 
`repository`.

`branch` was previously removed in favour of `pattern`.  Similarly, 
`pattern` becomes computed and takes the value of `branch`.  `branch` 
now becomes deprecated for removal in the next major release.

Fixes https://github.com/terraform-providers/terraform-provider-github/issues/566

### DIFF
--- a/github/resource_github_branch_protection.go
+++ b/github/resource_github_branch_protection.go
@@ -17,16 +17,26 @@ func resourceGithubBranchProtection() *schema.Resource {
 
 		Schema: map[string]*schema.Schema{
 			// Input
+			REPOSITORY: {
+				Type:     schema.TypeString,
+				Required: true,
+				ForceNew: true,
+			},
 			REPOSITORY_ID: {
 				Type:        schema.TypeString,
+				Computed:    true,
+				Description: "GraphQL `node_id` of a repository",
+			},
+			PROTECTION_BRANCH: {
+				Type:        schema.TypeString,
 				Required:    true,
-				ForceNew:    true,
-				Description: "",
+				Description: "`pattern` for a `BranchProtectionRule`",
+				Deprecated:  "use `pattern` after v4.0.0 instead",
 			},
 			PROTECTION_PATTERN: {
 				Type:        schema.TypeString,
-				Required:    true,
-				Description: "",
+				Computed:    true,
+				Description: "`pattern` for a `BranchProtectionRule`",
 			},
 			PROTECTION_IS_ADMIN_ENFORCED: {
 				Type:     schema.TypeBool,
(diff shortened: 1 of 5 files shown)


github/resource_github_branch_protection.go
```diff
@@ -17,16 +17,26 @@ func resourceGithubBranchProtection() *schema.Resource {
...
 			// Input
+			REPOSITORY: {
+				Type:     schema.TypeString,
+				Required: true,
+				ForceNew: true,
+			},
 			REPOSITORY_ID: {
+				Type:        schema.TypeString,
+				Computed:    true,
+				Description: "GraphQL `node_id` of a repository",
```
woodb, line 28: Nice!

github/resource_github_branch_protection.go
```diff
@@ -17,16 +17,26 @@ func resourceGithubBranchProtection() *schema.Resource {
...
+			},
+			PROTECTION_BRANCH: {
 				Type:        schema.TypeString,
 				Required:    true,
-				ForceNew:    true,
-				Description: "",
+				Description: "`pattern` for a `BranchProtectionRule`",
+				Deprecated:  "use `pattern` after v4.0.0 instead",
```
woodb, line 34: 👍

github/resource_github_branch_protection_test.go
```diff
@@ -15,22 +15,21 @@ func TestAccGithubBranchProtection(t *testing.T) {
...
+		  repository = github_repository.test.name
+		  branch     = "main"
 		}
-
 	`, randomID)
 
 		check := resource.ComposeAggregateTestCheckFunc(
+			resource.TestCheckResourceAttr(
+				"github_branch_protection.test", "branch", "main",
+			),
```
woodb, line 32: Thanks

website/docs/r/branch_protection.html.markdown
```diff
@@ -60,8 +60,8 @@ resource "github_team_repository" "example" {
 
 The following arguments are supported:
 
-* `repository_id` - (Required) The repository associated with this branch protection rule.
-* `pattern` - (Required) Identifies the protection rule pattern.
+* `repository` - (Required) The repository associated with this branch protection rule.
+* `branch` - (Deprecated) Identifies the protection rule pattern.
```
woodb, line 64:
```suggestion
* `branch` - (Deprecated) Identifies the protection rule pattern. In v4.0.0, will become `pattern`.
```

woodb (commented):
Looks good! There were some other breaking changes that we may want to revert, specifically around changes to `required_pull_request_reviews`
- `dismissal_users` -> `dismissal_restrictions`
- `dismissal_teams` -> `dismissal_restrictions`

Additionally, `dismissal_restrictions` is erroneously cited in the docs as `dismissal_actors` (see #569).

Should we make `dismissal_restrictions` a computed field w/ actor ID's from `dismissal_users` and `dismissal_teams`, adding a deprecation warning for those fields as well?

davidji99:
It would be great if the PR author could add in some state migrations so users that currently define `REPOSITORY_ID` don't need to `terraform state rm` and `terraform import` to move to the version that will require `REPOSITORY` attribute.

jcudit:
Closing this as we've discussed rolling forward with v4.0.0 instead.

PR #571, opened 2020-10-20
[Docs] dismissal_actors has been replaced by dismissal_restrictions in github_branch_protection

politician:
The example in the docs was good but not the reference


jcudit:
❤️  thanks for this.  A duplicate was merged.

PR #579, opened 2020-10-27
Use GetTeamBySlug instead of ListTeams in team data source

k24dizzle:
Instead of using `ListTeams` to iterate through every single team, use [GetTeamBySlug](https://github.com/google/go-github/blob/master/github/teams.go#L122) to find the team that matches a given slug.

### DIFF
--- a/github/data_source_github_team.go
+++ b/github/data_source_github_team.go
@@ -2,11 +2,9 @@ package github
 
 import (
 	"context"
-	"fmt"
 	"log"
 	"strconv"
 
-	"github.com/google/go-github/v32/github"
 	"github.com/hashicorp/terraform-plugin-sdk/helper/schema"
 )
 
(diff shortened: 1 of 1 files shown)


k24dizzle:
New to contributing to this repo, so forgive me if this info is located somewhere.

I see the acceptance tests are failing on this PR, but I was hoping to just run the tests related to this file (`data_source_github_team_test.go`) locally.

Is there an easy way to just test my changes without triggering the entire acceptance test suite locally as well?

jcudit:
@k24dizzle thanks for putting this PR together.  The change looks 👌  but will do some extra validation outside of this PR to make sure all is well.

Apologies for the state of acceptance testing.  I'll have to do more reading up on why this behaviour is happening.  Seems to be an issue with forks and the workflows configured for this project 🙃 .  As for local testing, take a look at the [`CONTRIBUTING.md`](https://github.com/terraform-providers/terraform-provider-github/blob/master/CONTRIBUTING.md) that was recently merged.  

I will also get another branch started that pulls in the commits in this PR, which should get the test suite running properly against this diff.

jcudit (approved):
This tested well over [here](https://github.com/terraform-providers/terraform-provider-github/actions/runs/358341625).  Will follow up with a PR that adds the latest testing suite to this resource.

PR #601, opened 2020-11-12
Update team.html.markdown

tonglil:
Note this field is only available for [GH Enterprise _Server_](https://docs.github.com/en/enterprise-server@2.22/rest/reference/enterprise-admin#ldap), not [GH Enterprise (the pricing plan)](https://github.com/pricing).


jcudit (approved):
Thanks!  Appreciate this correction 😄

PR #317, opened 2020-01-05
Fix issues with auto_init destroying repositories

majormoses:
From github API perspective setting or modifying `auto_init` only makes sense in the context of creating a new repository. Changing it after is ignored by the github API which is the behavior we should (and use to) match. The only scenario where this makes sense to assume that changing this intends on a destructive action (blowing up a github repo is not something that should be easy accidentally) is when you use the `terraform taint` command but for very different reasons. This change is related to several PRs and issues that subverts the communities expectations with no real or perceived value.

For further context please see #155 #154 https://github.com/sous-chefs/terraform-github-org/pull/111 #135 

fixes #155 

Signed-off-by: Ben Abrams <me@benabrams.it>

PR #609, opened 2020-11-19
remove `ForceNew` on `template*` as they are concerns only at creation time

majormoses:
There are a number of parameter resources that have been marked as `ForceNew: true` out of a desire in "correctness" by those that do not actually understand how these resources are used in the real world as well as the damage that can be done. No one wants to blow up a repository to change something like this, if they need to there is a mechanism built into terraform called [taint](https://www.terraform.io/docs/commands/taint.html). While there are some things that make sense for using `ForceNew` a repository for source control on properties that the API will ignore outside of creation is not one of them.

This is applying the same changes as #317 and inspired by https://github.com/terraform-providers/terraform-provider-github/issues/155#issuecomment-729901509

Signed-off-by: Ben Abrams <me@benabrams.it>

### DIFF
--- a/github/resource_github_repository.go
+++ b/github/resource_github_repository.go
@@ -105,13 +105,11 @@ func resourceGithubRepository() *schema.Resource {
 			},
 			"license_template": {
 				Type:     schema.TypeString,
-				Optional: true,
-				ForceNew: true,
+				Optional: true
 			},
 			"gitignore_template": {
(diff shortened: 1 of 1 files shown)


jcudit (approved):
🙇🏾 for cleaning up this behaviour.  This will make the provider safer to work with going forward.

Issue #597, opened 2020-11-11
Can't import github_branch_protection using repository:branch id


k24dizzle:
### Terraform Version
v4.0.0

### Affected Resource(s)
- `github_branch_protection`

### Terraform Configuration Files
```hcl
provider "github" {
  organization = "kzhaotest"
}

resource "github_branch_protection" "test" {
  repository_id  = "branch_protection"
  pattern        = "main"
  enforce_admins = false

  required_pull_request_reviews {
    dismiss_stale_reviews           = true
    required_approving_review_count = 1
  }
}
```

### Debug Output
https://gist.github.com/k24dizzle/db8134b8b00171f7b6dbd4c38d4886b4

I tried importing the branch protection, but the provider couldn't find the resource. But when I applied the changes, I ran into errors saying the branch protection rule for that branch already existed.

### Expected Behavior
I expected terraform to import the state of the branch_protection of my `main` branch in this repo: https://github.com/kzhaotest/test as described in the documentation here: https://www.terraform.io/docs/providers/github/r/branch_protection.html#import

### Actual Behavior
```
kzhao@kzhao-mbp151 ~/other/branch_protection (main) $ terraform import github_branch_protection.test test:main
github_branch_protection.test: Importing from ID "test:main"...
github_branch_protection.test: Import prepared!
  Prepared github_branch_protection for import
github_branch_protection.test: Refreshing state... [id=test:main]

Error: Cannot import non-existent remote object

While attempting to import an existing object to
github_branch_protection.test, the provider detected that no object exists
with the given id. Only pre-existing objects can be imported; check that the
id is correct and that it is associated with the provider's configured region
or endpoint, or use "terraform apply" to create a new remote object for this
resource.
```

### Steps to Reproduce
`terraform init`
`terraform import github_branch_protection.test test:main` 

### Notes

I did notice in other experiments that I was able to import a `github_branch_protection` resource by instead of using `repo_name:branch_name` using the `graphql node id` of the branch protection rule.

But I assume that this overall issue is related to https://github.com/terraform-providers/terraform-provider-github/pull/593

Interestingly enough, even though https://github.com/kzhaotest/test has a branch protection rule for the `main` branch, I wasn't able to get the id of the rule in the [graphql explorer](https://developer.github.com/v4/explorer/), so I wasn't able to test this assumption out here.

jcudit:
[...]

PR #610, opened 2020-11-24
Fix unable to resolve node id for branch_protection

Polygens:
After updating to version `4.0.1` we were experiencing issues with the `github_branch_protection` resource, any changes or new created branch protections would fail with the following error:
```
Error: Could not resolve to a Repository with the name 'org-name/MDEwOlJlcG9zaXRvcnkzOTgy'.
```

After looking deeper in the issue we found a check in the code looking for node id of 32 characters in length. To my knowledge the node id is never this long and as it is base64 encoded can be variable in length. Proposing to remove this check and replace it with a check if it is valid base64 instead.

This was only tested on a private hosted Github Enterprise.

### DIFF
--- a/github/util_v4_repository.go
+++ b/github/util_v4_repository.go
@@ -38,11 +38,6 @@ func getRepositoryID(name string, meta interface{}) (githubv4.ID, error) {
 }
 
 func repositoryNodeIDExists(name string, meta interface{}) (bool, error) {
-	// Quick check for node ID length
-	if len(name) != 32 {
-		return false, nil
-	}
-
 	// API check if node ID exists
 	var query struct {
 		Node struct {
--- a/github/util_v4_repository.go
+++ b/github/util_v4_repository.go
@@ -2,6 +2,7 @@ package github
 
 import (
 	"context"
+	"encoding/base64"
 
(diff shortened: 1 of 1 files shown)


Polygens:
Similar issue was mentioned in https://github.com/terraform-providers/terraform-provider-github/issues/597

jcudit (approved):
🙇

kzw:
how can we prioritize this patch?  this is blocking our upgrade to 4+

guillermotti:
When are you planning to release this bug fix? Thanks!

jcudit:
👋  sorry for the delay here.  Will merge this shortly and aim to release this week.

jcudit (approved):

majormoses:
@jcudit given that this is blocking some upgrades can we please get a hotfix release out?

faultylee:
Turn out testing with base64 decode alone is not enough.`abcdefhg` is a valid base64 encoded string, and we hit this issue today. After some testing, any 4 & 9 char string is base64 decodeable. See https://github.com/integrations/terraform-provider-github/issues/597#issuecomment-766705593

PR #628, opened 2020-12-09
Remove github.com/hashicorp/terraform from dependencies

cappyzawa:
I removed the dependency of  github.com/hashicorp/terraform that used only to use `acctest`.


jcudit:
Thanks for this.  Glad we can move on to using `github.com/hashicorp/terraform-plugin-sdk/` fully.

PR #635, opened 2020-12-21
change private to visibility

shogo82148:
private is deprecated.
same as https://github.com/terraform-providers/terraform-provider-github/pull/553


jcudit (approved):
Thanks for keep our docs correct!  Very much appreciated 🙇🏾

PR #644, opened 2020-12-30
Use commit SHA to lookup commit info in github_repository_file resource

shoekstra:
Currently the provider loops through commits in a repo until it finds the most recent commit containing the managed file. This is inefficient and could lead to you being rate limited if managing a few files that were updated a long time ago.

This PR fixes that by storing the commit SHA when updating the file and using that SHA to lookup the commit info instead of looping through all commits.


jcudit (approved):
❤️  the documentation fixes as well.  Will get this added to the next release.

shoekstra:
Super. Thanks @jcudit, appreciate it.

PR #663, opened 2021-01-19
Fix link to Milestones page

dievri:
Update outdated link in README.md


jcudit (approved):
Thanks for this.  I'll also check with the rest of the community about these broken links.  🤞🏾  that there is a solution that can automatically perform redirects for us.

PR #673, opened 2021-01-26
changelog: manually fix links to issues

docwhat:
They were pointing to the old archived repo, which doesn't have newer issues.

I suspect there is some automation that needs to be updated to the new repo location.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,15 +2,15 @@
 
 BUG FIXES:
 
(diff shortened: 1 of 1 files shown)


docwhat:
The older ones don't need to be updated if we don't want; the old repo is still there, just archived. 

Only newer issues need to be updated.

jcudit (approved):

Issue #500, opened 2020-06-26
github_team.id can't be used in a for_each expression to create multiple github_team_repository resources


jspiro:
### Terraform Version
0.12.28

### Affected Resource(s)
- github_team
- github_team_repository

### Terraform Configuration Files
Not working:
```hcl
data "github_team" "writers" {
  slug = "writer-team"
}

resource "github_team_repository" "writers" {
  for_each   = data.github_team.writers.id
  # or for multiple teams something like:
  # for_each   = { for obj in [data.github_team.writers] : obj.id => obj.id }
  team_id    = each.value
  repository = "repo"
  permission = "push"
}
```

Working:
```hcl
data "github_team" "writers" {
  slug = "foo-bar"
}

resource "github_team_repository" "writers" {
  team_id    = data.github_team.writers.id
  repository = "repo"
  permission = "push"
}
```

### Expected Behavior
I've been trying to do a simple thing: Provide a list (whether data or resource) of teams to github_team_repository using 0.12 `for_each` syntax. I expect that the `for_each` syntax would work.

### Actual Behavior
```
  on modules/github/repository/main.tf line 58, in resource "github_team_repository" "writers":
  XX:   for_each   = data.github_team.writers.id

The "for_each" value depends on resource attributes that cannot be determined
until apply, so Terraform cannot predict how many instances will be created.
To work around this, use the -target argument to first apply only the
resources that the for_each depends on.
```

### Steps to Reproduce
1. `terraform plan`

### Important Factoids
I built a team module that handles membership, and a repo module that hooks up those teams to admin/writer/reader roles. Passing one team into the other is a natural thing to do.

But TF cannot seem to get the dependencies right. It doesn't infer that the data needs to be looked up first, or that it's a computed value, or the team needs to be created first (if using github_team resources) when given in a `for_each`, but it works perfectly if I provide the team id directly.

I've tried to provide the ID a dozen different ways, through different vars, locals, lists, sets, you name it. I've tried elaborate and explicit layers of `depends_on` in both variables, resources, and the like. The above is the simplest reproduction.

### Workaround
If I create the teams first, and then hook them up to the repo, no problem.

But it's not really possible to run this in a CI environment without creating the teams first in one PR, then the repos in another – that's two different PRs, two different applications–I can't get my team to accept the workaround, they'd rather dump terraform.

Ultimately, this is annoying and should work, but maybe I'm missing something obvious. At this point I've taken it personally and have spent hours trying to figure it out 😩 

[...]

PR #676, opened 2021-01-26
Add Example For `github_team_repository`

jcudit:
Also documents a limitation with `for_each` in this scenario.

/cc https://github.com/integrations/terraform-provider-github/issues/500


jspiro (approved):
Thank you for doing this!

Issue #662, opened 2021-01-18
Unable to create Organization repos in GHE


jonesy1234:
### Terraform Version
```
Terraform v0.14.4
+ provider registry.terraform.io/hashicorp/github v4.3.0
+ provider registry.terraform.io/integrations/github v4.3.0
```

### Affected Resource(s)
- github_repository

### Terraform Configuration Files
```hcl
provider "github" {
  organization = "myorgname"
  token        = "**************************"
  base_url     = "https://github.enterprise.instance/"
}

resource "github_repository" "this" {
  name         = "shouldbeinorg"
  description  = "shouldbeinorg"

}
```

### Expected Behavior
Repository should be created in the organization and not the account of the PAT.

### Actual Behavior
Repository is created under the user associated with the PAT i.e. not in the org.

### Steps to Reproduce
1. `terraform apply`

### Important Factoids
Just to be clear this is for GitHub Enterprise, I've not verified behaviour in github.com

### References
This is possibly a similiar issue - https://github.com/integrations/terraform-provider-github/issues/639

jurgen-weber-deltatre:
I just got access to our GHE today and I thought I was doing it wrong! I can do it fine using the same PAT onthe CLI but in teh provider... no go.

the thing I am finding:

```
Error: POST https://api.github.com/user/repos: 401 Requires authentication []

  on modules/github/repositories.tf line 1, in resource "github_repository" "this":
   1: resource github_repository this {
```

that post is incorrect, when I do it successfully on thye CLI:

```
curl -XPOST -H "Accept: application/vnd.github.v3+json" -u USERNAME:PAT "https://api.github.com/orgs/{org}/repos" -d '{"name":"REPONAME","private":true,"visibility":"internal"}'

**Comment by jurgen-weber-deltatre** [comment]
so reading the code, I am not sure what the 'Organization' paramter in the provider config is doing....

but if I updated the provider config with

```
provider "github" {
  owner        = ORG
  token        = trimspace(data.aws_kms_secrets.tf_admin.plaintext["github"])
}
```

I get the following error:
```
Error: POST https://api.github.com/orgs/ORG/repos: 403 You need admin access to the organization before adding a repository to it. []
```

At least now I am getting the same URL I constructed while using CURL.

But still, something is up with the signing and auth of the provider, since I can create using CURL and not TF.

**Comment by jurgen-weber-deltatre** [comment]
I got it to work, use the above provider and:

```
resource github_repository this {
[...]

PR #680, opened 2021-01-29
visibility should be set on create and removed on update

jurgenweber:
https://github.com/integrations/terraform-provider-github/issues/662

You can find a lot of my journey in the above ticket and with the help of some at [google/go-github](https://github.com/google/go-github/pull/1786) and github support I have found my problem.

Having the visibility removed on create, meant I could not create internal repos (when I do not have permissions to create public repos in my org) and the comment I removed 

```
        // Auth issues (403 You need admin access to the organization before adding a repository to it.)
        // are encountered when the resources is created with the visibility parameter. As
        // resourceGithubRepositoryUpdate is called immediately after, this is subsequently corrected.
```

is the exact issue and problem I was facing, but this is suggesting that it should solve it. Since you can not update the visibility of a repo (edit: I later learned that this is not true, you can edit the visibility of repos), leaving it to update will always error also, as per the other comment:

```
// The endpoint will throw an error if trying to PATCH with a visibility value that is the same
```

this patch will always fail, so it should never run.

this PR solves this repo create problem, by ensuring that Visibility is set on create and never updates.

I also added in the example/demo I was using to test my provider locally.

Thanks

Note; If there are circumstances in which a repo can change its visibility do we know what they all are and we can update this accordingly.

### DIFF
--- a/github/resource_github_repository.go
+++ b/github/resource_github_repository.go
@@ -271,11 +271,6 @@ func resourceGithubRepositoryCreate(d *schema.ResourceData, meta interface{}) er
 	repoReq := resourceGithubRepositoryObject(d)
 	owner := meta.(*Owner).name
 
-	// Auth issues (403 You need admin access to the organization before adding a repository to it.)
-	// are encountered when the resources is created with the visibility parameter. As
-	// resourceGithubRepositoryUpdate is called immediately after, this is subsequently corrected.
-	repoReq.Visibility = nil
-
 	repoName := repoReq.GetName()
 	ctx := context.Background()
 
@@ -462,10 +457,9 @@ func resourceGithubRepositoryUpdate(d *schema.ResourceData, meta interface{}) er
 
 	repoReq := resourceGithubRepositoryObject(d)
 
+	// The visibility of a repo can not be changed once created
 	// The endpoint will throw an error if trying to PATCH with a visibility value that is the same
-	if !d.HasChange("visibility") {
-		repoReq.Visibility = nil
-	}
+	repoReq.Visibility = nil
 
 	// Can only set `default_branch` on an already created repository with the target branches ref already in-place
 	if v, ok := d.GetOk("default_branch"); ok {
--- a/github/resource_github_repository.go
+++ b/github/resource_github_repository.go
@@ -457,9 +457,17 @@ func resourceGithubRepositoryUpdate(d *schema.ResourceData, meta interface{}) er
 
 	repoReq := resourceGithubRepositoryObject(d)
 
-	// The visibility of a repo can not be changed once created
-	// The endpoint will throw an error if trying to PATCH with a visibility value that is the same
-	repoReq.Visibility = nil
+	if d.HasChange("visibility") {
+		// The endpoint will throw an error if this repo is being created and the old value is ""
+		o, n := d.GetChange("visibility")
+		log.Printf("[DEBUG] Old Value %v New Value %v", o, n)
+		if o.(string) == "" {
+			repoReq.Visibility = nil
+		}
+	} else {
+		// The endpoint will throw an error if trying to PATCH with a visibility value that is the same
+		repoReq.Visibility = nil
+	}
 
 	// Can only set `default_branch` on an already created repository with the target branches ref already in-place
 	if v, ok := d.GetOk("default_branch"); ok {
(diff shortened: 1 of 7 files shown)


jurgenweber:
I looked into this a bit more, it seems you can change the visibility of a repo, the probably is... On create a change from the TF SDK is computed, from "" to "internal". I added this test in.

The problem now is that even though this is created as visibility Internal and the update happens with no change.. THe repo is public!? Seems like my org's ability to make public's repos changed over the week. :)

jurgenweber:
All in all, this requires a bit of testing to handle all of the scenarios.

Do we have access to a private user, an org with the ability to create public repos, an org _with out_ the ability to create public repos.

Thanks

jcudit:
Spent some time with this branch and tested a few of the scenarios mentioned.  The experience was quite buggy so I'd like to defer this work to batch it in with related improvements to the repository resource.  I believe this is an important problem to fix, but want to ship a few outstanding PRs before coming back to this one.

jurgen-weber-deltatre:
Ok, how did you test them? What were the bugs. I have some capacity so I am happy to continue with it and get it to where it is needed.

jcudit:
I was working out of the example in this PR but kept hitting errors when deleting resources.  I attempted this against a GHES instance using a token with full permissions to a `terraformtesting` organization.  Also tried to create a user repository instead of an organization repository.  It seemed to be getting tripped up on disabling vulnerability alerts 😕 .  Have you seen this behaviour?  

```sh
$ terraform apply -auto-approve   -var "owner=${GITHUB_OWNER}"   -var "github_token=${GITHUB_TOKEN}"
github_repository.internal: Creating...

Error: DELETE https://<host>/api/v3/repos/terraformtesting/mtribes-jurgen/vulnerability-alerts: 404 Not Found []

  on main.tf line 1, in resource "github_repository" "internal":
   1: resource github_repository internal {
```

I hit my time limit while troubleshooting, but aim to pick this back up now that v4.4.0 is shipped.  Was planning to knock out a couple of vulnerability related bugs to better understand that code path before addressing whatever the bug is here that is blocking the example from running for me.

jurgenweber:
odd, I had no issues with any repo features, just the create process.

lneuhaus:
Is someone still working on this? Can I help, e.g. by testing and reporting? One of my projects is blocked by this fix getting merged.

jcudit:
A few of us are going to pair on reviews this week.  Aiming to get this one tested out during the session and return here with next steps.  🤞🏾 that things just work and we can get to shipping.

jcudit (commented):
Gave this another pass and think we should get some tests around repository visibility included with this improvement.  Seems like something that deserves more confidence as we make changes in this area.  d0869153 is a start that can be cherry-picked.  Is there any interest in picking things up from here?

jurgenweber:
Yeah, up to you guys. I will be honest and say. I don't have the space for this at the moment.. So feel free to butcher close and carry on with this PR as you see fit.

kfcampbell:
I'm taking a few days off later in the week so I'll poke at this a bit and start by cherry-picking @jcudit's test case. I'll post an update here in a few days!

github/resource_github_repository.go
```diff
@@ -462,8 +457,15 @@ func resourceGithubRepositoryUpdate(d *schema.ResourceData, meta interface{}) er
...
-	if !d.HasChange("visibility") {
+	if d.HasChange("visibility") {
+		// The endpoint will throw an error if this repo is being created and the old value is ""
+		o, n := d.GetChange("visibility")
+		log.Printf("[DEBUG] Old Value %v New Value %v", o, n)
+		if o.(string) == "" {
+			repoReq.Visibility = nil
+		}
+	} else {
+		// The endpoint will throw an error if trying to PATCH with a visibility value that is the same
```
tibbes, line 468:
```suggestion
	// The endpoint will throw an error if trying to PATCH with a visibility value that is the same
	if !d.HasChange("visibility") || d.IsNewResource() {
```

tibbes: @kfcampbell, if you're looking at this today, you might find this simplification useful.

jcudit:
Superseded by https://github.com/integrations/terraform-provider-github/pull/746.  

Huge thanks again @jurgen-weber-deltatre for getting this effort kicked off ❤️ .  Hoping this tests well within the community as it has been a long-standing pain point for the project.

jurgenweber:
no problems. Glad to hear you got somewhere.

Issue #480, opened 2020-05-31
github_repository_collaborator unable to validate `triage` and `maintain` permissions


HorizonNet:
### Terraform Version

v0.12.24

### Provider Version

v2.8.0

### Affected Resource(s)

- github_repository_collaborator

### Terraform Configuration Files
```hcl
data "github_membership" "test-user" {
  username = "testuser"
}

resource "github_repository" "my-repo" {
  name        = "my-repo"
  private     = true
  description = ""
}

resource "github_repository_collaborator" "my-repo-test-user" {
  repository = github_repository.my-repo.name
  username   = data.github_membership.test-user.username
  permission = "maintain"
}
```

```sh
terraform import github_repository_collaborator.my-repo-test-user my-repo:testuser
```

### Expected Behavior

After importing an existing collaborator with *Maintain* permissions the collaborator should have `maintain` permissions in the state.

### Actual Behavior

After importing an existing collaborator with *Maintain* permissions the collaborator gets `push` permissions in the state. Running `terraform plan` afterwards notes a replacement.

```
# github_repository_collaborator.my-repo-test-user must be replaced
-/+ resource "github_repository_collaborator" "my-repo-test-user" {
      ~ id            = "my-repo:testuser" -> (known after apply)
      + invitation_id = (known after apply)
      ~ permission    = "push" -> "maintain" # forces replacement
        repository    = "my-repo"
        username      = "testuser"
    }
```

Running `terraform apply` in the end leads to

```
Error: User testuser is already a collaborator
```

Overall, the error is similar to #469, but the actual problem seems to be the import to the state in this situation.

### Steps to Reproduce

1. `terraform import github_repository_collaborator.my-repo-test-user my-repo:testuser`
2. `terraform plan`
3. `terraform apply`

anGie44:
hi @HorizonNet, thanks for writing-up this issue! I was able to reproduce and looking into this more, I think we're bound by the limitations of the github API, such that the permission levels provided on request to create a github collaborator are not always equal to the permission levels returned from the github API.

[...]

PR #683, opened 2021-01-29
Add Diff Suppression Option To `repository_collaborator`

jcudit:
Adds a new option to enable diff suppression when permission is set to `triage` or `maintain`.  This is a quality of life workaround until the GitHub API has better support for what we're trying to do with this resource.  Full details in linked issue.

/cc https://github.com/integrations/terraform-provider-github/issues/480


github/resource_github_repository_collaborator.go
```diff
@@ -158,6 +172,10 @@ func resourceGithubRepositoryCollaboratorRead(d *schema.ResourceData, meta inter
 	return nil
 }
 
+func resourceGithubRepositoryCollaboratorUpdate(d *schema.ResourceData, meta interface{}) error {
```
jcudit, line 175: Adding an update function was required if we are to add the `permission_diff_suppression` option without being `ForceNew`.  This enables a smoother upgrade option as users can turn on this workaround without triggering a recreate.  Downside is we'll have to keep this no-op update in mind if we ever want to update actual state.

PR #684, opened 2021-01-29
Improve handling of getRepositoryID

k24dizzle:
- If we can't find the repository by node ID, then failover to try to find it by repository name


github/util_v4_repository.go
```diff
@@ -3,22 +3,20 @@ package github
...
 func getRepositoryID(name string, meta interface{}) (githubv4.ID, error) {
 
-	// Interperet `name` as a node ID and return
-	exists, err := repositoryNodeIDExists(name, meta)
+	// Interpret `name` as a node ID
+	exists, nodeIDerr := repositoryNodeIDExists(name, meta)
 	if exists {
 		return githubv4.ID(name), nil
 	}
-	if err != nil {
```
k24dizzle, line 17: As described here: https://github.com/integrations/terraform-provider-github/issues/597#issuecomment-766705593 there are some repo names that are base64 decodable (`wiki`, `abcdefgh `) and as a result an error would be returned here. Instead of returning an error, go ahead and try finding the repo by repo name anyways as fallback logic.

jcudit (approved):
Thanks for the fix!  Aiming to get this released this week.

github/util_v4_repository.go
```diff
@@ -30,9 +28,13 @@ func getRepositoryID(name string, meta interface{}) (githubv4.ID, error) {
 	}
 	ctx := context.Background()
 	client := meta.(*Owner).v4client
-	err = client.Query(ctx, &query, variables)
-	if err != nil {
-		return nil, err
+	nameErr := client.Query(ctx, &query, variables)
+	if nameErr != nil {
+		if nodeIDerr != nil {
```
bateller, line 33: This doesn't seem to be failing nicely when branch protection doesn't exist

PR #685, opened 2021-01-30
github/config: Fix detection of individual, non-org accounts

### LINKED ISSUE #678: v4.3.1 breaks owner provider configuration
### Terraform Version
```
Terraform v0.14.5
+ provider registry.terraform.io/integrations/github v4.3.1
```

### Affected Resource(s)
 - github_repository
 - all

```hcl
terraform {
  required_providers {
    github = {
    }
  }
}

provider "github" {
  owner = "sh0shin"
  # token is provided via GITHUB_TOKEN environment variable
}

resource "github_repository" "sick" {
  name        = "sick"
  description = "My sick git repository"
  visibility  = "public"
}
```

### Debug Output
https://gist.github.com/sh0shin/8faca30a9f162939fa50fcc097868112

### Expected Behavior
Repository `sick` should be created.

### Actual Behavior
```
Error: GET https://api.github.com/orgs/sh0shin: 404 Not Found []

  on main.tf line 11, in provider "github":
  11: provider "github" {
```

### Steps to Reproduce
 * `terraform apply`

issyl0:
- Previously, whenever an individual user tried to interact with their repos, the provider would return an error, rendering v4.3.1 unusable:

  ```
  ➜ terraform plan
  Error: GET https://api.github.com/orgs/issyl0: 404 Not Found []
  on /Users/issyl0/repos/terraform/github.tf line 1, in provider "github":
  ```

- Reproduction steps and a log of manual tests I did: https://gist.github.com/issyl0/cd61e4cb59de2c2e1e8f45e3cf7c12f5.

- `ConfigureOwner` works such that if the `owner.name` is not blank (ie, the user had specified `owner = <username>` in their Terraform file), the code progresses to check if the `owner.name` is an org. Importantly, that check (prior to this change) returned an error if `owner.name` was not an org. That final error meant it was impossible to run if not using an organisation account.

----

- This is a partial revert of #668. Interestingly, `github/config_test.go` didn't fail at all before (and I couldn't figure out how to make a better test). Reverting it all also fixed the problem, but the `ConfigureOwner` method appeared to do the right thing for me on its own. **I also don't have an org account to test this against to confirm regression of the issue that PR was fixing.**

- Fixes #678.

----

- As I was preparing to spend part of my Saturday using this provider to configure my personal account repos with consistent settings, I noticed it was broken so I thought I'd have a go at figuring out how! This is only my second Terraform provider change ever, so if things could be better, I'd appreciate some pointers for improvements! As always, thanks for building this provider!

### DIFF
--- a/github/config.go
+++ b/github/config.go
@@ -115,8 +115,6 @@ func (c *Config) ConfigureOwner(owner *Owner) (*Owner, error) {
 				owner.id = remoteOrg.GetID()
 				owner.IsOrganization = true
 			}
-		} else {
-			return nil, err
 		}
 	}
 


jcudit (approved):
🙇🏾  appreciate the fix here.  A few thoughts:
- This would be a full revert to the previous behaviour as it looks like the `Clients()` function is unused and can be deleted.
- Reverting back to the previous implementation makes sense given the impact of this bug
- Throwing a dedicated error when the `BaseURL` may be an alternate fix to #668, which we can pursue after this is released

Will wrap all of this into a follow-up issue as `github/config.go` could use some refreshing on the testing and implementation fronts.

issyl0:
@jcudit Sounds good, thanks for the review!

Issue #687, opened 2021-02-01
Feature Request: Add github_app_installation_repository resource


k24dizzle:
I was wondering if we could create a new resource to help manage app installations to repositories in an organization.

It would be similar to `github_team_repository`, taking in an app `installation_id` and a `repository_id` to manage this relationship.

https://docs.github.com/en/rest/reference/apps#add-a-repository-to-an-app-installation

k24dizzle:
Went ahead and gave it a shot in the PR above

gmile:
PR above got merged. Shall this be closed now?

Issue #514, opened 2020-07-15
Feature Request > GitHub App authentication support


kuwas:
### Terraform Version
v0.13.0-beta3

### Affected Resource(s)
- provider configuration

### Terraform Configuration Files
```hcl
provider "github" {
  app_id          = var.app_id
  private_key_pem = var.private_key_pem
}
```

### Debug Output
n/a, this feature doesn't exist

### Panic Output
n/a, this feature doesn't exist

### Expected Behavior
Terraform authenticates as a GitHub App, this removes the need of making a bot GitHub account for terraform usage

### Actual Behavior
n/a, this feature doesn't exist

### Steps to Reproduce
n/a, this feature doesn't exist

### Important Factoids
- GitHub Apps has higher API limits than a Personal Access Token
  - https://developer.github.com/apps/building-github-apps/understanding-rate-limits-for-github-apps/#server-to-server-requests

### References
- https://developer.github.com/apps/building-github-apps/authenticating-with-github-apps/#authenticating-as-a-github-app
- https://github.com/google/go-github/issues/1069

dutsmiller:
I would add that it would be nice to be able to use repository deployment keys in the same fashion.  While this would limit the over functionality of the provider, it would be nice to not have to tie the use of the provider to a specific user account.

patrickmarabeas:
It's been on my list to port over - but you can use this provider to auth via Github App, and feed the token into the official Github Terraform provider:

Provider schema: https://github.com/patrickmarabeas/terraform-provider-github-v4/blob/master/github-v4/provider.go#L31

Access the token: https://github.com/patrickmarabeas/terraform-provider-github-v4/blob/master/github-v4/data_source_github_token.go

jcudit:
Very interested in seeing this one ship.  Thanks for getting the discussion started!

patrickmarabeas:
I'll start working on integrating the functionality: https://github.com/terraform-providers/terraform-provider-github/pull/613

jspiro:
Is there a simple string I can generate by hand that encodes OAuth2 info in it that will drop-in work in the existing provider? The docs say
> token - (Optional) A GitHub OAuth / Personal Access Token. When not provided or made available via the GITHUB_TOKEN environment variable, the provider can only access resources available anonymously.
But I can't figure out what to give it. We're not OAuth2 knowledgable here, so any help or examples would be deeply appreciated. We're hitting our API limits constantly and need to switch over.

[...]

Issue #509, opened 2020-07-07
[Feature Request] Manage GitHub Apps via Terraform


neurotronix:
Hey Team,

I was looking for a way to manage GitHub apps [1] via Terraform but quickly realized there is none. It would be great to be able to setup a GitHub app, here are some exemplary settings one might want to manage:
- app name
- permissions
- webhook URL and secret
- event subscriptions

Cheers!

### Terraform Version
```
Terraform v0.12.28
```

### References
[1] https://developer.github.com/apps/about-apps/

kfcampbell:
> It would be great to be able to setup a GitHub app

I've been playing around with this a little bit, and the public API support for GitHub Apps isn't where it needs to be in order to support standard CRUD operations on a GitHub App. I've spoken with the team internally and they're aware of the issue but it isn't an immediate priority. 

The GraphQL API doesn't support App operations at the moment.

The [auth](https://github.com/google/go-github/blob/7b61934d204908a0e393382443513aeaf4601cbf/scrape/scrape.go#L123) used for creating Apps on the `go-github` package uses screen-scraping, which requires a One Time Password and which might rule out using it as part of the Terraform provider; I have some reservations about the stability of such a package. 

In any event, I wanted to create a proof of concept, so I created a sample app with hard-coded values but I haven't been able to successfully get an App created. When skipping the OTP, I get a `406 Not Acceptable` each time, and when attempting to provide a OTP through my standard 2FA app, the client panics when authenticating.

There might be something I'm missing on how the screen-scraping package of the `google/go-github` module is supposed to work. If there is, I'd love to hear about it!

In the meantime, we'll look into what operations we can support given an existing App.

kfcampbell:
I've investigated a little bit more on how to programmatically support the requested GitHub Apps operations. 

> * app name

I can't find anything in the API documentation about changing the name. I don't think the API that works behind the UI is public.
 
> * permissions

Can be managed when [creating an installation](https://docs.github.com/en/rest/reference/apps#create-an-installation-access-token-for-an-app), but updating doesn't appear to be supported in the public API either.

> * webhook URL and secret

[REST docs are here](https://docs.github.com/en/rest/reference/apps#update-a-webhook-configuration-for-an-app), but that functionality doesn't appear to be available in `go-github` yet.

> * event subscriptions

I believe these are managed through the same API as permissions above.

It definitely feels like we're working without tools on this feature a bit.

jcudit:
[...]

PR #690, opened 2021-02-03
Add github_app_installation_repository resource

k24dizzle:
Example Usage:
```
resource "github_app_installation_repository" "example" {
    repository      = "testrepo"
    installation_id = "1234567"
}
```

Gave https://github.com/integrations/terraform-provider-github/issues/687 a shot. However, while this implementation works in my local testing (I was able to install an existing app to existing repos), this PR is incomplete.

I was going to write acceptance tests for this PR, but realized that there is no way to spin up an organization that already has an app installed. We would need to write a corresponding resource that installs github apps in an organization, but I don't believe that there is an easy way to do that through the Github API.

Any ideas on potential next steps/feedback would be greatly appreciated!


jcudit:
Looking great so far!  A couple of suggestions ahead of merging this:
- Since this is currently difficult to acceptance test, we could benefit from a new example in the `examples/` directory
- We'll need to document this similar to other resources (see `website/docs/r` and `website/github.erb`)

❤️  the direction here.  Feels like we're closer to shipping https://github.com/integrations/terraform-provider-github/issues/509 and https://github.com/integrations/terraform-provider-github/issues/514 with this feature in place.  Thanks for driving this!

k24dizzle:
Thanks for the comments @jcudit, went ahead and implemented your feedback. Let me know if I can do anything else/if you have more feedback!

k24dizzle:
Hi @jcudit I was wondering if there were any updates to this PR, feel free to let me know if there's any more work to be done on this

jcudit:
Thanks for the ping 🙇🏾 .  A few of us are going to pair on reviews this week.  Aiming to get this one tested out during the session and will likely ship with the next release.

jcudit:
A few of us are going to pair on reviews this week.  Aiming to get this one tested out during the session and return here with next steps.  🤞🏾 that things just work and we can get to shipping.

github/resource_github_app_installation_repository.go
```diff
@@ -0,0 +1,146 @@
...
+
+	"github.com/google/go-github/v32/github"
+	"github.com/hashicorp/terraform-plugin-sdk/helper/schema"
+)
+
+func resourceGithubAppInstallationRepository() *schema.Resource {
+	return &schema.Resource{
+		Create: resourceGithubAppInstallationRepositoryCreate,
+		Read:   resourceGithubAppInstallationRepositoryRead,
+		Delete: resourceGithubAppInstallationRepositoryDelete,
```
jcudit, line 16: 🤔 would we benefit from an `Update` function here as well? Or is creating a new App installation usually non-disruptive?

website/docs/r/app_installation_repository.html.markdown
```diff
@@ -0,0 +1,49 @@
...
+# Create a repository.
+resource "github_repository" "some_repo" {
+  name = "some-repo"
+}
+
+resource "github_app_installation_repository" "some_app_repo" {
+  # The installation id of the app (in the organization).
+  installation_id    = "1234567"
+  repository         = "${github_repository.some_repo.name}"
+}
```
jcudit, line 32: Can we get a test that exercises this configuration?  Open to updates in HCL syntax as well.

jcudit (commented):
Excited to see this one land.  Thanks for getting it this far and being patient on the feedback 🙇🏾 .

website/docs/r/app_installation_repository.html.markdown
jcudit: 🤦🏾  I now see an earlier comment where I requested an example over an acceptance test.  I'll give testing a shot and plan to move forward with this with just the example if the test does not pan out.

jcudit (approved):
This tested well locally and I've got eaaaf6561722b11dd8fe8c86e0bbfee7733874c3 queued up to commit after this merges as-is.  Thanks again for this contribution and your patience up to this point @k24dizzle.

Issue #675, opened 2021-01-26
Modify `github_team_repository` to accept team slug as well as team id


k24dizzle:
We have a use case where we would like users to add teams to their repos, and having them being able to do so through a slug rather than an id would make the configuration more readable.

Would it be feasible to modify the `github_team_repository` resource to take in a team slug rather than a team id? 
https://registry.terraform.io/providers/integrations/github/latest/docs/resources/team_repository

This would be a similar change to the `github_branch_protection` resource taking in either a Node ID or name of a repository.

I was thinking about adding a piece of logic that would try to find the team ID if the given team ID is not an integer (in which case we assume that the user has passed in a slug instead of an ID).

*There is an edge case where a user tries to add a team repository resource with a team name composed of an integer (123), in which case the provider would assume that the user passed in a team id rather than a team slug.

usmonster:
Hello! Hasn't this been closed by #693?

Also, has there been any discussion or thought of applying this as a generic pattern, i.e. accepting a slug/name anywhere an ID is currently required, wherever possible? I have a specific use case in mind, but it would be nice to do this consistently. (I might open another issue or PR if it makes sense to anyone else but myself.) Thanks!

PR #693, opened 2021-02-05
Modify github_team_repository to accept slug as a valid team_id as well

k24dizzle:
Attempt to address: https://github.com/integrations/terraform-provider-github/issues/675

Was tested locally, and I attempted to modify the acceptance tests to test this new behavior.

However, I wasn't able to get the acceptance tests to run locally, even after creating a new org called `kzhaotest` and setting the `GITHUB_TOKEN` and `GITHUB_ORGANIZATION` env variables:
```
kzhao@kzhao-mbp151 ~/go/src/github.com/k24dizzle/terraform-provider-github (teamid) $ TF_LOG=DEBUG TF_ACC=1  go test -v   ./... -run ^TestAccGithubTeamRepository_basic
=== RUN   TestAccGithubTeamRepository_basic
2021/02/04 16:18:07 [ERROR] Github API Request error: &http.badStringError{what:"unsupported protocol scheme", str:""}
2021/02/04 16:18:07 [TRACE] Acquiring lock for GitHub API request (%!q(<nil>))
2021/02/04 16:18:07 [TRACE] Releasing lock for GitHub API request (%!q(<nil>))
    TestAccGithubTeamRepository_basic: resource_github_team_repository_test.go:17: Skipping because GITHUB_OWNER "kzhaotest" is a user, not an organization.
--- SKIP: TestAccGithubTeamRepository_basic (0.00s)
PASS
ok  	github.com/terraform-providers/terraform-provider-github/github	(cached)
```

Open to any feedback/comments

### DIFF
--- a/github/resource_github_team_repository.go
+++ b/github/resource_github_team_repository.go
@@ -22,10 +22,10 @@ func resourceGithubTeamRepository() *schema.Resource {
 
 		Schema: map[string]*schema.Schema{
 			"team_id": {
-				Type:         schema.TypeString,
-				Required:     true,
-				ForceNew:     true,
-				Description:  "ID or slug of team",
+				Type:        schema.TypeString,
+				Required:    true,
+				ForceNew:    true,
+				Description: "ID or slug of team",
 			},
 			"repository": {
 				Type:     schema.TypeString,
@@ -58,6 +58,9 @@ func resourceGithubTeamRepositoryCreate(d *schema.ResourceData, meta interface{}
 	// The given team id could be an id or a slug
 	givenTeamId := d.Get("team_id").(string)
 	teamId, err := getTeamID(givenTeamId, meta)
+	if err != nil {
+		return err
+	}
 
 	orgName := meta.(*Owner).name
 	repoName := d.Get("repository").(string)
--- a/github/resource_github_team_repository.go
+++ b/github/resource_github_team_repository.go
@@ -25,7 +25,7 @@ func resourceGithubTeamRepository() *schema.Resource {
 				Type:         schema.TypeString,
 				Required:     true,
 				ForceNew:     true,
-				ValidateFunc: validateTeamIDFunc,
+				Description:  "ID or slug of team",
 			},
 			"repository": {
 				Type:     schema.TypeString,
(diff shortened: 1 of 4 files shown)


github/resource_github_team_repository.go
```diff
@@ -128,7 +129,11 @@ func resourceGithubTeamRepositoryRead(d *schema.ResourceData, meta interface{})
 	}
 
 	d.Set("etag", resp.Header.Get("ETag"))
-	d.Set("team_id", teamIdString)
+	if d.Get("team_id") == "" {
+		// If team_id is empty, that means we are importing the resource.
+		// Set the team_id to be the id of the team.
+		d.Set("team_id", teamIdString)
+	}
```
jcudit, line 136: This seems like the riskiest bit.  What was your experience leading up to this diff?

jcudit (commented):
Looks good!  Had one question but seems safe to 🚢 

This resource had not gotten it's tests updated to work with the latest test suite.  Please cherry-pick a122f316 to get tests passing.

github/resource_github_team_repository.go
k24dizzle: In my testing, I didn't have this if statement at first.

I tried setting the `team_id` to be a team slug, and was able to plan and apply the following config to create a new team repo connection:
```
resource "github_team_repository" "test" {
    team_id = "sampleteam"
    repository = "testteamrepo"
    permission = "push"
}
```

When I tried another plan after that I expected that terraform would say that there would be no changes necessary. However, I got the following output:
```

  # github_team_repository.test must be replaced
-/+ resource "github_team_repository" "test" {
      ~ etag       = "W/\"6469d39bbad910892556bd4bb130c23ab7ffb655ec6a312d592f2e76736d67c9\"" -> (known after apply)
      ~ id         = "4466024:testteamrepo" -> (known after apply)
        permission = "push"
        repository = "testteamrepo"
      ~ team_id    = "4466024" -> "123" # forces replacement
    }

```

When terraform refreshed the state, using this `resourceGithubTeamRepositoryRead` function, it would replace the `team_id` from a slug to the actual team id, making it think that it would need to replace this resource.

So I added this if statement to avoid this situation. The only situation I could think of where we would want to actually set this `team_id` field here, would be when we were importing a team_repo resource.

k24dizzle:
Also thanks for updating the tests! That helps a lot! I went ahead and cherry-picked your commit into this PR

github/resource_github_team_repository.go
jcudit: Appreciate the detail here, thanks.

Issue #616, opened 2020-12-02
Do not destroy the repository when you change the repository name


smorimoto:
Is this intentional?

jcudit:
Unintentional.  We should support renaming a repository without having to destroy it.  Will track this for the next batch of `repository` resource updates.

smorimoto:
I also changed the local name at the same time, is that probably the cause? Ideally, even if you change both the local name and the repository name, it should not be destroyed.

k24dizzle:
Hi, I went ahead and gave this a shot here: https://github.com/integrations/terraform-provider-github/pull/699. Would appreciate any feedback!

smorimoto:
Cool, I will take a look at it when I have much spare time 🙂

majormoses:
This is actually quite an old issue and something that was not exactly intentional and more a set of unfortunate set of limitations of the (documented) api. See #55 for more information.

k24dizzle:
Seems like the main issue here is that the `github_repository` resource uses the `name` as the id of the resource, which we want to change, rather than the `repo_id` which shouldn't change. 

I wonder how far we could get by removing the `ForceNew` attribute from the `name` field. In my testing (in my PR above), removing that allows you to rename a repository. Behind the scenes, it will assign the resource a new id, since there is a new name, but are there any implications of changing the id of a terraform resource?

cc: @jcudit

jcudit:
Removing (or making configurable) the `ForceNew` is a reasonable first step to get this functionality.  My worry too is with unforeseen effects of changing the ID of a `repository` resource.  I recommend we enable this behaviour defaulted to off so that consumers can test or ignore the feature at their judgement.  

Aiming to land some suggestions on the attached PR to keep it moving.

devinrsmith:
I think this issues has been fixed by #699 

I've planned out a run, and now I see `github_repository` being updated in-place :)

Unfortunately, I think there are a lot of other resources that reference the repository name, and would be better served by referencing a repository id (#744).

From our own usage, we see the following resources planned to be destroyed during a repo rename: `github_branch`, `github_branch_default`, `github_branch_protection_v3`, `github_repository_collaborator`, and `github_team_repository`. (There are probably other relevant resources that need to be updated, that we don't use.)

I'm assuming I'll need to go through a rename and manual import step instead of a normal terraform apply to resolve this currently.

Caleb-Hoffman:
Hi @kfcampbell, just curious if there are any updates on this?

[...]

PR #699, opened 2021-02-11
Do not destroy the repository when you change the repository name

k24dizzle:
Inspired by the discussion in here: https://github.com/integrations/terraform-provider-github/issues/616, the change for this was easier than I expected. I just made it so that changing the `name` field wouldn't force a new resource.

I tested this change locally using the following terraform config:
```
  name        = "renamerepo"
  description = "testing renaming a repo"
  auto_init   = true
}
```

After planning and applying the config above, I went ahead and changed the config to:
```
  name        = "renamerepo2"
  description = "testing renaming a repo"
  auto_init   = true
}
```
The following [gist](https://gist.github.com/k24dizzle/5734e616c8131cdd43fa424a1b53672e) contains the resulting plan and apply of the provider `before` and `after` my changes.

I also went ahead and added a unit test for this:
```
    --- PASS: TestAccGithubRepositories/updates_a_repositories_name_without_error (10.03s)
        --- SKIP: TestAccGithubRepositories/updates_a_repositories_name_without_error/with_an_anonymous_account (0.00s)
        --- SKIP: TestAccGithubRepositories/updates_a_repositories_name_without_error/with_an_individual_account (0.00s)
        --- PASS: TestAccGithubRepositories/updates_a_repositories_name_without_error/with_an_organization_account (10.03s)
```

### DIFF
--- a/github/resource_github_repository_test.go
+++ b/github/resource_github_repository_test.go
@@ -68,6 +68,69 @@ func TestAccGithubRepositories(t *testing.T) {
 
 	})
 
+
+	t.Run("updates a repositories name without error", func(t *testing.T) {
+
+		oldName := fmt.Sprintf(`tf-acc-test-rename-%[1]s`, randomID)
+		newName := fmt.Sprintf(`%[1]s-renamed`, oldName)
+
+		config := fmt.Sprintf(`
+			resource "github_repository" "test" {
+			  name         = "%[1]s"
+			  description  = "Terraform acceptance tests %[2]s"
+			}
+		`, oldName, randomID)
+
+		checks := map[string]resource.TestCheckFunc{
+			"before": resource.ComposeTestCheckFunc(
+				resource.TestCheckResourceAttr(
+					"github_repository.test", "name",
+					oldName,
+				),
+			),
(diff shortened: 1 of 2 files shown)


github/resource_github_repository_test.go
```diff
@@ -68,6 +68,69 @@ func TestAccGithubRepositories(t *testing.T) {
...
+			"after": resource.ComposeTestCheckFunc(
+				resource.TestCheckResourceAttr(
+					"github_repository.test", "name",
+					newName,
+				),
+			),
+		}
+
+		testCase := func(t *testing.T, mode string) {
+			resource.Test(t, resource.TestCase{
```
k24dizzle, line 100: Is there a way to add a check in here that the existing resource wasn't destroyed/a new resource wasn't created?

jcudit: We manually verified that the rename operation did not cause a repository deletion by viewing audit logs for our test org:

CleanShot 2021-04-09 at 10 32 45@2x

jcudit (approved):
Given this added test and some manual verification, this safe to be released for further testing in the community.

PR #714, opened 2021-03-01
validate actions secret names

bendrucker:
This PR implements the secret naming rules described in the GitHub docs:

https://docs.github.com/en/actions/reference/encrypted-secrets#naming-your-secrets

Currently, if a user supplies an invalid secret name, it will be rejected at apply time:

```
422 Failed to add secret. Name is invalid []
```

Instead, this makes naming violations raise a clear error at validate/plan time. I've added unit tests against the validation function, similar to how other shared validation functions are handled. Happy to add a full acceptance test for an invalid case if that's of interest.

### DIFF
--- a/github/resource_github_actions_organization_secret.go
+++ b/github/resource_github_actions_organization_secret.go
@@ -20,9 +20,10 @@ func resourceGithubActionsOrganizationSecret() *schema.Resource {
 
 		Schema: map[string]*schema.Schema{
 			"secret_name": {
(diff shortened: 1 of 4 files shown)


issyl0:
Thanks for this PR, @bendrucker! I really like this feature - adding error messages is a positive step forward in user friendliness, and the implementation looks OK to me.

jcudit (approved):

PR #717, opened 2021-03-03
ci: Switch from Travis to GitHub Actions

issyl0:
- The Travis CI YAML specified Ubuntu Trusty as a runner OS, which is very end-of-life, so I switched those checks to GitHub Actions for consistency with the other jobs here (acceptance tests, etc).
- You can see a test workflow run at https://github.com/issyl0/terraform-provider-github/runs/2024800600.
- I didn't know if we wanted to do matrix builds to test multiple versions of Go, but that can be a goal for the future if we do. The development instructions for this provider still say Go 1.13.

----

- I've deleted the `.travis.yml` file, but obviously its absence once this hits master will affect CI runs on non-rebased existing PRs. Not sure how to avoid that?


jcudit (approved):
In the past the ecosystem has been cautious about moving forward Go versions and this project specifically has bumped versions when needing to stay compatible with our dependencies.  Open to bumping versions to take advantage of new language features or gain other benefits, but won't block this PR on it.

To be extra cautious we could run both the Action and Travis for a few releases.  Happy to revert if this causes friction for someone down the line as well.  Will pause on merging for now to allow for more feedback to come in from the community, but overall onboard with this change.

PR #725, opened 2021-03-11
Add github_teams data source

reinoudk:
The `github_teams` data source is useful to get all teams of an organization without having prior knowledge of their names/slug. 
One of the use cases is to synchronize these teams and their members to some external system.

This uses the GraphQL API and returns all teams of an organization including their members in a single request per 100 teams.

### DIFF
--- a/github/data_source_github_organization_teams.go
+++ b/github/data_source_github_organization_teams.go
@@ -1,7 +1,6 @@
 package github
 
 import (
-	"fmt"
 	"github.com/hashicorp/terraform-plugin-sdk/helper/schema"
 	"github.com/shurcooL/githubv4"
 	"log"
@@ -94,9 +93,7 @@ func dataSourceGithubOrganizationTeamsRead(d *schema.ResourceData, meta interfac
 		variables["cursor"] = githubv4.NewString(query.Organization.Teams.PageInfo.EndCursor)
 	}
 
-	// Use the organization ID and value of rootTeamsOnly to create a unique ID.
-	// This supports the edge case where the same data source is used twice: once with and once without rootTeamsOnly=true.
-	d.SetId(fmt.Sprintf("%s-%t", string(query.Organization.ID), rootTeamsOnly))
+	d.SetId(string(query.Organization.ID))
 	d.Set("teams", teams)
 
 	return nil
--- a/github/data_source_github_organization_teams.go
+++ b/github/data_source_github_organization_teams.go
@@ -1,11 +1,10 @@
 package github
 
 import (
+	"fmt"
 	"github.com/hashicorp/terraform-plugin-sdk/helper/schema"
 	"github.com/shurcooL/githubv4"
 	"log"
-	"strconv"
-	"time"
 )
 
 func dataSourceGithubOrganizationTeams() *schema.Resource {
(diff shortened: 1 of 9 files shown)


majormoses:
When I read the data source name I make a leap that I can ask for specific teams when this only looks to solve the "all" use case. I think that should be in the name of the data source if that is the scope of this source.

There is a data source for a single `github_team`, it does not provide a list of child teams or a way to traverse it. With a data source named as such I would expect that I can optionally provide a specific parent to start from and walk all the child objects.

reinoudk:
@majormoses Good point. Maybe it would be good to discuss this in #726 ?

reinoudk:
I've changed the name of the data source to `github_organization_teams`, based on the discussion in #726. I'd like to add a `github_team_child_teams` data source as well, but will create a different PR for that.

jcudit:
Seems on the right track from what I'm reading over in the related issue and from the code itself.  Can we also get an update to `website/github.erb` to add a link out to the new docs that are added?

reinoudk:
Ah, somehow missed that. Fixed.

majormoses (approved):
LGTM

github/data_source_github_organization_teams.go
```diff
@@ -0,0 +1,136 @@
...
+		additionalTeams := flattenGitHubTeams(query)
+		teams = append(teams, additionalTeams...)
+
+		if !query.Organization.Teams.PageInfo.HasNextPage {
+			break
+		}
+		variables["cursor"] = githubv4.NewString(query.Organization.Teams.PageInfo.EndCursor)
+	}
+
+	d.SetId(strconv.FormatInt(time.Now().Unix(), 10))
```
jcudit, line 98: Interesting approach here.  🤔  I'm unsure if this will cause unwanted behaviour and have not seen this pattern used before.  👍🏾  on trying it but we'll need to keep an eye out for fixes that may need to occur due to choosing what seems to be a non-deterministic ID.  

I see a similar data source pins to the overall org ID:

```go
// github/data_source_github_organization.go
d.SetId(strconv.FormatInt(organization.GetID(), 10))
```

jcudit (approved):
Added a comment about the ID chosen, but overall looks like a quality contribution!  Will pause for a couple of days in case there is more feedback but plan on shipping this as-is with the next release.

github/data_source_github_organization_teams.go
reinoudk: Sorry, added another change. In the previous commit I added the value of `root_teams_only` to the ID so it would be unique between multiple instances of the data source (edge case). But that is of course not necessary, so I simplified that again.

github/data_source_github_organization_teams.go
```diff
@@ -95,7 +94,9 @@ func dataSourceGithubOrganizationTeamsRead(d *schema.ResourceData, meta interfac
 		variables["cursor"] = githubv4.NewString(query.Organization.Teams.PageInfo.EndCursor)
 	}
 
-	d.SetId(strconv.FormatInt(time.Now().Unix(), 10))
+	// Use the organization ID and value of rootTeamsOnly to create a unique ID.
+	// This supports the edge case where the same data source is used twice: once with and once without rootTeamsOnly=true.
```
jcudit: Nice find, thanks for getting this in.

PR #734, opened 2021-03-18
Give a helpful error message when trying to import a github_branch that does not exist

tibbes:
I noticed that in issue #636, errors are not reported properly when importing a `github_branch` resource. In particular, the case where the branch does not exist on GitHub is not handled by the provider and the user gets a Terraform error instead:

```
Error: nil entry in ImportState results. This is always a bug with
the resource that is being imported. Please report this as
a bug to Terraform.
```

I've updated this message to:

```
Error: Repository tftest does not have a branch named nosuchbranch.
```


jcudit (approved):
Good thought!  Hopefully saves someone trouble down the line.

PR #735, opened 2021-03-19
Correct documentation of provider configuration, and deprecate organization argument

tibbes:
While investigating #662, I found that the code handling the `owner` and `organization` arguments to the GitHub provider could be simplified, and also that the documentation was incorrect.

Although the docs say that `owner` is for user ids (like `torvalds`) and `organization` is for organization ids (like `github`), in fact, either type of value can be put in either argument.

Given this, I suggest that we deprecate `organization`, and just use `owner`.

The main change in this pull request is to make `organization` and `GITHUB_ORGANIZATION` deprecated. Making arguments deprecated should only me done as part of a minor (or higher) release, according to the Hashicorp [versioning guidelines](https://www.terraform.io/docs/extend/best-practices/versioning.html#example-minor-number-increments).

In addition, the documentation says that `owner` can be set via the environment variable `GITHUB_OWNER`, and `organization` can be set via the environment variable `GITHUB_ORGANIZATION`. However, in reality, _both_ environment variables (with `GITHUB_ORGANIZATION` taking priority) were used as the default value of _both_ `owner` and `organization`. This had the odd side-effect of making the precedence of these methods inconsistent. From high to low precedence, the order is:

- setting `organization` in the provider configuration
- setting the `GITHUB_ORGANIZATION` environment variable
- setting the `GITHUB_OWNER` environment variable
- setting `owner` in the provider configuration

That is, `organization` overrides `GITHUB_ORGANIZATION` (as I'd expect), but `GITHUB_OWNER` overrides `owner` (that seems backwards).

Changing this behaviour seems desirable, but as it would be a backwards incompatible change, the Hashicorp [versioning guidelines](https://www.terraform.io/docs/extend/best-practices/versioning.html#example-major-number-increments) suggest we should only do so as part of a major release. Therefore, I *have not* changed the configuration precedence in this pull request, but I have put a warning in the documentation saying we might do so in future.


majormoses (approved):
Not much of a go developer, LGTM. It's a good solution to a bad problem, I would ideally like to see this release go out and then decide when we are willing to force the breaking change. I honestly don't see a big reason to wait long, if we have breaking changes that affect a small area it's much easier to review changelogs and upgrade then it is with large releases with multiple breaking changes.

jcudit (approved):
I think this is a necessary change to get into the project to clean up from past mistakes.  A huge thank you @tibbes for such a readable fix.  I feel this will remove more frustration than it causes, so lets ship and track the removal of the backwards compatibility for the next major release in the coming months.

Changes to this part of the code base have historically not gone smoothly, so I'll be keeping an eye out for related bugs after this is released.  Thanks to anyone who can quickly point out fixes as we upgrade this area. 🚀

Issue #286, opened 2019-10-18
[Feature Request] Pull Requests and `git` branch publishes


jcudit:
### Overview

👋 from GitHub's datacenter team.  We are requesting guidance on functionality for this pair of features:

- Add limited support for the [Git Data API](https://developer.github.com/v3/git/)
  - [ ] Assemble a git tree from one or many [templatefile](https://www.terraform.io/docs/configuration/functions/templatefile.html) outputs
  - [ ] [Publish](https://github.com/google/go-github/blob/v28.0.1/github/git_trees.go#L75-L99) an assembled git tree (☝️)
  - [ ] Read a branch for the purposes of associating it to a PR (👇)

- Add limited support for the [Pull Requests API](https://developer.github.com/v3/pulls/)
  - [ ] [Create](https://github.com/google/go-github/blob/v28.0.1/github/pulls.go#L253-L274)  a new Pull Request
  - [ ] [Merge](https://github.com/google/go-github/blob/v28.0.1/github/pulls.go#L38) a Pull Request when [mergable](https://github.com/google/go-github/blob/v28.0.1/github/pulls.go#L38)

- Package features into a quality 🚢 
  - [ ] Acceptance testing
  - [ ] Documentation external to the resource code itself
  - [ ] Lint for existing style and standards
  - [ ] Review for breaking changes to components that may already be in use in the wild
    - [ ] config files
    - [ ] Terraform states

cc https://github.com/terraform-providers/terraform-provider-github/issues/254 (related solution to initializing a repo with Terraform)

### Suggested UX

```hcl
resource "github_repository_content" "actions" {

  repository = "${local.repository_name}"

  # Optional Pull Requesst overrides of sane defaults
  pull_request {
    title = "Create standard Actions for this repo"
    target_branch    = "master"
    working_branch   = "actions-by-terraform"

    body = "${local.create_actions_body}"
    issue = "${local.create_actions_issue}"
  }

  # Declare content as a map of paths to rendered templates
  content = {
    ".github/workflows/build.yml" = templatefile("path/to/build.yml", { foo = "bar" }),
    ".github/workflows/release.yml" = templatefile("path/to/release.yml", { foo = "bar" }),
  }
}
```

This resource block accomplishes the following:
- Consumes a list of files (`content` map)
- Obtains a local clone of the repository specified by `repository`
- Renders all content to the repository worktree with a base of `target_branch`
- Commits changes (if any) and publishes a Pull Request using the `working_branch`

[...]

PR #739, opened 2021-03-27
Support Pull Requests

marcinwyszynski:
# Use case

Frankly, I only came here to get a list of open pull requests from Terraform (the `github_repository_pull_requests` data source), but I wouldn't be able to test it properly without the ability to create PRs from Terraform, so here we go. My immediate use case for this feature is test/QA/preview (every company has their own name) environments automatically created from open Pull Requests and kept alive while these remain open. It goes something like this:

```hcl
data "github_repository_pull_requests" "this" {
  base_repository = "infra"
  base_ref        = "main"
}

module "pr_stacks" {
  for_each = toset([ for pr in data.github_repository_pull_requests.this.results : pr.head_ref ])

  source = "./source"
  branch = each.key
}
```

The `source` module can then contain the definition of a Terraform project in form of a [`spacelift_stack`](https://github.com/spacelift-io/terraform-provider-spacelift/#spacelift_stack-resource) or a [`tfe_workspace`](https://registry.terraform.io/providers/hashicorp/tfe/latest/docs/resources/workspace), pointing to the head branch of the PR. I'm sure there are plenty other use cases, but this is what I personally need it for.

# Notes

Below are some of the notes I've been taking when researching and working on this diff. Hopefully they can shed some light on my decisions and help with the review.

## Forks are supported

I assumed that the resource should support pull requests between forks. Not sure how frequent that use case would be in practice, but the extra cost was negligible. In the API, the "owner" field can be optionally supplied to explicitly indicating the owner of the repo to which we're trying to upstream our changes. I also called a field called "repository" in most other resources "base_repository" to make a clear distinction between (potentially) 2 repositories involved in creating a PR.

## Deleting Pull Requests

It's not entirely clear how to treat PR deletion according to Terraform's CRUD semantics. The approach I've taken here is to close the PR unless it's already closed or merged in GitHub. Merging it (a "positive" action) feels intuitively wrong in what effectively is a destructor (a "negative" action).

## Drafts

Creating drafts is not supported. While the v3 API call itself allows you mark the PR as draft when creating it, it does not allow you to change this setting afterwards. Hence, Terraform cannot manage the lifecycle of this field, and I decided to make it computed instead.


jcudit:
🎉  excited for this one, thanks for the contribution!

I'm overall 👍🏾  on the data sources, but would like to think more about the new pull request resource.  This project has considered [adding a resource to manage Pull Requests](https://github.com/integrations/terraform-provider-github/issues/286) in the past but shied away from it in favour of the `github_repository_file` resource.  I suspect adding this resource will give users an attractive option and we'll end up getting more bug reports and feature requests to handle creative edge cases.  Feels like something that comes with a heavier maintenance burden than other types of resources.  Not against it, just hoping to hear more from the community around requirements so we can minimize re-work when initially rolling this out.

I've added this to the milestone after next and hope to hear more opinions around the implementation when revisiting.  🙇🏾  for the patience while this one works its way through the release process.

marcinwyszynski:
Thanks @jcudit, please LMK if I can be of any assistance.

github/data_source_github_repository_pull_requests.go
```diff
@@ -0,0 +1,227 @@
...
+			},
+			"results": {
+				Type:     schema.TypeList,
+				Computed: true,
+				Elem: &schema.Resource{
+					Schema: map[string]*schema.Schema{
+						"number": {
+							Type:        schema.TypeInt,
+							Computed:    true,
+							Description: "Per-repository, monononically increasing ID of this PR",
```
jcudit, line 59:
```suggestion
							Description: "Per-repository, monotonically increasing ID of this PR",
```

github/resource_github_repository_pull_request.go
```diff
@@ -0,0 +1,294 @@
...
+	}
+
+	d.Set("number", pullRequest.GetNumber())
+
+	if head := pullRequest.GetHead(); head != nil {
+		d.Set("head_ref", head.GetRef())
+		d.Set("head_sha", head.GetSHA())
+	} else {
+		// Totally unexpected condition. Better do that than segfault, I guess?
+		log.Printf("[WARN] Head branch missing, expected %s", d.Get("head_ref"))
```
jcudit, line 183: 👍🏾  agreed.

github/resource_github_repository_pull_request.go
```diff
@@ -0,0 +1,294 @@
...
+	}
+
+	return fmt.Errorf(strings.Join(errors, ", "))
+}
+
+func resourceGithubRepositoryPullRequestDelete(d *schema.ResourceData, meta interface{}) error {
+	// It's not entirely clear how to treat PR deletion according to Terraform's
+	// CRUD semantics. The approach we're taking here is to close the PR unless
+	// it's already closed or merged. Merging it feels intuitively wrong in what
+	// effectively is a destructor.
```
jcudit, line 259: 👍🏾  lets not merge by default.  Perhaps we can provide an option to do so in the future if this becomes a common ask.

jcudit (approved):
Thanks again for making these available for everyone!  Excited to see how these resources get used in the wild.

PR #740, opened 2021-03-27
Fix handling Action secrets

marcinwyszynski:
Sorry folks, not entirely sure how to properly write test cases to capture the bug and the fix. Hope though that at least the problem becomes immediately clear from the fix :)

### DIFF
--- a/github/resource_github_actions_organization_secret.go
+++ b/github/resource_github_actions_organization_secret.go
@@ -131,8 +131,29 @@ func resourceGithubActionsOrganizationSecretRead(d *schema.ResourceData, meta in
 	}
 
 	d.Set("plaintext_value", d.Get("plaintext_value"))
-	d.Set("updated_at", secret.UpdatedAt.String())
 	d.Set("created_at", secret.CreatedAt.String())
+
+	// This is a drift detection mechanism based on timestamps.
+	//
+	// If we do not currently store the "updated_at" field, it means we've only
+	// just created the resource and the value is most likely what we want it to
+	// be.
+	//
+	// If the resource is changed externally in the meantime then reading back
+	// the last update timestamp will return a result different than the
+	// timestamp we've persisted in the state. In that case, we can no longer
+	// trust that the value (which we don't see) is equal to what we've declared
+	// previously.
+	//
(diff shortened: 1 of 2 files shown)


github/resource_github_actions_organization_secret.go
```diff
@@ -131,9 +131,30 @@ func resourceGithubActionsOrganizationSecretRead(d *schema.ResourceData, meta in
...
+	// timestamp we've persisted in the state. In that case, we can no longer
+	// trust that the value (which we don't see) is equal to what we've declared
+	// previously.
+	//
+	// The only solution to enforce consistency between is to mark the resource
+	// as deleted (unset the ID) in order to fix potential drift by recreating
+	// the resource.
+	if updatedAt, ok := d.GetOk("updated_at"); ok && updatedAt != secret.UpdatedAt.String() {
+		log.Printf("[WARN] The secret %s has been externally updated in GitHub", d.Id())
+		d.SetId("")
```
jcudit, line 153: So we prefer the manually updated secret over what we declare via HCL?  This may conflict with the expectations of some users as terraform configuration is authoritative for some deployments.  

I can see an additional configuration parameter being useful here, but am curious to get more opinions on how to proceed.

jcudit (commented):
Thanks for the detailed comments -- I agree this is a difficult one to test and we'd likely have to hit the GitHub API outside of our normal testing helpers to reproduce the scenario.

I've left a comment with a suggested path we can take from here and will look out for feedback.

github/resource_github_actions_organization_secret.go
marcinwyszynski: No, it's the exact opposite - we actually rewrite the manually updated one with what's declared by HCL.
Notice how we unset the ID here if the resource changes externally, meaning that it will be recreated while it currently isn't. This change is fixing the bug that would go against the very expectations you're talking about.

jcudit (approved):

github/resource_github_actions_organization_secret.go
jcudit: Thanks for the additional explanation.  Clearer to me now on second read 🙇🏾

PR #745, opened 2021-03-30
Add import function for github_actions_organization_secret resource

xun-guo-anzx:
## Use Cases
Importer function will be useful when migrating existing Github organisational secrets to Terraform resources. This will allow importing existing organisation secrets into Terraform state along with it's visibility status and list of repositories when visibility is `selected`.

One defect of this implementation is it won't fetch value for `plaintext_value` for the secret to the state file. So would recommend using this along with resources ignoring changes for `plaintext_value`.

## Changes include:
* Create importer function for resource github_actions_organization_secret
* Import organization secrets along with selected repository IDs list if visibility is `selected`

### DIFF
--- a/github/resource_github_actions_organization_secret.go
+++ b/github/resource_github_actions_organization_secret.go
@@ -17,6 +17,12 @@ func resourceGithubActionsOrganizationSecret() *schema.Resource {
 		Read:   resourceGithubActionsOrganizationSecretRead,
 		Update: resourceGithubActionsOrganizationSecretCreateOrUpdate,
 		Delete: resourceGithubActionsOrganizationSecretDelete,
+		Importer: &schema.ResourceImporter{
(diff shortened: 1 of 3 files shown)


jcudit (commented):
Thanks for the contribution!  A couple of things:

- Can you cherry-pick 0d5bf215 to pull in tests and docs for this?
- I think 👇🏾  is OK to document and defer until we see friction reported by the community.
  > One defect of this implementation is it won't fetch value for `plaintext_value` for the secret to the state file. So would recommend using this along with resources ignoring changes for `plaintext_value`.

xun-guo-anzx:
Thanks @jcudit for update and recommendation. I've updated the branch with cherry-picked commit.

jcudit (approved):

PR #757, opened 2021-04-12
team: mark slug as computed when name changes

bendrucker:
This configures Terraform to mark `github_team.*.slug` as computed any time `name` changes, ensuring that it shows any resource that is dependent on `slug` as requiring an update.

I spotted this need when using `terraform import` with an existing team that had a different name than what was declared in configuration. Other resources in the same configuration ([`team_sync_group_mapping`](https://registry.terraform.io/providers/integrations/github/latest/docs/resources/team_sync_group_mapping) in my case) that depended on that slug still showed the current slug in the plan. Without this hint from `CustomizeDiff`, Terraform can't know that the computed value of `slug` will be different after the name is updated and the value in state (from the import) should not be used.

I used `customdiff.Sequence` as a general pattern here so that additional CustomizeDiffFuncs can be added for other fields. However, it could easily be removed too.

A test that has another resource that depends on `slug` provides coverage here. Using another team and sticking a `slug` in the description was convenient. Without the customized diff, the test fails:

```
==> Checking that code complies with gofmt requirements...
TF_ACC=1 go test $(go list ./... |grep -v 'vendor') -v -run TestAccGithubTeamUpdateName -timeout 120m
=== RUN   TestAccGithubTeamUpdateName
=== RUN   TestAccGithubTeamUpdateName/marks_the_slug_as_computed_when_the_name_changes
=== RUN   TestAccGithubTeamUpdateName/marks_the_slug_as_computed_when_the_name_changes/with_an_anonymous_account
    resource_github_team_test.go:199: anonymous account not supported for this operation
=== RUN   TestAccGithubTeamUpdateName/marks_the_slug_as_computed_when_the_name_changes/with_an_individual_account
    resource_github_team_test.go:203: individual account not supported for this operation
=== RUN   TestAccGithubTeamUpdateName/marks_the_slug_as_computed_when_the_name_changes/with_an_organization_account
    testing.go:654: Step 1 error: errors during apply:

        Error: Provider produced inconsistent final plan

        When expanding the plan for github_team.other to include new values learned so
        far during apply, provider "github" produced an invalid new value for
        .description: was cty.StringVal("tf-acc-kpzaa"), but now
        cty.StringVal("tf-acc-updated-kpzaa").

        issue tracker.

--- FAIL: TestAccGithubTeamUpdateName (8.71s)
    --- FAIL: TestAccGithubTeamUpdateName/marks_the_slug_as_computed_when_the_name_changes (8.71s)
        --- SKIP: TestAccGithubTeamUpdateName/marks_the_slug_as_computed_when_the_name_changes/with_an_anonymous_account (0.00s)
        --- SKIP: TestAccGithubTeamUpdateName/marks_the_slug_as_computed_when_the_name_changes/with_an_individual_account (0.00s)
        --- FAIL: TestAccGithubTeamUpdateName/marks_the_slug_as_computed_when_the_name_changes/with_an_organization_account (8.71s)
FAIL
FAIL	github.com/terraform-providers/terraform-provider-github/github	8.970s
FAIL
make: *** [testacc] Error 1
```

With the custom diff the test passes.

### DIFF
--- a/github/resource_github_team.go
+++ b/github/resource_github_team.go
@@ -7,6 +7,7 @@ import (
 	"strconv"
 
 	"github.com/google/go-github/v32/github"
+	"github.com/hashicorp/terraform-plugin-sdk/helper/customdiff"
 	"github.com/hashicorp/terraform-plugin-sdk/helper/schema"
 	"github.com/shurcooL/githubv4"
 )
@@ -20,6 +21,12 @@ func resourceGithubTeam() *schema.Resource {
 		Importer: &schema.ResourceImporter{
 			State: schema.ImportStatePassthrough,
 		},
+
+		CustomizeDiff: customdiff.Sequence(
+			customdiff.ComputedIf("slug", func(d *schema.ResourceDiff, meta interface{}) bool {
+				return d.HasChange("name")
+			}),
+		),
 
 		Schema: map[string]*schema.Schema{
 			"name": {
--- a/github/resource_github_team_test.go
+++ b/github/resource_github_team_test.go
@@ -151,3 +151,61 @@ func TestAccGithubTeamRemovesDefaultMaintainer(t *testing.T) {
 	})
 
 }
+
+func TestAccGithubTeamUpdateName(t *testing.T) {
+	randomID := acctest.RandStringFromCharSet(5, acctest.CharSetAlphaNum)
+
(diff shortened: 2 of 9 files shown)


jcudit (approved):

jcudit:
Much appreciation for getting this edge case covered.  I feel blind to this use case and am thankful to have contributions like this 🎉

PR #775, opened 2021-04-30
feat: add github_organization_member_privileges resource

Trois-Six:
This PR adds the `github_organization_member_privileges` resource to manage member privileges in the settings of an organization.
Destroying this resource won't do anything (just removing the resource from the state), as these settings are settings of an organization, and they always exist if an organization exists. For the very same reasons `create` and `update` do the very same things except that `create` creates the Id of the resource.
The official GitHub REST API is not up to date with what we can do in the current UI of github.com and the go client is even less up to date. I implemented what's supported in the current go client.


jcudit (commented):
This is on the right track, thanks for getting things started! Some feedback at this point:
- We'll need some documentation on the new resource.  See `website/docs/r/...` for examples as well as `website/github.erb` to link out to the new page.
- The test case looks valid but will not pass given how this project runs acceptance tests.  See `github/resource_github_team_test.go` as an example format that will pass as we need to specify the different scenarios a test will run as (anyonymous, individual, org).

jcudit:
Tested on a v2.21 GHES deployment and ran into the following failed test output:

>     testing.go:654: Step 0 error: Check failed: Check 1/1 error: got defaultRepoPermission ""; want "none"

🤔 still investigating, but feel free to send ideas.

nickfloyd:

PR #778, opened 2021-05-06
Missing GitHub repository data source attributes

cnuss:
Greetings! 

I've found in one of my projects that the Data Source for GitHub repositories continually needs to be refreshed, for example:

```
Terraform will perform the following actions:

  # module.bootstrap.module.github_config_files_public_websites["cdn"].data.github_repository.repository will be read during apply
  # (config refers to values not yet known)
 <= data "github_repository" "repository"  {
      ~ allow_merge_commit = true -> (known after apply)
      ~ allow_rebase_merge = true -> (known after apply)
      ~ allow_squash_merge = true -> (known after apply)
      ~ archived           = false -> (known after apply)
      ~ default_branch     = "main" -> (known after apply)
      + description        = (known after apply)
      ~ git_clone_url      = "git://github.com/futz-dev/cdn-web-cdn.git" -> (known after apply)
      ~ has_downloads      = false -> (known after apply)
      ~ has_issues         = false -> (known after apply)
      + has_projects       = (known after apply)
      ~ has_wiki           = false -> (known after apply)
      + homepage_url       = (known after apply)
      ~ html_url           = "https://github.com/futz-dev/cdn-web-cdn" -> (known after apply)
      ~ http_clone_url     = "https://github.com/futz-dev/cdn-web-cdn.git" -> (known after apply)
      ~ id                 = "cdn-web-cdn" -> (known after apply)
      - name               = "cdn-web-cdn" -> null
      ~ node_id            = "MDEwOlJlcG9zaXRvcnkzNjQ1MjI2MDI=" -> (known after apply)
      + pages              = (known after apply)
      ~ private            = true -> (known after apply)
      ~ repo_id            = 364522602 -> (known after apply)
      ~ ssh_clone_url      = "git@github.com:futz-dev/cdn-web-cdn.git" -> (known after apply)
      ~ svn_url            = "https://github.com/futz-dev/cdn-web-cdn" -> (known after apply)
      ~ topics             = [] -> (known after apply)
      ~ visibility         = "private" -> (known after apply)
        # (1 unchanged attribute hidden)
    }
```

As you can see in the output, the following attributes are causing a refresh to be made:
 - `name`
 - `description`
 - `has_projects`
 - `homepage_url`
 - `pages`

Upon diving into these I made the following fixes:
 - `name`: Should be `name` from the API, not the calculated `name`/`full_name` setting.
 - `description` is set to null when unset from the API
 - `has_projects`: Was missing from `dataSourceGithubRepositoryRead(...)`
[...]

PR #780, opened 2021-05-10
fix(data/repository): set name and full_name attributes as computed

yann-soubeyrand:
This avoids propagating null values when the evaluation of the data source is deferred.

In some rare situations where one is using the repository data source on a not yet existing repository (which will be created during terraform apply), the evaluation of this data source will be deferred. However, a resource using only its full_name attribute (if the name argument was set) or its name attribute (if the full_name argument was set), will immediately get a null value since these attributes are not set as computed. Setting these attributes to computed solves the problem.

### DIFF
--- a/github/data_source_github_repository.go
+++ b/github/data_source_github_repository.go
@@ -17,11 +17,13 @@ func dataSourceGithubRepository() *schema.Resource {
 			"full_name": {
 				Type:          schema.TypeString,
 				Optional:      true,
+				Computed:      true,
 				ConflictsWith: []string{"name"},
 			},
 			"name": {
 				Type:          schema.TypeString,
 				Optional:      true,
+				Computed:      true,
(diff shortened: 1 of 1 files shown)


cnuss:
@yann-soubeyrand looks like I submitted additional fixes in #778, should we combine forces?

yann-soubeyrand:
@cnuss sorry, I didn’t see your PR. I’ve no strong opinion for or against joining forces, as long as these changes end up in a released provider ;-)

jcudit:
Closing in favour of https://github.com/integrations/terraform-provider-github/pull/778.  @yann-soubeyrand please let me know if the new release does not fix your use case and I can get this one re-opened.

yann-soubeyrand:
@jcudit #778 doesn’t seem to include the changes of this PR, can we reopen this PR?

jcudit (approved):

PR #801, opened 2021-05-23
docs: improve examples and authentication info

loozhengyuan:
This PR improves the documentation examples by splitting into pre/post-0.12 code snippets so it will be clearer to new users on what the expected syntax should be. A new authentication section is added now that there is an option to choose between two authentication options. The arguments within the `app_auth` configuration block introduced in #753 is also expanded in detail.


jcudit (approved):
Quality PR - thanks so much for this!

Issue #803, opened 2021-05-25
Allow GitHub App PEM data to be passed directly


gusnuf:
The new GitHub app authentication feature launched in 0.4.10 is fantastic (thanks @alloveras), but it forces the application's PEM data to be available to Terraform as a file, the path of which must be passed via the `GITHUB_APP_PEM_FILE` environment variable.  This makes it unusable where the execution environment is not under the user's control, notably Terraform Cloud and Terraform Enterprise.

`GITHUB_APP_PEM_FILE` is already marked as containing sensitive data, so the simplest option is to allow PEM data to be passed in via this same variable.  PEM data is easily detectable as it has `-----` as a standard preamble.  If this preamble is seen, just use the data directly and don't bother trying to open the file and read its contents.

jspiro:
See #804!

gusnuf:
Fixed by #804

PR #804, opened 2021-05-27
[#803] Allow GitHub App PEM data to be passed directly

jspiro:
This is a generic (but breaking) solution to #803 that supports the current functionality via Terraform's `file` function to emulate previous behavior (wherein you provide a path to a PEM file).

This is cleaner and simpler than providing more named options for file contents, and the original behavior has only been around for a few days or weeks, so it's a good time to break it (and it's trivial to emulate with `file()`).

We've tested it with our real workspace in Terraform Cloud. It's also the only way we can use this feature, because that private key cannot be secured any other way and has root access to everything.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,9 @@
+## Unreleased
+
+BREAKING CHANGES:
+
+- Allow GitHub App PEM data to be passed directly ([#803](https://github.com/integrations/terraform-provider-github/issues/803))
+
 ## 4.10.1 (May 25, 2021)
 
 BUG FIXES:
--- a/github/apps.go
+++ b/github/apps.go
@@ -4,23 +4,20 @@ import (
 	"crypto/x509"
 	"encoding/json"
 	"encoding/pem"
+	"errors"
 	"fmt"
-	"gopkg.in/square/go-jose.v2"
-	"gopkg.in/square/go-jose.v2/jwt"
 	"io/ioutil"
 	"net/http"
 	"time"
+
+	"gopkg.in/square/go-jose.v2"
+	"gopkg.in/square/go-jose.v2/jwt"
 )
 
-// GenerateOAuthTokenFromApp generates a GitHub OAuth access token from a set of valid GitHub App credentials. The
-// returned token can be used to interact with both GitHub's REST and GraphQL APIs.
-func GenerateOAuthTokenFromApp(baseURL, appID, appInstallationID, appPemFile string) (string, error) {
-	pemData, err := ioutil.ReadFile(appPemFile)
-	if err != nil {
-		return "", err
-	}
-
-	appJWT, err := generateAppJWT(appID, time.Now(), pemData)
+// GenerateOAuthTokenFromApp generates a GitHub OAuth access token from a set of valid GitHub App credentials.
+// The returned token can be used to interact with both GitHub's REST and GraphQL APIs.
+func GenerateOAuthTokenFromApp(baseURL, appID, appInstallationID, pemData string) (string, error) {
+	appJWT, err := generateAppJWT(appID, time.Now(), []byte(pemData))
 	if err != nil {
 		return "", err
 	}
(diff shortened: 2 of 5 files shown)


gusnuf:
I like it, definitely not backwards compatible but the feature was only released last Friday.

jspiro:
Also a reminder we need a maintainer to approve Actions to run.

jspiro:
And it would be awesome if someone wanted to write docs on how to make the GitHub Application for this. We hacked ours together and got it working, but had to give it perms to everything since we weren't sure what was absolutely needed or not needed–can't say we understand it well enough to document it.

jspiro:
@jcudit I see this is in the 5.0 milestone–when is that targeted to land?

Might this not be better to merge this ASAP before too many people use the existing method of PEM, since it is not backwards-compatible, and the old method has only been out two weeks? The changes are very simple and more general than the original approach, while still supporting the original approach trivially.

jcudit (approved):

jcudit:
Agreed, lets run with v4.11.0 and hope that friction is minimal when releasing breaking changes in a minor version 🤞🏾

jspiro:
Cool, thanks for everyone's eyes and understanding! We're very much looking forward to using this too :-)

As this is my first contribution, I'm assuming you'll handle the merge when you're ready! 🖖🏻

jodok:
can someone point me to the documentation on how to set the environment variable for `GITHUB_APP_PEM_FILE` in terraform cloud? i've tried without newlines, quoted, without the begin and end line,... but always receive an `│ Error: No decodeable PEM data found` error.

jspiro:
> can someone point me to the documentation on how to set the environment variable for `GITHUB_APP_PEM_FILE` in terraform cloud? i've tried without newlines, quoted, without the begin and end line,... but always receive an `│ Error: No decodeable PEM data found` error.

It looks like someone removed my documentation for this: https://github.com/integrations/terraform-provider-github/pull/804/files#diff-4839085b71de83d7294472f5f8fa6652faf4b0b24df211c8eaf945ac96933aa5R109

As for setting it, what have you tried? Can you try it with a variable first as above?

shrink:
@jspiro are you using a Terraform Variable in Terraform Cloud, or an Environment Variable? As far as I can tell, it's impossible to use `GITHUB_APP_PEM_FILE` with Terraform Cloud at the moment, so I _think_ you're actually using a Terraform Variable?

@jodok I've had the same issue and _as far as I know_ there's no userland workaround, but I've created a PR for what I think will fix it: #931

jspiro:
@shrink We use a variable, just as you describe, in Terraform Cloud.

hargut:
> can someone point me to the documentation on how to set the environment variable for `GITHUB_APP_PEM_FILE` in terraform cloud? i've tried without newlines, quoted, without the begin and end line,... but always receive an `│ Error: No decodeable PEM data found` error.

Have seen the same problem, needed to use `\\n` within the `GITHUB_APP_PEM_FILE` environment instead of `\n` to get it working.

SanderBlom:
> > can someone point me to the documentation on how to set the environment variable for `GITHUB_APP_PEM_FILE` in terraform cloud? i've tried without newlines, quoted, without the begin and end line,... but always receive an `│ Error: No decodeable PEM data found` error.
> 
> Have seen the same problem, needed to use `\\n` within the `GITHUB_APP_PEM_FILE` environment instead of `\n` to get it working.

For anyone else who is interested you can convert the file using this basic nodejs script:
```JS
import { readFile } from 'fs/promises';
import { resolve } from 'path';

async function convertPemFile(filePath) {
    try {
        const pemContent = await readFile(resolve(filePath), 'utf8');

        const formattedPem = pemContent.trim().replace(/\r?\n/g, '\\n');

        console.log(`GITHUB_APP_PEM_FILE="${formattedPem}"`);
    } catch (error) {
        console.error(`Error reading the PEM file: ${error.message}`);
        process.exit(1);
    }
}

// Get the file path from command-line arguments
const filePath = process.argv[2];

if (!filePath) {
    console.error('Usage: node convert-pem.mjs <path-to-pem-file>');
    process.exit(1);
}

convertPemFile(filePath);
```

Issue #652, opened 2021-01-08
Need instructions for migrating from hashicorp/github to integrations/github


docwhat:
Now that GitHub has taken over ownership of the `github` provider we need some instructions for upgrading.

I tried the obvious:

1. update `required_providers`:
    1. change `source` attribute to `integrations/github`.
    2. change `version` to `4.2.0` (from `4.1.0` in my case)
2. `terraform state replace-provider registry.terraform.io/hashicorp/github registry.terraform.io/integrations/github`
3. `terraform init`
    - This shows it installing both hashicorp 4.1.0 and integrations 4.2.0
4. `terraform plan`

Unfortunately, it then seems to hang forever.

TRACE shows that it is trying to access github.com (it should be using our enterprise server) without specifying the org using the hashicorp 4.1.0 provider:

```
[DEBUG] plugin.terraform-provider-github_v4.1.0_x4: GET /orgs//teams/sre-team HTTP/1.1
[DEBUG] plugin.terraform-provider-github_v4.1.0_x4: Host: api.github.com
```

There is exactly _one_ `GET` using the 4.2 provider:

```
[DEBUG] plugin.terraform-provider-github_v4.2.0: GET /api/v3/orgs/the-org-on-our-ghe HTTP/1.1
[DEBUG] plugin.terraform-provider-github_v4.2.0: Host: github.example.com
```

No matter what I do, terraform insists on using the official hashicorp provider.

```sh
$ terraform -version
Terraform v0.14.4
+ provider registry.terraform.io/hashicorp/github v4.1.0
+ provider registry.terraform.io/integrations/github v4.2.0
```

frbayart:
On my side I make:

1. edit all .TF files (plan and modules too)
2. terraform state replace-provider kensu.io/kensu/github integrations/github
3. terraform init
4. terraform plan

And it works; at the beginning I didn't update my modules so I got some confusion....
(kensu.io/kensu/github is fork of hashicorp/github with last features suggested on the repo)

jcudit:
🤔  maybe we document this as a breaking change?
I can get the CHANGELOG updated if it helps boost visibility of the `state replace-provider` procedure above.

frbayart:
Also @docwhat if the problem persists, it's probably useful to check `terraform providers`

You have details per plan and modules, output is like:
```
Providers required by configuration:
.
├── provider[registry.terraform.io/integrations/github] 4.2.0
[...]

PR #816, opened 2021-06-12
Helpful note on replace-provider

glerb:
Solving this problem took me hours, and would've taken more but for https://github.com/integrations/terraform-provider-github/issues/652


glerb:
And the problem is actually still not completely solved. I get a diff on plan for every resource created from a module (I only have one module). I've added the `terraform { required_providers {}}` block in the module.

jcudit (approved):
❤️  thank you and apologies that you lost time on this one.

Hoping this note helps others avoid the same in the future.

PR #818, opened 2021-06-13
docs: add callout on github app incompatibility

### LINKED ISSUE #817: docs: improve GitHub Apps compatibility info
With #753, users can use GitHub Apps as an authentication method but this is not compatible with all API endpoints. The provider documentation does include this disclaimer but does not indicate which data sources or resources are incompatible with it. If the auth method is incompatible with a particular data source or resource, it should be indicated accordingly so users can be better informed before they select a particular auth method or when they encounter API-related errors when using the GitHub Apps auth method.

loozhengyuan:
Fixes #817

### DIFF
--- a/website/docs/r/app_installation_repository.html.markdown
+++ b/website/docs/r/app_installation_repository.html.markdown
@@ -7,6 +7,8 @@ description: |-
 
 # github_app_installation_repository
 
+~> **Note**: This resource is not compatible with the GitHub App Installation authentication method.
+
 This resource manages relationships between app installations and repositories
 in your GitHub organization.
 


loozhengyuan:
I've added a callout for the `github_app_installation_repository` resource for now but will be looking at the rest of the data sources and resources to see what else uses endpoints that are incompatible with GitHub Apps. If anyone knows of any, feel free to indicate here.

jcudit (approved):
Good information to surface 👍🏾

loozhengyuan:
@jcudit Quick update, I did a quick run through of the docs and it seems that this is the only resource that is incompatible. If you know of more, I can add on to this. Otherwise, I think this can be considered as ready.

PR #819, opened 2021-06-13
Fix a possible typo on setDismissalActorIDs

jarylc:
## Description
A simple PR for a typo.

There seems to be a typo in the setDismissalActorIDs func, this PR is to simply fix that if it was really unintended.

## Reason
While using the provider, my team and I have realized that `github_branch_protection:push_restrictions` does not change on the first terraform application and somehow only gets queued for change on the second time. It was a really weird bug, and it might be due to this.

If this PR is a non-issue, I will open an issue with more descriptions regarding it.

### DIFF
--- a/github/util_v4_branch_protection.go
+++ b/github/util_v4_branch_protection.go
@@ -190,7 +190,7 @@ func setDismissalActorIDs(actors []DismissalActorTypes) []string {
 			pushActors = append(pushActors, a.Actor.Team.ID.(string))
 		}
 		if a.Actor.User != (Actor{}) {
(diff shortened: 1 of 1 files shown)


jcudit (approved):
Looks correct.  Hoping this ends up being the fix!

usmonster:
I just found this after opening my PR (#824), which I believe is a slightly more complete fix, especially since the initial error has since propagated to another part of the file.

jarylc:
@jcudit I believe there's a more complete fix for this in this PR by @usmonster
https://github.com/integrations/terraform-provider-github/pull/824

Feel free to close this and work on that instead.

PR #856, opened 2021-07-20
Add admin property to organization data source.

Olin-Vorechovsky:
Current implementation place members and admins into members property of data_source_github_organization resource.

There is no way how to determine, if it is a member or admin.

Upon this I have added a new property `admins` with organization admins only.
This PR also fix that there are only members in `members` property.

Doc were updated for these two properties.

### DIFF
--- a/github/data_source_github_organization.go
+++ b/github/data_source_github_organization.go
@@ -1,6 +1,7 @@
 package github
 
 import (
+	"context"
 	"log"
 	"strconv"
 
@@ -34,6 +35,13 @@ func dataSourceGithubOrganization() *schema.Resource {
 				Computed: true,
 			},
 			"repositories": {
+				Type:     schema.TypeList,
+				Computed: true,
+				Elem: &schema.Schema{
+					Type: schema.TypeString,
+				},
+			},
+			"admins": {
 				Type:     schema.TypeList,
 				Computed: true,
 				Elem: &schema.Schema{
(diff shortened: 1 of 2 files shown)


github/data_source_github_organization.go
```diff
@@ -1,10 +1,11 @@
 package github
 
 import (
+	"context"
 	"log"
 	"strconv"
 
-	"github.com/google/go-github/v36/github"
+	"github.com/google/go-github/v35/github"
```
jcudit, line 8:
```suggestion
	"github.com/google/go-github/v36/github"
```

Can we get this aligned to the latest version the project is using?

jcudit (commented):
Looking good so far!  Almost ready to 🚀

Olin-Vorechovsky:
I have found the way how to deal with original `members` property so my PR is no more needed.

```
data "github_organization" "organization" {
  name = local.owner
}

data "github_membership" "memberships" {
  for_each = toset(data.github_organization.organization.members)

  username = each.value
}

locals {
  all_members = data.github_membership.memberships
  members = [for a in local.all_members : a.username if a.role == "member"]
  admins = [for a in local.all_members : a.username if a.role == "admin"]
}

output "len_all" {
  value = length(local.all_members)
}

output "members" {
  value = local.members
}

output "admins" {
  value = local.admins
}
```

PR #864, opened 2021-07-30
Manage organizations by Enterprise Account

Olin-Vorechovsky:
Hello,
would like to manage GitHub organizations (creating, renaming, deleting) by terraform.
I have not found `resource_github_organization` resource to create/update/delete organization by terraform.

Found that it is possible by GitHub API, but you need `enterprise account` (in oposit of `organization account` used for most cases) to do so: 
https://docs.github.com/en/enterprise-server@3.1/rest/reference/enterprise-admin

This is my very first terraform provider resource/test implementation and after running acceptance tests

`TF_LOG=DEBUG TF_ACC=1 go test -v   ./... -run TestAccGithubOrganization$`

I'm not able to pass this on public github.com. 
> Error: POST https://api.github.com/admin/organizations: 404 Not Found []

Is this a limitation of public GitHub and creating organizations is possible only on Enterprise GitHub? 

There are few more questions, I'm not sure about:
* Do github provider support authentication by Enterprise Account? Where to pass this, if GITHUB_OWNER does not fit to it?
* How to test against GHE, specially where to pass `base_url`
* How to proper test Enterprise Account use cases in acceptance tests

Many thanks for collaborating on this.

### DIFF
--- a/github/resource_github_organization.go
+++ b/github/resource_github_organization.go
@@ -40,17 +40,17 @@ func resourceGithubOrganizationCreate(d *schema.ResourceData, meta interface{}) 
 	client := meta.(*Owner).v3client
 
 	ownerName := meta.(*Owner).name
-	login := github.String(d.Get("login").(string))
-	admin := github.String(d.Get("admin").(string))
+	organizationName := github.String(d.Get("login").(string))
+	enterpriseAdmin := github.String(d.Get("admin").(string))
 
 	newOrganization := github.Organization{
-		Login: login,
+		Login: organizationName,
 	}
 
 	ctx := context.Background()
 
-	log.Printf("[DEBUG] Creating organization: %s (%s)", *login, ownerName)
-	githubOrganization, _, err := client.Admin.CreateOrg(ctx, &newOrganization, *admin)
+	log.Printf("[DEBUG] Creating organization: %s (%s)", *organizationName, ownerName)
+	githubOrganization, _, err := client.Admin.CreateOrg(ctx, &newOrganization, *enterpriseAdmin)
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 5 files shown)


github/resource_github_organization.go
```diff
@@ -0,0 +1,129 @@
...
+			return err
+		}
+	}
+
+	d.SetId(strconv.FormatInt(org.GetID(), 10))
+	return resourceGithubOrganizationRead(d, meta)
+}
+
+func resourceGithubOrganizationDelete(d *schema.ResourceData, meta interface{}) error {
+	// no github api for that?
```
jcudit, line 127: 🙃 this is unfortunate.  Maybe the v4 API has support here?  If not, we should add logging / documentation around this limitation and then possibly drop the resource from Terraform state.

jcudit (commented):
Interesting implementation so far!  This resource seems like one of the more difficult ones to implement given the state of available API actions related to org management.  Additionally, authorization / authentication options may limit this resource to be used in a specific use case.  That said, I'm onboard with getting this shipped if it helps out your situation.  We'd need to be clear with limitations and uses of this resources in the docs that ship with this, though.

I've got this queued up for testing against an enterprise deployment.  When testing in these scenarios, there is the `GITHUB_BASE_URL` environment variable available to point the provider to a deployment other than github.com.  Give that a shot if you have an enterprise deployment.

github/resource_github_organization.go
wernerb: in v4 there is create: https://docs.github.com/en/graphql/reference/mutations#createenterpriseorganization
And: https://docs.github.com/en/graphql/reference/mutations#removeenterpriseorganization

However there is no update mutation. The v4 api might still be better because create has multiple admins you can bootstrap in, while v3 only has a single person. 
The downside is, there is no renaming (as far as I can see) an enterprise organization in v3. Honestly I can live with organizations being create/destroy only. As far as I can see in v3 you can't update the admin user as well.

nickfloyd:

PR #870, opened 2021-08-03
Handle repo rename edge case for team repository deletions

k24dizzle:
There is an apply `error` that can happen occasionally when you have a plan that:
- Renames a `github_repository`
- Destroys and attempts to recreate the associated `github_team_repository`'s to connect to the new repo name.

Most times, such a plan works:
1. Destroy `github_team_repository`'s connected to the repo
2. Rename the `github_repository`
3. Recreate the `github_team_repository`'s under the repo's new name

However, sometimes, steps 1 and 2 overlap in which case you'll get a 404 when you try to destroy a `github_team_repository` that is connected to a repo that has just been renamed.

This GIST contains a sample `plan` and `apply` of this edge case:
https://gist.github.com/k24dizzle/36fbc3921c7a074168bee287263df02b

See this repo for example tf config that could run into this error: https://github.com/kzhaotest/team_repo_edge_case

This PR takes advantage of a [quirk of the github API](https://developer.github.com/changes/2015-04-17-preview-repository-redirects/) to fix this edge case. If you try to get a repo that has just been renamed, you can still access it via the API under the old name, to get the new name of the repo. Then you can use the new name to delete the `github_team_repository`.

### DIFF
--- a/github/resource_github_team_repository.go
+++ b/github/resource_github_team_repository.go
@@ -210,6 +210,23 @@ func resourceGithubTeamRepositoryDelete(d *schema.ResourceData, meta interface{}
 
 	log.Printf("[DEBUG] Deleting team repository association: %s (%s/%s)",
 		teamIdString, orgName, repoName)
-	_, err = client.Teams.RemoveTeamRepoByID(ctx, orgId, teamId, orgName, repoName)
+	resp, err := client.Teams.RemoveTeamRepoByID(ctx, orgId, teamId, orgName, repoName)
+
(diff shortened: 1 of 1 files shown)


k24dizzle:
@jcudit would really appreciate a review on this PR :)

k24dizzle:
@jcudit any update for this one?

jcudit (approved):

jcudit:
Thanks for getting this written out and patched up!

Issue #748, opened 2021-03-31
Unable to list the branches for a repository


nadenf:
### Affected Resource(s)
- github_branches
- github_repository

Feature request for either a: 
1) New data source called github_branches which lists all of the branches for a repository.
2) branches attribute for github_repository.

lubo:
Although `branches` attribute has been added to `github_repository`, it does not work as expected as shown in https://github.com/integrations/terraform-provider-github/pull/892#issuecomment-947950349.

PR #892, opened 2021-08-27
Add `branches` attribute to `github_repository`

mikeroll:
Addresses #748. 

Adding `branches` to both the resource and the data source.

### DIFF
--- a/github/data_source_github_repository.go
+++ b/github/data_source_github_repository.go
@@ -81,6 +81,22 @@ func dataSourceGithubRepository() *schema.Resource {
 				Type:     schema.TypeBool,
 				Computed: true,
 			},
+			"branches": {
+				Type:     schema.TypeList,
+				Computed: true,
+				Elem: &schema.Resource{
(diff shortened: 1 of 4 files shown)


jcudit (commented):
Can we get a test added or modified to incorporate the new field?  Otherwise looks 👌🏾 with the docs update and implementation!

mikeroll:
@jcudit I've only seen tests around for more advanced stuff like resource modification and nontrivial logic, so I figured we don't normally test simple things like this. I'll have another look though!

jcudit (approved):
Looks good after verifying locally:

```sh
STATE:

data.github_repository.test:
  ID = tf-acc-hgsf1
  provider = provider.github
  allow_merge_commit = true
  allow_rebase_merge = true
  allow_squash_merge = true
  archived = false
  branches.# = 1
  branches.0.name = main
  branches.0.protected = false
```

mikeroll:
Thank you for finalising this, I got distracted by other things unfortunately.

Issue #868, opened 2021-08-03
[Feature] Add `require conversation resolution before merging` support to github_branch_protection resource


nantiferov:
Hi there,

GitHub recently added new rule to branch protection, named `Require conversation resolution before merging`.
go-github already support it: https://github.com/google/go-github/issues/1945, https://github.com/google/go-github/pull/1959

Would be nice to have it in terraform github_branch_protection resource

nantiferov:
Afaik this feature was merged in https://github.com/integrations/terraform-provider-github/pull/904
And supported now by `github_branch_protection` resource, option `require_conversation_resolution`
https://github.com/integrations/terraform-provider-github/blob/main/website/docs/r/branch_protection.html.markdown

I will close this issue as done.

PR #893, opened 2021-08-31
Bump `go-github` to v37

kfcampbell:
In order to work on #868, we must be on v38 of the `google/go-github` client. We're currently 2 major versions behind, so this PR will need to be followed by another one before #868 is unblocked. 

[API Changes](https://github.com/google/go-github/releases/tag/v37.0.0) include only one breaking change: `Modify GetBranch to handle redirects`. I've chosen to default that to `true` in [our code here](https://github.com/integrations/terraform-provider-github/pull/893/files#diff-d8f242ff994e8b18a0065c9551db2856bd04f91d84e3491f64c4f489f60f14fdR334). Please let me know if that should change!


github/resource_github_repository_file.go
```diff
@@ -331,7 +331,7 @@ func resourceGithubRepositoryFileDelete(d *schema.ResourceData, meta interface{}
 // checkRepositoryBranchExists tests if a branch exists in a repository.
 func checkRepositoryBranchExists(client *github.Client, owner, repo, branch string) error {
 	ctx := context.WithValue(context.Background(), ctxId, buildTwoPartID(repo, branch))
-	_, _, err := client.Repositories.GetBranch(ctx, owner, repo, branch)
+	_, _, err := client.Repositories.GetBranch(ctx, owner, repo, branch, true)
```
jcudit, line 334: Agreed on `true` as a sane default.

jcudit (approved):

PR #900, opened 2021-09-07
feat: Add data source github_users

pascal-hofmann:
This pull request adds a `github_users` datasource that allows to check the existence of multiple users at once.

`login` and `node_id` of users that could be found can be accessed via `logins` / `node_ids`.

All unknown logins can be accessed via `unknown_logins`.

Background: Login to our companies internal GitHub Enterprise instance is done via Microsoft Azure AD and users get automatically created once they login for the first time. Unfortunately there is currently no way to check whether a user already exists via terraform. The existing `github_user` data source abort planning when unknown users are encountered.

In order to manage the users team membership via terraform we thus currently have to maintain a separate list of users that exist/do not exist in our GitHub Enterprise instance. This data source would solve this problem, as it offers a way to check for the existence of a given user without bailing out.

Technical details:
The new datasource uses the GraphQL API in order to avoid doing multiple requests.


jcudit (commented):
Looking good so far.  Can we get the linter warning fixed up?

pascal-hofmann:
> Looking good so far. Can we get the linter warning fixed up?

This should be fixed now. I chose to just ignore the error, because Write never returns one according to the Hash interface.

PR #901, opened 2021-09-08
Upgrade to go-github v38

kfcampbell:
In order to work on #868, we need to be on `go-github` v38 minimum. This PR is the next logical step after #893, and will unblock #868 entirely. 

The [release notes](https://github.com/google/go-github/releases/tag/v38.0.0) mention two breaking API changes, one of which is a deprecation and doesn't affect us. The [other one](https://github.com/google/go-github/pull/1960) adds a `ListOptions` parameter to the `GetCommit` function. The [go-github example](https://github.com/google/go-github/pull/1960/files#diff-5a4050371a643451820966b4f24acd7ce008aebe3a69631a73a9412660d67fe7R128) defaults to passing `nil`, which I suggest we do as well and I've chosen to implement [here](https://github.com/integrations/terraform-provider-github/compare/main...kfcampbell:go-github-v38?expand=1#diff-d8f242ff994e8b18a0065c9551db2856bd04f91d84e3491f64c4f489f60f14fdR251).


jcudit (approved):
Thanks for this!

Issue #606, opened 2020-11-13
github_branch_protection still slow with version 4.0.0


SanderKnape:
As was reported earlier with version 3.1.0, the performance of the `github_branch_protection` resource has slowed down significantly now that this resource is using v4 of the GitHub API. It makes using the github provider unusable when working with dozens of repositories.

Is there a way to keep using the v3 version of the API somehow?

jcudit:
There was some discussion around a new resource pinned at the v3 / REST implementation over [here](https://github.com/terraform-providers/terraform-provider-github/issues/566#issuecomment-722004719). Still unclear on what an acceptable migration path would be though if we went ahead with this approach. Any ideas?

🤔 there is also the optimization approach in case anyone can recommend ways to improve the GraphQL implementation.

SanderKnape:
Hmm, I understand the problem. I'm not sure what the best approach would be. Adding the `github_branch_protection_v3` resource sounds like a reasonable option. As long as I can use future versions of this provider with the branch protection performance of API v3, I don't have a strong preference for a specific implementation. I can't comment on which implementation would be the easiest to maintain moving forward.

lijok:
Reason for its slowness: https://github.com/integrations/terraform-provider-github/issues/849

akloss-cibo:
Oof... this is still super annoying.  Like really.  A lot.

kfcampbell:
`github_branch_protection_v3` still uses the REST API, which may help. I'd be very receptive to thoughts on how to improve the experience using the GraphQL API.

akloss-cibo:
It's still slow.  We pretty heavily use/abuse `-target` to avoid planning around github things because of how much longer it takes to run the plans.

PR #904, opened 2021-09-14
Feature: branch protection conversation resolution

josh-barker:
This PR addresses #868 and updates the following resources:

- github_branch_protection
- github_branch_protection_v3

You may find it easier to review commit by commit.


go.mod
```diff
@@ -10,7 +10,7 @@ require (
 	github.com/hashicorp/hcl/v2 v2.3.0 // indirect
 	github.com/hashicorp/terraform-config-inspect v0.0.0-20191212124732-c6ae6269b9d7 // indirect
 	github.com/hashicorp/terraform-plugin-sdk v1.7.0
-	github.com/shurcooL/githubv4 v0.0.0-20201206200315-234843c633fa
+	github.com/shurcooL/githubv4 v0.0.0-20210725200734-83ba7b4c9228
```
jcudit, line 13: Can we get some insight into why this dependency is changing?  Not a blocker, but hoping to understand what we're currently missing.

jcudit (commented):
Looking like a promising contribution so far with the following caveats:

- Can we get some clarity on the v4 dependency update?
- Can we drop the edits to the `_v3` resource? It has been in a frozen state since released so that we do not diverge from the recommended resource.

go.mod
josh-barker: @jcudit The structs/variables for Conversation Resolution do not exist with the older version of `githubv4`.

josh-barker:
@jcudit There are existing issues relating to performance of the branch protection resource using GraphQL.
For example https://github.com/integrations/terraform-provider-github/issues/606.

Therefore, I've implemented the configuration in both resources so that people have the option of choosing the more performant resource.

Personally, I can't use the `github_branch_protection` as I am managing too many repositories.

How did you want to proceed?

jcudit:
I think we can ship to both resources in this case, but overall discourage adding more to the `github_branch_protection_v3` resource.  Aiming to send this out in an upcoming release.

jcudit (approved):
Tests well, thanks for the contribution!

mambax:
Sorry but quick question. The [docs state](https://github.com/integrations/terraform-provider-github/blob/main/website/docs/r/branch_protection.html.markdown) that this new `require_conversation_resolution ` is `optional`.

Yet, after updating the provider our plans break:
```
github_team_repository.qa-microservice-streaming-api-omega-admin: Refreshing state... [id=9733:qa-microservice-streaming-api]

Error: Field 'requiresConversationResolution' doesn't exist on type 'BranchProtectionRule'

  with github_branch_protection.actions,
  on 30_actions.tf line 19, in resource "github_branch_protection" "actions":
  19: resource "github_branch_protection" "actions" {
```

Is that an issue or wanted behaviour?

Issue #649, opened 2021-01-04
Feature Request: Support automerge


mallyvai:
Github now supports auto-merge as a repository setting. Would be great to support it!

https://docs.github.com/en/free-pro-team@latest/github/collaborating-with-issues-and-pull-requests/automatically-merging-a-pull-request

kpfleming:
As is often the case, this new GitHub feature is not exposed via the API (neither REST nor GraphQL). Once it is available in one of the APIs, then the Go client library for that API will need to have the new setting added... and after that, this provider can be updated. It's unfortunately a long cycle.

entropitor:
The API is actually already setting the value to false when you call it 😭 Is it still not there, or just not documented?

(If you have a repository resource and you change anything else, and the toggle was true, it will be false after apply)

emmekappa:
It's now exposed through a GraphQL mutation

https://github.community/t/enabling-auto-merge-on-a-pr-using-the-api/150797/5

majormoses:
The next step is getting an issue and merged functionality into https://github.com/google/go-github so we can get ready to consume it.

jcudit:
🤔 does `go-github` leverage the GraphQL API?  If not, we'd be able to use our GraphQL client to toggle this configuration directly.

majormoses:
Oh I completely forgot we dont rely as heavily on that library anymore.

RTodorov:
@jcudit apparently there is a lib for the graphql API already added to the project: github.com/shurcooL/githubv4

I noticed that the graphql lib (v4) is not used in the repository logic just yet. If one wants to suggest a PR for adding the auto-merge using v4, would that be ok?

s4nji:
> It's now exposed through a GraphQL mutation
> 
> Github thread: https://github.community/t/enabling-auto-merge-on-a-pr-using-the-api/150797/5
> 
> GraphQL doc reference: https://docs.github.com/en/graphql/reference/mutations#enablepullrequestautomerge

This seems to just enable/disable automerge on specific PR; for this to work you have to [allow automerge at repository level](https://docs.github.com/en/github/administering-a-repository/managing-auto-merge-for-pull-requests-in-your-repository), and after a brief look at the graphql explorer, it does not seem to be exposed via the v4 API yet.

jcudit:
@RTodorov yes! PRs are welcome in general.  However, as pointed out above, the v4 mutation does not apply to the repository level.  We will need to wait on better support.

[...]

PR #821, opened 2021-06-13
New Resource: github_actions_runner_group

soulshake:
(Disclaimer: this is my first time working on a Terraform provider, and this PR is still a work in progress)

This PR is intended to add support for self-hosted runner groups in GitHub Actions as described in https://docs.github.com/en/rest/reference/actions#self-hosted-runner-groups.

### My questions

1. The [docs][1] have a documented parameter `visibility` which should accept `private` as an option. But I can't get a runner group with `visibility = private` through any combination of `POST` or `PATCH` calls I've tried (details below). Should I still attempt to add support for this?
2. The docs do _not_ have a documented `allows_public_repositories` parameter, but providing this parameter seems to have an effect. Should I add support for this?

#### 1. visibility = private?

I added tests, but the one that tests `visibility = true` is failing.

<details>
<summary>
Output of failing test for a runner group with `visibility = private`
</summary>

```
$ TF_LOG= TF_ACC=1  go test -v   ./... -run ^TestAccGithubActionsRunnerGroup
?   	github.com/terraform-providers/terraform-provider-github	[no test files]
=== RUN   TestAccGithubActionsRunnerGroup_all
=== PAUSE TestAccGithubActionsRunnerGroup_all
=== RUN   TestAccGithubActionsRunnerGroup_private
=== PAUSE TestAccGithubActionsRunnerGroup_private
=== RUN   TestAccGithubActionsRunnerGroup_selected
=== PAUSE TestAccGithubActionsRunnerGroup_selected
=== CONT  TestAccGithubActionsRunnerGroup_all
=== CONT  TestAccGithubActionsRunnerGroup_selected
=== CONT  TestAccGithubActionsRunnerGroup_private
--- PASS: TestAccGithubActionsRunnerGroup_all (10.01s)
=== CONT  TestAccGithubActionsRunnerGroup_private
    testing.go:654: Step 0 error: Check failed: Check 2/2 error: got runnerGroup visibility "all"; want "private"
--- FAIL: TestAccGithubActionsRunnerGroup_private (17.92s)
=== CONT  TestAccGithubActionsRunnerGroup_selected
    testing.go:654: Step 1 error: ImportStateVerify attributes not equivalent. Difference is shown below. Top is actual, bottom is expected.
        
        (map[string]string) {
        }
        
        
        (map[string]string) (len=2) {
         (string) (len=25) "selected_repository_ids.#": (string) (len=1) "1",
[...]

PR #905, opened 2021-09-15
Upgrade go-github to v39

lauraseidler:
This should allow moving forward with #649 (I'll provide another PR). 

Release: https://github.com/google/go-github/releases/tag/v39.0.0 - the only breaking changes are in relation to [`ListRepositoryAccessRunnerGroup`](https://github.com/google/go-github/pull/2086), which this provider isn't using as far as I can see. 

Note: I'm not really a Go pro, so in case I messed something up, please let me know so I can learn :)


kfcampbell (approved):
This looks reasonable to me!

xun-guo-anzx:
Hi @lauraseidler, with [this change](https://github.com/integrations/terraform-provider-github/issues/821) in release 4.15.0, I think [ListRepositoryAccessRunnerGroup](https://github.com/google/go-github/pull/2086) is referenced in [resource_github_actions_runner_group](https://github.com/integrations/terraform-provider-github/blob/main/github/resource_github_actions_runner_group.go#L172). I assume this needs to be updated now?

lauraseidler:
Yeah, looks like it. Though I'm unsure what to pass there - I guess technically, this call should be looped to make sure to fetch all pages? 

Or we pass the default from GitHub (30 per page, page 1), as that would keep the current behaviour as is, and then full pagination can be implemented separately?

Thoughts @xun-guo-anzx?

kfcampbell:
@xun-guo-anzx what's the behavior if `nil` is passed for options to `ListRepositoryAccessRunnerGroup`?

xun-guo-anzx:
> @xun-guo-anzx what's the behavior if `nil` is passed for options to `ListRepositoryAccessRunnerGroup`?

It should work the same as current behaviour, which only fetches the first 30 repository from the list

lauraseidler:
I've rebased this now and am passing `nil` to `ListRepositoryAccessRunnerGroup` here: https://github.com/integrations/terraform-provider-github/pull/905/files#diff-3a784b7896a69029f7895122c10d1858ade59409457828a7c3f4e0a17ae1207aR172

Personally, I would stick to this as the scope of this PR, as it should keep the behaviour exactly the same as before and only does the update, which should unblock both #649 and also implementing pagination for this in a follow up PR.

Does that work @xun-guo-anzx @kfcampbell?

kfcampbell:
Yep that sounds good to me!

jcudit (approved):

PR #907, opened 2021-09-21
Make rate-limiting configurable

### LINKED ISSUE #845: Disable RateLimitTransport
Either completely disable it or at least give people the option to disable it through config
Imported some 3000 github resources into terraform yesterday, and then spent the better half of today trying to figure out why my terraform runs started taking 15-20 minutes
Commented this line out https://github.com/integrations/terraform-provider-github/blob/master/github/config.go#L34, and runs went down to 20-30 seconds with parallelism of 25
Increasing parallelism above that triggers github api abuse mechanism
The PR that introduced this mechanism https://github.com/integrations/terraform-provider-github/pull/145 acknowledges the slowdown but shrugs the reduced chance of triggering the abuse mechanism a win
Then someone raised a follow-up pr giving users the option to disable this transport, which didn't go anywhere. https://github.com/integrations/terraform-provider-github/pull/235
Please reconsider your stance on this. A 45x slowdown is unacceptable.

Thank you

kfcampbell:
Fixes #845.

This PR provides a configurable delay to sleep between writes for our rate-limiting construct. Inspired by our very talented coworker Dave Cheney's ["Functional options for friendly APIs"](https://dave.cheney.net/2014/10/17/functional-options-for-friendly-apis) blog post. The default is the same as before, which is sleeping 1000ms between write operations. 

Example terraform:

```hcl
terraform {
  required_providers {
    github = {
    }
  }
}

provider "github" {
  owner = "kfcampbell-terraform-provider"
  write_delay_ms = 50 # sleep for 50ms in between requests
}

resource "github_repository" "writedelay-test-repo-0" {
  name        = "writedelay-test-repo-0"
  visibility  = "private"
  description = "A test repository for looking at write delays."
}

```

Example log output showing new sleeping times:

```log
2021-09-21T21:41:48.125-0700 [DEBUG] provider.terraform-provider-github: 2021/09/21 21:41:48 [DEBUG] Sleeping 50ms between write operations
```

Example error output when a negative value is given: 
image

Example error output when a string is given:
image

Example error output when a cartoonishly large number (9999999999999999999999) is given: 
image

---
@kfcampbell's TODO list:
Some things that need to be addressed before making ready for review:
- [x] Validate that the integer is positive
- [x] Is it desirable to give a default in both the schema and in the transport?
    - I think this is necessary.
- [x] Manual testing
- [x] Look into seeing if this is something we can test as part of an integration test
    - Existing `transport_test.go` tests don't test any of the behavior with write delays
    - I'm not quite sure how I'd go about adding this but I'm receptive to ideas if others have them.

### DIFF
--- a/github/provider.go
+++ b/github/provider.go
@@ -228,6 +228,9 @@ func providerConfigure(p *schema.Provider) schema.ConfigureFunc {
 		}
 
 		writeDelay := d.Get("write_delay").(int)
+		if writeDelay <= 0 {
+			return nil, fmt.Errorf("write_delay must be greater than 0ms")
+		}
 		log.Printf("[DEBUG] Setting write_delay to %d", writeDelay)
 
 		config := Config{
--- a/github/provider.go
+++ b/github/provider.go
@@ -164,7 +164,8 @@ func init() {
 		"app_auth.id":              "The GitHub App ID.",
 		"app_auth.installation_id": "The GitHub App installation instance ID.",
 		"app_auth.pem_file":        "The GitHub App PEM file contents.",
-		"write_delay_ms":           "Amount of time in milliseconds to sleep in between writes to GitHub API.",
+		"write_delay_ms": "Amount of time in milliseconds to sleep in between writes to GitHub API. " +
+			"Defaults to 1000ms or 1s if not set.",
 	}
 }
 
--- a/github/provider.go
+++ b/github/provider.go
@@ -164,7 +164,7 @@ func init() {
 		"app_auth.id":              "The GitHub App ID.",
 		"app_auth.installation_id": "The GitHub App installation instance ID.",
 		"app_auth.pem_file":        "The GitHub App PEM file contents.",
-		"write_delay_ms":           "Amount of time in milliseconds to sleep in between writes to GitHub REST API.",
+		"write_delay_ms":           "Amount of time in milliseconds to sleep in between writes to GitHub API.",
 	}
 }
 
--- a/github/provider.go
+++ b/github/provider.go
@@ -43,11 +43,11 @@ func Provider() terraform.ResourceProvider {
 				Default:     false,
 				Description: descriptions["insecure"],
 			},
-			"write_delay": {
+			"write_delay_ms": {
 				Type:        schema.TypeInt,
 				Optional:    true,
 				Default:     1000,
-				Description: descriptions["write_delay"],
+				Description: descriptions["write_delay_ms"],
 			},
 			"app_auth": {
 				Type:        schema.TypeList,
@@ -164,7 +164,7 @@ func init() {
 		"app_auth.id":              "The GitHub App ID.",
 		"app_auth.installation_id": "The GitHub App installation instance ID.",
 		"app_auth.pem_file":        "The GitHub App PEM file contents.",
-		"write_delay":              "Amount of time in milliseconds to sleep in between writes to GitHub REST API.",
+		"write_delay_ms":           "Amount of time in milliseconds to sleep in between writes to GitHub REST API.",
 	}
 }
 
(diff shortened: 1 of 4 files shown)


kfcampbell:
Questions for reviewers:

- Is `write_delay` an appropriate name? Should I use `write_delay_ms` or something else entirely?
- Are there ways of integration testing this behavior? There's no prior art.
    - Maybe we could do something where we invoke a timer and artificially make ourselves rate-limited? I'm not quite seeing how we'd do this.
    - Maybe it doesn't need it? 
 - Are there additional docs updates that are missing from this PR?

github/provider.go
```diff
@@ -157,6 +164,7 @@ func init() {
 		"app_auth.id":              "The GitHub App ID.",
 		"app_auth.installation_id": "The GitHub App installation instance ID.",
 		"app_auth.pem_file":        "The GitHub App PEM file contents.",
+		"write_delay_ms":           "Amount of time in milliseconds to sleep in between writes to GitHub REST API.",
```
jcudit, line 167:
```suggestion
		"write_delay_ms":           "Amount of time in milliseconds to sleep in between writes to GitHub API.",
```

Does it apply to both REST and GraphQL API types? I believe this gets invoked whenever we detect a mutating HTTP verb.

Also may want to mention our default.

github/transport.go
```diff
@@ -64,8 +64,8 @@ func (rlt *rateLimitTransport) RoundTrip(req *http.Request) (*http.Response, err
 	// If you're making a large number of POST, PATCH, PUT, or DELETE requests
 	// for a single user or client ID, wait at least one second between each request.
 	if rlt.delayNextRequest {
-		log.Printf("[DEBUG] Sleeping %s between write operations", writeDelay)
-		time.Sleep(writeDelay)
+		log.Printf("[DEBUG] Sleeping %s between write operations", rlt.writeDelay)
```
jcudit, line 67:
```suggestion
		log.Printf("[DEBUG] Sleeping %d between write operations", rlt.writeDelay)
```

Do we need to specify the integer format?

github/transport.go
```diff
@@ -113,20 +113,39 @@ func (rlt *rateLimitTransport) RoundTrip(req *http.Request) (*http.Response, err
...
 	ctx := req.Context()
 	log.Printf("[TRACE] Releasing lock for GitHub API request (%q)", ctx.Value(ctxId))
 	rlt.m.Unlock()
 }
 
-func NewRateLimitTransport(rt http.RoundTripper) *rateLimitTransport {
-	return &rateLimitTransport{transport: rt}
+type RateLimitTransportOption func(*RateLimitTransport)
+
+// NetRateLimitTransport takes in an http.RoundTripper and a variadic list of
```
jcudit, line 130:
```suggestion
// NewRateLimitTransport takes in an http.RoundTripper and a variadic list of
```

website/docs/index.html.markdown
```diff
@@ -107,6 +107,8 @@ The following arguments are supported in the `provider` block:
   * `installation_id` - (Required) This is the ID of the GitHub App installation. It can sourced from the `GITHUB_APP_INSTALLATION_ID` environment variable.
   * `pem_file` - (Required) This is the contents of the GitHub App private key PEM file. It can also be sourced from the `GITHUB_APP_PEM_FILE` environment variable.
 
+* `write_delay_ms` - (Optional) The number of milliseconds to sleep in between write operations in order to satisfy the GitHub REST API rate limits. Defaults to 1000ms or 1 second if not provided.
```
jcudit, line 110:
```suggestion
* `write_delay_ms` - (Optional) The number of milliseconds to sleep in between write operations in order to satisfy the GitHub API rate limits. Defaults to 1000ms or 1 second if not provided.
```

Applies to both APIs?

jcudit (approved):
This is on the right track.  Added some cosmetic suggestions but likely to ship in the upcoming release.

github/provider.go
kfcampbell: Thanks, I've incorporated your suggestion and also added the default (same as before, 1 second in between requests).

github/transport.go
kfcampbell: I find the string formatting more readable, although I think this is a somewhat subjective opinion. Here's a quick example:

```go
package main

import (
        "log"
        "time"
)

func main() {
        writeDelay := time.Duration(50) * time.Millisecond
        log.Printf("Sleeping %s string seconds between write operations", writeDelay)
        log.Printf("Sleeping %d digit seconds between write operations", writeDelay)
}
```

This program produces the following output: 

image

I think for this type of unstructured log output, it's nice to have the human-readable suffix attached rather than default to the Go nanosecond unit, which is hard to parse at a glance. What do you think?

github/transport.go
kfcampbell: Oops, thank you!

website/docs/index.html.markdown
kfcampbell: Yep!

github/transport.go
jcudit: Agreed, nice touch.

jcudit (approved):
Excited to see this one land, thanks!

Issue #796, opened 2021-05-20
continual changes detected due to changing etag


StephenWithPH:
### Terraform Version
```
Terraform v0.15.4
on linux_amd64
+ provider registry.terraform.io/integrations/github v4.9.4
```

### Affected Resource(s)
- `github_repository`

### Expected Behavior
After a successful `terraform apply`, subsequent `terraform plan` or `terraform apply` should detect no changes and have nothing to do.

### Actual Behavior
Every `terraform plan` or `terraform apply` after creating the resource shows:

```
Note: Objects have changed outside of Terraform

Terraform detected the following changes made outside of Terraform since the last "terraform apply":

  # github_repository.foo has been changed
  ~ resource "github_repository" "foo" {
      ~ etag                   = <some value> -> <some other value>
        id                     = "foo"
        name                   = "foo"
        # (23 unchanged attributes hidden)
    }
```

### Steps to Reproduce
Please list the steps required to reproduce the issue, for example:
1. `terraform apply` to create a GitHub repo
2. `terraform apply` again and see the above warning regarding changes to `etag`

### Important Factoids
Because `etag` is not exposed on the `github_repository` resource, it's not possible to mark it to be ignored.

sdomula:
I ran into the same problem after upgrading to Terraform 0.15.4. Deleing the repositories from the state with `terraform state rm` and then importing them with `terraform import` did not solve this issue. The ETag is still reported to be changed outside of Terraform.

mbainter:
JIC - I also experience this problem on 0.15.4 and integrations/github v4.10

loozhengyuan:
I just encountered this after upgrading from `0.15.0` to `1.0.0`. Interestingly, **no changes were proposed** even when Terraform detected a drift on the `etag` field:

```console
$ terraform plan
terraform -chdir=terraform/github plan 
Acquiring state lock. This may take a few moments...
...

Note: Objects have changed outside of Terraform

Terraform detected the following changes made outside of Terraform since the last "terraform apply":

  # github_repository.foo has been changed
  ~ resource "github_repository" "foo" {
      ~ etag                   = <some value> -> <some other value>
        id                     = "foo"
        name                   = "foo"
        # (23 unchanged attributes hidden)
    }

Unless you have made equivalent changes to your configuration, or ignored the relevant attributes using ignore_changes, the following plan may include actions to undo or
respond to these changes.

[...]

PR #909, opened 2021-09-21
fix(r/repository): suppress `etag` changes

jcudit:
/cc https://github.com/integrations/terraform-provider-github/issues/796

### DIFF
--- a/github/resource_github_repository.go
+++ b/github/resource_github_repository.go
@@ -207,6 +207,9 @@ func resourceGithubRepository() *schema.Resource {
 			"etag": {
 				Type:     schema.TypeString,
 				Computed: true,
+				DiffSuppressFunc: func(k, old, new string, d *schema.ResourceData) bool {
+					return true
+				},
 			},
 			"template": {
 				Type:     schema.TypeList,


kfcampbell (approved):
I haven't been able to reproduce the changing `etag` issue recently with several repository creations. That being said, I can't remember ever caring about the `etag` output and it's good with me to suppress it by default.

k24dizzle:
Is anyone else running into this following error? 

```
Error: Internal validation of the provider failed! This is always a bug
with the provider itself, and not a user issue. Please report
this bug:

1 error occurred:
	* resource github_repository: etag: DiffSuppressFunc is for suppressing differences between config and state representation. There is no config for computed-only field, nothing to compare.
```

starhawking:
> Is anyone else running into this following error?
> 
> ```
> Error: Internal validation of the provider failed! This is always a bug
> with the provider itself, and not a user issue. Please report
> this bug:
> 
> 1 error occurred:
> 	* resource github_repository: etag: DiffSuppressFunc is for suppressing differences between config and state representation. There is no config for computed-only field, nothing to compare.
> ```

I just ran into this myself! Everything worked again when I backed out this change

starhawking:
Is there a reason testing was turned off for releases? There isn't much in the commit message when it was turned off, and they caught this error

https://github.com/integrations/terraform-provider-github/commit/554df9fc34c56dd74c77e237bfd12beac803b529

```
=== RUN   TestProvider/has_an_implementation
--- FAIL: TestProvider (0.00s)
    --- FAIL: TestProvider/runs_internal_validation_without_error (0.00s)
    --- PASS: TestProvider/has_an_implementation (0.00s)
```

PR #913, opened 2021-09-24
Skip If-None-Match header if empty

tbh1:
Avoid sending empty If-None-Match header when a resource's ETAG is not defined, which may cause issues with enterprise WAFs.


jcudit (approved):
Makes sense, thanks for catching this 🙇🏾

Issue #908, opened 2021-09-21
Branch protection rule failed to be created with new format Github repository node_id


xun-guo-anzx:
With Github migrating the repository's GraphQL global node ID to new format, branch protection rule resource cannot be created with `node_id` because the new format of node ID is no longer encoded by Base64.

Branch protection rule setup with repository name is still working correctly without being affected.

### Terraform Version

0.14.11

### Affected Resource(s)

- github_branch_protection

### Terraform Configuration Files
```hcl
resource "github_repository" "new_example_repo" {
  name        = "new-example-repo"
  description = "New Example Repo"
}

resource "github_branch_protection" "example" {
  repository_id = github_repository.new_example_repo.node_id

  pattern          = "main"
  enforce_admins   = true
  allows_deletions = true
}
```

### Expected Behavior

Branch protection rule should be created in the new repository.

### Actual Behavior

Error reported as following:
```
Error: Could not resolve to a Repository with the name 'owner/R_xxXXXXxxXx'.

  on main.tf line 7, in resource "github_branch_protection" "example":
   7: resource "github_branch_protection" "example" {
```

### Steps to Reproduce

1. `terraform apply`

logicbomb421:
I'm also having this issue, but it seems weirdly intermittent. My module uses the repository `node_id`, which works for most repos, but every now and then I get the "could not resolve repository owner/<node_id>" error. If I can provide any additional information please let me know.

rwhittlemoj:
Also hit this issue.  Using this as a workaround for now:
`repository_id = can(base64decode(github_repository.repository.node_id)) ? github_repository.repository.node_id : github_repository.repository.name`

dimisjim:
Same problem here, with another one which relates to the direct push whitelist users.

It perpetually shows it as a diff, and tries to add them each time. Applies say that are successful, but afterwards the diff appears again:
![image](https://user-images.githubusercontent.com/6890249/136029876-7e12e2f4-bad5-40fb-b225-4a72202264b9.png)

ayk33:
The workaround mentioned doesn't work if your `github_repository` resource has a `for_each` loop in it. It attempts to change all of the existing branch protections `repository_id` to `github_repository.repository.name`.
Can we get some traction on this issue?

kylekurz:
+1 on this, if you rename a repository, the `name` field will change, but the `node_id` should remain the same. Would be vastly preferable to continue using `node_id` everywhere.

[...]

PR #914, opened 2021-09-25
Properly handle the new style of Node IDs in the GitHub GraphQL API

starhawking:
Properly handle branches using the new style of Node IDs in the GraphQL API

https://github.blog/2021-02-10-new-global-id-format-coming-to-graphql/

Addresses https://github.com/integrations/terraform-provider-github/issues/908

### DIFF
--- a/github/util_v4_repository.go
+++ b/github/util_v4_repository.go
@@ -2,7 +2,6 @@ package github
 
 import (
 	"context"
-	"encoding/base64"
 	"errors"
 
 	"github.com/shurcooL/githubv4"
@@ -41,11 +40,6 @@ func getRepositoryID(name string, meta interface{}) (githubv4.ID, error) {
 }
 
 func repositoryNodeIDExists(name string, meta interface{}) (bool, error) {
-	// Check if the name is a base 64 encoded node ID
-	_, err := base64.StdEncoding.DecodeString(name)
-	if err != nil {
-		return false, nil
-	}
 
 	// API check if node ID exists
 	var query struct {
@@ -58,7 +52,7 @@ func repositoryNodeIDExists(name string, meta interface{}) (bool, error) {
 	}
 	ctx := context.Background()
 	client := meta.(*Owner).v4client
-	err = client.Query(ctx, &query, variables)
+	err := client.Query(ctx, &query, variables)
 	if err != nil {
 		return false, err
 	}
--- a/github/util_v4_repository.go
+++ b/github/util_v4_repository.go
@@ -2,6 +2,7 @@ package github
 
 import (
 	"context"
+	"encoding/base64"
 	"errors"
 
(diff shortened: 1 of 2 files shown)


majormoses (approved):
LGTM

majormoses:
@jcudit can we get some eyes? This bug is popping up for all new repositories while existing seems to work.

github/util_v4_repository.go
```diff
@@ -41,11 +40,6 @@ func getRepositoryID(name string, meta interface{}) (githubv4.ID, error) {
 }
 
 func repositoryNodeIDExists(name string, meta interface{}) (bool, error) {
-	// Check if the name is a base 64 encoded node ID
-	_, err := base64.StdEncoding.DecodeString(name)
-	if err != nil {
-		return false, nil
-	}
```
jcudit, line 48: Are there any backwards compatibility risks with removing this? Aiming to read more on this front, but hesitating to merge this without more clarity around if this will be a breaking change for some.

starhawking: @jcudit That's what I wanted to verify with mocking out the Github API in the unit tests I wrote. I saw in the edit history that every version that accepted a node_id had this local check before calling out to the github api. Originally, it was checking the length. It was checking for base64, but I never saw a version without a pre-check, so it wasn't clear if it was originally put in to address an issue, or if the idea was to simply save an api call.

If you can think of any other cases that'd be worth testing for, I'd be more than happy to add them.

jcudit: > originally put in to address an issue

Yes, the aim was to allow passing in a repository name to improve the configuration experience.  See https://github.com/integrations/terraform-provider-github/pull/593 for details on this.

jcudit: c5a7393 suggests we check for both new and old formats.  How does everyone feel about this approach?  Seems like a 👎🏾 from a performance / rate limiting perspective but 👍🏾  for reliability.

github/util_v4_repository.go
```diff
@@ -16,6 +16,12 @@ func getRepositoryID(name string, meta interface{}) (githubv4.ID, error) {
 		return githubv4.ID(name), nil
 	}
 
+	// Interpret `name` as a legacy node ID
+	exists, _ = repositoryLegacyNodeIDExists(name, meta)
```
kfcampbell, line 20: What's the reason the error is ignored here? This function already returns an error, so perhaps it's best to add a condition like the following:

```suggestion
	exists, err = repositoryLegacyNodeIDExists(name, meta)
	if err != nil {
	    return nil, err
	}
```

kfcampbell: Ohhh, I see, it's to continue to check the non-legacy case. Please disregard me! This looks good.

bateller (approved):

jcudit (approved):

PR #931, opened 2021-10-07
fix: Replace `\n` strings with new lines in pem_file

shrink:
#804 added environment variable input for `pem_file` but golang's pem decode functionality only works when the string contains _actual_ new lines. Terraform Cloud does not permit new lines in Environment Variables, therefore it is currently not possible to use the `GITHUB_APP_PEM_FILE` Environment Variable on Terraform Cloud.

Unfortunately, there's no test coverage for the functionality in question so while this is, in principle, a very simple change, I'm not entirely sure if there are side effects I have not considered.

https://play.golang.org/p/h3Pn97nqF8j

### DIFF
--- a/github/provider.go
+++ b/github/provider.go
@@ -3,8 +3,8 @@ package github
 import (
 	"fmt"
 	"log"
-	"time"
 	"strings"
+	"time"
 
 	"github.com/hashicorp/terraform-plugin-sdk/helper/schema"
 	"github.com/hashicorp/terraform-plugin-sdk/terraform"
--- a/github/provider.go
+++ b/github/provider.go
@@ -218,6 +218,13 @@ func providerConfigure(p *schema.Provider) schema.ConfigureFunc {
 			}
 
 			if v, ok := appAuthAttr["pem_file"].(string); ok && v != "" {
+				// The Go encoding/pem package only decodes PEM formatted blocks
+				// that contain new lines. Some platforms, like Terraform Cloud,
+				// do not support new lines within Environment Variables.
+				// Any occurrence of \n in the `pem_file` argument's value 
+				// (explicit value, or default value taken from 
+				// GITHUB_APP_PEM_FILE Environment Variable) is replaced with an
+				// actual new line character before decoding.
 				appPemFile = strings.Replace(v, `\n`, "\n", -1)
 			} else {
 				return nil, fmt.Errorf("app_auth.pem_file must be set and contain a non-empty value")
--- a/github/provider.go
+++ b/github/provider.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"log"
 	"time"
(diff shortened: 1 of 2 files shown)


github/provider.go
```diff
@@ -217,7 +218,7 @@ func providerConfigure(p *schema.Provider) schema.ConfigureFunc {
 			}
 
 			if v, ok := appAuthAttr["pem_file"].(string); ok && v != "" {
-				appPemFile = v
+				appPemFile = strings.Replace(v, `\n`, "\n", -1)
```
kfcampbell, line 221: Do you think it's worth writing a comment here to explain this behavior? I think it's not super obvious and probably worth documenting in some manner.

github/provider.go
```diff
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"log"
 	"time"
+	"strings"
```
kfcampbell, line 7: Something about this line is causing a formatting error. Can you run `gofmt` to fix it?

kfcampbell (commented):
I had a couple comments...do you mind helping me understand these changes a little better?

github/provider.go
shrink: I have added a comment to the code with some context. 

I chose to add new line replacement in the provider's configuration code because it felt most appropriate as it is dealing with how the provider is configured, however, it could be more clear if the behaviour was added to the part of the provider where the PEM is consumed, i.e: https://github.com/integrations/terraform-provider-github/blob/main/github/apps.go#L19

```diff
func GenerateOAuthTokenFromApp(baseURL, appID, appInstallationID, pemData string) (string, error) {
+       pemData = strings.Replace(pemData, `\n`, "\n", -1)
	appJWT, err := generateAppJWT(appID, time.Now(), []byte(pemData))
```

_Personally_ I think it belongs in the configuration code but I am happy to move it into the consuming code (either [`apps.go@GenerateOAuthTokenFromApp`](https://github.com/integrations/terraform-provider-github/blob/main/github/apps.go#L19) or [`apps.go@generateAppJWT`](https://github.com/integrations/terraform-provider-github/blob/main/github/apps.go#L75)) if you think that will be clearer about the behaviour and intent.

kfcampbell: Thanks for doing that! I agree with you that this change belongs at the configuration level, so all consumers can trust they have the correctly-escaped version of the certificate. 

I was a little concerned that this might break existing certiricate handling somehow, but I believe that the `base64`-encoded nature of the certificates guarantees that there won't be `\n` string literals in the cert that we accidentally over-replace, so I think this is pretty safe.

kfcampbell (approved):
My approval doesn't grant anything to this PR, but FWIW @jcudit I think this can be merged.

jcudit (approved):

Issue #929, opened 2021-10-06
Error when `owner` isn't specified and using GitHub App authentication


dortort:
Hi there,

Thank you for opening an issue. Please note that we try to keep the Terraform issue tracker reserved for bug reports and feature requests. For general usage questions, please see: https://www.terraform.io/community.html.

### Terraform Version
```
Terraform v0.13.2
+ provider registry.terraform.io/integrations/github v4.16.0
```

### Affected Resource(s)
- provider

### Terraform Configuration Files
```hcl
provider "github" {
  app_auth {
    id              = var.github_app_id
    installation_id = var.github_app_installation_id
    pem_file        = var.github_app_pem_file
  }
}
```

### Debug Output
https://gist.github.com/dortort/010645cc4b3762fbccd12fb2dc7800a9

### Panic Output
If Terraform produced a panic, please provide a link to a GitHub Gist containing the output of the `crash.log`.

### Expected Behavior
No error should occur as long as only [GitHub App accessible resources](https://docs.github.com/en/rest/overview/endpoints-available-for-github-apps) are used.

### Actual Behavior
Produced error:
```
Error: GET https://api.github.com/user: 403 Resource not accessible by integration []
```

### Steps to Reproduce
1. Create config with `github` provider and any resource or data bloc
2. Provide GitHub App auth details for provider and do _not_ specify `owner`
3. Run `terraform apply`

### Important Factoids
Possibly relevant: the GitHub App whose credentials were provided is _not_ configured to "Request user authorization (OAuth) during installation"
![image](https://user-images.githubusercontent.com/1169721/136274450-9ba2606b-6848-4862-b8ec-f2ebf02533c9.png)

Simply adding the [optional `owner` argument](https://registry.terraform.io/providers/integrations/github/latest/docs#owner) resolved the issue:
```hcl
provider "github" {
  owner = "MyOrg"
  app_auth {
    id              = var.github_app_id
    installation_id = var.github_app_installation_id
    pem_file        = var.github_app_pem_file
  }
}
```

tekumara:
Adding `owner` when using a github app resolved this for me.

PR #932, opened 2021-10-07
Determine default owner from App authentication

dortort:
Determine `owner` from App authentication (`app_auth`) if none specified.

Suggested [fix](https://github.com/integrations/terraform-provider-github/issues/929#issuecomment-937721191) by @shrink to https://github.com/integrations/terraform-provider-github/issues/929.


github/provider.go
```diff
@@ -239,13 +249,52 @@ func providerConfigure(p *schema.Provider) schema.ConfigureFunc {
...
+				}
+
+				v3client, err := c.NewRESTClient(RateLimitedHTTPClient(client, c.WriteDelay))
+
+				if err != nil {
+					return nil, err
+				}
+
+				var o Owner
+				o.v3client = v3client
```
jcudit, line 276: Can this be simplified to calling `v3client` directly?  It looks like we allocate an `Owner` here but do not use it meaningfully.

github/provider.go
```diff
@@ -239,13 +249,52 @@ func providerConfigure(p *schema.Provider) schema.ConfigureFunc {
...
+				ts := oauth2.StaticTokenSource(
+					&oauth2.Token{AccessToken: token},
+				)
+
+				client := oauth2.NewClient(ctx, ts)
+				c := Config{
+					Token:      token,
+					BaseURL:    baseURL,
+					Insecure:   insecure,
+					WriteDelay: time.Duration(writeDelay) * time.Millisecond,
```
jcudit, line 266: Seems reasonable to omit configuring this (and the refactoring to define `writeDelay` earlier) as we only perform a read request to resolve the owner's login.

jcudit (commented):
This area is one of the larger pain points of the provider so this review has more scrutiny than normal.

Added a couple of suggestions. We may also benefit from a documentation update as noted [here](https://github.com/integrations/terraform-provider-github/issues/929#issuecomment-937721191):


> owner - (Optional) This is the target GitHub organization or individual user account to manage. For example, torvalds and github are valid owners. It is optional to provide this value and it can also be sourced from the GITHUB_OWNER environment variable. When not provided and a token is available, the individual user account owning the token will be used. **When not provided and no token is available, the provider may not function correctly.**

jcudit:
Seems like a good PR to ship with the upcoming v5 changes.  Will circle back once the auth fixes are more solid and try to tie this effort in.

tekumara:
I would use this!

tekumara:
Still hoping this can be reviewed 🙏

deiga:
Hey @dortort 👋

Thank you very much for your contribution!
I'm sorry it's been stuck for so long. 

Would you be available to rebase and resolve the merge conflicts?

Then we could try to actually get this landed ☺️

deiga:
Hey there,

Thanks for the contribution and sorry that you've had to endure such a long wait time.

We've been discussing this internally and we want to take the provider auth config into a fully explicit model (https://github.com/integrations/terraform-provider-github/issues/3116)
Your change adds more implicit and hidden behaviour and thus goes against the plans we have.

PR #951, opened 2021-10-18
fix regex for topic validation

iwarapter:
There is a length requirement for topic validation otherwise the following error is thrown:
```
 422 Validation Failed [{Resource: Field: Code: Message:Topics must start with a lowercase letter or number, consist of 35 characters or less, and can include hyphens.}]
```

### DIFF
--- a/github/resource_github_repository.go
+++ b/github/resource_github_repository.go
@@ -178,7 +178,7 @@ func resourceGithubRepository() *schema.Resource {
 				Optional: true,
 				Elem: &schema.Schema{
 					Type:         schema.TypeString,
-					ValidateFunc: validation.StringMatch(regexp.MustCompile(`^[a-z0-9][a-z0-9-]*$`), "must include only lowercase alphanumeric characters or hyphens and cannot start with a hyphen"),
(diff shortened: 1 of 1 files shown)


github/resource_github_repository.go
```diff
@@ -178,7 +178,7 @@ func resourceGithubRepository() *schema.Resource {
 				Optional: true,
 				Elem: &schema.Schema{
 					Type:         schema.TypeString,
-					ValidateFunc: validation.StringMatch(regexp.MustCompile(`^[a-z0-9][a-z0-9-]*$`), "must include only lowercase alphanumeric characters or hyphens and cannot start with a hyphen"),
+					ValidateFunc: validation.StringMatch(regexp.MustCompile(`^[a-z0-9][a-z0-9-]{1,34}$`), "must include only lowercase alphanumeric characters or hyphens and cannot start with a hyphen and consist of 35 characters or less"),
```
danielpops, line 181: @jcudit FYI there is a bug here, the `{1,34}` only applies to the second set, so this ends up incorrectly failing for single-character topics (e.g. `c`)

Issue #880, opened 2021-08-13
github_branch_protection does not enable "Require status checks to pass before merging"


gertjanmaas:
### Terraform Version
```
➜ terraform --version
Terraform v1.0.3
on darwin_amd64
+ provider registry.terraform.io/hashicorp/github v4.13.0
+ provider registry.terraform.io/integrations/github v4.13.0
```

### Affected Resource(s)
- `github_branch_protection`

### Terraform Configuration Files
```hcl
resource "github_branch_protection" "main_branch" {
  repository_id  = github_repository.my_test_repo.name

  pattern                 = "main"
  enforce_admins   = true

  required_status_checks {
    strict  = true
  }

  required_pull_request_reviews {
    required_approving_review_count = 1
  }
}
```

### Expected Behavior
"Require status checks to pass before merging" and "Require branches to be up to date before merging" options in branch protection should be enabled.

### Actual Behavior
"Require status checks to pass before merging" is disabled and "Require branches to be up to date before merging" is enabled.

### Steps to Reproduce
1. Use the configuration above
2. `terraform apply`

### Important Factoids
We upgraded from github provider `3.0.0` to `4.13.0` a while ago and recently noticed that this configuration behaves differently. 
In 3.0.0 setting `strict = true` in `required_status_checks` would enable both "Require status checks to pass before merging" and "Require branches to be up to date before merging".
In 4.13.0 it only enables "Require branches to be up to date before merging", which is weird because that option in the UI is hidden behind "Require status checks to pass before merging".

Note that the intention here is to require all status checks to pass before merging. Passing an empty `contexts = []` in the `required_status_checks` block does not seem to help and we're applying these branch protection rules on multiple repositories where we don't know the context names.

ShaunEgan:
Can confirm, I am experiencing the same issue with the `4.13.0` version.

martinmigasiewicz-tomtom:
I am facing this problem as well on 4.14 and realized that it works if I go back to `github_branch_protection_v3`.

kfcampbell:
I've confirmed this behavior. Using the above HCL produces a weird state in the GitHub UI where status checks aren't required before merging:
![image](https://user-images.githubusercontent.com/9327688/136291394-0b101d19-de47-411f-b7d6-654de453c1ab.png)

But upon clicking that checkbox, the "up to date" checkbox appears, already checked: 
[...]

Issue #572, opened 2020-10-20
v3.1.0 github_branch_protection does not apply required_status_checks when set to false


wilmer-mendez-globant:
### Terraform Version
`v0.13.3`

### Affected Resource(s)
Please list the resources as a list, for example:
- github_branch_protection

### Terraform Configuration Files
```hcl
locals {
  repo_settings    = jsondecode(file(var.filename))
  branches_created = toset(local.repo_settings.additional_branches)
  repo_perms       = merge(var.org_perms, local.repo_settings.teams)
}

resource "github_branch_protection" "set_rules" {
  for_each = local.repo_settings.branch_protection

  repository_id          = github_repository.create_repo.node_id
  pattern                = each.key
  enforce_admins         = each.value.enforce_admins
  require_signed_commits = each.value.require_signed_commits

  required_pull_request_reviews {
    required_approving_review_count = each.value.required_pull_request_reviews.required_approving_review_count
    dismiss_stale_reviews           = each.value.required_pull_request_reviews.dismiss_stale_reviews
    require_code_owner_reviews      = each.value.required_pull_request_reviews.require_code_owner_reviews
  }

  required_status_checks {
    strict   = "false"
  }
}

```

### Expected Behavior
Terraform should have updated the statefile with the required_status_checks attribute set to false. 

### Actual Behavior
Terraform does not update the statefile so tf plan or tf apply show required_status_checks add on each run.

onurg:
basically we have templates which is where read this as json and `strict` can be true or false - rules are applied based on that and status_context is passed in as well.  Right now, it modifies it in each run unless there are contexts set !  Behavior started after upgrade from 2.9.0 to 3.1.0

bharathkkb:
I am also seeing this. There is a permadiff for the `required_status_checks` block. 
```
Terraform v0.13.4
+ provider registry.terraform.io/hashicorp/github v3.1.0
```

jcudit:
Thanks for reporting and confirming this behaviour. The switch over from the REST implementation to the GraphQL implementation for this resource has been problematic on various fronts. A revert is being discussed, but it seems the feature of having a wildcard pattern is very much wanted by the community.

Will keep this big in mind as we discuss next steps on making it less buggy.

guillermotti:
Any news or actions to be taken on this? It's so annoying this issue. Also happens when we try to add an App as `push_restrictions`

```
[...]

PR #954, opened 2021-10-21
Fix strict branch protection

kfcampbell:
Strict protection on v4 branch protection doesn't work at the moment. This PR makes minor code changes and removes the diff suppression func in order to correct strict branch protection. 

The research to get to this point is chronicled in #880. 

The TL;DR is that by removing the diff suppression and making a minor code change to the branch protection processing, we can restore strict branch protection with the graphQL client.

### DIFF
--- a/github/resource_github_branch_protection.go
+++ b/github/resource_github_branch_protection.go
@@ -85,9 +85,8 @@ func resourceGithubBranchProtection() *schema.Resource {
 				},
 			},
 			PROTECTION_REQUIRES_STATUS_CHECKS: {
-				Type:             schema.TypeList,
-				Optional:         true,
-				DiffSuppressFunc: statusChecksDiffSuppression,
+				Type:     schema.TypeList,
+				Optional: true,
 				Elem: &schema.Resource{
 					Schema: map[string]*schema.Schema{
 						PROTECTION_REQUIRES_STRICT_STATUS_CHECKS: {
--- a/github/resource_github_branch_protection.go
+++ b/github/resource_github_branch_protection.go
@@ -85,8 +85,9 @@ func resourceGithubBranchProtection() *schema.Resource {
 				},
 			},
 			PROTECTION_REQUIRES_STATUS_CHECKS: {
-				Type:     schema.TypeList,
-				Optional: true,
+				Type:             schema.TypeList,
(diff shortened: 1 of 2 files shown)


kfcampbell:
Alright I've spent more time here and done an expanded testing matrix on both the current release and this feature branch. The results are as follows:

### main branch

- branch protection, without required status checks
	- same weird state with the hidden checked box
	- churn: initial push_restrictions = [] then nothing
- branch protection, with required status checks, strict = false
	- initial apply is broken, no required status checks, same weird state with the hidden checked box
	- churn: initial push_restrictions = [] then nothing, still broken
- branch protection, with required status checks, strict = false, with contexts
	- initial apply works
	- churn: initial push_restrictions = [] then nothing
	- if contexts are present but empty, same broken behavior as above is exhibited
- branch protection, with required status checks, strict = true
	- initial apply is broken, no required status checks, same weird state with the hidden checked box
	- churn: initial push_restrictions = [] then nothing, still broken
- branch protection, with required status checks, strict = true, with contexts
	- initial apply works
	- churn: initial push_restrictions = [] then nothing
	- if contexts are present but empty, same broken behavior as above is exhibited
	
### bugfix branch

- branch protection, without required status checks
	- same weird state with the hidden checked box
	- churn: initial push_restrictions = [] then nothing
- branch protection, with required status checks, strict = false
	- initial apply works
	- churn: initial push_restrictions = [] and contexts = [] then nothing
- branch protection, with required status checks, strict = false, with contexts
	- initial apply works
	- churn: initial push_restrictions = [] then nothing
	- if contexts are present but empty, there's churn: initial push_restrictions = [] and contexts = [] then nothing
- branch protection, with required status checks, strict = true
	- initial apply works
	- churn: initial push_restrictions = [] and contexts = [] then nothing
- branch protection, with required status checks, strict = true, with contexts
	- initial apply works
	- churn: initial push_restrictions = [] then nothing

In addition, @jcudit and I have done some internal investigation and it looks like the weird behavior with the hidden checkbox already checked is by design, or at least it's been that way from the beginning. 

TL;DR this PR is ready for review! The bug is fixed on this branch and there's no regression on #572.

jcudit (approved):

PR #961, opened 2021-10-30
Fix go module path

turkenh:
### Description of the change

This PR fixes go module path per https://golang.org/ref/mod#module-path:

> If the module is released at major version 2 or higher, the module path must end with a major version suffix like /v2. This may or may not be part of the subdirectory name. 

Appends `/v4` since the current major is `v4`.

Also fixes the repository path which seems to be legacy after migrating from https://github.com/hashicorp/terraform-provider-github. 

### How has this code been tested

1. Run `make build` 
2. Verify the code compiles and builds the plugin binary


jcudit:
After reviewing this we agree this is the right move from a Go project standpoint and would like to move forward with the changes.

Hesitating here a bit to consider backwards compatibility and will return with more details.  In the meantime, can you provide some more information around what obstacle this creates for your use case?

turkenh:
Thanks for reviewing this @jcudit.

In [Terrajet](https://github.com/crossplane-contrib/terrajet) project, we are working on generating [Crossplane](https://github.com/crossplane/crossplane) providers leveraging existing Terraform providers. We are importing the provider package for its schema and this issue prevents us from importing the latest version. We are aware that it is not the suggested way and [planning to change](https://github.com/crossplane-contrib/terrajet/issues/131) though.

Given that this is unsupported, I couldn't insist on merging this. However, apart from our use case, I believe it is the correct way and could just be merged if it doesn't break anything :)

jcudit (approved):
🤞🏾 this works as intended.  Will keep an eye out for use cases we haven't considered that this may break 👀

jcudit:
I'm having trouble running a basic test locally with this branch as-is:

```
$ TF_ACC=1 go test -v -count=1 -timeout 30m -run ^TestAccGithubBranchProtection/configures_default_settings_when_empty$ github.com/integrations/terraform-provider-github/github
cannot find package "." in:
        /Users/jcudit/integrations/terraform-provider-github/vendor/github.com/terraform-providers/terraform-provider-github/github
```

May not be as straightforward as I'd hoped.  Aiming to revisit as part of the v5 release.

kfcampbell (approved):
I'm unable to reproduce the errors @jcudit was seeing earlier; tests run for me locally with these changes. In addition, I've asked my coworker Alan about the safety of doing this, and I was reassured that it's okay. 

I'll merge this shortly!

PR #965, opened 2021-11-02
Add validate func for repository name

KeisukeYamashita:
## WHY

Because the GitHub has the repository name length limit.
If we can configure this validate func, it will give the developer early failure in the plan rather the apply making the experience better.


majormoses (approved):
LGTM

jcudit (approved):

Issue #294, opened 2019-11-12
request: support null for required_approving_review_count


skeggse:
### Terraform Version

```
Terraform v0.12.13
+ provider.github v2.2.1
```

### Affected Resource(s)

```
github_branch_protection.required_pull_request_reviews.required_approving_review_count
```

### Terraform Configuration Files

```hcl
resource "github_branch_protection" "default" {
  repository = github_repository.repo.name
  branch     = "master"

  enforce_admins = true

  required_status_checks {
    strict = true
  }

  required_pull_request_reviews {
    required_approving_review_count = null
  }
}
```

### Debug Output

N/a

### Panic Output

N/a

### Expected Behavior
What should have happened?

The provider should have configured the branch protection without any required reviewers. The upstream service supports `null` for the `required_approving_review_count` field, and assigns it accordingly if the containing `required_pull_request_reviews` block is omitted.

### Actual Behavior
What actually happened?

The provider configured branch protection with one reviewer, per the [provider-configured default](https://github.com/terraform-providers/terraform-provider-github/blob/885cb9495f0d4c3b20e48b52a034e215ec81cab8/github/resource_github_branch_protection.go#L104).

### Steps to Reproduce
Please list the steps required to reproduce the issue, for example:
1. `terraform apply`

### Important Factoids
Are there anything atypical about your accounts that we should know? For example: Running in EC2 Classic? Custom version of OpenStack? Tight ACLs?

Nope.

### References
https://github.com/terraform-providers/terraform-provider-github/pull/181

iniinikoski:
Also bumped into this... :|

rradillen:
3 years later I am still running into this issue: PR 971 could also provide an acceptable fix for this bug

kfcampbell:
Closing as #971 has been merged and released.

skeggse:
thank you!

PR #971, opened 2021-11-09
Allow to set required_approving_review_count=0

jtyr:
When using this branch protection settings:

```
[X] Require a pull request before merging
    [ ] Require approvals
    [X] Dismiss stale pull request approvals when new commits are pushed
    [ ] Require review from Code Owners
    [ ] Restrict who can dismiss pull request reviews
```
The following Terraform code is trying to change the `required_approving_review_count` value to `1`:

```terraform
resource "github_branch_protection" "main" {
  repository_id = github_repository.this.name
  pattern       = data.github_repository.this.default_branch

  required_pull_request_reviews {
    dismiss_stale_reviews = true
  }
}
```

This PR allows to set the `required_approving_review_count` to `0` as that seems to be the default value if _Require a pull request before merging_ is checked.

### DIFF
--- a/github/resource_github_branch_protection.go
+++ b/github/resource_github_branch_protection.go
@@ -67,7 +67,7 @@ func resourceGithubBranchProtection() *schema.Resource {
 							Type:         schema.TypeInt,
 							Optional:     true,
 							Default:      1,
-							ValidateFunc: validation.IntBetween(1, 6),
+							ValidateFunc: validation.IntBetween(0, 6),
 						},
 						PROTECTION_REQUIRES_CODE_OWNER_REVIEWS: {
 							Type:     schema.TypeBool,
--- a/github/resource_github_branch_protection_v3.go
+++ b/github/resource_github_branch_protection_v3.go
@@ -101,7 +101,7 @@ func resourceGithubBranchProtectionV3() *schema.Resource {
 							Type:         schema.TypeInt,
 							Optional:     true,
 							Default:      1,
-							ValidateFunc: validation.IntBetween(1, 6),
+							ValidateFunc: validation.IntBetween(0, 6),
 						},
 					},
 				},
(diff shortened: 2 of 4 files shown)


majormoses:
While I agree with the technical correctness I think we want to do this in a major release as it breaks existing expectations for anyone using it. I think I will likely need to update some internal modules within my org.

jspiro:
Simply remove the default and this can be merged as a minor release. I actually disagree with changing the default–no reviews should be an intentional setting of zero, and maintains backwards-compatibility.

jtyr:
@jspiro Do you want me to change the `Default` back to `1` or remove the `Default` completely?

jspiro:
Well, I'm not a maintainer, but I've been using this for a few years and was a contributor at least once :-)

My intuition says removing the default is actually the best course of action, I like that–the count should be intentional IMO. It doesn't change previous behavior by accident, it'll result in errors where missing that are easy to fix.

BUT: To get this merged ASAP, leaving the previous default but allowing it to be set to zero may satisfy the maintainers releasing it as a minor release, since it only adds new functionality. If it was set to zero before it would have failed validation.

jtyr:
I have reverted the `Default` value back to `1`.

jspiro:
I think we should merge this!

(It occurs to me we may need one test to show it can be set to zero, depending whether we have tests for this functionality already. It's NBD to me but might be to maintainers.)

cooperwalbrun:
Is this still targeted to be merged for the v5.0.0 milestone? I recently ran into the issue mentioned in #294 and am curious what the resolution was regarding the major/minor discussion above, since it seems this change will also address #294. I agree with @jspiro that this would be nice to have as soon as possible, especially if it is non-breaking (per @jtyr's change to the default).

jcudit:
Looks good for a minor release, thanks for all the discussion that went into adjust this.  Aiming to get this released ahead of v5 🙇🏾

majormoses (approved):
LGTM

Thanks for fixing this quirky behavior.

k24dizzle:
oops I made my own PR to fix this here before realizing this PR existed: https://github.com/integrations/terraform-provider-github/pull/1042

I added a couple of unit tests to test this new behavior, but looking forward for either one of these PRs to get merged in.

jcudit (approved):

Issue #395, opened 2020-03-26
[Feature Request] Authoritative repository access management


eriksw:
Currently, there appears to be no way to authoritatively manage repository permissions using the GitHub terraform provider.

My use case is that there are repositories that have been created over the years, that have been imported into management by terraform, but these repositories have miscellaneous grants of read/admin/write permissions to arbitrary teams and users. (Most commonly, 'admin' to the user that initially created the repo, but there's also cases like read access being granted to a contract pentester that nobody remembered to revoke, etc...)

`github_team_repository` and `github_repository_collaborator` exist, but they are only additive: the absence of a resource setting the access for a team/user will not remove that user's access.

It would be great if there was an equivalent to [google_project_iam_policy](https://www.terraform.io/docs/providers/google/r/google_project_iam.html) — something where I can declare what users and teams should have what levels of access, and that upon apply will remove all other access.

maroux:
Just opened a PR for this issue, still working on tests.

maroux:
@jcudit could someone take a look at the linked PR and let me know if this is something you'd be open to merging?

maroux:
cc @kfcampbell maybe?

The PR is now complete with tests.

eriksw:
@maroux Is your PR intended to cover team access as well, or just individual users?

What I'm hoping will eventually be part of this provider is a resources where there's one instance of that resource and it covers both team and user(/collaborator) access in an authoritative manner. (Everything that's not a GitHub App and not inherited from the organization level.)

maroux:
@eriksw both teams and individual users. Example config:
```
resource "github_repository_collaborators" "a_repo_collaborators" {
  repository = "our-cool-repo"

  user {
    permission = "admin"
    username  = "SomeUser"
  }
  
  team {
    permission = "pull"
    team_id = "SomeTeam"
  }
}
```

eriksw:
Please don't close this until an authoritative resource is available.

PR #975, opened 2021-11-10
Add new resource github_team_members to allow authoritative team management

stawik-mesa:
Relates to #395 

Add new resource to allow authoritative team memberships.

Usage:

```terraform
resource "github_team_members" "test_team_memberships" {
  team_id  = github_team.test_team.id
  members {
    username = "octocat"
    role     = "member"
  }

  members {
    username = "octocat2"
    role     = "maintainer"
  }
}
```

### DIFF
--- a/github/resource_github_team_members_test.go
+++ b/github/resource_github_team_members_test.go
@@ -207,7 +207,7 @@ resource "github_team_members" "test_team_members" {
 		role     = "%s"
 	}
 
-	depends_on = [github_membership.github_membership]
+	depends_on = [github_membership.test_org_membership]
 }
 `, username, randString, username, role)
 }
--- a/github/resource_github_team_members_test.go
+++ b/github/resource_github_team_members_test.go
@@ -2,7 +2,6 @@ package github
 
 import (
 	"context"
-	"errors"
 	"fmt"
 	"strconv"
 	"testing"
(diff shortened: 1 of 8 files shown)


majormoses:
I like the idea, I am not sure if the name should be so similar that someone might accidentally typo the wrong one. Maybe something more explicit would be better as this deviates from the 1:1 relationship we have with the api to resource now.

Some suggestions, feel free to mix/match as I could not say I found a name I loved:
- `github_team_membership_removes_undefined`
- `github_team_members_remove_unmanaged`
- `github_group_only_these_members`
- `github_group_remove_unmanaged_users`
- `github_group_terraform_managed_only`

stawik-mesa:
Do you really think it's necessary to such a long name?

I would prefer `github_team_members` (differs from `github_team_membership`).

I also added a detailed description in the resource documentation.

website/docs/r/team_members.html.markdown
```diff
@@ -0,0 +1,74 @@
...
+
+resource "github_team_members" "some_team_members" {
+  team_id  = github_team.some_team.id
+
+  members {
+    username = "SomeUser"
+    role     = "maintainer"
+  }
+
+  members {s
```
jcudit, line 46:
```suggestion
  members {
```

website/docs/r/team_members.html.markdown
```diff
@@ -0,0 +1,74 @@
...
+---
+
+# github_team_members
+
+Provides a GitHub team members resource.
+
+Authoritative. Sets the team members for the team and removes all unmanaged users from the team. This resource allows you to manage users from teams in your organization. When applied,
+the users will be added to the team. If the user hasn't accepted their invitation to the
+organization, they won't be part of the team until they do. When
+destroyed, the users will be removed from the team.
```
jcudit, line 15: Can we reword this slightly?

```suggestion
Provides a GitHub team members resource.

This resource allows you to manage members of teams in your organization. It sets the requested team members for the team and removes all unmanaged users.

When applied, if the user hasn't accepted their invitation to the organization, they won't be part of the team until they do.

When destroyed, all users will be removed from the team.
```

jcudit (commented):
I think this resource would make the overall use case much easier to configure, but dilutes the project slightly with yet another way to accomplish the task.

As for naming, I wish `github_team_membership` was already available or `github_team` could have taken the `members` block.  I'm open to not block on the naming if we can get clearer documentation around when to use this resource versus the others.

Finally, the tests need to be reworked slightly to match the existing pattern.  See [`resource_github_team_test`](https://github.com/integrations/terraform-provider-github/blob/main/github/resource_github_team_test.go) for a relevant example, particularly the anonymous / individual / organization casing.

stawik-mesa:
@jcudit Thanks for the comments. I updates the documentation and reworked the test case.

github/resource_github_team_members_test.go
```diff
@@ -0,0 +1,213 @@
...
+}
+
+resource "github_team_members" "test_team_members" {
+  team_id  = "${github_team.test_team.id}"
+	members {
+		username = "%s"
+		role     = "%s"
+	}
+
+	depends_on = [github_membership.github_membership]
```
jcudit, line 210:
```suggestion
	depends_on = [github_membership.test_org_membership]
```

jcudit (approved):
Gave this a spin locally but found it difficult to test given the whole invite-another-collaborator flow.  Looks good on another read through though and feel we can ship and iterate if needed.

🙇🏾  for the docs updates, hoping they reduce confusion over time.

stawik-mesa:
@jcudit Do you already have a timeline for the merge? Thanks for the reviews and approval ;)

kfcampbell (approved):
@jcudit and I have read through this together and we'll get it released shortly!

Issue #980, opened 2021-11-18
Authorization fixes: vNext proposal


kfcampbell:
# Authorization Superissue, v5.0.0 proposal

Recently we've had a lot of papercuts with the provider's authentication mechanism. I've performed a bunch of testing (see results below) and I've come up with a list of behavior that I think is broken in the current release that I think we can fix in the next major version. 

### Terraform Version
```
Terraform v1.0.11
on linux_amd64
```

### Affected Resource(s)
Multiple resources: this issue concerns our authorization policies for the terraform provider

### Terraform Configuration Files

Here's an example that creates a repository, used in attempting to reproduce #876.
```hcl
terraform {
  required_providers {
    github = {
      source = "integrations/github"
      version = "4.18.0"
    }
  }
}

provider "github" {
  owner = "kfcampbell-terraform-provider"
  token = "ghp_personal_token_redacted"
}


resource "github_repository" "app_repo" {
  name        = "876-repro"
  description = "App repository"

  visibility = "private"
  delete_branch_on_merge = "true"
}
```

### Behavior

See [these issues](https://github.com/integrations/terraform-provider-github/issues?q=is%3Aopen+is%3Aissue+label%3Aauthentication) for a non-comprehensive list of current authorization issues.

I have responded to a few of these threads (#876, #878, #655) with reproduction cases.

I've performed a pretty comprehensive test of authentication results

### Release v5.0.0 proposed authorization changes: 

- Deprecate GITHUB_ORGANIZATION environment variable (and `organization` provider block configuration) in favor of GITHUB_OWNER
- If no token is present, still process owner variables so that unauthenticated requests may be made against public org-owned repos
- Improve error handling and messaging around missing tokens/owner configuration (currently it's pretty difficult to figure out this is a problem and act accordingly)
- Tweaks to avoid pointless requests (for example, when no token or owner is configured, we try a pointless request to e.g. `api.github.com//example-repo`, which obviously won't succeed)
- Docs overhaul:
    - Clarify how provider authentication works in the case where both the integrations and hashicorp providers are used (I keep returning to @tibbes's excellent comment [here](https://github.com/integrations/terraform-provider-github/issues/655#issuecomment-804419698) about it)
    - Specify in our docs exactly how the decision tree works for each form of authentication (app, provider PAT config, environment variable PAT config) and when/how they can be combined
- Return an error in some cases where our requests fail but the provider doesn't return an error code (see scenarios described below)

### What's next

[...]

Issue #977, opened 2021-11-15
app_auth credentials expire after an hour


jcogilvie:
### Terraform Version
Run `terraform -v` to show the version. If you are not running the latest version of Terraform, please upgrade because your issue may have already been fixed.

0.13.5

### Affected Resource(s)
All github resources that require authentication.

### Terraform Configuration Files
```hcl
provider "github" {
  owner = "MyOrg"
  app_auth {
     # correct values are set on the command line
  }
}
```

### Debug Output
```
Mon, 15 Nov 2021 00:50:18 GMT
/home/runner/work/_temp/82078d0e-8462-4088-a73b-519aff5c7a56/terraform-bin refresh --parallelism 50
Mon, 15 Nov 2021 00:50:46 GMT
module.my_module.data.github_team.core: Refreshing state... [id=...34]
[...many successful objects being refreshed...]
Mon, 15 Nov 2021 01:50:48 GMT
aws_iam_user_policy.tf_user: Refreshing state... [id=my-repo:state-access]
Mon, 15 Nov 2021 01:50:48 GMT
Error: GET https://api.github.com/repos/MyOrg/my-repo: 403 API rate limit of 60 still exceeded until 2021-11-15 02:50:44 +0000 UTC, not making remote request. [rate reset in 59m57s]
```

### Panic Output
None

### Expected Behavior
Authentication should be valid for as long as it takes terraform to run.

### Actual Behavior
Credentials time out after an hour.  Installation Access Tokens are valid for [exactly one hour](https://docs.github.com/en/developers/apps/building-github-apps/authenticating-with-github-apps#authenticating-as-an-installation):

> Installation access tokens have the permissions configured by the GitHub App and expire after one hour.  

### Steps to Reproduce
Please list the steps required to reproduce the issue, for example:
1. `terraform refresh`

### Important Factoids
We have a *very* large set of repositories managed by the github provider.  In attempting to switch from OAuth token authentication to GH Apps authentication, we have discovered that after approximately an hour (+ a few seconds usually) we start getting 403s to resources that are expressily allowed by our GH App permissions.

jcogilvie:
My issue is specific to logging in as an app installation with the GitHub App `app_auth` [field](https://registry.terraform.io/providers/integrations/github/latest/docs#github-app-installation) of the provider, given an app ID, an app installation ID, and a PEM file.  

[...]

Issue #655, opened 2021-01-12
Data source github_team returns 404


eleni-salamani:
Using the github_team data source with the latest provider version returns 404.

### Terraform Version
✗ terraform -v
Terraform v0.13.6
+ provider registry.terraform.io/hashicorp/aws v2.48.0
+ provider registry.terraform.io/hashicorp/github v4.1.0
+ provider registry.terraform.io/hashicorp/null v2.1.2
+ provider registry.terraform.io/integrations/github v4.2.0

### Affected Resource(s)
- github_team data source

### Terraform Configuration Files
```hcl
data "github_team" "team_details" {
  for_each = var.teams

  slug = each.key

}

resource "github_team_repository" "repository_team" {
  for_each = var.teams

  repository = github_repository.repository.name
  team_id    = data.github_team.team_details[each.key].id
  permission = each.value
}
```

and the provider configuration
```hcl
terraform {
  required_providers {
    github = {
      source = "integrations/github"
      version = "4.2.0"
    }
  }
}

provider "github" {
  version = "4.2.0"

  organization = "myorg"
  token        = "blabla"
}
```

### Debug Output
https://gist.github.com/eleni-salamani/bf0a8d0173700c95e6451ff6c78600a7

### Expected Behavior
The data source should return the team information so that it can be used in further resources.

### Actual Behavior
The call returns 404. It seems the API call is not built correctly as the organization is not set.

```
Error: GET https://api.github.com/orgs//teams/b2b: 404 Not Found []

  on modules/repository/main.tf line 61, in data "github_team" "team_details":
  61: data "github_team" "team_details" {
```

https://github.com/integrations/terraform-provider-github/blob/master/github/data_source_github_team.go#L57

### Steps to Reproduce
Please list the steps required to reproduce the issue, for example:
1. Add the github_team data source. 
2. Make sure the organization information is set in the provider
3. Issue `terraform apply`

erumble:
I'm experiencing the same issue. It looks like the org information is not being pulled from the provider's configuration, but from the token. My config looks like:

```
# Personal access token set by `GITHUB_TOKEN` env var
provider "github" {
  organization = "my-org"
}

data "github_team" "owner" {
  slug = "my-team"
}
```

This should hit the endpoint:
```
https://api.github.com/orgs/my-org/teams/my-team
```

But it is instead hitting:
```
https://api.github.com/orgs/my-user/teams/my-team
```

[...]

PR #987, opened 2021-11-26
Auth changes for vNext

kfcampbell:
Part of #980. 

TODO: 

- [ ] Can we do a lookup and obviate the need for `GITHUB_TEST_ORGANIZATION`?
- [ ] Ensure `skipUnlessMode` does the right thing with organization, user, and anonymous tests
    - I don't _think_ this is the case now since I'm seeing a bunch of requests to `GET /orgs/kfcampbell HTTP/1.1` in the tests
- [ ] Perform same array of testing matrix as I did with the main branch earlier

This WIP PR aims to:

- Deprecate GITHUB_ORGANIZATION environment variable (and organization provider block configuration) in favor of GITHUB_OWNER
- If no token is present, still process owner variables so that unauthenticated requests may be made against public org-owned repos
- Improve error handling and messaging around missing tokens/owner configuration (currently it's pretty difficult to figure out this is a problem and act appropriately
- Tweaks to avoid pointless requests (for example, when no token or owner is configured, we try a pointless request to e.g. api.github.com//example-repo, which obviously won't 
- Docs 
    - Clarify how provider authentication works in the case where both the integrations and hashicorp providers are used (I keep returning to @tibbes's excellent comment here about it)
    - Specify in our docs exactly how the decision tree works for each form of authentication (app, provider PAT config, environment variable PAT config) and when/how they can be combined
    - Return an error in some cases where our requests fail but the provider doesn't return an error code

Note that there is another small behavior change: I doubt anyone is using this, and I don't think it's particularly good practice, but I believe it used to be possible to run actions on behalf of both an organization and an individual account, using the same token, in the same terraform file. After this v5.0.0 change is released, this will no longer be true, and setting different owners for a token will require different provider configuration. This change is intentional and (I think) a good thing.


github/provider_utils.go
```diff
@@ -55,6 +54,14 @@ func skipUnlessMode(t *testing.T, providerMode string) {
 		} else {
 			t.Log("GITHUB_TOKEN and GITHUB_OWNER environment variables should be set")
 		}
+		// TODO(kfcampbell): this is a problem. how are we going to know it's an organization
+		// in acceptance tests?
+		// ideas:
+		// - we could perform a lookup online to determine the type
```
jcudit, line 60: I like this option most given we cache and reuse the result as aggressively as possible.  Debug logs suggest we currently resolve an owner to an organization multiple times.

website/docs/index.html.markdown
```diff
@@ -111,15 +109,8 @@ The following arguments are supported in the `provider` block:
...
-For backwards compatibility, if more than one of `owner`, `organization`,
-`GITHUB_OWNER` and `GITHUB_ORGANIZATION` are set, the first in this
-list takes priority.
+## Authenticating with multiple providers
 
-1. Setting `organization` in the GitHub provider configuration.
-2. Setting the `GITHUB_ORGANIZATION` environment variable.
-3. Setting the `GITHUB_OWNER` environment variable.
-4. Setting `owner` in the GitHub provider configuration.
+When authenticating with multiple providers, it is possible to fail to authenticate both of them correctly and experience unexpected behavior. This is especially fraught when using both `integrations/github` and `hashicorp/github`. As [this GitHub comment](https://github.com/integrations/terraform-provider-github/issues/655#issuecomment-804419698) from @tibbes details, [local names](https://www.terraform.io/docs/language/providers/requirements.html#local-names) must be used when using two providers with the same name. Otherwise, as @tibbes mentions, the `provider "github"` block is sent to one of the providers but not the other, which means that the other provider is working in anonymous mode with no token, no owner, and no organization.
```
kfcampbell, line 114: @tibbes I completely stole most of this content from your excellent comment [here](https://github.com/integrations/terraform-provider-github/issues/655#issuecomment-804419698) since I think it should be added to our official docs. 

Is this form of attribution okay with you? Would you prefer that I revert this and that you make a commit yourself? Is there another way to go about this? I'm not quite sure what the best path forward is, but I think your comment is amazing and should be on the website, so I'd like to help make that happen somehow.

tibbes: Yes, what you've added is absolutely fine by me!

kfcampbell: Sounds good, thank you! I'll resolve this comment for now, but please let me know if you change your mind or if you'd like to alter this content at all. Thanks!

jcogilvie:
Will this address #977?

nickfloyd:

arledesma:
@kfcampbell if this is still valid then can you update the title to reference vNEXT to align with your new major version strategy? I understand that the branch name would look awkward but the title is misleading as there is already a v5.

posquit0:
I hope that this feature will be available in 2023. 🥲

kfcampbell:
> if this is still valid then can you update the title to reference vNEXT to align with your new major version strategy?

Done! 

> I hope that this feature will be available in 2023. :smiling_face_with_tear:

I apologize for the slow rate of our feature development here. If you'd like to take this branch and run with it, please feel free!

PR #993, opened 2021-12-01
Update go-github to v41

kfcampbell:
We're a few minor versions and 2 major versions behind. Luckily, the [recent releases](https://github.com/google/go-github/releases) have all been pretty small changes and the breaking changes don't affect us. Let's update the library now so we don't fall further behind in the future!


jcudit (approved):

go.mod
```diff
@@ -5,7 +5,8 @@ go 1.16
 require (
 	github.com/client9/misspell v0.3.4
 	github.com/golangci/golangci-lint v1.25.1
-	github.com/google/go-github/v39 v39.0.0
+	github.com/google/go-github/v39 v39.2.0
```
jcudit, line 8: Should this be removed in favour of the version below?

kfcampbell: @jcudit I'm so sorry, I left out the second step! :facepalm: It's all updated now and is ready for re-review. Please forgive me for the hassle.

jcudit (approved):

PR #996, opened 2021-12-03
fix topic regex to allow single characters

sdomula:
The regex introduced with #951 does not match with single alphanumeric character topics, e.g., `"r"` and `"2"`. But those are valid topics.


danielpops:
Fixes #999

danielpops (approved):
I'm probably not qualified to approve, but this lgtm! :)

jcudit (approved):

lopopolo:
I ran into this on https://github.com/topics/c.

PR #1020, opened 2021-12-23
feat(branch-protection): Push/Reviewer actors can be specified by name 

dion-gionet:
Changes the github_branch_protection resource.

This enables us to use the branch protection resource without using a data block just to retrieve the node ID for users/teams. Node IDs can still be used as an input but you can now use a prefix to define a username or team name

The / prefix defines a user (Ex.: /testuser), the orgname/ prefix defines a team. (Ex.: testorg/team)

### DIFF
--- a/github/util_v4_branch_protection.go
+++ b/github/util_v4_branch_protection.go
@@ -3,6 +3,8 @@ package github
 import (
 	"context"
 	"fmt"
+	"log"
+	"strings"
 
 	"github.com/hashicorp/terraform-plugin-sdk/helper/schema"
 	"github.com/shurcooL/githubv4"
@@ -11,27 +13,27 @@ import (
 type Actor struct {
 	ID   githubv4.ID
 	Name githubv4.String
+	Slug githubv4.String
+}
+
+type ActorUser struct {
+	ID    githubv4.ID
+	Name  githubv4.String
+	Login githubv4.String
 }
 
 type DismissalActorTypes struct {
 	Actor struct {
-		Team Actor `graphql:"... on Team"`
-		User Actor `graphql:"... on User"`
-	}
-}
-
-type BypassPullRequestActorTypes struct {
-	Actor struct {
-		Team Actor `graphql:"... on Team"`
-		User Actor `graphql:"... on User"`
+		Team Actor     `graphql:"... on Team"`
+		User ActorUser `graphql:"... on User"`
 	}
 }
 
 type PushActorTypes struct {
 	Actor struct {
-		App  Actor `graphql:"... on App"`
-		Team Actor `graphql:"... on Team"`
-		User Actor `graphql:"... on User"`
+		App  Actor     `graphql:"... on App"`
+		Team Actor     `graphql:"... on Team"`
+		User ActorUser `graphql:"... on User"`
 	}
 }
 
@@ -46,9 +48,6 @@ type BranchProtectionRule struct {
 	ReviewDismissalAllowances struct {
 		Nodes []DismissalActorTypes
 	} `graphql:"reviewDismissalAllowances(first: 100)"`
-	BypassPullRequestAllowances struct {
-		Nodes []BypassPullRequestActorTypes
-	} `graphql:"bypassPullRequestAllowances(first: 100)"`
 	AllowsDeletions                githubv4.Boolean
 	AllowsForcePushes              githubv4.Boolean
 	BlocksCreations                githubv4.Boolean
@@ -74,7 +73,6 @@ type BranchProtectionResourceData struct {
 	AllowsForcePushes              bool
 	BlocksCreations                bool
 	BranchProtectionRuleID         string
-	BypassPullRequestActorIDs      []string
 	DismissesStaleReviews          bool
 	IsAdminEnforced                bool
 	Pattern                        string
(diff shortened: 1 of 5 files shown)


dion-gionet:
@kfcampbell Do you have any feedback on this?

github/util_v4_branch_protection.go
```diff
@@ -223,60 +232,89 @@ func branchProtectionResourceData(d *schema.ResourceData, meta interface{}) (Bra
...
+	data, _ := branchProtectionResourceData(d, meta)
+	orgName := meta.(*Owner).name
+l:
 	for _, a := range actors {
-		if a.Actor.Team != (Actor{}) {
-			bypassActors = append(bypassActors, a.Actor.Team.ID.(string))
+		for _, v := range data.BypassPullRequestActorIDs {
+			if (a.Actor.Team.ID != nil && a.Actor.Team.ID.(string) == v) || (a.Actor.User.ID != nil && a.Actor.User.ID.(string) == v) {
+				bypassActors = append(bypassActors, v)
+				continue l
```
kfcampbell, line 267: Is there another way to do this without the continue to the row label? I don't love this approach.

github/util_v4_branch_protection.go
```diff
@@ -223,60 +232,89 @@ func branchProtectionResourceData(d *schema.ResourceData, meta interface{}) (Bra
...
 	}
-
-	return pushActors
+	return dismissalActors
 }
 
-func setBypassPullRequestActorIDs(actors []BypassPullRequestActorTypes) []string {
+func setBypassPullRequestActorIDs(actors []BypassPullRequestActorTypes, d *schema.ResourceData, meta interface{}) []string {
 	bypassActors := make([]string, 0, len(actors))
+	data, _ := branchProtectionResourceData(d, meta)
```
kfcampbell, line 260: Why aren't we handling the error here?

github/util_v4_branch_protection.go
```diff
@@ -223,60 +232,89 @@ func branchProtectionResourceData(d *schema.ResourceData, meta interface{}) (Bra
 	return data, nil
 }
 
-func setDismissalActorIDs(actors []DismissalActorTypes) []string {
-	pushActors := make([]string, 0, len(actors))
+func setDismissalActorIDs(actors []DismissalActorTypes, d *schema.ResourceData, meta interface{}) []string {
+	dismissalActors := make([]string, 0, len(actors))
+	data, _ := branchProtectionResourceData(d, meta)
```
kfcampbell, line 237: Why aren't we handling the error here?

github/util_v4_branch_protection.go
```diff
@@ -223,60 +232,89 @@ func branchProtectionResourceData(d *schema.ResourceData, meta interface{}) (Bra
...
-			bypassActors = append(bypassActors, a.Actor.Team.ID.(string))
+		for _, v := range data.BypassPullRequestActorIDs {
+			if (a.Actor.Team.ID != nil && a.Actor.Team.ID.(string) == v) || (a.Actor.User.ID != nil && a.Actor.User.ID.(string) == v) {
+				bypassActors = append(bypassActors, v)
+				continue l
+			}
 		}
-		if a.Actor.User != (Actor{}) {
-			bypassActors = append(bypassActors, a.Actor.User.ID.(string))
+		if a.Actor.Team.Slug != "" {
```
kfcampbell, line 270: Is this a breaking change for folks who weren't using the newly-introduced `/` character functionality?

github/util_v4_branch_protection.go
```diff
@@ -366,3 +404,71 @@ func getBranchProtectionID(repoID githubv4.ID, pattern string, meta interface{})
...
+	}
+
+	return actors, nil
+}
+
+// Given a string that is either a username or team slug, return the
+// node id of the user or team it is referring to. Team slugs must be provided
+// with the organization name as prefix (Ex.: exampleorg/exampleteam). Usernames
+// must be provided with the "/" prefix otherwise getNodeIDv4 assumes that
+// the provided string is a node ID.
```
kfcampbell, line 426: Is this specification from the GitHub API/documented anywhere public?

github/util_v4_branch_protection.go
dion-gionet: Added error handling

github/util_v4_branch_protection.go
dion-gionet: I've changed the checks and removed the labels, is this method acceptable?

kfcampbell (approved):
Thanks for updating this! Integration tests for this branch protection resource are broken on the main branch and failing here as well with this error:

```
    testing.go:705: Step 0 error: errors during apply:
        
        Error: Could not resolve to a node with the global id of '1234'
        
          on /tmp/tf-test2340247973/main.tf line 8:
          (source code not available)
```

So I don't think this is a regression. Thank you for the contribution.

github/util_v4_branch_protection.go
kfcampbell: Yep, thank you!

github/util_v4_branch_protection.go
kfcampbell: Got it, thanks!

github/util_v4_branch_protection.go
kfcampbell: I was thinking more of the `/` prefix vs. infix behavior. Regardless, I think it's intuitive enough.

kfcampbell:
@dion-gionet Are you okay if I merge/release this PR today?

PR #1022, opened 2021-12-30
Add support for disabling the use of the vulnerability management endpoint

enieuw:
This PR adds a flag to disable reading the GetVulnerabilityAlerts alerts endpoint on refresh which prevents terraform from running least privileges in the plan phase. 

The endpoint `GET /repos/:owner/:repo/vulnerability-alerts` needs administration :write permissions when using a Github App to deploy. In the plan phase this is unwanted because it forces us to expose high privileged secrets to branch builds rather than just to the main build.

### DIFF
--- a/github/resource_github_repository.go
+++ b/github/resource_github_repository.go
@@ -202,6 +202,10 @@ func resourceGithubRepository() *schema.Resource {
 				Type:     schema.TypeBool,
 				Optional: true,
 			},
+			"ignore_vulnerability_alerts_during_read": {
+				Type: schema.TypeBool,
+				Optional: true,
+			},
 			"full_name": {
 				Type:     schema.TypeString,
 				Computed: true,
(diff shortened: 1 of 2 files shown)


jcudit (commented):
I really like this PR as it opens us up to more use cases and I've had friction with inconsistent vulnerability alert support across environments in the past.

I suggest we generalize `ignore_vulnerability_alerts_during_read` to something like `privileged` so we can reuse the flag for other cases that may arise or already exist.

Keeping this queued for an upcoming release for now in case there is more input from the community.  Overall on board though and anticipate this shipping this quarter.

majormoses:
I think if we want to use a generic and overloaded term such as `privileged` then we should explicitly define what this means. Both org and repository admins are privileged users in different contexts if this is generic this will make it hard to enable someone who wants to only use features that work with a repository admin but not an org admin.

kfcampbell:
I'm a bit confused why the checks aren't running on this PR. Maybe they'll trigger with a comment?

kfcampbell (approved):
I'm dumb; the checks weren't running because they didn't exist at the point when this PR was opened. I've now updated this branch with `main` and made a small change for `gofmt` purposes. 

I agree with @majormoses regarding the term privileged, and I think it makes sense to take this as-is now and adjust the flag later if need be.

PR #1026, opened 2022-01-06
test: add manual test runs

jcudit:
This tunes our acceptance testing in a couple of ways:
- reduces frequency of automated runs to weekly
- re-instates a manual option that allows for labels applied to PRs to trigger tests

See https://github.com/jcudit/terraform-provider-github/actions/workflows/dotcom-acceptance-tests-manual.yml for example test runs and https://github.com/jcudit/terraform-provider-github/pull/1 as an example of labelling.

### DIFF
--- a/.github/workflows/dotcom-acceptance-tests-all.yml
+++ b/.github/workflows/dotcom-acceptance-tests-all.yml
@@ -8,7 +8,7 @@ on:
     #         │ │ │ ┌───────────── month (1 - 12 or JAN-DEC)
     #         │ │ │ │ ┌───────────── day of the week (0 - 6 or SUN-SAT)
     #         * * * * *
-    - cron:  '0 0 * * *'
+    - cron:  '0 0 * * 3'
 
 jobs:
 
--- a/.github/workflows/dotcom-acceptance-tests-manual.yml
+++ b/.github/workflows/dotcom-acceptance-tests-manual.yml
@@ -1,0 +1,90 @@
+name: Dotcom Acceptance Tests (manual)
+
+on:
+  pull_request:
+    types: [labeled]
+
+jobs:
+
+  acceptance-tests-anonymous:
+    runs-on: ubuntu-latest
+    if: contains(join(github.event.pull_request.labels.*.name, ', '), 'test')
+    steps:
+      - name: Parse Args
+        id: args
+        run: |
+          echo "::set-output name=run_allowed::$(
+            jq -rc .label.name $GITHUB_EVENT_PATH | cut -d/ -f 2
+          )"
+      - name: Checkout
+        uses: actions/checkout@v2
+        with:
+          ref: ${{ github.event.pull_request.head.ref }}
+          fetch-depth: 2
+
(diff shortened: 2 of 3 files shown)


.github/workflows/dotcom-acceptance-tests-all.yml
```diff
@@ -8,7 +8,7 @@ on:
     #         │ │ │ ┌───────────── month (1 - 12 or JAN-DEC)
     #         │ │ │ │ ┌───────────── day of the week (0 - 6 or SUN-SAT)
     #         * * * * *
-    - cron:  '0 0 * * *'
+    - cron:  '0 0 * * 3'
```
kfcampbell, line 11: Is the goal on this change to save on rate limits? Overall GitHub resources/churn?

.github/workflows/dotcom-acceptance-tests-manual.yml
```diff
@@ -0,0 +1,109 @@
...
+
+on:
+  pull_request:
+    types: [labeled]
+
+jobs:
+
+  acceptance-tests-anonymous:
+    runs-on: ubuntu-latest
+    if: contains(join(github.event.pull_request.labels.*.name, ', '), 'test')
```
kfcampbell, line 11: Do you think it's worth making this a little stricter? It looks like a label containing `fittest` or `attested` or `testify` would trigger this. Maybe these are weird hypotheticals I'm constructing and this won't be worth thinking about in practice.

.github/workflows/dotcom-acceptance-tests-manual.yml
```diff
@@ -0,0 +1,109 @@
...
+        id: args
+        run: |
+          echo "::set-output name=run_allowed::$(
+            jq -rc .label.name $GITHUB_EVENT_PATH | cut -d/ -f 2
+          )"
+      - name: Checkout
+        uses: actions/checkout@v2
+        with:
+          ref: ${{ github.event.pull_request.head.ref }}
+          fetch-depth: 2
```
kfcampbell, line 23: Out of curiosity, what's the reason for `fetch-depth: 2` here?

kfcampbell (commented):
I had a couple questions but overall this looks good!

.github/workflows/dotcom-acceptance-tests-all.yml
jcudit: Yes!

.github/workflows/dotcom-acceptance-tests-manual.yml
jcudit:
```suggestion
    if: contains(join(github.event.pull_request.labels.*.name, ', '), 'test/')
```

kfcampbell (approved):
Nice work! This looks great.

.github/workflows/dotcom-acceptance-tests-manual.yml
jcudit: Will revisit here with more context, seems architectural though.

PR #1027, opened 2022-01-06
feat: add `tree` data source

jasonwalsh:
Not sure if this is a desired data source, but I had a use case where I needed to find specific files in a GitHub repository using this provider and there was no such way. I ended up using the `http` data source to issue a request to the [Get a tree](https://docs.github.com/en/rest/reference/git#get-a-tree) endpoint, but decided to write a data source for it.

This is my first time contributing to a Terraform provider, so if I need to make any changes to the data source code or the acceptance tests, please let me know.

This pull request adds a new data source `github_tree` which returns a single tree given a SHA1 value. See [Get a tree](https://docs.github.com/en/rest/reference/git#get-a-tree) from the GitHub documentation.

An example is as follows:

```hcl
resource "github_repository" "this" {
  name = "my-repo"
}

data "github_branch" "this" {
  branch     = "main"
  repository = github_repository.this.name
}

data "github_tree" "this" {
  recursive  = true
  repository = github_repository.this.name
  tree_sha   = data.github_branch.this.sha
}

output "entries" {
  value = data.github_tree.this.entries
}

```

Which would output something like:

```hcl
entries = tolist([
  {
    "mode" = "100644"
    "path" = ".gitignore"
    "sha" = "7a3e2fd0945d0099d4f7604518b7e863c57069c0"
    "size" = 716
    "type" = "blob"
  },
  {
    "mode" = "100644"
    "path" = "LICENSE"
    "sha" = "4158a3a44c13ca0bcb249210565ae0cd718a2057"
    "size" = 1074
    "type" = "blob"
  },
  {
    "mode" = "100644"
    "path" = "README.md"
    "sha" = "5314ef162955d8e2bff4a6f5a565e14f30a5aa6c"
    "size" = 71
    "type" = "blob"
  },
])
```

### DIFF
--- a/github/data_source_github_tree.go
+++ b/github/data_source_github_tree.go
@@ -10,10 +10,6 @@ func dataSourceGithubTree() *schema.Resource {
 	return &schema.Resource{
 		Read: dataSourceGithubTreeRead,
 		Schema: map[string]*schema.Schema{
-			"owner": {
-				Type:     schema.TypeString,
-				Required: true,
-			},
 			"recursive": {
 				Type:     schema.TypeBool,
 				Default:  false,
@@ -23,7 +19,7 @@ func dataSourceGithubTree() *schema.Resource {
 				Type:     schema.TypeString,
 				Required: true,
 			},
-			"tree": {
+			"entries": {
 				Type:     schema.TypeList,
 				Computed: true,
 				Elem: &schema.Resource{
@@ -60,7 +56,7 @@ func dataSourceGithubTree() *schema.Resource {
 }
 
 func dataSourceGithubTreeRead(d *schema.ResourceData, meta interface{}) error {
-	owner := d.Get("owner").(string)
+	owner := meta.(*Owner).name
 	repository := d.Get("repository").(string)
 	sha := d.Get("tree_sha").(string)
 	recursive := d.Get("recursive").(bool)
@@ -74,7 +70,7 @@ func dataSourceGithubTreeRead(d *schema.ResourceData, meta interface{}) error {
 		return err
 	}
 
-	entries := make([]interface{}, 0)
+	entries := make([]interface{}, 0, len(tree.Entries))
 
 	for _, entry := range tree.Entries {
 		entries = append(entries, map[string]interface{}{
@@ -87,7 +83,7 @@ func dataSourceGithubTreeRead(d *schema.ResourceData, meta interface{}) error {
 	}
 
 	d.SetId(tree.GetSHA())
-	d.Set("tree", entries)
+	d.Set("entries", entries)
 
 	return nil
 }
--- a/github/data_source_github_tree.go
+++ b/github/data_source_github_tree.go
@@ -1,0 +1,93 @@
+package github
+
+import (
+	"context"
+
+	"github.com/hashicorp/terraform-plugin-sdk/helper/schema"
+)
+
+func dataSourceGithubTree() *schema.Resource {
+	return &schema.Resource{
+		Read: dataSourceGithubTreeRead,
+		Schema: map[string]*schema.Schema{
+			"owner": {
+				Type:     schema.TypeString,
+				Required: true,
+			},
+			"recursive": {
+				Type:     schema.TypeBool,
+				Default:  false,
+				Optional: true,
+			},
+			"repository": {
+				Type:     schema.TypeString,
+				Required: true,
+			},
+			"tree": {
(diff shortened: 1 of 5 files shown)


kfcampbell:
Interesting! Do you mind explaining a little bit more about the use case for this? 

Also, would you mind providing relevant updates to the [website docs](https://github.com/integrations/terraform-provider-github/tree/main/website/docs/d)?

Also also: it looks like the build is red right now due to a `go fmt` error; let me know if you'd like me to push the fix or you'd rather. Thanks!

jasonwalsh:
Hey @kfcampbell, 

My use case may be really specific, so if it doesn't create much value to add this data source, then it's no problem.

Our team uses env0 to manage Terraform operations like `apply` and `plan`.  env0 uses [templates](https://docs.env0.com/docs/create-your-first-template), which are containers for Terraform configurations that are associated with an SCM like GitHub or GItLab.

Regarding my use case; I wrote a Terraform module for creating projects in env0 using their official [provider](https://registry.terraform.io/providers/env0/env0/0.2.24). The module reads a YAML file which contains an array of desired projects to create in env0.

For example, the configuration file could look something like the following:

```yaml
- repository: hashicorp/hashicat-aws
  terraform_version: 1.1.2
  variables:
    - category: env
      key: AWS_DEFAULT_REGION
      value: us-east-1
```

Which would create a new project named `hashicat-aws` and then use the [Get a tree](https://docs.github.com/en/rest/reference/git#get-a-tree) endpoint to discover all of the files ending in `.tf` or `.tf.json` in that repository.

The Terraform config for discovering these files looks like the following:

```hcl
locals {
  // Filters the elements in the tree to only include files ending in `.tf` or `.tf.json`.
  templates = toset([
    for element in jsondecode(data.http.this.body)["tree"] :
    dirname(element.path)
    if element.type == "blob" && length(regexall("\\w+(.tf|.tf.json)$", element.path)) > 0
  ])
}

// Fetches the tree for a specific commit. By default the latest commit of
// the default branch (e.g., `main`).
data "http" "this" {
  request_headers = {
    Accept        = "application/vnd.github.v3+json"
    Authorization = "token ${var.token}"
  }

  url = "https://api.github.com/repos/${local.owner}/${local.repository.name}/git/trees/${data.github_branch.this.sha}?recursive=true"
}

// Creates an env0 template for each directory that contains `.tf` or `.tf.json` files.
// For example, the `hashicorp/hashicat-aws` repository contains `.tf` files in the
// root of the repository, so a single env0 template is created: `env0_template.this["."]`.
resource "env0_template" "this" {
  for_each = local.templates

  description            = "Managed by Terraform"
  github_installation_id = 21147410
  name                   = basename(each.value) == "." ? env0_project.this.name : basename(each.value)
  path                   = each.value
  repository             = local.repository.html_url
  revision               = local.repository.default_branch
  terraform_version      = var.terraform_version
  type                   = "terraform"
}

```

In order to run Terraform, I need to input the `token` variable, so the command looks like:

```
terraform apply -var="token=$GITHUB_TOKEN"
```

Since this new data source would be part of the GitHub provider, it would remove the need to define a `token` variable since that can be configured in the provider using the `token` argument of by reading the value of the `GITHUB_TOKEN` environment variable.

I hope I was able to explain this well. If something doesn't make sense, please let me know and I'll do my best to explain further.

I'll also run `go fmt` and update the website docs in separate commits.

Thank you!

jasonwalsh:
When I run `go fmt` against `github/data_source_github_tree.go` it does not complain, but the `make lint` target continues to fail. What am I missing?

kfcampbell:
@jasonwalsh I ran `make fmt` in [this commit](https://github.com/integrations/terraform-provider-github/pull/1055/commits/1d8d302d58add2b5a357e017e86e6e149f17c7e5) and the checks passed. Perhaps you could try the same command here? 

I can be convinced to include the data source when the docs are updated! Thanks for contributing.

jasonwalsh:
@kfcampbell ran `make fmt` and updated the docs. Let me know if there is anything else I should do. Thanks for your quick responses!

kfcampbell (approved):
I'm sorry, I missed one more thing earlier! Do you mind linking to the new docs page you added on [this doc](https://github.com/integrations/terraform-provider-github/blob/main/website/github.erb)? That way the sweet new docs page you added will be discoverable. Thank you!

jasonwalsh:
@kfcampbell done

PR #1031, opened 2022-01-12
Make data_github_repository work with non-existing repositories

### LINKED ISSUE #933: github_repository needs to not fail if repository doesn't exist
**Affected Resource(s)**
github_repository

For a new repository the following will fail the entire plan/apply. It would be nice if we return null so that we could use that to conditionally check if repository exist or not.

Our use case is to create the repository only if does not exist.

```
Terraform Configuration Files
data "github_repository" "existing_repository" {
  full_name   = "${var.github_org}/${var.repo}"
}
```
Error Output
GET https://api.github.com/repos/<org>/<repo> 404 Not Found []

Reference
[We have this for branch here](https://github.com/integrations/terraform-provider-github/issues/747)

tobiassjosten:
Fixes #933 – github_repository needs to not fail if repository doesn't exist.

This is my first stab at working with a Terraform provider, so bear with me. :)

I began by updating the tests and even with this new condition and no changes to the code, they still worked. So I'm not sure how useful they actually are. As for the implementation itself, I referenced c240bccda92652f74cffc98983aa3b0709b7b404 for the logic but since I don't trust the tests too much I'm not entirely sure how to best verify the change. Please advice!

### DIFF
--- a/github/data_source_github_repository.go
+++ b/github/data_source_github_repository.go
@@ -7,7 +7,7 @@ import (
 	"net/http"
 	"strings"
 
-	"github.com/google/go-github/v41/github"
+	"github.com/google/go-github/v42/github"
 	"github.com/hashicorp/terraform-plugin-sdk/helper/schema"
 )
 
--- a/github/data_source_github_repository.go
+++ b/github/data_source_github_repository.go
@@ -208,7 +208,7 @@ func dataSourceGithubRepositoryRead(d *schema.ResourceData, meta interface{}) er
 	if err != nil {
 		if err, ok := err.(*github.ErrorResponse); ok {
 			if err.Response.StatusCode == http.StatusNotFound {
-				log.Printf("Error reading GitHub branch reference %s/%s: %s", owner, repoName, err)
+				log.Printf("[DEBUG] Missing GitHub repository %s/%s", owner, repoName)
 				d.SetId("")
 				return nil
 			}
--- a/github/data_source_github_repository.go
+++ b/github/data_source_github_repository.go
@@ -3,8 +3,11 @@ package github
 import (
 	"context"
 	"fmt"
+	"log"
+	"net/http"
 	"strings"
 
+	"github.com/google/go-github/v41/github"
 	"github.com/hashicorp/terraform-plugin-sdk/helper/schema"
 )
 
@@ -203,6 +206,13 @@ func dataSourceGithubRepositoryRead(d *schema.ResourceData, meta interface{}) er
 
 	repo, _, err := client.Repositories.Get(context.TODO(), owner, repoName)
 	if err != nil {
+		if err, ok := err.(*github.ErrorResponse); ok {
+			if err.Response.StatusCode == http.StatusNotFound {
+				log.Printf("Error reading GitHub branch reference %s/%s: %s", owner, repoName, err)
+				d.SetId("")
+				return nil
+			}
+		}
 		return err
 	}
 
--- a/github/data_source_github_repository_test.go
+++ b/github/data_source_github_repository_test.go
@@ -55,16 +55,22 @@ func TestAccGithubRepositoryDataSource(t *testing.T) {
 
 		config := map[string]string{
 			"no_match_name": `
-				data "github_repository" "no_match_name" {
+				data "github_repository" "test" {
 					name = "owner/repo"
 				}
 			`,
 			"no_match_fullname": `
-				data "github_repository" "no_match_fullname" {
+				data "github_repository" "test" {
 					full_name = "owner/repo"
 				}
 			`,
 		}
+
+		check := resource.ComposeTestCheckFunc(
+			resource.TestCheckNoResourceAttr(
+				"data.github_branch.test", "id",
+			),
+		)
 
 		testCase := func(t *testing.T, mode string) {
 			resource.Test(t, resource.TestCase{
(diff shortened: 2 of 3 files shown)


github/data_source_github_repository.go
```diff
@@ -4,8 +4,10 @@ import (
 	"context"
 	"fmt"
 	"log"
+	"net/http"
 	"strings"
 
+	"github.com/google/go-github/v41/github"
```
jcudit, line 10:
```suggestion
	"github.com/google/go-github/v42/github"
```

Had a recent bump to v42 over in https://github.com/integrations/terraform-provider-github/pull/1035.  This may need an update.

jcudit (commented):
Looks close! 

As for testing, I think it makes sense to delete the `TestAccGithubRepositoryDataSource/raises_expected_errors_when_querying_for_a_repository` test given the nature of this change.  Feel free to 🔥 it on your next pass.  All other tests in that file were passing for me.

tobiassjosten:
Thank you so much for the feedback, @jcudit! I'll have another stab at it next week. 👍 

Although, I think I tweaked that specific test so as to verify that the data resource works with non-existent repos. Perhaps it should be renamed? Or do you not see a big need for it in the test suite?

github/data_source_github_repository_test.go
```diff
@@ -55,29 +55,35 @@ func TestAccGithubRepositoryDataSource(t *testing.T) {
 
 		config := map[string]string{
 			"no_match_name": `
-				data "github_repository" "no_match_name" {
+				data "github_repository" "test" {
```
kfcampbell, line 58: Do you mind explaining the reasoning behind this change?

github/data_source_github_repository_test.go
```diff
@@ -55,29 +55,35 @@ func TestAccGithubRepositoryDataSource(t *testing.T) {
...
-				data "github_repository" "no_match_fullname" {
+				data "github_repository" "test" {
 					full_name = "owner/repo"
 				}
 			`,
 		}
 
+		check := resource.ComposeTestCheckFunc(
+			resource.TestCheckNoResourceAttr(
+				"data.github_branch.test", "id",
```
kfcampbell, line 71: I'm not actually sure what this is checking. Do you know?

github/data_source_github_repository.go
```diff
@@ -205,6 +207,13 @@ func dataSourceGithubRepositoryRead(d *schema.ResourceData, meta interface{}) er
 	log.Printf("[DEBUG] Reading GitHub repository %s/%s", owner, repoName)
 	repo, _, err := client.Repositories.Get(context.TODO(), owner, repoName)
 	if err != nil {
+		if err, ok := err.(*github.ErrorResponse); ok {
+			if err.Response.StatusCode == http.StatusNotFound {
+				log.Printf("Error reading GitHub branch reference %s/%s: %s", owner, repoName, err)
```
kfcampbell, line 212: What do you thinking about giving this a log level of `[DEBUG]` similar to line 207 above?

github/data_source_github_repository_test.go
```diff
@@ -55,29 +55,35 @@ func TestAccGithubRepositoryDataSource(t *testing.T) {
```
kfcampbell, line 55: The name of this test should be changed if it's kept, maybe to something like `t.Run("does not error when a repository is not found", ...`.

I agree with @jcudit that I think the test should probably be removed. What do you think?

kfcampbell (commented):
This change might be worthy of a new major version; it feels backwards-incompatible although it's more permissive. 

@tobiassjosten I agree, the tests are somewhat confusing and not always reliable. I've found I get the most confidence from writing Terraform configuration and using the [instructions in our Contributing.md](https://github.com/integrations/terraform-provider-github/blob/main/CONTRIBUTING.md) to validate that the changes made are working.

github/data_source_github_repository_test.go
tobiassjosten: It's been too long (sorry about that) but I think it was because of the following assertion. I could only find how to create one check, so I named both configurations the same. If that makes sense?

Although, now that I read through it and see your other comment, I think I need to get a better idea of how this testing framework actually works. :)

github/data_source_github_repository_test.go
tobiassjosten: I think I was under the impression that it made sure that the repository doesn't exist (e.g. has no ID). Probably, my idea was that if that holds true and the test doesn't crash (from the missing actual repository) then it proves that the feature works. But, yeah, this is far from clear and definitely needs reworking.

github/data_source_github_repository.go
tobiassjosten: That makes sense. I took this from the branch equivalent, so I'll go ahead and update both places. Also, the message isn't correct. I'll change it to better reflect the 404 that we know is the cause here.

github/data_source_github_repository_test.go
tobiassjosten: Right now it doesn't make much sense, hehe. I'm reluctant to remove it entirely, though, since we're changing the functionality here. Give me a little more time to dig into Terraform testing and I'll see if I can come up with something to verify that this change works.

github/data_source_github_repository.go
tobiassjosten: I just rebased. 👍

tobiassjosten:
I couldn't quite figure out what to test with this change and with both @jcudit and @kfcampbell recommending deleting the one affected test, that's what I did. I'll admit I still haven't run through the testing steps in CONTRIBUTING.md so that's what I'll do next. The code should be done, though.

tobiassjosten:
Acceptance tests are working fine and I've rebased and updated the dependencies to be in line with the evolving main branch. Are we good to merge? :)

kfcampbell (approved):
Yep I think this is good to go, thank you!

Issue #353, opened 2020-02-12
Support modifying `Allow forking` option for resource_github_repository


ganhuang:
Currently the the option `Allow forking` is not supported in the resource resource_github_repository. But it could be done via UI.

#### Affected Resource(s)
Please list the resources as a list, for example:
- resource_github_repository

benj-fletch:
@ganhuang I don't believe that this is possible to be configured via the [Github API](https://developer.github.com/v3/repos/) for repositories? Do you know another way that this could be supported?

deppy:
@benj-fletch https://developer.github.com/v3/repos/forks/#create-a-fork

kfcampbell:
We could use Terraform to create a fork, certainly, but read/update/delete operations still aren't supported as far as I can see.

timhirsh:
It looks like the ask in this issue was misunderstood (and the issue prematurely closed due to staleness).  The `allow_forking` option is part of the "update repository" api here: https://docs.github.com/en/free-pro-team@latest/rest/repos/repos?apiVersion=2022-11-28#update-a-repository

@kfcampbell, https://github.com/integrations/terraform-provider-github/pull/1033 appears to have implemented this but it was reverted shortly after.

kfcampbell:
@timhirsh you're correct; I'll reopen and pin this issue.

deiga:
Resolved by https://github.com/integrations/terraform-provider-github/pull/3056

PR #1033, opened 2022-01-13
Support for allow_forking on a repository/update to go-github v42

diogopms:
Support enabling/disabling allow_forking on a repository as per #353
[More info on the Github Public API](https://docs.github.com/en/rest/reference/repos#branches)
image

Waiting from go-github to create a release on this PR:

google/go-github#2256


diogopms:
This PR is blocked by updating the github-go library and support the field (https://github.com/google/go-github/issues/2255)

kfcampbell:
@diogopms I've merged the latest `main` in, and chosen to default `allow_forking` to true, which I believe matches the behavior shown in the API docs you linked above. 

Are you okay with these changes as-is? If so, I can get them merged. If not, let's talk about it!

diogopms:
👍🏻  looks good let's merge it , thanks for fixing it 🙏🏻

jaredfholgate:
@kfcampbell and @diogopms This change is exhibiting odd behaviour and may need to be fixed or reverted. As per issue #1136. Also, for some reason the website docs have not been updated to reflect this new field, even though they are changed in this push. Happy to help if I can?

kfcampbell:
@jaredfholgate I've reverted this PR and published [v4.24.1](https://github.com/integrations/terraform-provider-github/releases/tag/v4.24.1) with only that change in it. If you'd like to take a stab at fixing the behavior, I'd be happy to review it!

martinm82:
What would be the desired behaviour? Should the implementation first check whether forking has been disabled on org level and in such a case do nothing? Is that feasible? I am not familiar with this provider but would really like to have the allow_forking support in the provider.

PR #802, opened 2021-05-24
Modify github_team to accept slug as a valid parent_team_id

n0rad:
Inspired by #693, This PR add support for declaring `parent_team_id` as a string on `github_team`.

This is especially useful when declaring teams in a loop and so cannot reference the other team object .

PR #1039, opened 2022-01-17
Remove team from state if deletion failed and it does not exist

cytopia:
# Remove team from state if deletion failed and it does not exist

I experienced this behaviour while continuing the work on team slug: https://github.com/integrations/terraform-provider-github/pull/802 So in order not to submit all changes at once, I will make this a separete PR (it does affect current provider as well)

* **Type**: bugfix (similar to this one for repositories: https://github.com/integrations/terraform-provider-github/pull/1031)

## Current state

When deleting a team on GitHub, the API will automatically delete all of its child team. In case we want to remove a team with the terraform provider which parent team has already been deleted, the resource will fail and report back with a `404 - team not found` (or slightly similar).

## Improvement

When deleting a team with this terraform provider and it fails, we try to fetch it and check to see if it returns a 404. If the team we tried to delete does not exist, we will simply remove it from TF state and continue grafefully.

## Notes on code

As I am pretty new to golan, please give me any hints on styling, comments or best-practices which I might have done wrongly.

### DIFF
--- a/github/resource_github_team.go
+++ b/github/resource_github_team.go
@@ -243,6 +243,29 @@ func resourceGithubTeamDelete(d *schema.ResourceData, meta interface{}) error {
 
 	log.Printf("[DEBUG] Deleting team: %s", d.Id())
 	_, err = client.Teams.DeleteTeamByID(ctx, orgId, id)
+	/*
+		When deleting a team and it failed, we need to check if it has already been deleted meanwhile.
+		This could be the case when deleting nested teams via Terraform by looping through a module
+		or resource and the parent team might have been deleted already. If the parent team had
+		been deleted already (via parallel runs), the child team is also already gone (deleted by
+		GitHub automatically).
+		So we're checking if it still exists and if not, simply remove it from TF state.
+	*/
+	if err != nil {
+		// Fetch the team in order to see if it exists or not (http 404)
+		_, _, err = client.Teams.GetTeamByID(ctx, orgId, id)
+		if err != nil {
+			if ghErr, ok := err.(*github.ErrorResponse); ok {
+				if ghErr.Response.StatusCode == http.StatusNotFound {
+					// If team we failed to delete does not exist, remove it from TF state.
+					log.Printf("[WARN] Removing team: %s from state because it no longer exists",
+						d.Id())
+					d.SetId("")
+					return nil
+				}
+			}
+		}
+	}
 	return err
 }
 


github/resource_github_team.go
```diff
@@ -243,6 +243,29 @@ func resourceGithubTeamDelete(d *schema.ResourceData, meta interface{}) error {
...
+		So we're checking if it still exists and if not, simply remove it from TF state.
+	*/
+	if err != nil {
+		// Fetch the team in order to see if it exists or not (http 404)
+		_, _, err = client.Teams.GetTeamByID(ctx, orgId, id)
+		if err != nil {
+			if ghErr, ok := err.(*github.ErrorResponse); ok {
+				if ghErr.Response.StatusCode == http.StatusNotFound {
+					// If team we failed to delete does not exist, remove it from TF state.
+					log.Printf("[WARN] Removing team: %s from state because it no longer exists",
```
cytopia, line 261: The following warning does not show up with `TF_LOG=WARN`. It only shows up in `TF_LOG=DEBUG`. Not sure if this is a general issue with this provider or just with me doing something wrong in the code:

jcudit (commented):
Looking close!  This seems difficult to get a test case set up for so maybe we add additional documentation around this case for awareness.

Aiming to ship in an upcoming release.

github/resource_github_team.go
```diff
@@ -243,6 +243,29 @@ func resourceGithubTeamDelete(d *schema.ResourceData, meta interface{}) error {
 
 	log.Printf("[DEBUG] Deleting team: %s", d.Id())
 	_, err = client.Teams.DeleteTeamByID(ctx, orgId, id)
+	/*
+		When deleting a team and it failed, we need to check if it has already been deleted meanwhile.
+		This could be the case when deleting nested teams via Terraform by looping through a module
+		or resource and the parent team might have been deleted already. If the parent team had
+		been deleted already (via parallel runs), the child team is also already gone (deleted by
+		GitHub automatically).
+		So we're checking if it still exists and if not, simply remove it from TF state.
```
jcudit, line 252: This is a really good comment to add to our docs on this resource as well.

github/resource_github_team.go
cytopia: @jcudit the only thing I'd need some help on is the logging level as the above is only showing up in `TG_LOG=DEBUG`. Example line below:
```
2022-01-22T10:47:11.813Z [DEBUG] provider.terraform-provider-github_v4.19.1-fl.4: 2022/01/22 10:47:11 [WARN] Removing team: 5592187 from state because it no longer exists
```
However, when looking at a valid `WARN` message, it looks like this:
```
2022-01-22T10:47:04.978Z [WARN]  ValidateProviderConfig from "provider[\"registry.terraform.io/flaconi/github\"]" changed the config value, but that value is unused
```

Any idea on what I'm doing wrong here?

cytopia:
> This seems difficult to get a test case set up for

I haven't checked yet how the internal tests work. Two integration tests I can think of might be written as so:

**1. Positive check:**

1. **Terraform Run 1:** Provision team with child team
2. **GH API call:** Remove parent team
3. **Terraform Run 2:** Remove child team and ensure the `WARN` message exists in the output


**2. Negative check:**

1. **Terraform Run 1:** Provision team with child team
2. **Terraform Run 2:** Remove child team and ensure the `WARN` message does not exist
3. **Terraform Run 3:** Remove team and ensure the `WARN` message does not exist

github/resource_github_team.go
jcudit: I am stumped.  This looks very similar to what I'm seeing in [`hashicorp/terraform`](https://github.com/hashicorp/terraform/search?p=2&q=warn), so 😕 why it is misbehaving for you.

Does `TF_LOG=JSON` do anything different? Perhaps `TG_LOG` was used in error?

kfcampbell (approved):
Let's go ahead and ship this, and we can respond to feedback as it comes up.

PR #1040, opened 2022-01-18
Upgrade github.com/ulikunitz/xz to v0.5.10

kfcampbell:
This will fix a Dependabot security alert we're seeing. Advisories: [here](https://github.com/advisories/GHSA-q6gq-997w-f55g) and [here](https://github.com/advisories/GHSA-25xm-hr59-7c27).


jcudit (approved):
Seems harmless for our use case but 👍🏾 to keeping our vulnerabilities minimal.  Thanks!

PR #1043, opened 2022-01-21
test: checkout pull request via sha instead of ref

jcudit:
Reconfigures the manual acceptance run Action to checkout using `sha` instead of `ref`.


jcudit:
Testing above shows this works, but is flaky.  Good enough feedback for reviewing PRs without having to run tests locally, though!

kfcampbell (approved):
:shipit: :shipit: :shipit:

PR #1048, opened 2022-01-25
Small CI cleanup

kfcampbell:
There's a couple of small fixes to our CI checks I'd like to make:
- `make build` was running `go install`, not `go build`
- `make test` ran a fmtchecker but not our actual unit tests, which are trivial to run
- Our CONTRIBUTING.md was recommending running unit tests, which we don't really encourage for local development validation
- `tools.go` was improperly formatted; not sure why our CI check didn't catch it
- Removed the `-mod=vendor` flag
    - In [this doc](https://go.dev/ref/mod), see the section that reads: "At go 1.14 or higher, automatic vendoring may be enabled. If the file vendor/modules.txt is present and consistent with go.mod, there is no need to explicitly use the -mod=vendor flag."
- Updated to the most recent version of Go for the Actions runner
    - The project remains low in version and able to be importable by a wide range of Go versions.
    - This should give us quicker build times and smaller binary sizes while retaining backards compatibility. 

I also added `testdata/` to the .gitignore file based on a personal preference, but I'm happy to revert that if desired.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,6 +1,9 @@
 name: GitHub Actions CI
 
-on: [push, pull_request]
+on:
+  push:
+    branches: [main]
+  pull_request: {}
 
 jobs:
   ci:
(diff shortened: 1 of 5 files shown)


jcudit (approved):
Wish I could approve it twice 🙇🏾

PR #1054, opened 2022-02-01
feat: add configurable read_delay_ms

morremeyer:
This adds a configurable delay to all non-write API operations.

I set the default to be `0` so that this is a non-breaking minor version bump.

## Rationale

For modules using this provider, it is fairly easy to hit the rate limit for a token/user on state refresh already (see e.g. #998).
This PR will allow users to configure a backoff that will result in slower, but successful runs.

## Default value considerations

A better default value that we could aim to incorporate in the next major release would be 720ms: Possibly, the most common scenario is the use with a PAT that has 5000 requests per hour. With an hour having 3600 seconds, that leads to 720ms delay between requests to reach that rate limit with read requests only.

As we pause for 1s between writes anyway, 720 milliseconds would therefore be guaranteed to not hit the rate limit if the account is not used for anything else.

### DIFF
--- a/github/config.go
+++ b/github/config.go
@@ -20,6 +20,7 @@ type Config struct {
 	BaseURL    string
 	Insecure   bool
 	WriteDelay time.Duration
+	ReadDelay  time.Duration
 }
 
 type Owner struct {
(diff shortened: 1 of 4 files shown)


morremeyer:
I just ran this code against our setup locally on a 2019 MacBook Pro (2,6 GHz 6-Core Intel Core i7) with `read_delay_ms = 0`, it ran for 8 minutes 39 seconds.

With `read_delay_ms = 720`, it ran for 21 minutes and 3 seconds.

However, as the run makes a total of 1837 API requests, it could only run 2 times per hour, while the execution speed without a delay would allow for more than 6 full runs per hour.

Especially when making/planning for a lot of changes repeatedly, setting a backoff therefore is preferable to me.

kfcampbell:
This seems reasonable to me, and is a fun extension of the earlier work on write delays. I've added it to a milestone for future release.

jcudit (approved):

kfcampbell (approved):
This looks good to me, thank you!

morremeyer:
@kfcampbell Do I need to do anything else on this or did you just push it to get 4.21 out of the door?

morremeyer:
@kfcampbell No worries, just wanted to make sure I didn’t forget anything. Thank you!

jspiro:
Nice feature: Perhaps another way to make it a default is to add a requests/hour configuration and set this dynamically based on it (e.g. 5000 for PAT and 15000 for App).

Issue #1057, opened 2022-02-07
Environment ID gets set incorrectly on update


aceresia-bg:
When updating a resource of type "github_repository_environment" the ID of the resource is broken by the following line.

https://github.com/integrations/terraform-provider-github/blob/bc38f5b42a142ecec8e8b338a47cc2d159a0ad53/github/resource_github_repository_environment.go#L183

The ID is expected to be in the format of `<repo name>:<environment name>`, However this line sets it to `<repo name>:<numerical id>`.  This causes the following debug error:

`2022-02-07T09:15:29.422-0500 [DEBUG] provider.terraform-provider-github_v4.20.0: 2022/02/07 09:15:29 [WARN] Removing repository environment <repo name>:<numerical id> from state because it no longer exists in GitHub`

Also the following error is produced:
╷
│ Error: Provider produced inconsistent result after apply
│ 
│ When applying changes to github_repository_environment.repo_environment,
│ provider "provider[\"registry.terraform.io/integrations/github\"]" produced
│ an unexpected new value: Root resource was present, but now absent.
│ 
│ This is a bug in the provider, which should be reported in the provider's
│ own issue tracker.

PR #1058, opened 2022-02-07
Environment ID gets set incorrectly on update

aceresia-bg:
The environment ID is updated in an incompatible way when updating the resource.  Issue #1057


kfcampbell:
@aceresia-bg I hope you don't mind, I've updated the PR to remove the now unnecessary import and I'll merge it now! Thanks for the contribution.

PR #1061, opened 2022-02-10
Documentation: Add id to github_user data dource

### LINKED ISSUE #1004: Documentation: github_user doesn't expose User ID as an attribute
The data resource github_user, sets the user ID here:
https://github.com/integrations/terraform-provider-github/blob/main/github/data_source_github_user.go#L133

But it is not exposed as an attribute one can use.

kangaechu:
Fixes #1004


kfcampbell:
Thank you!

PR #1066, opened 2022-02-13
Allow for retrieving latest prerelease as data source.

youcandanch:
### Description
Being able to retrieve the latest release from a given repository is awesome, but there's circumstances in which it'd be beneficial to grab the latest prerelease. In a continuous deployment workflow where there's a "prerelease" like environment (staging, preproduction, prerelease, rc, etc.) it'd be great to be able to pull the tag information directly from the provider, rather than having to skip out to the API to do it.

This PR introduces that functionality by allowing for passing `latest_prerelease` as a valid value to `retrieve_by` on the release data source. It's a little more involved than retrieving just the latest release, because the GitHub API at the moment doesn't allow for allowing prereleases as a flag on `/latest`. Instead, we're retrieving a few pages worth of releases, filtering out the prereleases, and selecting the most recently published one. 

This _does_ mean that there's a pretty substantial performance difference, as there's both more data to process and more requests being made.

### DIFF
--- a/github/data_source_github_release.go
+++ b/github/data_source_github_release.go
@@ -30,6 +30,7 @@ func dataSourceGithubRelease() *schema.Resource {
 				Type:     schema.TypeString,
 				Required: true,
 				ValidateFunc: validation.StringInSlice([]string{
+					"latest_prerelease",
 					"latest",
 					"id",
 					"tag",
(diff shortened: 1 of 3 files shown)


youcandanch:
@kfcampbell anything I need to do on this to get it into a reviewable state? Haven't done work on the provider before so wanted to make sure I'm not missing anything. Thanks in advance!

github/data_source_github_release.go
```diff
@@ -108,8 +109,38 @@ func dataSourceGithubReleaseRead(d *schema.ResourceData, meta interface{}) error
...
 	switch retrieveBy := strings.ToLower(d.Get("retrieve_by").(string)); retrieveBy {
+	case "latest_prerelease":
+		// The GitHub API doesn't specify a way to just load the most recent prerelease,
+		// so we'll load all releases and determine which one is the most recent
+		// prerelease after.
+		log.Printf("[INFO] Refreshing GitHub latest prerelease from repository %s", repository)
+		var releases []*github.RepositoryRelease
+		nextPage := 1
+		// TODO: 10 is sort of arbitrary here -- what's the best way to allow configurability
+		// for this to prevent the provider from becoming glacially slow on large repos?
```
kfcampbell, line 121: This is an excellent question. Do you know what the max number of releases we can retrieve per page with the API?

github/data_source_github_release.go
```diff
@@ -108,8 +109,38 @@ func dataSourceGithubReleaseRead(d *schema.ResourceData, meta interface{}) error
...
+						release = rel
+					}
+				}
+			}
+			if response.NextPage > nextPage {
+				nextPage = response.NextPage
+			} else {
+				break
+			}
+		}
```
kfcampbell, line 143: I'm a little uncomfortable with this logic as-is. What do you think about refactoring it to a function like `GetLatestRelease` that we could write a unit test for?

kfcampbell (commented):
I had a couple of thoughts, and I approved the Action to run. Thanks for creating this!

Ideally we'd have a documentation update to go on the website for the `latest_prerelease` functionality as well.

github/data_source_github_release.go
youcandanch: 100 is the limit.

github/data_source_github_release.go
youcandanch: Yeah, I like that. Will take a stab at it this weekend.

youcandanch:
Still haven't had a moment to swing back to this, but I did open a post on the GitHub community for something similar: https://github.community/t/release-api-should-include-a-way-to-retrieve-the-latest-prerelease/239836

It'd be more ideal to handle it at the API level, I think, though that's an ask for core GitHub team and I imagine a fairly large one to think through.

nickfloyd:

PR #1084, opened 2022-03-06
Add a data source for refs.

youcandanch:
### Description
We're currently lacking a way to retrieve some basic information on a given ref within the Terraform provider. A crisp use case for this might be "I have a release that's based off of a tag, and I'd like to be able to retrieve the SHA1 for the HEAD commit in order to use elsewhere." This PR introduces a `github_ref` data source, which pretty shamelessly steals from the existing `github_branch` data source. I considered lifting out some of the code into a shared function, but the API calls are lightweight enough where I don't think there's that much harm from a few lines of duplication. 

Tests are currently passing and documentation has been added to the markdown.

### DIFF
--- a/github/data_source_github_ref.go
+++ b/github/data_source_github_ref.go
@@ -1,0 +1,67 @@
+package github
+
+import (
+	"context"
+	"log"
+	"net/http"
+
+	"github.com/google/go-github/v42/github"
+	"github.com/hashicorp/terraform-plugin-sdk/helper/schema"
+)
+
+func dataSourceGithubRef() *schema.Resource {
+	return &schema.Resource{
+		Read: dataSourceGithubRefRead,
+
+		Schema: map[string]*schema.Schema{
+			"repository": {
+				Type:     schema.TypeString,
+				Required: true,
+				ForceNew: true,
+			},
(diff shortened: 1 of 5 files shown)


youcandanch:
@kfcampbell this one's probably simpler to land then #1066, which I haven't had as much time to work on. Any chance you could kick the workflows for this as well? Thanks so much!

kfcampbell:
@youcandanch Done! 

Do you also mind linking to the sweet new docs page you just created in the [.erb file](https://github.com/integrations/terraform-provider-github/blob/main/website/github.erb#L14)? Thanks!

youcandanch:
@kfcampbell all set!

kfcampbell (approved):
Thank you for the contribution! I'll shoot to get this released at the end of this week.

youcandanch:
@kfcampbell looks like an older commit of mine made it up here, which means there's some errors in the ref code. I'm gonna fix it, but I'm also gonna glance and see why tests were able to pass, as I don't think there's a way they _should_ be passing right now. Issue opened here: https://github.com/integrations/terraform-provider-github/issues/1108

PR #702, opened 2021-02-12
Use pagination to retrieve all team members

adebasi:
This PR addresses the issue: https://github.com/integrations/terraform-provider-github/issues/701

When using the data source `github_team` a maximum of 30 members is retrieved. With this change, the `per_page` value gets increased to 100 and pagination is used to get all team members.

PR #1092, opened 2022-03-17
fix: use pagination to fetch all team members

carocad:
## If applied, this pull request will...

- use pagination on the new github_team_members resource
- compute a new etag based on the etag of each request

## Why is this change needed?

- Because it currently fetches only 30 members so terraform tries to change the members on each run even though nothing changed

## References to relevant tickets, articles or other resources

- https://github.com/integrations/terraform-provider-github/pull/702
  - I used the same code from that PR

### DIFF
--- a/github/resource_github_team_members.go
+++ b/github/resource_github_team_members.go
@@ -205,17 +205,52 @@ func resourceGithubTeamMembersRead(d *schema.ResourceData, meta interface{}) err
 		ctx = context.WithValue(ctx, ctxEtag, d.Get("etag").(string))
 	}
 
+	etags := make([]string, 0)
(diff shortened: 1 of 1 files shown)


github/resource_github_team_members.go
```diff
@@ -239,7 +274,7 @@ func resourceGithubTeamMembersRead(d *schema.ResourceData, meta interface{}) err
 	}
 
 	// Combine etag of both requests
```
pascal-hofmann, line 276:
```suggestion
	// Combine etag of all requests
```

kfcampbell (approved):
This is great, thank you! I'll shoot to get this released at the end of this week.

PR #1099, opened 2022-03-31
Fix typo in branch protection's `pull_request_bypassers` test

reedloden:
s/dismiss_stale_reviews/pull_request_bypassers/


reedloden:
I'm slightly concerned that this worked until now. Is this test actually testing the right thing?

kfcampbell:
@reedloden the short answer is that our tests are in an unhealthy state, as I've hinted at in #1414. These tests are not run automatically on PRs due to resource constraints. There was an attempt at labelling automation that could trigger test runs, though I don't have access to any of the test accounts configured. There's a long-overdue overhaul in testing waiting to be done, and it's particularly difficult in this instance due to the surface area of the provider and the resource costs of integration testing. 

In any event, thank you for noticing and fixing this particular instance!

Issue #1045, opened 2022-01-21
feature request: Add support for 'public' attribute on github pages configuration


pkwarren:
We'd like to use the Github pages functionality of the terraform provider, however currently it doesn't support the github pages 'public' option: https://github.com/google/go-github/blob/178169fc04bc4a05ca48f98709cdd02ab5b556e3/github/repos_pages.go#L21

### Terraform Version
v1.1.4

### Affected Resource(s)
- github_repository

### Terraform Configuration Files
```hcl
resource "github_repository" "example" {
  name        = "example"
  description = "My awesome web page"
  private = true
  pages {
    source {
      branch = "main"
      path   = "/docs"
    }
    # TODO: Doesn't work
    public = false
  }
}
```

### Expected Behavior
Should be able to configure 'public = false' to limit visibility of github pages.

### Actual Behavior
Option not yet supported.

pkwarren:
It looks like this will require fixes to upstream GitHub module: https://github.com/google/go-github/issues/2261

btkostner:
Looks like it has been included upstream and will be included in the next release

btkostner:
go-github has released v43.0.0 which includes the fix here https://github.com/google/go-github/releases/tag/v43.0.0

btkostner:
This issue should not yet be closed

kfcampbell:
Oh I'm sorry! Thanks for catching this @btkostner.

btkostner:
Currently waiting for PR to merge

ViacheslavKudinov:
We need this feature! 🙏🏻

dweaver33:
![image](https://user-images.githubusercontent.com/6692418/230620467-24a28414-f802-4dc8-a7e9-8dbb25414bbf.png)

calebworks:
Any updates on this?

ViacheslavKudinov:
Hello,
any updates or was it implemented ?

Luis-3M:
What else is needed to make this feature available?

btkostner:
It looks like the upstream lower level API library has this feature released now. Someone just needs to open the PR for it in the terraform provider.

lbartolotti-gocity:
I could use this feature too 👍

BartolottiLuca:
Hello, I've give this feature request a go, anyone know if I have to poke someone to get it reviewed or is just a matter of waiting? 
https://github.com/integrations/terraform-provider-github/pull/2540

[...]

PR #1112, opened 2022-04-12
feat: add repository pages public attribute

btkostner:
This adds a `public` attribute to repository pages. This controls the accessibility of pages. See [the GitHub API docs](https://docs.github.com/en/rest/reference/pages#update-information-about-a-github-pages-site--parameters) for more information. Here is the code in [go-github](https://github.com/google/go-github/blob/master/github/repos_pages.go#L21). Closes https://github.com/integrations/terraform-provider-github/issues/1045

> **Note**: I'm not a golang programmer, so I'm not 100% sure this is everything that is required


kfcampbell (commented):
@btkostner sorry for the delay! Somehow I hadn't seen this yet. Would it be possible to alter the test(s) for this resource to include this new functionality? I'm thinking of some kind of addition of an appropriate test in [this file](https://github.com/integrations/terraform-provider-github/blob/main/github/resource_github_repository_test.go).

github/resource_github_repository_test.go
```diff
@@ -813,6 +813,52 @@ func TestAccGithubRepositoryPages(t *testing.T) {
...
+				pages {
+					source {
+						branch = "main"
+					}
+					public = false
+				}
+			}
+		`, randomID)
+
+		check := resource.ComposeTestCheckFunc(
```
kfcampbell, line 830: Thank you for adding a test to capture this behavior! Is this check passing for you? When I run it, I'm getting `Check failed: Check 1/1 error: github_repository.test: Attribute 'pages.0.public' expected "false", got "true"`. 

It looks like we're only getting the "public" option, never setting it anywhere. Wouldn't we need some logic in the Create and Update functions to set this attribute?

Issue #988, opened 2021-11-27
Allow custom role names for permission argument


emtunc:
As per https://docs.github.com/en/rest/reference/teams#add-or-update-team-repository-permissions - custom role names (https://docs.github.com/en/enterprise-cloud@latest/organizations/managing-peoples-access-to-your-organization-with-roles/managing-custom-repository-roles-for-an-organization) are now permitted when adding/updating permissions for teams.

The Terraform provider still only expects _pull, triage, push, maintain, or admin_

For the avoidance of doubt, I have tested setting permissions with a custom role name (it seems to be case sensitive) via the GH API and it works.

Would someone be so kind as to allow string values for the permission argument please?

joshuahancox:
I raised a PR here, based on another - could anyone help in getting this reviewed/merged? 😃 

https://github.com/integrations/terraform-provider-github/pull/1113

joshuahancox:
@kfcampbell could we merge/release this?

grant-adarga:
Both of the linked PRs are ready to go :shipit: (sorry about the delay in updating mine @joshuahancox)

PR #1038, opened 2022-01-17
feat(github_team_repository): allow for custom repository roles

grant-adarga:
Allow custom repository roles to be passed to the `github_team_repository` resource by removing the string validation on the `permission` argument.

Update the default permission to `push`, which is the default for the [newer API resource](https://docs.github.com/en/rest/reference/teams#add-or-update-team-repository-permissions--parameters); previously the API pulled the default from the (deprecated) `permission` parameter of the `team` resource. This may cause an unintended change for some provider users, so I'll leave this up for the maintainers to decide whether to include or not. 

I also opted to leave the tests unchanged as `"pull", "triage", "push", "maintain", "admin"` are still valid test cases.

Closes: https://github.com/integrations/terraform-provider-github/issues/988

PR #1113, opened 2022-04-12
feat(github_team_repository): allow for custom repository roles

joshuahancox:
This PR is based on the PR here: 

https://github.com/integrations/terraform-provider-github/pull/1038/files

I am waiting for this functionality and so would like to get it merge ASAP, if possible without waiting for a major version.

The previous PR is still valid since changing the default is a breaking change that can be merged before the next major release.

Previous PR details:
> Allow custom repository roles to be passed to the github_team_repository resource by removing the string validation on the permission argument.
> 
> Update the default permission to push, which is the default for the [newer API resource](https://docs.github.com/en/rest/reference/teams#add-or-update-team-repository-permissions--parameters); previously the API pulled the default from the (deprecated) permission parameter of the team resource. This may cause an unintended change for some provider users, so I'll leave this up for the maintainers to decide whether to include or not.
> 
> I also opted to leave the tests unchanged as "pull", "triage", "push", "maintain", "admin" are still valid test cases.
> 

Closes: https://github.com/integrations/terraform-provider-github/issues/988

### DIFF
--- a/github/resource_github_team_repository.go
+++ b/github/resource_github_team_repository.go
@@ -36,7 +36,6 @@ func resourceGithubTeamRepository() *schema.Resource {
 				Type:         schema.TypeString,
 				Optional:     true,
 				Default:      "pull",
-				ValidateFunc: validateValueFunc([]string{"pull", "triage", "push", "maintain", "admin"}),
 			},
 			"etag": {
 				Type:     schema.TypeString,
--- a/website/docs/r/team_repository.html.markdown
+++ b/website/docs/r/team_repository.html.markdown
@@ -44,7 +44,7 @@ The following arguments are supported:
 * `team_id` - (Required) The GitHub team id or the GitHub team slug
 * `repository` - (Required) The repository to add to the team.
 * `permission` - (Optional) The permissions of team members regarding the repository.
-  Must be one of `pull`, `triage`, `push`, `maintain`, or `admin`. Defaults to `pull`.
+  Must be one of `pull`, `triage`, `push`, `maintain`, `admin` or the name of an existing [custom repository role](https://docs.github.com/en/enterprise-cloud@latest/organizations/managing-peoples-access-to-your-organization-with-roles/managing-custom-repository-roles-for-an-organization) within the organisation. Defaults to `pull`.
 
 
 ## Import


joshuahancox:
@kfcampbell do you mind taking a look at this one. Apologies for the trying to hurry things along but this is a functionality I would like to use.

github/resource_github_team_repository.go
```diff
@@ -36,7 +36,6 @@ func resourceGithubTeamRepository() *schema.Resource {
 				Type:         schema.TypeString,
 				Optional:     true,
 				Default:      "pull",
-				ValidateFunc: validateValueFunc([]string{"pull", "triage", "push", "maintain", "admin"}),
```
reedloden, line 39: Is there a way to pull all possible roles from GitHub and then validate against that?

joshuahancox: I am sure we could use [this API](https://docs.github.com/en/rest/orgs/custom-roles) for that. 

Do you know of an example where something similar is done before? Or could you suggest where I should create the function. First time contributor :)

joshuahancox: Seems like we do some other validation in `utils.go` (e.g. validateValueFunc and validateTeamIDFunc) but I don't see anyone calling back to github api here.

joshuahancox: I think we probably need to wait for this: https://github.com/google/go-github/pull/2336? But I am also a little out of my depth here.

btsteve: The go-github functionality for merged a little while ago.

kfcampbell: It's merged but unfortunately has not been released yet; the last release of the library was March 1st. We'll have to wait for their next official release to use the functionality in [go-github#2336](https://github.com/google/go-github/pull/2336).

@joshuahancox is correct and thus far we've chosen to hard-code those validation functions rather than use extra API calls. I'm a little hesitant to change that approach here, especially as we tend to see API overuse on more intensive Terraform runs.

reedloden: https://github.com/integrations/terraform-provider-github/pull/1138 will bump `go-github` to include the necessary functionality.

stefansedich:
Just saw this replacement PR for this functionality, but +10000 from me to get this one in, it is something that is currently blocking us assigning users a custom role with dependabot permissions.

joshuahancox:
Is there anything else I can do to help get this in? :)

github/resource_github_team_repository.go
kfcampbell: I've just updated and merged #1138!

kfcampbell:
@joshuahancox thank you for the offer! The best way to speed this up would be to use the GitHub API to validate the roles created as suggested [here](https://github.com/integrations/terraform-provider-github/pull/1113#discussion_r855624189). I've merged the PR that updates `google/go-github` to the required version.

joshuahancox:
> @joshuahancox thank you for the offer! The best way to speed this up would be to use the GitHub API to validate the roles created as suggested [here](https://github.com/integrations/terraform-provider-github/pull/1113#discussion_r855624189). I've merged the PR that updates `google/go-github` to the required version.

Thanks @kfcampbell, @reedloden, 

My idea would have been to include a `validateTeamRepositoryPermissionFunc()` in `util.go` and validate from the API inside that function, and then use that as the `ValidateFunc` in the `resourceGithubTeamRepository()` function inside `resource_github_team_repository.go`.

However I can't pass the `meta interface{}` required to interact with the API due to the type `SchemaValidateFunc` requiring only a single interface (field value) and a single string (field name). 

I can't see any other example of interacting with the api for validation in the codebase, so unsure where I should implement this. Could you give a little guidance please?

Apologies for the amount of help needed.

joshuahancox:
Also, just realised that there is a problem with the function from go-github - so have raised a pr there too.

https://github.com/google/go-github/pull/2370/files

reedloden:
@joshuahancox Don't forget you need to check against the superset of default roles + any custom roles. As currently written, the proposed change only checks against custom roles, no?

joshuahancox:
@reedloden yes of course, updated thankyou!

Any idea about my other questions regarding the type `SchemaValidateFunc`? What should I do there?

github/util.go
```diff
@@ -192,3 +192,40 @@ func validateSecretNameFunc(v interface{}, keyName string) (we []string, errs []
...
+	valid := false
+	for _, role := range roles {
+		if role == roleName {
+			valid = true
+			break
+		}
+	}
+
+	if !valid {
+		return nil, []error{fmt.Errorf("A custom role with the name %s does not exist in this GitHub organisation", roleName)}
```
reedloden, line 227:
```suggestion
		return nil, []error{fmt.Errorf("A role with the name %s does not exist in this GitHub organisation", roleName)}
```

joshuahancox:
Thanks @reedloden I committed your further suggestion. Any idea on how to handle my other question?

reedloden:
> Thanks @reedloden I committed your further suggestion. Any idea on how to handle my other question?

Sorry, I'm just an outside lurker, so not sure of the best approach. One of the core developers would likely have a better idea on how to handle it.

joshuahancox:
> > Thanks @reedloden I committed your further suggestion. Any idea on how to handle my other question?
> 
> Sorry, I'm just an outside lurker, so not sure of the best approach. One of the core developers would likely have a better idea on how to handle it.

No worries! Thankyou for the help so far anyway. @kfcampbell could you advise me a little further? Or do you know someone that I could @ for some guidance?

https://github.com/integrations/terraform-provider-github/pull/1113#issuecomment-1134360137

joshuahancox:
@jcudit @radeksimko @anGie44 @kfcampbell 

Does anyone have any idea what I should do here? We would like to validate the permission that we are setting on the team/repository resource here against custom roles returned from the github list custom roles API + and the known existing roles, however the type `SchemaValidateFunc` does not allow for passing the `meta interface{}` required to interact with the API.

I can't see a way to do the validation, should we validate in a different way, or remove the validation altogether (as in the original PR)?

anGie44:
Hi @joshuahancox , afik `SchemaValidateFunc` doesn't give you access to the `meta` unfortunately but a `CustomizeDiff` does, with the limitation that it returns an apply-time error I believe instead of at plan-time like when using the validatefunc so that may affect the UX as users usually like to know input errors before apply. You could potentially do something like the following where you transform the current method already in `util.go`:

```go
func resourceGithubTeamRepository() *schema.Resource {
	return &schema.Resource{
		...

		CustomizeDiff: customdiff.Sequence(
			resourceTeamRepositoryPermissionCustomizeDiff,
		),
	}
}

func RepoRoles_Values() []string {
	return []string{
		"pull",
		"triage",
		"push",
		"maintain",
		"admin",
	}
}

func resourceTeamRepositoryPermissionCustomizeDiff(diff *schema.ResourceDiff, meta interface{}) error {
	if roleName := diff.Get("permission").(string); roleName != "" {
		roles := RepoRoles_Values()

		orgName := meta.(*Owner).name
		ctx := context.Background()

		customRoles, _, err := client.Organizations.ListCustomRepoRoles(ctx, orgName)

		if ghErr, ok := err.(*github.ErrorResponse); ok && ghErr.Response.StatusCode == http.StatusNotFound {
			return nil
		}

		if err != nil {
			return fmt.Errorf("Error in response from github while checking for existing custom roles. Error: %s", err)
		}

		for _, role := range customRoles.CustomRepoRoles {
			roles = append(roles, *role.Name)
		}

		valid := false
		for _, role := range roles {
			if role == roleName {
				valid = true
				break
			}
		}

		if !valid {
			return fmt.Errorf("A role with the name %s does not exist in this GitHub organisation", roleName)
		}
	}

	return nil
}
```

The error handling there is just as an example as I'm not too sure if you'd want to always error out when no custom roles are available. Hope this helps though!

kfcampbell:
I apologize I haven't been around much. @grant-adarga has updated his PR to remove the validation entirely, which I'll handle merging/releasing as-is for now. 

I agree it's obnoxious we can't get client/owner information in `SchemaValidationFunc`s. It would be great to have access to the meta there. In the meantime, probably the best course of action would be bumping to the next Terraform SDK version and use a [SchemaValidateDiagFunc](https://pkg.go.dev/github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema#SchemaValidateDiagFunc) to bubble up a warning to the user if their role isn't a default one and hope they have a custom one configured.

PR #1109, opened 2022-04-09
fix: remove incorrect required schema key on ref data source

youcandanch:
In #1084, I introduced a data source for refs, but it looks like the merged version is off an older commit I had pushed, which includes a required schema key on the data source that isn't used by the ref API. This PR introduces the correct code (along with a quick typo fix on the documentation). 

### Test output:
```
--- PASS: TestAccGithubRefDataSource (24.72s)
    --- PASS: TestAccGithubRefDataSource/queries_an_existing_branch_ref_without_error (12.40s)
        --- SKIP: TestAccGithubRefDataSource/queries_an_existing_branch_ref_without_error/with_an_anonymous_account (0.00s)
        --- PASS: TestAccGithubRefDataSource/queries_an_existing_branch_ref_without_error/with_an_individual_account (6.42s)
        --- PASS: TestAccGithubRefDataSource/queries_an_existing_branch_ref_without_error/with_an_organization_account (5.98s)
    --- PASS: TestAccGithubRefDataSource/queries_an_invalid_ref_without_error (12.31s)
        --- SKIP: TestAccGithubRefDataSource/queries_an_invalid_ref_without_error/with_an_anonymous_account (0.00s)
        --- PASS: TestAccGithubRefDataSource/queries_an_invalid_ref_without_error/with_an_individual_account (6.28s)
        --- PASS: TestAccGithubRefDataSource/queries_an_invalid_ref_without_error/with_an_organization_account (6.03s)
PASS
ok      github.com/integrations/terraform-provider-github/v4/github     24.745s
```

PR #1118, opened 2022-04-19
Fix typo in github_ref docs

k24dizzle:
Screen Shot 2022-04-18 at 6 50 23 PM

https://registry.terraform.io/providers/integrations/github/latest/docs/data-sources/ref

Noticed a typo here ^


youcandanch:
Hey @k24dizzle -- caught that one while I was fixing the provider itself in #1109, so this should be resolved in 4.24.0.

PR #1119, opened 2022-04-19
feat: Add a data source for custom GET requests to REST API.

galargh:
###### Description

I wanted to use https://registry.terraform.io/providers/integrations/github/latest/docs/data-sources/repository_file to check if a specific file exists or not. I realised it was not possible because `terraform apply` would fail if a file does not exist because the underlying API call receives 404.

That's how I got an idea to introduce a general purpose API call data source which issues a GET request to a GitHub API endpoint and returns a response.

###### Example

```tf
data "github_rest_api" "check_if_file_exists" {
  endpoint = "repos/ORG/REPO/contents/FILE?ref=BRANCH"
}

output "does_file_exist" {
  value = data.github_rest_api.check_if_file_exists.code == 200 ? "Yes! It does!" : "No :("
}

```

###### Testing
- [x] Ran `TF_ACC=1 go test -v ./... -run ^TestAccGithubRestApiDataSource`
- [x] Installed the provider (`go install`), added it to my project and successfully used the newly added data source


kfcampbell:
I like the idea! What are your thoughts on allowing query parameters and request bodies as well?

kfcampbell (approved):

galargh:
> I like the idea! What are your thoughts on allowing query parameters and request bodies as well?

Oh, sorry, I missed that comment before. Thank you so much for all the reviews and merges ❤️ 

I thought that for a data source, all we want to do is GET requests, so I don't think "body" would be that useful. At least, I'm not aware of any GitHub API GET endpoints that make use of it.

As for query params, in this iteration, they can be provided inline as part of the endpoint string. Making them a separate input could be nice though.

Pagination support could be another useful extension here.

kfcampbell:
No worries! Agreed about query params as a separate input and pagination support. Should you feel motivated to open a PR to add them, I would gladly review it. 

Otherwise, I'm planning on cutting a new version of the provider later today that will ship with this feature.

PR #1149, opened 2022-05-15
Improve `github_repository` import docs

Blacksmoke16:
Add example of an import with `for_each` resource

Took me a while to figure out the exact syntax for a `for_each` context, so hopefully this'll help someone in the future.

### DIFF
--- a/website/docs/r/repository.html.markdown
+++ b/website/docs/r/repository.html.markdown
@@ -160,8 +160,18 @@ The following additional attributes are exported:
 
 ## Import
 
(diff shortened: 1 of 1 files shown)


shoekstra:
Hi @Blacksmoke16! This looks more like a Terraform specific thing than something to add to these docs. How to import a resource when using a `for_each` is considered assumed knowledge, and isn't specific to importing a GitHub resource. The import docs exist to show how to specify the resource you want to import.

I'd be in favour of skipping this from the docs and keeping them consistent with other providers.

PR #1151, opened 2022-05-15
Fix example `github_team_members` import resource address

Blacksmoke16:
Pretty sure this should be `github_team_members` as the singular version didn't work.


kfcampbell (approved):
Good catch, thank you!

PR #1158, opened 2022-05-21
fix: default_branch 404 when repository was deleted

mariux:
This fixes this situation when a repository got deleted and the default branch of the repository can not be refreshed and fails with a 404 due to the missing repository.

In this situation we can assume the default branch vanished with the repository and thus just mark it as gone and not fail with a 404 not found.


CameronHall:
Big +1 on this. We're hitting this same issue and it's _very annoying_.

mariux:
what do we need to get this merged?

PR #1162, opened 2022-05-23
Add only_protected_branches in data source github_repository

bpaquet:
This will allow to retrieve only the protected branches instead of the first 30.


kfcampbell (approved):
Thank you for the contribution!

PR #1173, opened 2022-06-03
Support import via team name

phy1729:
Rather than make the user lookup the team name via the API, use the existing `getTeamID` function to do it for them.


phy1729:
I would still like this feature. I've rebased to resolve a minor conflict that arose.

kfcampbell:
I've approved the Actions run, though it appears there's additional conflicts now.

phy1729:
Rebased for that conflict too.

kfcampbell (approved):

PR #1175, opened 2022-06-06
Add optional repository IDs flag

david-bain:
Note: [PR 1177](https://github.com/integrations/terraform-provider-github/pull/1177), which also migrates the repositories data source to use graphql is preferred over this commit.

Repository IDs are required for some resources. Allowing the github_repositories data source to return the IDs in addition can eliminate the extra step of querying each repository for its ID.

- Add `include_repo_id` optional input flag to return repo IDs
- Add `results_per_page` optional input flag to allow customising the number of results returned per API call (default is 100, which remains the same as prior)
- Add test to confirm


website/docs/d/repositories.html.markdown
```diff
@@ -25,10 +26,11 @@ data "github_repositories" "example" {
 The following arguments are supported:
 
 * `query` - (Required) Search query. See [documentation for the search syntax](https://help.github.com/articles/understanding-the-search-syntax/).
-
-* `sort` - (Optional) Sorts the repositories returned by the specified attribute. Valid values include `stars`, `fork`, and `updated`. Defaults to `updated`.
+* `sort` - (Optional) Sorts the repositories returned by the specified attribute. Valid values include `stars`, `fork`, `name`, and `updated`. Defaults to `updated`.
```
kfcampbell, line 29: `name` isn't actually present in the validation function above. Is it allowed by the API?

david-bain: I wrote this a long time ago so I'm trying to see why I modified that, but I believe it is a mistake that probably carried over as at some point I was going to add a resort to allow name but decided against it to keep it the same as current. I will update to resolve, removing 'name'.

david-bain:
@kfcampbell If you would prefer to merge this as it is less of a change than #1177, I am happy to do so (although I will add the optional `results_per_page` input)

kfcampbell:
@david-bain I'd like to do this PR and then #1177 in a major version update (which I'd like to get in the habit of cutting more frequently). 

Are you comfortable with the state of this PR as-is or would you like to add the results_per_page item here before merging?

david-bain:
Let me add results_per_page as I believe this one as is could fall over with the default value for some people. Give me a few days.

nickfloyd (approved):
❤️

PR #1176, opened 2022-06-06
Add support for a summary view of org teams.

david-bain:
Add a flag input to optionally skip returning the repositories and members on calling the github_organization_teams data source.
This improves performance of the graphql call when only summary information is needed.
Especially important on orgs with large numbers of teams/members.

- Add `summary_only` optional flag input to skip returning members and repositories
- Add 'results_per_page' optional int to reduce the results per page returned.
- Update docs to reflect change


david-bain:
Bumping, hopefully this can get in soon as we cannot use the current data source given the number of teams and members we have.
@kfcampbell are you able to review?

kfcampbell (approved):
Apologies for the delay! This looks good, thank you for contributing. I'll get it released soon.

PR #1177, opened 2022-06-06
Migrate data_source_github_respositories to graphql

david-bain:
Leverage graphql to optimise the API calls to just the required information. Duplicates the functionality of the existing data source.

Additionally, repository IDs are required for some resources. Allowing the github_repositories data source to return the IDs in addition can eliminate the extra step of querying each repository for its ID.

* Migrate data_source_github_repositories to graphql
* Add optional include_repo_id input flag to return repo IDs
* Add optional input "results_per_page" to help alleviate 502 timeout errors
* update tests to validate

This PR replicates the additions of [PR 1175](https://github.com/integrations/terraform-provider-github/pull/1175) while also migrating to graphql.


david-bain:
Closing this for now as there seems to be a bug in the GraphQL github API that is causing duplicated as page boundaries. Once solved I will reopen

david-bain:
Looks like the duplication and missing values issue is also present on the rest version if you sort by stars or forks. Maybe also be present on updated by seemed more reliable.
I will continue to sit on this PR until GitHub gets back to me on if there is a confirmed issue on their side.

david-bain:
This change also solves https://github.com/integrations/terraform-provider-github/pull/1313

david-bain:
Bumping, hopefully this can get in soon as we cannot use the current data source given the number of teams and members we have, it is timing out before it returns.
@kfcampbell are you able to review?

kfcampbell:
Do you mind explaining a little more why you submitted the change to graphQL? What's the relative performance difference? How complex is your plan/how frequently are you running this data source?

david-bain:
> Do you mind explaining a little more why you submitted the change to graphQL? What's the relative performance difference? How complex is your plan/how frequently are you running this data source?

This all came about because when I ran the standard method, it timed out because too much data was being returned. The intention was to specifically target the information that the terraform module actually cares about using a graphql query.

I found that on the standard method used by #1175, I needed to modify the 'results_per_page' hardcoded value to limit the response to not cause a timeout, which is not ideal as that is more API calls. On this one, I didn't need to do that in the end, but I had already added it to try and work out why it was failing and decided to leave it in as someone else may find it useful instead of hardcoding.

I can update #1175 to include the `results_per_page` option but as the graphql version allowed us to return 100 results/page without timeouts I focused my development there as it was more responsive.

On another note, as the sort in the graphql version is implemented by this module, we can add additional methods, such as name etc, which may be useful? Happy to update that if desired as additional options.

kfcampbell:
Makes sense; thanks for the update. I'd like to make this change in a breaking version update, which I'd like to get in the habit of cutting more often.

laughedelic:
Wasn't this change planned for release?

Issue #1152, opened 2022-05-16
github_repository_file adds additional author and committer information when editing a file resulting in unsigned commits


pawnu:
### Terraform Version
Terraform v1.1.9

### Affected Resource(s)
- github_repository_file

### Debug Output
https://gist.github.com/pawnu/51b1d43ce618692446316515dc25fa8a

### Expected Behavior
When a file is edited without committer or author name and email information, the PUT request to edit file should not have committer or author information specified. This will allow github to sign the commit for you with your user/bot token.

### Actual Behavior
The PUT request adds author and committer name and email when editing the file. This results in unsigned commit as github will expect the author to sign the commit also.

See https://gist.github.com/pawnu/51b1d43ce618692446316515dc25fa8a#file-terraform_edit_file-txt line 235 vs https://gist.github.com/pawnu/51b1d43ce618692446316515dc25fa8a#file-terraform_create_file-txt line 282

### Steps to Reproduce
```
resource "github_repository_file" "tftestfile1" {
  repository          = data.github_repository.py-selenium-docker.name
  branch              = "main"
  file                = "tftestfile1"
  content             = "test file via terraform provider edit"
  commit_message      = "Managed by Terraform provider"
  overwrite_on_create = true
}
```

1. Create a file with github_repository_file resource without specifying `commit_author` and `commit_email`, and run apply. The commit which created the file is signed by your Github App/user and verified.
2. Make change to the same file content, and run apply again. The commit which edited the file is from `web-flow` user and unsigned. https://github.com/web-flow

I ran PUT requests with raw github API to create and edit file without committer and author information, and both result in signed commit.

CREATE
```
curl -i -X PUT   -H "Authorization: token ghs_y[...redacted...]"   -H "Accept: application/vnd.github.v3+json"   https://api.github.com/repos/pawnu/py-selenium-docker/contents/.github/workflows/testcreatefile   -d '{"message":"Create new file","content":"bXkgbmV3IGZpbGUgY29udGVudHM="}'
```
https://gist.github.com/pawnu/51b1d43ce618692446316515dc25fa8a#file-rest_api_create_file-txt

EDIT
```
[...]

PR #1179, opened 2022-06-06
Fix `github_repository_file`: Disable default compute of author and email

pawnu:
When creating a file without author and email information, github will add the following default author and email values.

```
      "commit_author": "GitHub",
      "commit_email": "noreply@github.com",
```

When the same file is edited, with Computed set to `false` and no author and email information provided, it will correctly exclude the author and email information during PUT request, i.e. set to null.
```
      - commit_author       = "GitHub" -> null
      - commit_email        = "noreply@github.com" -> null
```
This will allow github to sign your commit with your bot/user's access token. Otherwise, github will expect you to create a signed commit as above default user.

Fixes https://github.com/integrations/terraform-provider-github/issues/1152


pawnu:
I believe this will need more work as the state will save `commit_author`, and `commit_email` as null, and when comparing against the previous commit of the file in question, it will try to remove the default author and email information even when there has been no change to the file.

tjcorr:
> This means there will be a change every time you run terraform plan to remove the author/email information.

I'm wondering if this is actually the desired behavior? Let's think of an unusual but possible scenario. Let's say I create a file without specifying the commit author/email using terraform with credentials from `user A`. Terraform will appropriately not pass this information to GitHub and the commit will get signed by `user a` by default. Now later, without making any changes to the resource, I rerun terraform but now using credentials from `user b`.  What should happen? Technically the computed committer author/email has indeed changed and we should be updating the file with this new information.  I don't think that info is known until Terraform actually applies using whatever credentials are supplied hence why it should always show a change every time you re-run. I agree this seems weird since we didn't specify the committer information in the first place. However maybe it is required if we want to handle the scenario when a committer is explicitly specified and changed from one value to another?

Part of me feels the committer fields should always just be computed and not able to be specified by a user. Then we could safely ignore the fields and only update the file on a SHA change. This would break anyone manually specifying that information. Perhaps there is some value to being able to set the author/email to values like "Terraform Bot" or team email of those responsible for the automation . 

This leaves us trying to balance out which approach makes the most sense or if we need to put in some special logic when committer information isn't provided.

rwblokzijl:
Another option might be to add a `sign_commit` flag that is mutually exclusive with author/email.

kfcampbell:
I like the idea of adding a `sign_commit` flag. If the PR is updated to include it, I'll review and merge!

rwblokzijl:
@pawnu will you be able to fix this? I would, but I don't know anything about go...

pawnu:
I can try but I'm a beginner at Go too so it might take some time. Happy if someone takes it from here too as I want this to get fixed soon.

pawnu:
I think `commit_as_token` makes more sense? 

Also, I couldn't manage to create signed commit for deletion action even with the raw [github API](https://docs.github.com/en/rest/repos/contents#delete-a-file) by omitting committer/author information so I don't think it's possible with terraform at this instance.

rwblokzijl:
Thanks for the implementation :)

I think `sign_commit` is more clear to the user while `commit_as_token` exposes implementation details. As a user just wanting signed commits its unclear this flag would do that.

Regarding the deletion event not being signed: I think this is a bug at the github API. Personally i don't think it should stop us from implementing this flag. I submitted the following [bug report](https://github.com/community/community/discussions/29668)

pawnu:
I'd say this is still a workaround and we're not intentionally signing commits so I'm not sure. I'd expect a flag like `sign_commit` to take in your signature key to create sign commits as your specified commit author. Seems like this functionality does exist in go-github https://github.com/google/go-github/blob/master/github/git_commits.go#L91

rwblokzijl:
I really think this is an intended feature we're using.

Github always "verifies" your commits when doing things through their own verified services (clicking "update branch" on a PR, editing inside the browser, etc). If a commit if verified like this github add the following popup when clicking the vefified tag:
2022-08-16-232636_264x98_scrot
However, if the commit is made by a user and signed by a known GPG key it looks as follows:
2022-08-16-232732_255x92_scrot
Github marks a commit as verified iff it can verify without a doubt that the github user associated with the commit is actually the one that created it. This is either by them adding a GPG key to their personal account and using that to sign the commit. Or by editing a file <ins>while authenticated</ins> to github.

In our case <ins>we are authenticated</ins> to github (by API token). Thus github should verify the commit (unless the username/email doesn't match).

rwblokzijl:
Actually, given my previous comment I am now in favour of the original implementation. Just signing it by default if the author/email are omitted as this mirrors the github features. @kfcampbell would you be willing to merge given this argument or do you still thing a flag is needed?

kfcampbell:
> Actually, given @pawnu and my previous comments I am now in favour of the original implementation. Just signing it by default if the author/email are omitted as this mirrors the github features.

I believe that the current state of the PR reflects a flag being added, not the original implementation. @rwblokzijl to clarify: you're in favor of not merging this PR at all, and relying on the existing state of the code?

rwblokzijl:
No, i am in favor of the original state of the PR: If author and email are ommited, allow github to verify the commit.

pawnu:
> I believe this will need more work as the state will save `commit_author`, and `commit_email` as `null`, and when comparing against the previous commit of the file in question, it will try to change the upstream default author and email information to the local state `null` even when there has been no change to the file. This means there will be a change every time you run terraform plan to remove the author/email information.

So, I guess this `resourceGithubRepositoryFileRead` is used to read and set the state? It seems to resolve the above issue now.

kfcampbell:
I'm not sure I'm seeing the correct behavior when testing this locally. 

My Terraform file looks like this:

```
provider "github" {
}

terraform {
  required_providers {
    github = {
    }
  }
}


  name      = "repo-file-test-repo-1"
  auto_init = true
}

resource "github_repository_file" "test" {
  repository     = github_repository.test.name
  branch         = "main"
  file           = "test"
  content        = "baz"
  commit_message = "Managed by Terraform2"
  commit_author  = "Terraform User" # comment to test omission of commit_author
  commit_email   = "terraform@example.com" # comment to test omission of commit_email
}
```

On both the main branch and this feature branch, I created a repo with auto-init on, then created a file _without_ the included commit_author and commit_email, then edited that file afterward with a slightly different file content _with_ the commit_author and commit_email present. 

In both cases, the result was that omitting the commit author and email did _not_ cause GitHub to use itself for the committer. Instead, it used my local git credentials (same token as I provided in the `GITHUB_TOKEN` environment var), but picked up my commit_author and email from there. In neither case was the subsequent commit verified:

image

This change doesn't seem like it creates the desired functionality. What am I missing here?

pawnu:
@kfcampbell I tested this with github app so I was assuming it would be the same with PAT. I guess there is no way to sign a commit as a user with terraform github at present? 

I replicated your tests with my bot user and 2 files, and they seem to be working as expected.

**File1**
file1

**File2**
file2

kfcampbell:
:+1: it definitely seems to be an app-only feature. I think we could still merge this functionality if the docs and/or naming were updated to make that clear.

pawnu:
@kfcampbell Done, I have updated the docs but feel free to edit or clarify further. Thanks

website/docs/r/repository_file.html.markdown
```diff
@@ -47,9 +47,9 @@ The following arguments are supported:
 * `branch` - (Optional) Git branch (defaults to `main`).
   The branch must already exist, it will not be created if it does not already exist.
 
-* `commit_author` - (Optional) Committer author name to use.
+* `commit_author` - (Optional) Committer author name to use. **NOTE:** Github app users may omit author and email information so github can verify commits as the Github App. This is useful when a branch protection rule requires signed commits.
```
kfcampbell, line 50:
```suggestion
* `commit_author` - (Optional) Committer author name to use. **NOTE:** GitHub app users may omit author and email information so GitHub can verify commits as the GitHub App. This maybe useful when a branch protection rule requires signed commits.
```

website/docs/r/repository_file.html.markdown
```diff
@@ -47,9 +47,9 @@ The following arguments are supported:
...
 
-* `commit_email` - (Optional) Committer email address to use.
+* `commit_email` - (Optional) Committer email address to use. **NOTE:** Github app users may omit author and email information so github can verify commits as the Github App. This is useful when a branch protection rule requires signed commits.
```
kfcampbell, line 52:
```suggestion
* `commit_email` - (Optional) Committer email address to use. **NOTE:** GitHub app users may omit author and email information so GitHub can verify commits as the GitHub App. This may be useful when a branch protection rule requires signed commits.
```

github/resource_github_repository_file.go
```diff
@@ -84,13 +84,13 @@ func resourceGithubRepositoryFile() *schema.Resource {
 			"commit_author": {
 				Type:        schema.TypeString,
 				Optional:    true,
-				Computed:    true,
+				Computed:    false,
 				Description: "The commit author name, defaults to the authenticated user's name",
```
kfcampbell, line 88:
```suggestion
				Description: "The commit author name, defaults to the authenticated user's name. GitHub app users may omit author and email information so GitHub can verify commits as the GitHub App. ",
```

github/resource_github_repository_file.go
```diff
@@ -84,13 +84,13 @@ func resourceGithubRepositoryFile() *schema.Resource {
...
-				Description: "The commit author name, defaults to the authenticated user's name",
+				Computed:    false,
+				Description: "The commit author name, defaults to the authenticated user's name. GitHub app users may omit author and email information so GitHub can verify commits as the GitHub App. ",
 			},
 			"commit_email": {
 				Type:        schema.TypeString,
 				Optional:    true,
-				Computed:    true,
+				Computed:    false,
 				Description: "The commit author email address, defaults to the authenticated user's email address",
```
kfcampbell, line 94:
```suggestion
				Description: "The commit author email address, defaults to the authenticated user's email address. GitHub app users may omit author and email information so GitHub can verify commits as the GitHub App.",
```

kfcampbell (approved):
Thank you for following up!

PR #1185, opened 2022-06-10
Create data source for external groups

### LINKED ISSUE #1145: GitHub EMU external groups data source
Hi there,

### Terraform Version

```shell
Terraform v1.1.9
on linux_amd64
+ provider registry.terraform.io/integrations/github v4.25.0-alpha
```

### Affected Resource(s)

- github_emu_group_mapping

### Expected Behavior

The GitHub Terraform provider should provide a data source to retrieve the GitHub EMU external groups `group_id` attribute so we can use it with the [github_emu_group_mapping ](https://registry.terraform.io/providers/integrations/github/latest/docs/resources/emu_group_mapping) resource.

### Actual Behavior

Actually the GitHub Terraform provider only  provides the data source [github_organization_team_sync_groups ](https://registry.terraform.io/providers/integrations/github/latest/docs/data-sources/organization_team_sync_groups) that does not work with GitHub EMU external groups. An alternative is to retrieve the external groups with the [http data source](https://registry.terraform.io/providers/hashicorp/http/latest/docs/data-sources/http) and use the body of the HTTP response to create the external groups `group_id` attribute reference.

For example, we can create a data source and a local containing all the external groups attributes (group_name, group_id, updated_at) like that:

```hcl
data "http" "external_groups" {
  url = "https://api.github.com/orgs/${var.github_org}/external-groups"

  request_headers = {
    Accept        = "application/vnd.github.v3+json"
    Authorization = "Bearer ${var.github_token}"
  }
}

locals {
  external_groups = jsondecode(data.http.external_groups.body)["groups"]
}
```

This should give us what we need but it would be far way better to get the attributes reference from a provider data source instead of having to deal with the raw body HTTP response coming from the REST API.

kfcampbell:
Fixes #1145.

When given a Terraform snippet like the following: 

```
provider "github" {}

terraform {
  required_providers {
    github = {
    }
  }
}

locals {
  local_groups = "${data.github_external_groups.example_external_groups}"
}

output "groups" {
  value = local.local_groups
}

```

this PR will produce a state that looks like so: 

```
Outputs:

groups = {
    external_groups = [
        {
            group_id   = 0
            group_name = "Test Group 0"
            updated_at = "2021-02-22T19:25:55Z"
        },
        {
            group_id   = 1
            group_name = "Test Group 1"
            updated_at = "2021-02-22T19:25:55Z"
        }
    ]
    id              = "/orgs/TEST_ORG_NAME/external-groups"
}
```

This will be helpful in retrieving groups for use in the [EMU group mapping resource](https://registry.terraform.io/providers/integrations/github/latest/docs/resources/emu_group_mapping).


kfcampbell:
@micbegind would the state captured in the PR description fix your issue?

philthethrill99:
@kfcampbell Yes this would fix our issue

Issue #1183, opened 2022-06-07
resource_github_team_repository cannot handle custom role after creation


joshua-hancox:
## Affected Resource(s)

- resource_github_team_repository

## Expected Behavior
After creating a resource_github_team_repository with a custom role set in the permission field, making no changes, and re-running the plan, I expect no changes to be planned by terraform.

## Actual Behavior
After creating the resource and re-running the plan, terraform plan tries to change from the built in role that the custom role was based on, to the role that was originally set.

## Steps to Reproduce
e.g. config
```hcl
resource "github_team_repository" "custom_role" {
  team_id    = github_team.this.id
  repository = github_repository.this.name
  permission = "my-custom-role"
}
```
(assume that `my-custom-role` was created in github, and based on the built in maintainer role, with some added permissions)

1. Run terraform apply
2. Once the terraform apply has been applied, run terraform plan

The plan output will show:

```bash
  # github_team_repository.custom_role will be updated in-place
  ~ resource "github_team_repository" "custom_role" {
        id         = "5817097:my-repo-name"
      ~ permission = "maintainer" -> "my-custom-role"
    }
```

## Suggested fix

I think this issue comes from [this function](https://github.com/integrations/terraform-provider-github/blob/366aac0e7ff3a9c6f44c26c397ca4aefb1ec0803/github/util_permissions.go#L20) being called [here](https://github.com/integrations/terraform-provider-github/blob/main/github/resource_github_team_repository.go#L135) to guess what the permission is that is applied.

It assumes that there are only built in roles, which is no longer the case.

I have extended the Repository struct in go-github with [this PR](https://github.com/google/go-github/pull/2379) to include the role_name field in the repository that is returned(automatically includes the `.GetRoleName()` function), and also raised a [PR in this repo](https://github.com/integrations/terraform-provider-github/pull/1182) to bump to the latest version of that module.

Once that is complete I think we should be able to remove the function `getRepoPermission` altogether and amend `resource_github_team_repository` to do it's permission look up from:

```golang
	permName, permErr := getRepoPermission(repo.GetPermissions())
	if permErr != nil {
		return permErr
	}

	d.Set("permission", permName)
```
to
```hcl
	d.Set("permission", repo.GetRoleName())
```

If that makes sense, and we can merge the PR to bump the module then let me know and I can raise a new PR for the fix.

joshuahancox:
[...]

PR #1192, opened 2022-06-16
use GetRoleName() instead of permission mapping func

joshua-hancox:
Requires go-github version bump: https://github.com/integrations/terraform-provider-github/pull/1205

fixes: https://github.com/integrations/terraform-provider-github/issues/1183


kfcampbell (approved):
Thank you for doing this! Apologies for the delay.

csainty:
This breaks the resource for non-custom roles as they naming is different. `read` vs `pull`, `write` vs `push` etc.

Tenzer:
I'm seeing the same as @csainty. Confusingly, around half of the repositories we manage seem to use "push" and the other half "write". It could look like the oldest repositories we have wants to use "push", so I'm guessing it might have something to do with when the repository is created.

Not having it mapped to the same name makes it difficult to use `for_each` to assign a list of repositories to a group with the correct permissions.

schans:
Same here. This is a breaking change for use. If we keep it as is some 300 repos will be changed on every run, if we change push to write, some 450 repos will be changed every run. 

We should revert or create some option to have aliases or something.

schans:
Created a new ticket in https://github.com/integrations/terraform-provider-github/issues/1222 as comments here might get unnoticed.

PR #448, opened 2020-05-08
fix assets_url attribute typo

smiller171:
I noticed that the attribute `assets_url` was typod to be `asserts_url`. This PR is my attempt to fix that mistake without breaking anyone's existing Terraform code.

~~I'm not sure I've handled the deprecation message properly, but I'm happy to make any adjustments needed.~~

I definitely didn't do the deprecation message right and have removed it to allow the tests to pass. Any guidance on how to implement the deprecation message would be appreciated.

PR #1194, opened 2022-06-19
Fix asserts_url typo

shabbyrobe:
This is an attempt to revive #448, which seems to have stalled. The change is largely the same, but I have also followed the instructions for deprecation found [here](https://www.terraform.io/plugin/sdkv2/best-practices/deprecations#renaming-a-computed-attribute).


shabbyrobe:
@kfcampbell - hope you don't mind me pushing my luck and pinging you on this one too 😄

This feels like it's relatively benign, do you think this is something that could be merged?

kfcampbell (approved):
Thanks for bringing this up! This has always bothered me. 

I wouldn't mind cutting a major release to remove the deprecated version at some point in the near future as well. I'd like to get in the habit of releasing more often.

kfcampbell:
Integration tests are passing locally for me.

PR #1197, opened 2022-06-20
feat: Add child teams to github teams response

maxforasteiro:
## Description

This allows the community to use the organization-like structure that can be built on GitHub to manage their assets on terraform, avoiding duplication of structure and allowing teams to be managed in a single place

## Disclaimer

The `github/data_source_github_team_test.go` test didn't cover the success case, since this would require mocking some API calls and this would introduce a big change-set, I opt to not do it. If you know a better and easier way to implement a test to cover not only the new implementation but the rest of the code, I'll be more than happy to do 😄


maxforasteiro:
hey @kfcampbell, could you maybe take a look here? (sorry for pinging you directly - I just saw that you've been active on the repo for the last couple of weeks)

kfcampbell:
@maxforasteiro I'm sorry for the delay here, this slipped through the cracks for me. Is this functionality still something you're interested in getting in?

PR #1211, opened 2022-07-03
Send a `User-Agent` in GitHub API requests with `terraform-provider-github` and the version

timrogers:
This updates the transport used to make requests to the GitHub API to set a `User-Agent` request header with `terraform-provider- github` and the current version (e.g. `terraform-provider-github/4.26.1`).

Because this hooks in with the `RoundTrip` function, it can be expected to overwrite `User-Agent` values set elsewhere (e.g. as defaults or on a per-request basis by `go-github` or `shurcooL/githubv4`).

This change does mean that someone will have to update the `Version` constant with the new version when it is incremented. I have documented that in `RELEASE.md`.

This is only the second Go I've ever written, so I definitely won't be at all sensitive to feedback!

### DIFF
--- a/github/config.go
+++ b/github/config.go
@@ -32,9 +32,10 @@ type Owner struct {
 	IsOrganization bool
 }
 
-func RateLimitedHTTPClient(client *http.Client, writeDelay time.Duration, readDelay time.Duration) *http.Client {
+func CustomizedHTTPClient(client *http.Client, writeDelay time.Duration, readDelay time.Duration) *http.Client {
 
 	client.Transport = NewEtagTransport(client.Transport)
+	client.Transport = NewUserAgentTransport(client.Transport)
 	client.Transport = NewRateLimitTransport(client.Transport, WithWriteDelay(writeDelay), WithReadDelay(readDelay))
 	client.Transport = logging.NewTransport("Github", client.Transport)
 
@@ -49,7 +50,7 @@ func (c *Config) AuthenticatedHTTPClient() *http.Client {
 	)
 	client := oauth2.NewClient(ctx, ts)
 
-	return RateLimitedHTTPClient(client, c.WriteDelay, c.ReadDelay)
+	return CustomizedHTTPClient(client, c.WriteDelay, c.ReadDelay)
 }
 
 func (c *Config) Anonymous() bool {
@@ -58,7 +59,7 @@ func (c *Config) Anonymous() bool {
 
 func (c *Config) AnonymousHTTPClient() *http.Client {
 	client := &http.Client{Transport: &http.Transport{}}
-	return RateLimitedHTTPClient(client, c.WriteDelay, c.ReadDelay)
+	return CustomizedHTTPClient(client, c.WriteDelay, c.ReadDelay)
 }
 
 func (c *Config) NewGraphQLClient(client *http.Client) (*githubv4.Client, error) {
--- a/github/transport.go
+++ b/github/transport.go
@@ -13,8 +13,9 @@ import (
 )
 
 const (
-	ctxEtag = ctxEtagType("etag")
-	ctxId   = ctxIdType("id")
+	ctxEtag        = ctxEtagType("etag")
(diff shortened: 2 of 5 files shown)


github/config.go
```diff
@@ -32,9 +32,10 @@ type Owner struct {
 	IsOrganization bool
 }
 
-func RateLimitedHTTPClient(client *http.Client, writeDelay time.Duration, readDelay time.Duration) *http.Client {
+func CustomizedHTTPClient(client *http.Client, writeDelay time.Duration, readDelay time.Duration) *http.Client {
```
timrogers, line 35: I've elected to rename this as I think the new name better reflects what it does.

github/transport.go
```diff
@@ -13,8 +13,9 @@ import (
 )
 
 const (
-	ctxEtag = ctxEtagType("etag")
-	ctxId   = ctxIdType("id")
+	ctxEtag        = ctxEtagType("etag")
+	ctxId          = ctxIdType("id")
+	packageVersion = "4.26.1"
```
timrogers, line 18: Someone will have to update this when the version is bumped - perhaps there is a way to avoid that!

github/transport_test.go
```diff
@@ -45,6 +45,39 @@ func TestEtagTransport(t *testing.T) {
 	}
 }
 
+func TestUserAgentTransport(t *testing.T) {
+	ts := githubApiMock([]*mockResponse{
+		{
+			ExpectedUri: "/repos/test/blah",
+			ExpectedHeaders: map[string]string{
+				// TODO: Figure out if there is a way to make this test more dynamic
+				"User-Agent": "terraform-provider-github/4.26.1",
```
timrogers, line 54: What would be the best way for me to refer to the `packageVersion` from `transport.go` here?

github/transport_test.go
```diff
@@ -45,6 +45,39 @@ func TestEtagTransport(t *testing.T) {
 	}
 }
 
+func TestUserAgentTransport(t *testing.T) {
```
timrogers, line 48: I'd love to have some kind of integration test for REST and GraphQL that shows that the header is actually set, but I couldn't find a super natural place to do that. Any ideas?

timrogers:
@kfcampbell 👋🏻 Hope you're doing well! Would you be able to give this a quick review?

github/config.go
kfcampbell: I'm good with that!

github/transport_test.go
kfcampbell: Hmm...were we to do that, we'd need actual credentials querying an actual repo. I think both adding a new file or expanding the current transport tests would be appropriate. 

Docs about writing Terraform acceptance tests are [here](https://www.terraform.io/plugin/framework/acctests) or look at existing tests [here](https://github.com/integrations/terraform-provider-github/blob/main/github/resource_github_branch_protection_test.go#L13) as a model. 

Let me know if you'd like to pair on this!

github/version.go
```diff
@@ -0,0 +1,3 @@
+package github
+
+const Version = "4.26.1"
```
kfcampbell, line 3: I don't love the idea of introducing this file. I'd rather use a solution in which we use linker flags to set the version corresponding to the git tag in our build system. 

I'm imagining using the `ldflags` option in Goreleaser [here](https://goreleaser.com/customization/build/#builds) in order to auto-increment the build version. Does that make sense?

kfcampbell:
I've run the PR checks and gave a couple of ideas on how to change the PR. Please let me know if you'd like to pair on updating this!

jeremyhayes:
👋 @timrogers @kfcampbell Any chance of reviving this? Seems super valuable to be able to segment API callers using this provider vs hand-rolled uses of `go-github`.

kfcampbell:
@jeremyhayes I love the idea; we've unfortunately been rather resource-constrained recently and this hasn't been picked back up yet. PRs are appreciated!

PR #1216, opened 2022-07-08
Fix github_team_members: Support github_team_members role edits

csainty:
When a member's role is changed, only the delete operation is run.
The code clearly sets both `delete` and `create` on lines 140:141, but then uses `continue` to loop the iteration after running just one of them.

From what I can work out, the test case already checks this and is failing. I have therefore left it unchanged and it *should* start passing.

I say should because in my local environment I had to edit the test to not mess with org membership, otherwise my collaborator account was pending invitation to the org and failing the test because of that. I am thinking your test setup approves those invitations automatically.
With that change made this PR takes the test from `FAIL` to `PASS` for me.


csainty:
@stawik-mesa as the original author, are you willing to check my thinking here and confirm what I am seeing? 🙏

stawik-mesa:
@csainty I think you are right. Probably I had not enough sleep at the time I implemented it.

stawik-mesa (approved):

csainty:
> I think you are right. Probably I had not enough sleep at the time I implemented it.

Don't beat yourself up, simple mistake.

Now let's tag @jcudit and @kfcampbell as the original approvers and see if they can weigh in.

kfcampbell:
I've been able to reproduce the failure and subsequent test passing following @csainty's advice locally. I do not believe that the integration tests are being run correctly in CI. I'll get this merged shortly and released soon, and prioritize work on our integration testing setup.

PR #1219, opened 2022-07-08
feat: surpress ETag changes for member resources

niels-s:
Suppress the ETag changes for the `github_membership`, `github_team`, and `github_team_members`.
Following the example of
https://github.com/integrations/terraform-provider-github/pull/909.

Ignoring these changes with LifeCycle rules didn't work either.

I don't know if these changes cause any other impact. Besides, the changes won't show up in the plan anymore.
Since the referenced PR didn't add any tests for these changes, I omitted them too. Let me know if it's still desired to add tests anyway.


niels-s:
@kfcampbell I understand you are busy. But could I get some feedback on this PR if this is in line with expectations or not? 
Perhaps I can get the workflow for it approved so the test can run and I can fix any broken things?

kfcampbell (approved):
I am unable to reproduce changing etags locally, but I don't mind suppressing them by default following the model of previous maintainers.

kfcampbell:
Oooh actually, looks like I misunderstood! Tests are failing:

```
        provider_test.go:37: err: 4 errors occurred:
            	* resource github_membership: etag: DiffSuppressFunc is for suppressing differences between config and state representation. There is no config for computed-only field, nothing to compare.
            	* resource github_team: etag: DiffSuppressFunc is for suppressing differences between config and state representation. There is no config for computed-only field, nothing to compare.
            	* resource github_team_members: etag: DiffSuppressFunc is for suppressing differences between config and state representation. There is no config for computed-only field, nothing to compare.
            	* resource github_team_membership: etag: DiffSuppressFunc is for suppressing differences between config and state representation. There is no config for computed-only field, nothing to compare.
```

Looks like this isn't a generator of any diffs, so we can't suppress it. I guess that explains why I couldn't reproduce it.

PR #1220, opened 2022-07-15
feat(github_ip_ranges): support for Web and API IP Ranges

Jamian:
Adding the following attributes to the `github_ip_ranges` data resource:
* `web`
* `web_ipv4`
* `web_ipv6`
* `api`
* `api_ipv4`
* `api_ipv6`

Somebody else had an old PR for this here - https://github.com/integrations/terraform-provider-github/pull/483 which was blocked pending GraphQL API support.


kfcampbell:
Hmm, looks like `Web` doesn't exist on the github.com/google/go-github@v45.2.0 type `APIMeta`.

github/data_source_github_ip_ranges.go
```diff
@@ -22,7 +22,12 @@ func dataSourceGithubIpRanges() *schema.Resource {
 				Computed: true,
 				Elem:     &schema.Schema{Type: schema.TypeString},
 			},
-			"pages": {
```
kfcampbell, line 25: This pages attribute disappears, but is still present in the docs file `ip_ranges.html.markdown`. Did you mean to remove it?

github/data_source_github_ip_ranges.go
```diff
@@ -144,6 +169,16 @@ func dataSourceGithubIpRangesRead(d *schema.ResourceData, meta interface{}) erro
 		return err
 	}
 
+	cidrWebIpv4, cidrWebIpv6, err := splitIpv4Ipv6Cidrs(&api.Api)
```
kfcampbell, line 172: The properties present on the `ApiMeta` struct are defined [here](https://github.com/google/go-github/blob/master/github/misc.go#L142) and don't include `Api` or `Web`. As best I can tell looking at the [file history](https://github.com/google/go-github/commits/master/github/misc.go), they haven't included that information in the past, either. 

Perhaps there's another way of getting those CIDRs?

github/data_source_github_ip_ranges.go
Jamian: No you're quite right, good spot. I will make sure this is added back in!

github/data_source_github_ip_ranges.go
Jamian: @kfcampbell You're right. I've opened a PR in the Go provider to support exposing these `API` and `Web` blocks. Once that's in and a part of a release I'll revisit this.

Jamian: Change was merged, just waiting on https://github.com/integrations/terraform-provider-github/pull/1258 and I'll get any changes in and we can try this again 😄

kfcampbell: Approved/merged!

Jamian: @kfcampbell updates made, thanks for the reminder!

kfcampbell (approved):
I've locally confirmed acceptance testing is passing. Thank you for the contributions! This functionality will be released this week.

PR #1223, opened 2022-07-18
Remove mistakenly committed binary

kfcampbell:
This binary was mistakenly committed as part of #1162. I'm avoiding force-pushing right now as I don't want to disrupt a linear history for the project, though I can re-evaluate that decision if necessary.


james-callahan:
This is a pretty big binary to clone as part of this repo.
I suggest force-pushing sooner rather than later; otherwise you'll have people adding it back in PRs.

PR #1229, opened 2022-07-20
Docs: remove references to "name" parameter in repository_webhook, due to parameter removal

metalsp0rk:
When working with provider version 4.27.1 and following the github_repository_webhook guidelines, I get this error:
```
Error: "name": [REMOVED] The `name` attribute is no longer necessary.
```

This PR updates the docs to address that oversight.


kfcampbell:
Hmm...I think this is still valid. @metalsp0rk is this something you're still interested in getting in?

PR #1242, opened 2022-07-31
Add membership_type to data.github_team

bpaquet:
Add an option to `data.github_team` to query only direct member.

We need to use graphql for this because the Rest API does not offer the feature.


bpaquet:
@kfcampbell this PR is binary free :) :(

kfcampbell:
I've been able to see successful integration tests before and after this change. Thanks!

PR #1251, opened 2022-08-02
Update team_members.html.markdown

nikwotton:
Updating the documentation to reflect current behavior - if members is meant to be optional I can open an Issue instead, but either way the documentation does not match behavior right now


kfcampbell:
Thanks for contributing! We've corrected the docs in the meantime so I'm going to close this out.

PR #1255, opened 2022-08-12
Allow skip fetch members/repos for github_team data source

jeffgran:
Since a lot of resources (graphql-based ones) now require us to use the team's Node ID instead of just the slug, we often have to use data sources just to look up the ID of the team. However, this is an extreme performance drag because some teams have dozens or hundreds of repositories and members to look up page-wise. We don't need any of that, all we need is the Node ID. So we'd like an option to disable those lookups to speed this up and reduce the chance of rate limiting issues.

### DIFF
--- a/github/data_source_github_team.go
+++ b/github/data_source_github_team.go
@@ -18,6 +18,16 @@ func dataSourceGithubTeam() *schema.Resource {
 				Type:     schema.TypeString,
 				Required: true,
 			},
+			"fetch_members": {
+				Type:     schema.TypeBool,
+				Optional: true,
+				Default:  true,
+			},
+			"fetch_repositories": {
+				Type:     schema.TypeBool,
+				Optional: true,
+				Default:  true,
+			},
 			"name": {
 				Type:     schema.TypeString,
 				Computed: true,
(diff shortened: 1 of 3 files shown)


Esysc:
It looks interesting. We are reaching more and more the rate limits nowadays. @kfcampbell this PR is almost 1 year old, it would be possible to have it merged in a next release?

kfcampbell:
@jeffgran Do you have interest in resolving the merge conflicts here?

jeffgran-dox:
This has actually already been implemented in a different way. In the latest version of the provider we are using this:

```hcl
data "github_team" "foo" {
  summary_only = true
  slug = "foo"
}
```
The `summary_only` flag effectively does the same thing.

See: https://github.com/integrations/terraform-provider-github/pull/1379

Esysc:
> This has actually already been implemented in a different way. In the latest version of the provider we are using this:
....
> The `summary_only` flag effectively does the same thing.
> 
> See: #1379

Thank you @jeffgran-dox, I've missed that one.

kfcampbell:
Ahh thanks for pointing that out @jeffgran-dox! I'll close this now.

PR #1294, opened 2022-09-19
Remove mention of `name` from repository_webhook

abatilo:
According to:
https://github.com/integrations/terraform-provider-github/blob/06054eb8fdaa96354e4706583fc05c3d06d27760/github/resource_github_repository_webhook.go#L40

This field is no longer necessary. When I tried to create a webhook using provider version v5.0.0 using TF Cloud, the `Plan` failed and let me know that this field has been removed. The documentation still mentions it though, but I think it should be removed!


kfcampbell (approved):
@abatilo sorry this fell through the cracks for me! It's interesting that our [REST docs](https://docs.github.com/en/rest/webhooks/repos?apiVersion=2022-11-28#create-a-repository-webhook) still say `name` is an acceptable body parameter, though it probably should be removed there as well.

abatilo:
No worries! It happens. I'm just glad that this got merged at all :P

PR #1308, opened 2022-09-28
adding support to pass app_id

junli-lyft:
https://docs.github.com/en/rest/branches/branch-protection#set-status-check-contexts


MJavedAli:
Has anyone tested this, and that it works fine ? I would like see this PR getting merged.

junli-lyft:
Not really. We needed this change but now we had other priorities so the work as paused.
Feel free to take it away and continue.

Issue #1328, opened 2022-10-13
5.4.0 is much slower


bootswithdefer:
5.3.0 plan/apply takes ~25 minutes
5.4.0 plan/apply takes ~65 minutes

### Terraform Version
Terraform v1.3.2
on linux_amd64
+ provider registry.terraform.io/integrations/github v5.4.0

### Affected Resource(s)
Suspicions that it's github_branch_protection

### Terraform Configuration Files
Confidential, will provide if needed

### Debug Output
Will provide if needed

### Panic Output
N/A

### Expected Behavior
Plan takes 25 minutes

### Actual Behavior
Plan takes 65 minutes

### Steps to Reproduce
Both plan and apply take longer.

### Important Factoids
We have 3848 resources in our state, consisting of these types:
github_actions_organization_secret
github_branch_protection
github_membership
github_repository
github_repository_collaborator
github_repository_webhook
github_team
github_team_members
github_team_membership
github_team_repository

### References
#1020  may be the issue?

kfcampbell:
@dion-gionet do you have thoughts on potentially reducing the impact of the nested `for` loops introduced in #1020? I'm thinking it would be best to revert the PR for the incoming v5.5.0 release which would remove the immediate cause of the performance regression for the time being.

dion-gionet:
@kfcampbell mm, I'll have to look into it and come back to you. I'll make another PR when I have a solution

dion-gionet:
@kfcampbell I've opened https://github.com/integrations/terraform-provider-github/pull/1331 which should fix this whenever you have time to check it out!

kfcampbell:
@bootswithdefer I'm going to merge #1331 and cut a v5.7.0 release soon. If you get a chance, would you mind pulling in the latest version once it's out and retrying your plan?

bootswithdefer:
@kfcampbell 5.7.0 runs at the same speed as 5.3.0, thanks!

PR #1331, opened 2022-10-14
fix(branch-protection): Use more efficient function to fetch actors

dion-gionet:
Fixes https://github.com/integrations/terraform-provider-github/issues/1328

Brings back https://github.com/integrations/terraform-provider-github/pull/1020 but fixed the slowness that was caused by slow function calls (branchProtectionResourceData).

I've tested multiple scenarios from provider versions 5.4.0, 5.5.0 and my patch to make sure the performance regression is gone or at least very comparable to the speed of 5.5.0.
This was tested on a test org with over 160 repos (1 github_team_repository, 1 github_branch_default and 1 github_branch_protection resource per repo)

As you can see from the results below, the speed has greatly improved by using a trimmed down version of branchProtectionResourceData. You can see exactly what I changed on commit f308e306543a02675121c288896dc4c60a3c8605

# Test results
## Plan using node IDs
- Planning on 5.4.0

  oldpatch planids

- Planning with my patch

  newpatch planids

- Planning on 5.5.0

  5 5 0 planids

## Plan using user login name
- Planning on 5.4.0

  oldpatch planlogin

- Planning with my patch

  newpatch planlogin

## Apply change on github_branch_protection for all resources using node IDs and user login names
- Applying on 5.5.0 with node IDs

  5 5 0 applyids

- Applying with my patch with user login names. This is slower than applying with node IDs since it has to fetch the ID before applying

  newpatchapplylogin

- Applying with my patch with node IDs

  newpatchapplyids

cc @kfcampbell

### DIFF
--- a/github/resource_github_branch_protection.go
+++ b/github/resource_github_branch_protection.go
@@ -154,6 +154,27 @@ func resourceGithubBranchProtectionCreate(d *schema.ResourceData, meta interface
 	if err != nil {
 		return err
 	}
+
+	var reviewIds, pushIds, bypassIds []string
+	reviewIds, err = getActorIds(data.ReviewDismissalActorIDs, meta)
+	if err != nil {
+		return err
+	}
+
+	pushIds, err = getActorIds(data.PushActorIDs, meta)
+	if err != nil {
+		return err
+	}
+
+	bypassIds, err = getActorIds(data.BypassPullRequestActorIDs, meta)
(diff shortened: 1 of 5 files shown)


kfcampbell (approved):
Input on #1328 about the efficiency of these changes is appreciated! Thanks for the test data, @dion-gionet.

Issue #749, opened 2021-04-02
github_actions_organization_secret always applies the value, even if it's unchanged


greggilbert:
### Terraform Version
Terraform v0.13.5

### Affected Resource(s)
Please list the resources as a list, for example:
- github_actions_organization_secret

### Terraform Configuration Files
```hcl
resource "github_actions_organization_secret" "this" {
  for_each = toset(local.secrets)

  secret_name     = each.value
  visibility      = "selected"
  plaintext_value = "(placeholder)"   # we'll change these values outside of Terraform

  selected_repository_ids = [
    for repo in data.github_repository.repos : repo.repo_id
  ]

  lifecycle {
    ignore_changes = [
      plaintext_value
    ]
  }
}
```

### Expected Behavior
In this example, I'm giving providing a list of repositories that can access the secret. You can see I'm using `(placeholder)` as the plaintext value, with the idea that I would go into Github outside of TF and add in the real value once.

When I change the list of repositories, I expect that the real value I inputted would remain the same.

### Actual Behavior
However, it actually replaces the real value with the placeholder.

### Steps to Reproduce
1. Use the `github_actions_organization_secret` resource to create a value like `FOO` with a default value of `(placeholder)`
2. `terraform apply` the change/creation
3. Go into your Github organization and edit the secret to have a value like `Hello World`
4. In your Github Actions workflow, create a step like this:
    ```yml
      - name: DEBUG!
        env:
          MY_SECRET: ${{ secrets.FOO }}
        run: |
          echo ${MY_SECRET} | sed 's/./& /g' 
    ```
5. Push and/or trigger the Github Action
6. Note how the value that gets printed out is `(placeholder)` and not the expected `Hello World`

SanderKnape:
We're running into this issue as well. Any thoughts on how this could be fixed?

marcinwyszynski:
Well, this is probably linked to [my earlier fix](https://github.com/integrations/terraform-provider-github/pull/740). The whole idea of declaring resources with Terraform is for Terraform to make sure that the remote state matches the declared state. If you want Terraform to be blind to external changes, please explicitly set `ignore_changes` on this field.

jvanbrunschot:
I think that's the whole problem, that `ignore_changes` don't work.

marcinwyszynski:
@jvanbrunschot have you tried ignoring `updated_at`, too?

jvanbrunschot:
[...]

PR #1351, opened 2022-11-05
fix: destroy the resource on drift

yordis:
closes 749

Based on the request from @jcudit https://github.com/integrations/terraform-provider-github/issues/749#issuecomment-945144056 related to https://github.com/integrations/terraform-provider-github/issues/749#issuecomment-939018875


kfcampbell:
@yordis have you had a peek at our [CONTRIBUTING.md](https://github.com/integrations/terraform-provider-github/blob/main/CONTRIBUTING.md#quick-end-to-end-example) yet? There's some environment configuration you need but testing changes locally is very doable.

yordis:
@kfcampbell definitely helped a ton, my breakpoint is being hit now. Now trying to figure out how we suppose to setup the test case for it!

yordis:
I am trying to figure out what function would allow me to check if a new `id` was assigned to the resource ... any feedback welcome

yordis:
The thing is working! Just the testing environment is tricky! I am unsure how to create a scenario where the external resource change and whatnot ... 😢 

https://github.com/integrations/terraform-provider-github/assets/4237280/f7204e80-8413-4d3f-a1e2-7608e847115f

kfcampbell:
@yordis can you build a version of the provider locally and use it to manually test?

yordis:
@kfcampbell good idea!

yordis:
Hey I gave it a try to test it by building the local plugin, but I couldn't figure out how to make it work:

I did the following:

```txt
cat ~/.terraformrc         
provider_installation {
  dev_overrides {
    "integrations/github" = "/Users/ubi/Developer/github.com/integrations/terraform-provider-github/bin/"
  }

  direct {}
}
```

No luck, I tested manually as I showed before, so maybe try on your end if you do not mind.

kfcampbell:
@yordis can you try following the instructions [here](https://github.com/integrations/terraform-provider-github/blob/main/CONTRIBUTING.md#building-the-provider) to build the provider locally? You should be using a `dev.tfrc` file in order to specify the override.

yordis:
I tried that guideline multiple times, and it didn't work.


As shown in the video, I tested by running the code in the test environment. I would appreciate it if someone took the lead in helping to do an integration test out of the generated binary.

yordis:
keep

MagnusHJensen:
Hi @yordis 

We're running into the same problem just with regular action secrets.
I would like to offer my help in getting this fixed.

I can see your approach is to add a new `destroy_on_drift` property, but could we not just respect the already available `ignore_changes` array? And then check for either `plaintext_value` or `encrypted_value` appearing in that array, and not destroy it on drift, instead of introducing a new variable/config option? LMK what you think.

nbali:
this is funny, because we can't

calebworks:
Thanks for updating the documentation in this PR :+1:

PR #1120, opened 2022-04-20
Feature/include all branches

KensoDev:
When creating a repository from a template, you can pass in a boolean to
include all the branches in the template repo.

This is documented here: https://docs.github.com/en/repositories/creating-and-managing-repositories/creating-a-repository-from-a-template

In addition, the provider already supports that here: https://github.com/google/go-github/blob/3e8a7f0fbe0e4402f9ad5a80e0ab0a050786e0f8/github/repos.go#L441

Adding a variable under the template structure and then pass it in to
the provider function works.

This PR includes tests + docs for the new addition. It is backward compatible with the previous behavior

PR #1353, opened 2022-11-08
Include All Branches in template repositories

kfcampbell:
This PR picks up @KensoDev's changes from #1120 which I don't have permission to edit unfortunately.


kfcampbell:
@KensoDev thank you for contributing! Apologies for the delay.

PR #1357, opened 2022-11-13
Fix github.Timestamp to string

hiroygo:
`github.Timestamp` must be converted to `string` before set.


nickfloyd (approved):
Thanks for the contribution ❤️!

PR #1370, opened 2022-11-17
Updates to go 1.17 and bumps golangci-lint from 1.18.0 to 1.50.1

nickfloyd:
The dependency golangci-lint requires go v 1.17.  This PR bumps that version as well as updates the vulnerable package.

NOTE: I'm not sure if this is a breaking change, but since CI uses go 1.19 I think we're ok.  maintainers will need to update Go if they are currently @ 1.16

----

### Pull request checklist
- [ ] Tests for the changes have been added (for bug fixes / features) - N/A
- [ ] Docs have been reviewed and added / updated if needed (for bug fixes / features) - N/A, release notes will reflect this change
- [x] Added the appropriate label for the given change

- [x] No

 @kfcampbell please verify that this would not be considered a breaking change 

If `Yes`, what's the impact:  

* N/A

----


kfcampbell:
Confirming that switching to Go 1.17 will not be a breaking change for consumers of this plugin. Developers will need to have Go 1.17 at a minimum installed, which they should anyways. Go supports the last two minor versions, so 1.16 isn't receiving patches anymore. 

This lint failure is obnoxious. I can reproduce it locally using the currently-specified version of golangci/golangci-lint and the latest version. There's a thread about similar issues [here](https://github.com/golangci/golangci-lint-action/issues/448), [here](https://github.com/golangci/golangci-lint/issues/2788) and [here](https://github.com/golangci/golangci-lint/discussions/1920). Those threads seem to suggest that there's a missing dependency or perhaps a C library expected, though I don't see how that's possible given that the project builds just fine and the error is reproduceable with `CGO_ENABLED=0`. 

Another thing I tried is running `go clean -modcache` and then re-running `go mod tidy` and `go mod vendor` to see if we had a corrupted go.sum file somehow. Doing so does not show any file changes. I also thought the change might be related to Go 1.19 support, but that was added to golangci-lint in v1.48.0. 

I'll have to keep thinking about this one.

kfcampbell:
Aha! The error is a real one, but it's in the testing code and not the building code which is why builds work fine still. 

```
 sh$ make test
go test ./...
# github.com/hashicorp/terraform-plugin-sdk/internal/configs/configload
vendor/github.com/hashicorp/terraform-plugin-sdk/internal/configs/configload/loader_snapshot.go:52:30: cannot use fs (variable of type snapshotFS) as type afero.Fs in argument to configs.NewParser:
	snapshotFS does not implement afero.Fs (missing Chown method)
vendor/github.com/hashicorp/terraform-plugin-sdk/internal/configs/configload/loader_snapshot.go:57:32: cannot use fs (variable of type snapshotFS) as type afero.Fs in struct literal:
	snapshotFS does not implement afero.Fs (missing Chown method)
vendor/github.com/hashicorp/terraform-plugin-sdk/internal/configs/configload/loader_snapshot.go:200:18: cannot use snapshotFS{} (value of type snapshotFS) as type afero.Fs in variable declaration:
	snapshotFS does not implement afero.Fs (missing Chown method)
?   	github.com/integrations/terraform-provider-github/v5	[no test files]
FAIL	github.com/integrations/terraform-provider-github/v5/github [build failed]
FAIL
make: *** [GNUmakefile:27: test] Error 2
```

I'm still exploring how we might fix this.

kfcampbell:
Okay, so here's the deal: we're dependent on an old version of hashicorp/terraform-provider-sdk. We can and should update that in the near future. 

That version of hashicorp/terraform-provider-sdk is dependent on v1.4.1 of spf13/afero. v1.5.0 of spf13/afero [introduced the chown method](https://github.com/spf13/afero/compare/v1.4.1...v1.5.0) to an interface that the terraform-plugin-sdk depends on. 

We can pin to v1.4.1 of spf13/afero (released on 6 October 2020) by running `go get github.com/spf13/afero@v1.4.1` and subsequently `go mod tidy && go mod vendor`. In the near future, I think that's the best way forward until we can update hashicorp/terraform-provider-sdk to the next major version.

@nickfloyd would you like me to push the changes up to this branch or would you like to do it?

nickfloyd:
> @nickfloyd would you like me to push the changes up to this branch or would you like to do it?

Solid sleuthing!  Feel free to make the change given you're already in there.  Thanks for helping me get this one sorted!

kfcampbell (approved):

PR #1376, opened 2022-11-21
Add resource github_app_installation_repositories

david-bain:
This resource allows multiple repositories to be passed in; which greatly improves the performance of the resource compared to the single repository version when needing to control state of multiple app installations with multiple repos, required in larger organisations.

The optimisation occurs as only a single call to get the list of repos is required per installation per read, regardless of the number of repositories being added.

- Add resource_github_app_installation_repositories
- Add tests
- update docs

Resolves N/A

----

### Before the change?

* `resource_github_app_installation_repostiory` was a 1:1 installation:repo; however in reality, many repositories can be added to an installations allow-list. This meant that an app installation allowlist with say 5 repositories required 5 resources to be created/managed. Inside that resource, there was some generic calls that related more to the installation so if an app installation had multiple repositories allowed, those exact same API calls are repeated multiple times, e.g. `client.Apps.ListUserRepos()`.
* This made managing app allow-lists this way VERY slow

### After the change?

* Creates a new resource that allows multiple repositories to be defined against an installation in a 1:Many fashion. Following from the above, this means that the installation API calls are only called once per resource and as there are much less resources now due to the grouping of repositories, the terraform runs MUCH faster.
* In our organisation, it is 30x faster this way to plan

### Other information

* This resource has been running within our org on a fork for approximately 3 months and has been very successful.

----

### Pull request checklist

- [x] No

If `Yes`, what's the impact:  

* N/A

### Pull request type

 
 

- Bugfix: `Type: Bug`

----


website/docs/r/app_installation_repositories.html.markdown
```diff
@@ -0,0 +1,57 @@
...
+```
+
+## Argument Reference
+
+The following arguments are supported:
+
+* `installation_id`       - (Required) The GitHub app installation id.
+* `selected_repositories` - (Required) A list of repository names to install the app on.
+
+~> **Note**: Due to GitHub API limitations, deleting this resource will leave one repository with the app installed. Manually uninstall the app after deleting the resource.
```
kfcampbell, line 48: How is the remaining repository determined? Is it deterministic?

kfcampbell (commented):
Hi @david-bain, this is super cool and thanks for opening a PR to share it! I have a couple questions: 

- Do you mind adding a link to the new documentation in the `github.erb` file so the new doc is accessible? 
- What am I missing about running the integration tests? I've exported a valid `APP_INSTALLATION_ID` retrieved from my organization's app installation settings, but my tests fail with the following error: 

```
    testing.go:705: Step 0 error: errors during apply:
        
        Error: GET https://api.github.com/user/installations/131977/repositories?per_page=100: 404 Not Found []
```

Do you think this could be permission-related? 
- Would you mind putting a note at the top of the docs for the new resource and `resource_github_app_installation_repository` that distinguishes between the purpose of each resource and describes when you might want to use one over the other? 

Come to think of it, it may even be a good idea to deprecate `resource_github_app_installation_repository` soon in favor of the new resource.

david-bain:
Let me check into the above, I wrote this a while back on a fork so I need to refresh my memory regarding the remaining repo and tests.

github/resource_github_app_installation_repositories.go
```diff
@@ -0,0 +1,186 @@
...
+			if err != nil {
+				return err
+			}
+		}
+	}
+
+	// Remove repositories that existed on GitHub but not selectedRepositories
+	// There is a github limitation that means we can't remove the last repository from an installation.
+	// Therefore, we skip the first and delete the rest. The app will then need to be uninstalled via the GUI
+	// as there is no current API endpoint for [un]installation. Ensure there is at least one repository remaining.
```
david-bain, line 82: As per the comment here, I skip the first... which is base on the order of repositories that are returned by the the [list accessible repos API call](https://docs.github.com/en/rest/apps/installations?apiVersion=2022-11-28#list-repositories-accessible-to-the-user-access-token).
I guess I could sort it alphabetically so that it is always the first alphabetical by name that remains?

website/docs/r/app_installation_repositories.html.markdown
david-bain: See comment below, I'll update the docs to reflect

david-bain: @kfcampbell 
Re running the tests: I just pulled my branch and made sure to set `GITHUB_TOKEN`, `GITHUB_ORGANISATION`, and `APP_INSTALLATION_ID` and it ran fine.
I guess recheck your `APP_INSTALLATION_ID` is the correct one for your org specified in `GITHUB_ORGANISATION`? I was able to recreate your error when I deliberately put in an incorrect installation ID...

david-bain:
Docs updated.
I believe I have addressed all your points. Please let me know if there is anything else.

github/resource_github_app_installation_repositories.go
kfcampbell: This seems like a reasonable call. It's definitely an instance where GitHub's behavior doesn't fit a tidy CRUD model.

website/docs/r/app_installation_repositories.html.markdown
kfcampbell: Oh my gosh you're totally right! Thanks for the help troubleshooting the testing.

kfcampbell (approved):
Thanks for making changes on this! I'll get this merged now and released in the near future.

AlfredoBall:
* **Update**: I wrote this comment in ignorance of the true intention of the permissions model, Git Hub Apps are **not** designed to manage what they didn't create or what they are given by a user / PAT. I won't delete it though because I think it's still informative and even more so that I was **wrong**.

I had blinders on thinking about fixing what isn't broken.

I wasn't thinking as much about the permissions as I was the automation, which I was also wrong about.

----------------------------------------------------------------------------------------

# Architectural Post-Mortem: Structural Limitations of `github_app_installation_repositories`

> **Status:** **Unresolvable (Upstream API Constraint)**

---

## Summary

Managing GitHub App repository assignments through the `github_app_installation_repositories` resource using a machine-to-machine authentication model (`app_auth`, GitHub App installation tokens, or OpenID Connect) is not currently possible.

This limitation is **not** a defect in Terraform or the GitHub provider. Instead, it stems from how the GitHub REST API exposes repository installation management. The API endpoints required to explicitly add or remove repositories from an installation are only available through user-scoped routes that require user-delegated credentials.

As a result, a fully automated workflow using GitHub App authentication alone cannot perform the complete CRUD lifecycle for this resource.

---

## API Endpoint Constraints

The `github_app_installation_repositories` resource synchronizes repository assignments by calling GitHub REST API endpoints whose availability depends on the authentication context.

### Repository Assignment Endpoints

Explicitly assigning or removing repositories requires the following endpoints:

```text
PUT    /user/installations/{installation_id}/repositories/{repository_id}
DELETE /user/installations/{installation_id}/repositories/{repository_id}
```

These endpoints are intentionally exposed only under the `/user/` namespace.

According to GitHub's REST API documentation, they do **not** accept:

- GitHub App installation access tokens
- GitHub App user access tokens
- Fine-grained personal access tokens

Instead, they require a user-delegated authentication context (typically a classic Personal Access Token with the necessary repository permissions).

Consequently, requests made through `app_auth` fail with:

```text
403 Resource not accessible by integration
```

### Missing Organization-Scoped Mutation Endpoint

GitHub provides organization-scoped endpoints for retrieving installation information, but **no equivalent endpoint exists** for modifying repository assignments.

For example, the following route does **not** exist:

```text
PUT /orgs/{org}/installations/{installation_id}/repositories/{repository_id}
```

Because no organization-scoped mutation API exists, all repository assignment operations are forced through the user-scoped `/user/` endpoints, preventing automation using GitHub App installation credentials alone.

---

## Why `legacy_client = false` Does Not Help

Beginning with version **6.13.0**, the GitHub provider introduced the new client implementation via:

```hcl
legacy_client = false
```

This change modernized the provider's transport layer, including fixes for several issues such as stale ETag caching (HTTP 304 behavior) that could previously lead to state inconsistencies.

However, this architectural improvement does **not** affect authentication or API routing.

Regardless of which client implementation is used, the provider must still call the same `/user/installations/...` endpoints, where GitHub performs server-side authorization checks. Those requests are rejected before any provider-side logic can complete.

In other words, the limitation exists entirely on the GitHub API side rather than within the provider implementation.

---

## Relational Contradictions in Platform Scoping

The hard block on explicit repository mapping creates an operational contradiction where the GitHub App's app_auth can create or destroy repositories, where the repository is automatically assigned to a GitHub App's repository list upon creation with an GitHub App auth context, yet it cannot toggle repository visibility through the API. This creates a conflict where the token can delete assets, but not change its own scope without deleting the underlying resource.

```mermaid
graph TD
    classDef allowed stroke:#00cc66,stroke-width:2px,fill:#e6ffed;
    classDef blocked stroke:#ff3333,stroke-width:2px,fill:#ffeef0;

    Key["GitHub App Private Key (.pem)"] -->|Authentication Context| AppAuth["app_auth"]
    
    AppAuth -->|POST /orgs/org/repos| Create["CREATE A NEW REPOSITORY<br><br>ALLOWED<br>Backend hooks automatically<br>APPEND it to App's scope."]
    
    AppAuth -->|DELETE /repos/owner/repo| Destroy["DELETE ENTIRE REPOSITORY<br><br>ALLOWED<br>App can permanently erase<br>corporate code and history<br>from existence."]
    
    AppAuth -->|DELETE /user/installations/...| Toggle["SOFTLY REMOVE FROM APP ACCESS LIST<br><br>HARD BLOCKED<br>API throws 403 Forbidden.<br>App token is forbidden from<br>simply hiding the repository."]

    class Create allowed;
    class Destroy allowed;
    class Toggle blocked;
```

---

## Underlying Data Model

Conceptually, GitHub stores the relationship between an installation and its accessible repositories as a separate mapping rather than as an attribute of either object.

A repository contains metadata about the repository itself, while the installation maintains the list of repositories it can access.

Because this relationship belongs to the installation, modifying it requires installation-specific mutation endpoints. Since those endpoints are only exposed through the user-scoped API, repository assignments cannot currently be managed using installation tokens alone.

---

## An Interesting Asymmetry

This API design creates an operational asymmetry when compared with the standard `github_repository` lifecycle.

### Repository Creation

When a GitHub App creates a repository through:

```text
POST /orgs/{org}/repos
```

GitHub automatically associates the new repository with the creating installation.

No additional API call is required.

### Repository Deletion

The same GitHub App installation token can later delete that repository through:

```text
DELETE /repos/{owner}/{repo}
```

Removing the repository also removes its installation association.

### Repository Assignment

However, attempting to explicitly add or remove repository access while leaving the repository itself intact requires the user-scoped installation endpoints discussed above.

Thus, a GitHub App installation token can:

- Create a repository.
- Automatically gain access to it.
- Delete the repository.

…but **cannot explicitly change repository assignment for an existing repository.**

---

## Relevant Ecosystem References

The following discussions document various aspects of this API limitation.

### Terraform GitHub Provider

#### Issue #2103

*`github_app_installation_repository` does not work with fine-grained personal access tokens.*

#### Issue #2431

*Cross-organization repository lookup behavior when using user-scoped installation endpoints.*

### GitHub Community

#### Discussion [#24549](https://github.com/orgs/community/discussions/24549)

*Working across organizations with GitHub Apps.*

### Provider Modernization

#### PR #3448 — New GitHub Client

*Introduced the `legacy_client = false` architecture. This modernized the provider's transport layer but does not change the underlying GitHub REST API constraints affecting `github_app_installation_repositories`.*

---

## Conclusion

The current GitHub REST API does **not** expose an organization-scoped mechanism for modifying GitHub App repository assignments.

Because the only available mutation endpoints require a user-delegated authentication context, the `github_app_installation_repositories` resource cannot support a complete CRUD lifecycle using a pure machine-to-machine authentication model (`app_auth`).

Unless GitHub introduces new installation-scoped mutation endpoints or expands support for GitHub App installation tokens on the existing endpoints, this limitation cannot be resolved within the Terraform provider itself.

For environments standardized on GitHub App authentication, repository assignment should therefore be treated as an external bootstrap or administrative operation rather than infrastructure that can be fully managed through Terraform.

PR #1399, opened 2022-12-01
Fix: Ignore repository file author and email changes if set implicitly

### LINKED ISSUE #1345: Perpetual updates in github_repository_file
When upgrading from provider version 5.3.0 to 5.6.0, Terraform shows diff for all `github_repository_file` resources to remove `commit_author` and  `commit_email` from resource. After successful apply, next plan operation shows the same diff again. Reverting back to 5.3.0 solves the problem. Issue can be reproduces on 5.5.0 as well.

Tried removing resource from state and importing it back again - no luck.
Provider is initialised using a PAT token of separate service user in organisation.

### Terraform Version
Terraform v1.3.3
on darwin_arm64

### Affected Resource(s)
- github_repository_file

```hcl
resource "github_repository_file" "kustomize" {
  repository = local.fluxcd_config.github_main_repository_name
  file       = data.flux_sync.main.kustomize_path
  content    = local.kustomize_file
  branch     = local.fluxcd_config.github_main_repository_branch
}
```

### Expected Behavior
Changes to resource should be successfully applied.

### Actual Behavior
Changes are applied successfully and new empty commit is created. Next plan operation shows the diff again.

```hcl
  # module.eks.github_repository_file.kustomize will be updated in-place
  ~ resource "github_repository_file" "kustomize" {
      - commit_author       = "<REDACTED>" -> null
      - commit_email        = "<REDACTED>" -> null
        id                  = "<REDACTED>"
        # (8 unchanged attributes hidden)
    }
```

### References
This behaviour is most likely introduced in https://github.com/integrations/terraform-provider-github/pull/1179.

pawnu:
Resolves #1345

----

### Before the change?
When using a PAT without author and email information to manage github repository file, the author and email information is saved to state as null vs upstream as PAT author and email.

This leads to upstream vs local comparison and change due to differing values.

### After the change?
When using a PAT without author and email information to manage github repository file, the author and email information is saved to state as null vs upstream as PAT author and email.

However, it is not read for comparison by the provider if no author and email information is provided. In Github App's case upstream values would be github web-flow user details and on PAT's case it would be PAT owner details, in both cases both of these would be static as long as no author/email information is provided explicitly, so there's no need to register them as a change.

----

### Pull request checklist

- [x] No

If `Yes`, what's the impact:  

* N/A


kfcampbell (approved):
Thanks for submitting a fix here, @pawnu.

Issue #1307, opened 2022-09-28
Cannot create github_branch_protection_v3 after update to v5.3.0


jtsaito:
### Terraform Version
**v5.3.0**

### Affected Resource(s)

- github_branch_protection_v3
- github_branch_default
- github_branch_default

The issue occures when creating a new `github_branch_protection_v3` resource with attributes referenced from `github_branch_default` and `github_branch_protection_v3` resources.

### Debug Output

```
github_branch_protection_v3.<MY_REPO_NAME_REDACTED>: Creating...
╷
│ Error: PUT https://api.github.com/repos/<MY_ORG_REDACTED>/<MY_REPO_NAME_REDACTED>/branches/main/protection: 422 Invalid request.
│
│ No subschema in "anyOf" matched.
│ For 'properties/checks', nil is not an array.
│ Not all subschemas of "allOf" matched.
│ For 'anyOf/1', {"strict"=>true, "contexts"=>["build"], "checks"=>nil} is not a null. []
```

### Expected Behavior
The branch protection v3 resource should be created.

### Actual Behavior
Terraform errors out with message above.

#### Step 1

Create a repo with terraform apply.

```
resource "github_repository" "test-repo" {
  name        = "test-repo"
  description = "This repo is a test"

  has_downloads = true
  has_issues    = false
  has_projects  = false
  has_wiki      = false

  visibility = "private"

  auto_init = true

  vulnerability_alerts = false
}

resource "github_branch_default" "test-repo" {
  repository = github_repository.test-repo.name
  branch     = "main"
}
```

#### Step 2

Terraform apply the following change:

```
resource "github_branch_protection_v3" "test-repo" {
  repository     = github_repository.test-repo.name
  branch         = github_branch_default.test-repo.branch
  enforce_admins = true

  required_status_checks {
    strict   = true
    contexts = ["build"]
  }

  required_pull_request_reviews {
    dismiss_stale_reviews      = true
    require_code_owner_reviews = true
  }
}
```

### Important Factoids
This works up to GitHub provider version v5.2.0 without errors and errors in v5.3.0.

This seems to be linked to the go client update in the [release diff](https://github.com/integrations/terraform-provider-github/compare/v5.2.0...v5.3.0)  where the API Client is updated as follows.

https://github.com/integrations/terraform-provider-github/blob/6773d80600f761bf50b695f475c2d457d1151ee5/vendor/github.com/google/go-github/v47/github/repos.go#L908

From 

```
[...]

Issue #1212, opened 2022-07-04
[Feature Request] Add support for branch protection required_status_check 'checks' object


david-bain:
Github has updated protected branches to include an `app_id` along with the status_check `context` as a new object named `checks`. This has depreciated `context` at the higher level and is required to be able to set the source of a `required status check`. 
Is it possible to get an update to the branch_protection resource and utils to add support for this in the graphQL call?

I believe we could make the app_id optional as it is currently optional on the [REST API call](https://docs.github.com/en/rest/branches/branch-protection#update-status-check-protection), and therefore could come in as a minor update.

ardakuyumcu:
I took a stab at implementing this locally, but there are some decisions to be made around how to implement this new field in the state file. Below are my notes:

- The GitHub API does not allow you to pass both the `contexts` field and the `checks` field together in an update API call, you have to pass only one of the two.
```
Error: PUT https://api.github.com/repos/ardakuyumcuorg/tf-acc-test-6emb0/branches/main/protection: 422 Invalid request.

        No subschema in "anyOf" matched.
        More than one subschema in "oneOf" matched.
        Not all subschemas of "allOf" matched.
        For 'anyOf/1', {"strict"=>true, "contexts"=>["github/foo", "github/bar"], "checks"=>[{"context"=>"github/foo"}, {"context"=>"github/bar", "app_id"=>-1}]} is not a null. []
```

- Even if you set only the checks field in the Terraform resource definition, when reading back the resource from GitHub both the `contexts` field and the `checks` field are populated, as they are the same from GitHub's perspective. This causes issues with the terraform plan output when you try to apply the same resource definition twice:
```
➜  ~ curl -L -H "Authorization: token $GITHUB_TOKEN" -H "Accept: application/vnd.github.v3+json" https://api.github.com/repos/ardakuyumcuorg/tf-acc-test-oqozw/branches/main/protection
...
  "required_status_checks": {
    "url": "https://api.github.com/repos/ardakuyumcuorg/tf-acc-test-oqozw/branches/main/protection/required_status_checks",
    "strict": true,
    "contexts": [
      "github/foo"
    ],
    "contexts_url": "https://api.github.com/repos/ardakuyumcuorg/tf-acc-test-oqozw/branches/main/protection/required_status_checks/contexts",
    "checks": [
      {
        "context": "github/foo",
        "app_id": null
      }
    ]
  },
...
```

[...]

Issue #1147, opened 2022-05-12
unable to update branch_protections for github repos


balusarakesh:
Hi there,

We were trying to enable `enforce_admins = true` and we see a TF plan like the following but when we apply we get an error. As you can see below there are no changes to `required_status_checks` but we are not able to apply it.
```
  # module.repositories.github_branch_protection_v3.managed["REPO_NAME:BRANCH_NAME"] will be updated in-place
  ~ resource "github_branch_protection_v3" "managed" {
      ~ enforce_admins                  = false -> true
        id                              = "REPO_NAME:BRANCH_NAME"
        # (5 unchanged attributes hidden)


        # (2 unchanged blocks hidden)
    }
```

### Terraform Version
 - `1.0.9`
 - terraform-provider-github: `v4.24.1`

### Affected Resource(s)
Please list the resources as a list, for example:
- github_branch_protection_v3

If this issue appears to affect multiple resources, it may be an issue with Terraform's core, so please mention this.

### Terraform Configuration Files
```
resource "github_branch_protection_v3" "example" {
  repository     = github_repository.example.name
  branch         = "main"
  enforce_admins = true

  required_status_checks {
    strict   = true
    contexts = []
  }

}
```

### Debug Output
unfortunately I'm not allowed to share full debug output due to company policy, here's some relevant debug logs:
```
2022-05-12T09:24:17.198-0700 [DEBUG] provider.terraform-provider-github_v4.24.1: {
2022-05-12T09:24:17.198-0700 [DEBUG] provider.terraform-provider-github_v4.24.1:  "message": "Invalid request.\n\nNo subschema in \"anyOf\" matched.\nNo subschema in \"oneOf\" matched.\nNot all subschemas of \"allOf\" matched.\nFor 'anyOf/1', {\"strict\"=>true} is not a null.",
2022-05-12T09:24:17.198-0700 [DEBUG] provider.terraform-provider-github_v4.24.1:  "documentation_url": "https://docs.github.com/rest/reference/repos#update-branch-protection"
2022-05-12T09:24:17.198-0700 [DEBUG] provider.terraform-provider-github_v4.24.1: }

│ Error: PUT https://api.github.com/repos/ORG_NAME/REPO_NAME/branches/BRANCH_NAME/protection: 422 Invalid request.
│
│ No subschema in "anyOf" matched.
│ No subschema in "oneOf" matched.
│ Not all subschemas of "allOf" matched.
[...]

Issue #1414, opened 2022-12-06
[MAINT]: Document (and automate?) required resources to run integration tests


kfcampbell:
### Describe the need

Currently, our unit tests expect a variety of prerequisites to avoid skipping and run successfully. These include:

- An organization and a user account
- A template repository inside the organization
- A GitHub App installed at the organization level
- The GitHub App installed on at least one repo and the installation ID exported to a particular environment variable

The above list is non-exhaustive. When this issue is closed, we should at a minimum have documented all the criteria required to run unit tests. We should also have evaluated whether it's worth automating the creation of the prerequisites. 

### SDK Version

all versions

### API Version

N/A

### Relevant log output

```shell
N/A
```

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

wwsean08:
I'll do my best to work on untangling this over the next couple of weeks unless someone is actively working on it.  My plan is to create a `TESTING.md` file which can be linked to in `CONTRIBUTING.md`.  Long term that document can possibly include testing procedures, standards, and things of that nature.

kfcampbell:
<3 <3 <3 that would be wonderful @wwsean08! Your plan sounds excellent.

wwsean08:
Quick question, is there any way to determine if an account is an enterprise account?   `TestAccGithubOrganizationIpAllowListDataSource/queries_without_error/with_an_organization_account` this test manages an IP allow list which is limited to enterprise accounts?  If it is possible it might be worth adding special "enterprise" type tests.

kfcampbell:
> is there any way to determine if an account is an enterprise account? 

I'm not aware of anything, unfortunately. Perhaps it would be possible to figure this out through the [Get the authenticated user](https://docs.github.com/en/rest/users/users?apiVersion=2022-11-28) endpoint, or the [Return all of the user's organization memberships](https://docs.github.com/en/rest/guides/discovering-resources-for-a-user?apiVersion=2022-11-28#return-all-of-the-users-organization-memberships) workflow. 

> If it is possible it might be worth adding special "enterprise" type tests.

I agree; I think this would be super worthwhile!

wwsean08:
Turns out there actually is an `enterprise` constant and I just missed it, so I'll look at fixing the tests that are enterprise dependent.  I wasn't using my terminology correctly as I meant an org is enterprise, but it seems like we've got this resolved at this point.

stevehipwell:
[...]

PR #1415, opened 2022-12-07
feature/985 branch protection `checks`

### LINKED ISSUE #985: Add support for configuring checks to branch protection
### Affected Resource(s)
- github_branch_protection
- github_branch_protection_v3

### Description

The [GitHub API](https://docs.github.com/en/rest/reference/repos#update-branch-protection) supports setting up `checks` as well as `contexts` for branch protection. `checks` provide a few more configuration options (like enforcing that a check is set by a specific app)

image

Could you add support to the `github_branch_protection` and `github_branch_protection_v3` resources to configure `checks`?  Currently they only support `contexts`

Thanks

TheQueenIsDead:
Resolves: 
 - #985
 - #1147 
 - #1307
 - #1212

----

### Before the change?

* Checks were not able to be set 
* Setting required status checks on branch protection resources was broken

### After the change?

* Checks are able to be set via a ":" separated string
* Contexts are now passed to the API as checks and users can provision branch protections again

### Other information

This should render the ability to define checks in configuration such as this:
```tf
resource "github_branch_protection_v3" "bp" {

  repository = "terraform-template-module"
  branch     = "master"

  required_status_checks  {
    strict = true

    contexts = [
      "context1",
      "context2",
      "context3",
      "context4",
    ]

    checks = [
      "negative_one:-1",
      "undefined_comma:",
      "undefined",
      "zero_check:0",
      "app_id_context:1234567",
    ]
  }
}
```

----

### Pull request checklist
- [x] Added the appropriate label for the given change

- [x] No

If `Yes`, what's the impact:  

* N/A

### Pull request type

 
 

- Bugfix: `Type: Bug`

----


TheQueenIsDead:
This works in the sense that is creates and deletes branch protections :+1: 

However, there is something interesting happening in the diff, like so:
```
 # github_branch_protection_v3.bp will be updated in-place
  ~ resource "github_branch_protection_v3" "bp" {
        id                              = "terraform-template-module:master"

      ~ required_status_checks {
          ~ contexts       = [
              - "example",
              - "newcheck",
              - "noappidcontext",
            ]

          ~ check {
              ~ app_id  = -1 -> 1
            }
          - check {
              - app_id  = -1 -> null
              - context = "newcheck" -> null
            }

            # (1 unchanged block hidden)
        }
    }
```

This seems to suggest that despite not including a `contexts` array, they will become updated via the inclusion of `check` blocks.

Not too sure how to handle this currently, but will see what I can come up with.

github/resource_github_branch_protection_v3_utils.go
```diff
@@ -46,10 +46,24 @@ func flattenAndSetRequiredStatusChecks(d *schema.ResourceData, protection *githu
 			contexts = append(contexts, c)
 		}
 
+		var checks []interface{}
+		for _, chk := range rsc.Checks {
+			chkMap := make(map[string]interface{})
+			chkMap["context"] = chk.Context
+			chkMap["app_id"] = func() int {
+				if chk.AppID != nil {
```
TheQueenIsDead, line 54: If this is the ID from upstream via the SDK, it's likely to already be defaulted to a non-nil making this check redundant.

github/resource_github_branch_protection_v3.go
```diff
@@ -53,12 +53,30 @@ func resourceGithubBranchProtectionV3() *schema.Resource {
...
+							Elem: &schema.Resource{
+								Schema: map[string]*schema.Schema{
+									"context": {
+										Type:     schema.TypeString,
+										Required: true,
+									},
+									"app_id": {
+										Type:     schema.TypeInt,
+										Optional: true,
+										Default:  -1,
```
TheQueenIsDead, line 75: Upon reflection this should be nullable.
The API states the field can be _ommitted_ to allow the most recent app, **or** set to _-1_ to allow any app:
https://docs.github.com/en/rest/branches/branch-protection?apiVersion=2022-11-28#update-branch-protection
https://pkg.go.dev/github.com/google/go-github/v48@v48.1.0/github#RequiredStatusCheck

cfisher281:
Appreciate you doing this work @TheQueenIsDead.

TheQueenIsDead:
@kfcampbell , small question from me. I've added the deprecated flag to the contexts attribute, and the process for carrying out this deprecation is usually to release a minor with the flag, and then remove the code in the next major release version. This is documented by Hashicorp here:
https://developer.hashicorp.com/terraform/plugin/sdkv2/best-practices/deprecations

It makes a lot of sense, but given the change to the Github go client, I cannot implement this functionality nicely without removing part of the code that flattens the context array when it reads from the API. Without that removal, subsequent applies will show the contexts array being blitzed on each plan, despite it not being defined in Terraform:
```console
 ~ contexts       = [
              - "context_app_id_allow_all",
              - "context_no_app_id_most_recent_check",
              - "context_with_app_id_defined",
            ]
```

The question is, would you accept the removal of `contexts` in this PR (Jarring as on release it would give little time to warn users of deprecation), or is there perhaps another route to releasing this that would be preferred?

As it stands, it is not possible to only send `contexts` with the SDK :-( So these are interesting circumstances.

kfcampbell:
Hmm...none of those options are very desirable. 

I'm looking at the [API reference](https://docs.github.com/en/rest/branches/branch-protection?apiVersion=2022-11-28#update-branch-protection) and I see the following under `required_status_checks`: 

image

> Deprecated: The list of status checks to require in order to merge into this branch. If any of these checks have recently been set by a particular GitHub App, they will be required to come from that app in future for the branch to merge. Use checks instead of contexts for more fine-grained control.

Perhaps some of that behavior is due to the API changing to deprecate contexts as well. 

> The question is, would you accept the removal of contexts in this PR (Jarring as on release it would give little time to warn users of deprecation), or is there perhaps another route to releasing this that would be preferred?

For a breaking change like that, we would end up cutting a v6 version, which ideally we should not do that in this PR without a deprecation notice. 

Do you mind explaining a little bit more about this description? 

> given the change to the Github go client, I cannot implement this functionality nicely without removing part of the code that flattens the context array when it reads from the API. Without that removal, subsequent applies will show the contexts array being blitzed on each plan, despite it not being defined in Terraform:

Is there a way we can handle contexts when they're present in Terraform by warning of deprecation and then formatting them into the new `checks` array of objects?

TheQueenIsDead:
> Perhaps some of that behavior is due to the API changing to deprecate contexts as well.

I wouldn't be suprised! As identified in a previous issue, the SDK "fully deprecated" contexts as part of this PR, while I believe the API itself still supports it momentarily:
https://github.com/google/go-github/issues/2467#issuecomment-1250070578

> For a breaking change like that, we would end up cutting a v6 version, which ideally we should not do that in this PR without a deprecation notice.

I appreciate that, and am happy to help go down that route where possible 👍 However, it does feel a little ceremonious given `contexts` on the SDK are broken due to the above PR.

> Do you mind explaining a little bit more about this description?

Certainly! What I am trying to highlight is that the Github API seems to return a list of `contexts` even when only `checks` are provisioned. When this gets converted to TF schema, it assumes that on next apply the `contexts` array will be deleted, because it is not defined in our file.

I think you're right that we could simply convert contexts to checks (With a default of allow all apps, ie -1). We could possibly keep the schema and not try and map the response to a `contexts` array in that instance.
That feels a little dirty, but it could be a good midpoint to allowing users to keep doing what they're doing, and give them a little notice about moving to checks.

I'll have a play around :-)

TheQueenIsDead:
Hmm, super interesting. I have made a change to flatten and expand all contexts into checks, using the default -1 app_id.

On initial apply, everything looks ok. When a second apply is ran, the change output looks really messy. On the third apply in a row (With no changes), Terraform detects no changes. 

I would have expected it to detect no changes the first time around, but I am unsure why it creates a plan like this. Full example output can be seen in this gist:
https://gist.github.com/TheQueenIsDead/ea838b56322f4aa998cd1b84f1129cbe

TheQueenIsDead:
I have tried sorting the `contexts` and `checks` array alphabetically, but this did not resolve the weird diff. I feel that this may be due to the way Terraform does the diff (Which I am not too familiar with under the hood), but I am guessing it reconciles diffs based on it's own schema, where it has one schema based on current state, and another based on what the file definition says.

I also tried experimenting with `ConfigMode: schema.SchemaConfigModeAttr`, which did not seem to help.

I'm wondering if the 'advanced data structure' as an array is causing issues, I imagine if we used a map of strings `([]string{"context_name:-1"})` then we would get a better diff, though this is inflexible for representing new attributes that may be added to the API in the future.

TheQueenIsDead:
@kfcampbell , I have changed the approach from using resource blocks in the schema to instead using a list of `:` delimited strings. This yielded a much better diff, and has the added bonus that an end user can simply rename the `contexts` array and have a working config.

Will update the docs soon, not terribly sure about how to approach testing.

kfcampbell:
> I have changed the approach from using resource blocks in the schema to instead using a list of : delimited strings. This yielded a much better diff, and has the added bonus that an end user can simply rename the contexts array and have a working config.

:+1:, thank you for the persistence on this one. 

> Will update the docs soon

Docs changes look good. As a side note, since the current behavior is broken as you mentioned, that breaking change feels okay to release under v5.x.x.

> not terribly sure about how to approach testing.

Our testing in the project isn't in a very healthy state at the moment. While #1414 and other related issues are still open, I've been running integration tests manually before merging PRs :see_no_evil: :see_no_evil: :see_no_evil: 

You can find some documentation and an example test pattern to follow [here](https://github.com/integrations/terraform-provider-github/blob/main/CONTRIBUTING.md#local-development-iteration) in our CONTRIBUTING.md file. Please let me know if you have more specific questions!

TheQueenIsDead:
No worries! Thanks for bearing with my first time developing a provider.

> since the current behavior is broken as you mentioned, that breaking change feels okay to release under v5.x.x.

If we go with the list approach then we should still be able to use contexts and checks, meaning the change will not be breaking :-) 

Alrighty, so I created a new org and setup the tests on my forked repo's main. All of the tests pass except for the `required_status_check` test which fails with the expected schema violations caused by the [v48.1.0](https://github.com/TheQueenIsDead/terraform-provider-github/blob/0b3d26d9a28509eec7b4f1df37f0a57ca798260a/go.mod#L8) version of `go-github`
image


However, this is the `main` branch of **MY** fork, which is 13 commits behind the official repositories main:
https://github.com/TheQueenIsDead/terraform-provider-github/tree/main

When I run the tests on this PR branch (which is up to date with the latest version of the `integrations` organisation main branch) we're not as successful :see_no_evil: Lots each test for branch protections v3 fails on:
```
Error: json: cannot unmarshal object into Go struct field Protection.lock_branch of type bool
``` 

I wonder if this breakage was introduced by the recent update to the go-github provider version with support for `lock_branch` as introduced by this PR https://github.com/google/go-github/pull/2577 in [v48.2.0](https://github.com/google/go-github/releases/tag/v48.2.0)

I see that this was updated recently 
https://github.com/integrations/terraform-provider-github/pull/1425

@kfcampbell is there any chance that this has broken the integration tests on main for the `GithubBranchProtectionV3` tests?

TheQueenIsDead:
I have just synced my fork with master and I can confirm that the tests are not passing on main anymore. @kfcampbell does this require a issue to be raised? It slightly blocks me on testing unless I revert the version back

fproulx-boostsecurity:
Yay

gilacost:
Any updates on this? I was looking around and trying to contribute but I found out that this was already in place. It looks great! What does else do we need to do to get it merged? 
Please add also this [issue](https://github.com/integrations/terraform-provider-github/issues/1451) to the closing block in the PR description. 🤗

kfcampbell:
@TheQueenIsDead thank you for all your effort and clear communication! I really appreciate it. 

> is there any chance that this has broken the integration tests on main for the GithubBranchProtectionV3 tests?

Definitely, and there's also a chance that they haven't been exactly healthy for a long time. 

> does this require a issue to be raised?

Ideally, yes. We have a significant amount of testing debt to pay back on this project. I've opened #1414 to begin this effort, but it would be great to have an issue for this specific testing suite. 

Given the above and the substantial number of issues ready to be closed, I think it would be reasonable to merge this PR before full testing support is achieved.

@TheQueenIsDead do you believe this is ready for final review/merging?

github/resource_github_branch_protection_v3_utils.go
```diff
@@ -40,16 +40,34 @@ func buildProtectionRequest(d *schema.ResourceData) (*github.ProtectionRequest,
...
+			// Parse into contexts
 			contexts = append(contexts, c)
+			checks = append(contexts, c)
+		}
+
+		// Flatten checks
+		for _, chk := range rsc.Checks {
+			// Parse into contexts
+			contexts = append(contexts, chk.Context)
+			checks = append(contexts, fmt.Sprintf("%s:%d", chk.Context, chk.AppID)) //  buildTwoPartID(chk.Context, strconv.Itoa(int(*chk.AppID))))
```
kfcampbell, line 62: Can this end-of-line comment be removed?

github/resource_github_branch_protection_v3_utils.go
```diff
@@ -191,8 +209,56 @@ func expandRequiredStatusChecks(d *schema.ResourceData) (*github.RequiredStatusC
 			m := v.(map[string]interface{})
 			rsc.Strict = m["strict"].(bool)
 
+			// Initialise empty literal to ensure an empty array is passed mitigating schema errors like so:
+			// For 'anyOf/1', {"strict"=>true, "checks"=>nil} is not a null. []
+			rscChecks := []*github.RequiredStatusCheck{}
+
+			// TODO: Remove once contexts is deprecated
+			// Iterate and parse contexts into checks using -1 as default to allow checks from all apps.
```
kfcampbell, line 217: Is this -1 value documented on the GitHub API? Do you know where it comes from?

kfcampbell (commented):
Hi! I'm swinging by to check on this PR again as I'm really looking forward to having it in. @TheQueenIsDead do you mind giving these questions a look when you get a chance?

github/resource_github_branch_protection_v3_utils.go
TheQueenIsDead: Yes it is :-) 
https://docs.github.com/en/rest/branches/branch-protection?apiVersion=2022-11-28#update-branch-protection
image

TheQueenIsDead:
Ok, so I have created a new bug report for the unmarshalling error that is causing the tests to fail for this branch #1476.

> Given the above and the substantial number of issues ready to be closed, I think it would be reasonable to merge this PR before full testing support is achieved. @TheQueenIsDead do you believe this is ready for final review/merging?

In a perfect world I think the recent go-github provider breakage would be reverted so that the tests can pass. As mentioned in the issue, it seems to have broken unmarshalling branch protection objects, so even if this PR is merged and released, users still wont be able to benefit from it!

I'm happy with the changes proposed here, but I am also the author, hence the most biased party ;-) The lack of tests are the only thing making me wary however.
The diff was a little confusing when running an apply to get the contexts to be written as checks, but subsequent runs yeilded a decent diff, which I am happy about given my limited experience with programming and the Terraform SDK

kfcampbell:
> I have created a new bug report for the unmarshalling error that is causing the tests to fail for this branch https://github.com/integrations/terraform-provider-github/issues/1476.

This should be fixed. I believe the problem was due to [this upstream issue](https://github.com/google/go-github/issues/2612), fixed in [this google/go-github PR](https://github.com/google/go-github/pull/2606), released [here](https://github.com/google/go-github/releases/tag/v49.0.0), and added to the project as part of #1473. 

On the main branch, tests for this resource are failing for me as well :sob: but not for unmarshaling reasons. 

> it seems to have broken unmarshalling branch protection objects, so even if this PR is merged and released, users still wont be able to benefit from it!

I don't believe this is the case anymore! :tada: 

> The lack of tests are the only thing making me wary

Absolutely, I agree. However, given the resource's impaired state currently and the number of issues/engagement about it, it may be appropriate to merge this PR earlier as a mitigation and add/modify earlier content afterward. Issues saying "confusing diff" are generally better than issues saying "doesn't work" IMO.

ilmax:
I also bumped into this one, looking forward to seeing this merged :)

TheQueenIsDead:
Apologies for the delay, I've been struggling to find time to commit to other projects.

> On the main branch, tests for this resource are failing for me as well sob but not for marshalling reasons.

I'll have a quick look at the tests to see if I can identify the issue and resolve it here, otherwise I'm all for merging this to give a path forwards, with room to refine later on. Pragmatism 😊

kfcampbell:
@TheQueenIsDead thank you for all the work here! 

I realized I merged #1489; does that affect this PR/testing scenarios at all? If you're comfortable with the current state after #1489, I'll approve/merge this and cut a release soon.

TheQueenIsDead:
Awesome, that resolves all the issues that aren't to do with `required_status_checks`!
The only issue with this PR is that it takes one more apply before it is happy reconciling the diffs on future runs.

I'm happy to have this merged with the goal of improving it later :+1:

kfcampbell (approved):
Thanks for all the work on this, @TheQueenIsDead! I'm looking forward to having this in.

rpdelaney:
Any guess when we might get this in a release? :)

mbarany:
what if my status check context includes a colon?
Our CircleCi jobs for example `ci/circleci: unit_test` produces the following error:
```
Error: Could not parse unit_test:-1 as valid app_id
```

kfcampbell:
@mbarany do you mind creating an issue for that?

PR #1431, opened 2022-12-13
[Fix]: Advanced security settings for public repos.

### LINKED ISSUE #1419: [MAINT]: Public Repositories do not play well with Advanced Security settings
### Describe the need

First of all, thank you very much for adding the GitHub Advanced Security settings in terraform, see #1104 . This has been a boon to our organization, and we've used it to track and add settings to every repository using a common setup module.

Most of our repositories are private, but one of them is public. I've tried a dynamic setting to remove the `advanced_security` block from the public repository, but now, trying to remove it, I get the message `422 Advanced security is always available for public repos []`.

I do want to keep other security settings, like "Secret scanning", without having a required block specifying advanced security for the public repo. Please advise if there's a better way to go about this, otherwise, let me know if this is something that can be adjusted in the `security_and_analysis` block.

### SDK Version

_No response_

### API Version

_No response_

### Relevant log output

_No response_

### Code of Conduct

kuhlman-labs:
Resolves #1419 

----

### Before the change?

* Creating a public repo resource would error out if advanced security settings were set with the following message:

`"Advanced security is always available for public repos"` 

### After the change?

*  This change adds logic so that if a repos visibility is set to public it will not include the `advanced security` field in the payload but still allows setting the other fields.

### Other information

* 

----

### Pull request checklist

- [ ] No

If `Yes`, what's the impact:  

* N/A

### Pull request type

 
 

- Bugfix: `Type: Bug`

----


github/resource_github_repository.go
```diff
@@ -823,17 +823,19 @@ func flattenSecurityAndAnalysis(securityAndAnalysis *github.SecurityAndAnalysis)
...
 		return nil
 	}
 
 	securityAndAnalysis := input[0].(map[string]interface{})
 	update := &github.SecurityAndAnalysis{}
 
-	advancedSecurity := securityAndAnalysis["advanced_security"].([]interface{})[0].(map[string]interface{})
-	update.AdvancedSecurity = &github.AdvancedSecurity{
-		Status: github.String(advancedSecurity["status"].(string)),
+	if d.Get("visibility").(string) != "public" {
```
sethvargo, line 834: I'm not sure this is the correct approach. We should only apply the `advanced_security` settings if they were explicitly set in the HCL. Right now, setting the value in HCL has no effect (with no warning or anything). 

I recommend using `GetOk` instead of `Get` on "security_and_analysis" and only adding the value if set.

sethvargo (changes requested):

github/resource_github_repository.go
sethvargo: > The only time where the `advanced_security` field does not need to be set us for public repositories because it is enabled by default.

That is true as of today. It could change in the future and then suddenly things break in an unexpected way.

> In order to use the `GetOk` approach I would need to set that field as optional which may cause confusion for users in other scenarios.

I'm not sure why you'd need to set it as optional to use `GetOk`? I also think that's correct because it _is_ optional. It's not *required* for public or private repositories:

screenshot-20221213-dUPGmOyy@2x

> I think there is some value in seeing that the `advanced_security` field is set in the HCL even for public repos so practitioners can see at a glance that the repo is set to use advanced security.

This is all rooted in the assumption that "public == advanced security analysis", and I think that's a fundamentally flawed assumption.

kuhlman-labs: I think we're talking about two different things here.

> I'm not sure why you'd need to set it as optional to use GetOk? I also think that's correct because it is optional. It's not required for public or private repositories:

The block for `security_and_analysis` is currently optional, if it is not set in the HCL then we don't send the call to configure those settings. However, if you do set the `security_and_analysis` block in your HCL all of the subfields are required `advanced_security`, `secret_scanning`, and `secret_scanning_push_protection`. The issue when setting the `security_and_analysis` block for public repos is it would error out because by default `advanced_security` is enabled and sending that field in the payload would give the error message `"Advanced security is always available for public repos"`. However you can still configure the `secret_scanning`, and `secret_scanning_push_protection` features for a public repo. To do a GetOk check I would need to change the `advanced_security` field to optional which I think would cause some confusion in other scenarios as that field needs to be set to enabled to be able to enable the `secret_scanning`, and `secret_scanning_push_protection` fields.

sethvargo: I understand what you're saying, but I don't think this fixes the bug. See https://github.com/integrations/terraform-provider-github/blob/48619d4445cf87921850ef8425ee483defff915d/github/resource_github_repository.go#L635-L644, as an example. If the github_repository resource was already under management and the user deletes the entire `security_and_analysis` section, then Terraform will attempt to update the repo to disable Advanced Security (L638). But that's not valid for public repos.

My point is that you're headed down a path where you must explicitly check for repo visibility for any calls that modify `security_and_analysis`. Not only is this very error-prone, but I think it's the wrong design here.

kuhlman-labs: I've pushed a change that will add some logic to check the repo visibility when removing the `security_and_analysis` block and send the appropriate payload. 

Why do you see this as error prone? What approach would you recommend?

sethvargo: > Why do you see the check for visibility as error prone?

This comment thread is _literally_ an example of why it's error prone 😄. I think a better pattern is to centralize the logic for compiling the `SecurityAndAnalysis` block into a single function that takes the current state, previous state, and repo visibility into consideration when building the structure. Then use that function everywhere and throw a _ton_ of unit tests at it to make sure it behaves as expected under all possible conditions.

usmonster:
I guess this is superseded by #1489 and can be closed?

kfcampbell:
@usmonster I believe that is correct, though I'll leave it to @kuhlman-labs to close the PR.

Issue #876, opened 2021-08-09
Unable to create a repository in an organisation


jeroenjacobs79:
Hello,

I'm trying to create a repository in an organisation using my personal access token. My config is something like this:

```
provider "github" {
  owner = "headincloud"
}

resource "github_repository" "app_repo" {
  name        = "cluster-apps"
  description = "App repository"

  visibility = "private"
  delete_branch_on_merge = "true"
}

```

I have set the `GITHUB_TOKEN` to my PAT (which I granted all the permissions for now), but the repo still gets created under my own user-name, not onder my organisation.

Now, maybe creating organisation repos doesn't work with PAT's, and I should use OAuth? But I have no clue on how to proceed with this. When trying to create a new GitHub OAuth app, it asks for a homepage URL and callback URL but I have no clue on what I should put here for TerraForm. The documentation is not clear on this either...

So, what is the correct approach here? I believe the documentation could be improved on this subject.

Terraform version: 1.0.3
Github provider version: 4.13.0

jeroenjacobs79:
Update:

`gh repo create headincloud/my-test-repo` works just fine, so PAT's are allowed to create organisation repositories. 

So, the `owner` statement in the `provider` section seems to get ignored?

nicksantamaria:
I've encountered this same problem. Using the `owner` parameter in the `provider` block did not work, but setting the `GITHUB_OWNER` envvar did.

glennpratt:
Seeing the same issue, which appears to be a regression from previous versions

serain:
I faced this issue only when trying to create a repo in a child module. It doesn't pick up the org configuration from the parent module and you have to explicitly pass an aliased provider to the module.

mariodavid:
hi @serain ,

I'm facing the same issue within the module. Can you show an example of how you passed in the provider to the module? Just via a dedicated variable in the module or via the provider meta argument from the module (https://www.terraform.io/docs/language/meta-arguments/module-providers.html)?

Sinclert:
Same issue with current version (`4.17.0`). It seems the documentation block talking about the `owner` / `token` GitHub provider arguments, and how they could be replaced by the `GITHUB_OWNER` and `GITHUB_TOKEN` env. variables **is misleading**.

❌  - Does not work:
```hcl
provider "github" {
  owner = "my-org"
  token = "my-token"
}
```
✅  - Does work:

```hcl
provider "github" {}
```

```
export GITHUB_OWNER="my-org"
[...]

PR #1460, opened 2023-01-03
Document importance of submodule required_providers block

RulerOf:
Discussed over in #876, using implicit provider inheritance with this provider does not work, because terraform will prefer the `hashicorp/github` module by default.

Tagging @kfcampbell who requested the PR :)


kfcampbell (approved):
Thank you for doing this! I'm sure this will help prevent some confusion in the future.

PR #1465, opened 2023-01-06
fix: skip anonymous tests that can't be done anonymously

wwsean08:
This is in reference to #1414 but does not actually solve it.

----

### Before the change?

* The tests would fail due to the requirement of a token, however you can't have a token with anonymous tests.

### After the change?

* The tests no longer attempt to be run.

### Other information

While working on #1414 I found these tests which will never work without a feature change and test rework as well.

----

### Pull request checklist

- [X] No

If `Yes`, what's the impact:  

* N/A

### Pull request type

 
 

- Bugfix: `Type: Bug`

----

### TestAccGithubRepositoryFileDataSource
In this test it tries to create a repo and file which can then be queried, which is not possible with no access token as an anonymous test would have.

My original idea was to create a test with a static file, however that failed as there is no `owner` attribute on the data source so the owner would get set to an empty string even with `GITHUB_OWNER` or `GITHUB_ORGANIZATION` set.

### TestAccGithubTreeDataSource
In this test it tries to create a repo which can then be queried, which is not possible with no access token as an anonymous test would have.

My original idea was to create a test with a static repo, however that failed as there is no `owner` attribute on the data source so the owner would get set to an empty string even with `GITHUB_OWNER` or `GITHUB_ORGANIZATION` set.

### DIFF
--- a/github/data_source_github_repository_file_test.go
+++ b/github/data_source_github_repository_file_test.go
@@ -75,7 +75,7 @@ func TestAccGithubRepositoryFileDataSource(t *testing.T) {
 		}
 
 		t.Run("with an anonymous account", func(t *testing.T) {
-			testCase(t, anonymous)
+			t.Skip("anonymous account not supported for this operation")
 		})
 
 		t.Run("with an individual account", func(t *testing.T) {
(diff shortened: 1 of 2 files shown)


kfcampbell (approved):
Thank you!

Issue #1157, opened 2022-05-20
github_user_invitation_accepter requires invitation_id and fails if collaborators have already accepted


HariSekhon:
### Terraform Version

 v1.1.3

### Affected Resource(s)

- github_user_invitation_accepter

### Terraform Configuration Files

pretty much straight off this doc page:

https://registry.terraform.io/providers/integrations/github/latest/docs/resources/user_invitation_accepter

```hcl
resource "github_repository_collaborator" "my-ci" {
  repository = github_repository.repo.id
  username   = "my-ci"
  permission = "${var.ci_permission}"
}
provider "github" {
  alias = "invitee"
  token = var.ci_github_token
}
resource "github_user_invitation_accepter" "my-ci" {
  provider = github.invitee
  invitation_id = github_repository_collaborator.my-ci.invitation_id 
}
```

### Debug Output

https://gist.github.com/HariSekhon/22dd17a57951c8c8110d0ba10782da1e

### Expected Behavior
Ignore if the collaborator has already accepted

### Actual Behavior
Gets 404 errors or says field is not set when it's being passed `github_repository_collaborator.my-ci.invitation_id`

### Steps to Reproduce

1. `terraform apply`

morremeyer:
This is a bug we also encounter. It would be great if it could be resolved. I'll try to take a look at it later.

morremeyer:
I checked the code, specifically https://github.com/integrations/terraform-provider-github/blob/950473879572717866c7a8eb8bc1cc4346e8fe76/github/resource_github_user_invitation_accepter.go#L27-L44. 

This just passes back the errors from the API client if any occurs - I would therefore revise my `bug(?)` as to me, changing this would be adding new functionality - the current one works as designed.

To make usage of the `github_user_invitation_accepter` easier, I propose to add a flag to enable ignoring 404 errors. With that, the resource won't error out when an invitation has already been accepted.
I'm planning to contribute a PR for that next week.

kfcampbell:
@morremeyer that sounds awesome, thank you!

morremeyer:
I looked into implementing both my proposals. I came to the following conclusions:

Implementing the flag to enable ignoring 404 errors is technically easy, I started that in #1469. However, writing the tests would take me a considerable amount of time since this would be the first test using a client to manually perform an action for the collaborator in the tests. (see also https://github.com/integrations/terraform-provider-github/pull/1469#issuecomment-1385154011). 

[...]

PR #1469, opened 2023-01-09
feat: Add capability to ignore non-existing invitations

morremeyer:
Part of #1157 (resolves one of two cases).

----

### Before the change?

`github_user_invitation_accepter` errors when an invite was already accepted or declined.

### After the change?

The behaviour is configurable with the `ignore_not_found` argument. The argument defaults to `false`, which does not change default behaviour.

----

### Pull request checklist

- [X] No

### Pull request type

### DIFF
--- a/github/resource_github_user_invitation_accepter.go
+++ b/github/resource_github_user_invitation_accepter.go
@@ -3,8 +3,11 @@ package github
 import (
 	"context"
 	"fmt"
+	"log"
+	"net/http"
 	"strconv"
 
+	"github.com/google/go-github/v49/github"
 	"github.com/hashicorp/terraform-plugin-sdk/helper/schema"
 )
 
@@ -20,6 +23,11 @@ func resourceGithubUserInvitationAccepter() *schema.Resource {
 				Required: true,
 				ForceNew: true,
 			},
+			"ignore_not_found": {
+				Type:     schema.TypeBool,
+				Required: false,
+				ForceNew: false,
+			},
 		},
 	}
 }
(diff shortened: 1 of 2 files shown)


wwsean08:
So I'm working on reviewing tests currently to figure out the exact setups required to run all the acceptance tests.  So far I haven't seen any tests that would require two users/tokens, which it feels like this would.  I think it could be done as there are some one-off tests that require their own environment variables for things they require, which you could in theory use that pattern.

I _think_ this could be done by using two TestStep objects as opposed to one like most tests, where the first creates the invite (would look like a conventional TestStep you could see in most tests), then a second one that includes a [PreConfig](https://github.com/integrations/terraform-provider-github/blob/b0e28a4850694ad327473348ab073805166b3d1d/vendor/github.com/hashicorp/terraform-plugin-sdk/helper/resource/testing.go#L382) function.  In that function you would accept the invite using the GitHub client, then use the same config to validate that it ignores and doesn't error out.  

The only thing that I'm not 100% sure about is if you'll need to set [Destroy](https://github.com/integrations/terraform-provider-github/blob/b0e28a4850694ad327473348ab073805166b3d1d/vendor/github.com/hashicorp/terraform-plugin-sdk/helper/resource/testing.go#L421) to false in the first step, I think you will, but haven't tested it.

Note this is just my personal opinion based on my digging into testing lately and not any sort of official opinion as I'm not a maintainer or working for GitHub.  There might be a better way that I'm missing that someone else can offer.

morremeyer:
@wwsean08 Thank you for that explanation! I checked this out some more and think you are quite right.

However, there is no logic yet for a `testAccProvider` for the collaborator tests (like in https://github.com/integrations/terraform-provider-github/blob/89ef6659cf43afe6216793be8da15c09a5d37a8a/github/provider_test.go#L14), so implementing tests here would be time consuming for me as I don't know the code well enough yet.

Since #1488 resolves the problem we are currently having, I decided to close this PR so that someone with more experience can pick up the work here once they have time.

PR #1491, opened 2023-01-17
Set LDAP DN on team creation

### LINKED ISSUE #1490: Setting `ldap_dn` on `github_team` is inefficient, and requires site admin perms unnecessarily
The code in `resource_github_team.go` `resourceGithubTeamCreate` applies LDAP mapping for a team by:

- creating the team with no LDAP mapping
- calling `client.Admin.UpdateTeamLDAPMapping()` in a follow up call

This works, but has a couple of issues:

- The LDAP DN setting can be applied to the team in the payload of the initial creation ([ref docs](https://docs.github.com/en/enterprise-server@3.5/rest/teams/teams#create-a-team)) so making a second call to update it after the fact is unnecessary
- It requires that the PAT token the user uses to authenticate has the `site_admin` scope as the `UpdateTeamLDAPMapping` call is a call to an admin endpoint. To call the team creation endpoint you only need `admin:org` perms

I'd like to submit a PR to change this to add the LDAP DN (if set) to the original team creation call, and remove the `UpdateTeamLDAPMapping` call:

```
newTeam := github.NewTeam{
  Name:        name,
  Description: github.String(d.Get("description").(string)),
  Privacy:     github.String(d.Get("privacy").(string)),
// ADD THIS LINE:
  LDAPDN:      github.String(d.Get("ldap_dn").(string)),
}
```

This works for me locally

jMarkP:
Set `LDAPDN` on the `NewTeam` on creation, if it's not the empty string. This avoids an extra call to the GHE API, but more importantly means the user does not need to be a site admin to create a team with LDAP sync

Resolves #1490 

----

### Before the change?

* The provider used to call `client.Teams.CreateTeam()` and then follow up with `client.Admin.UpdateTeamLDAPMapping()`
* As well as being another call to the API, this requires that the user has the `site_admin` scope, but this is not required if the LDAP mapping is set on team creation

### After the change?

* Now the `LDAPDN` field is set in the call to `client.Teams.CreateTeam()` and the second call has been removed

### Other information

* This does introduce a potential imbalance for users who don't have site admin. They will be able to _create_ a `github_team` resource with an LDAP mapping, but won't be able to _update_ that mapping in the future. I'm not sure whether that's a blocker to this or not. Happy to hear your thoughts.
* I've not added tests. There are no tests for LDAP mappings at the moment, and I presume it would be difficult to add to the test suite without access to a GHE appliance setup with AD integration.

----

### Pull request checklist

- [x] No

If `Yes`, what's the impact:  

* N/A

### Pull request type

 
 

- Bugfix: `Type: Bug`

----


kfcampbell:
@jMarkP should the same change be replicated across to the "Update" function as well?

jMarkP:
> @jMarkP should the same change be replicated across to the "Update" function as well?

Unfortunately the team update endpoint doesn't allow you to update the LDAP DN of a team: https://docs.github.com/en/enterprise-server@3.5/rest/teams/teams#update-a-team

You have to use the [Update LDAP mapping admin endpoint](https://docs.github.com/en/enterprise-server@3.5/rest/enterprise-admin/ldap#update-ldap-mapping-for-a-team) which is what the update function already does

jMarkP:
Thanks for the merge @kfcampbell !

PR #1540, opened 2023-02-09
docs: fix erroneous github_repository_file docs

### LINKED ISSUE #1539: [DOCS]: `github_repository_file` data source `branch` argument docs are incorrect
### Describe the need

The docs for data source `github_repository_file` describe an optional `branch` argument with a note that if omitted, the repo's default branch will be used (and that it's usually `main`). However, when looking at the [code for the data source](https://github.com/integrations/terraform-provider-github/blob/2d1b724913b9432ac95645a8b3d57a105882e07c/github/data_source_github_repository_file.go#L31), it seems like the default for the argument is always going to be `main`.

### SDK Version

`5.17.0`

### API Version

_No response_

### Relevant log output

_No response_

### Code of Conduct

GilUpstream:
- Updates the docs with the correct default set for the branch argument

Resolves #1539 

----

### Before the change?

* Docs lead provider consumers to believe the default branch will be using when using the data source without a `branch` argument.

### After the change?

* Consumers of this provider will know that `main` will be used as the `branch` if it is not supplied.

### Other information

* Removed the note about the branch already existing as well since that is relevant to the resource for `github_repository_file` only.

----

### Pull request checklist

- [ ] No

If `Yes`, what's the impact:  

* N/A

### Pull request type

 
 

- Bugfix: `Type: Bug`

----


nickfloyd:
@GilUpstream Thanks for the update for clarity here.  This does make me wonder if this particular aspect of the provider is working as it should - i.e. It should use the default branch and not just `main` (which could be different from the default).  

@kfcampbell Do you have any thoughts or feels on this specific behavior?

kfcampbell:
Agreed that we should look up the default branch; I've created #1546 to track that work. In the meantime, this PR is an improvement in that the docs will match the behavior.

GilUpstream:
> @GilUpstream Thanks for the update for clarity here. This does make me wonder if this particular aspect of the provider is working as it should - i.e. It should use the default branch and not just `main` (which could be different from the default).
> 
> @kfcampbell Do you have any thoughts or feels on this specific behavior?

Agreed! I'm currently using the `github_repository` data source to look up the repo's default branch. My configs would be much simpler if the default branch was used.

PR #1502, opened 2023-01-20
feat(team_membership): allow id lookup from slug

jhoenzsch:
Resolves #1501

----

### Before the change?

* github_team_membership allows passing in the team slug as a team id.

### After the change?

* github_team_membership allows passing in the team slug as a team id.
----

### Pull request checklist
- [X] Tests for the changes have been added (for bug fixes / features)
- [X] Docs have been reviewed and added / updated if needed (for bug fixes / features)
- [X] Added the appropriate label for the given change

### Does this introduce a breaking change?

Please see our docs on [breaking changes](https://github.com/octokit/.github/blob/master/community/breaking_changes.md) to help!

- [ ] Yes (Please add the `Type: Breaking change` label)
- [X] No

If `Yes`, what's the impact:  

* N/A

### Pull request type

 
 

Please add the corresponding label for change this PR introduces:
- Bugfix: `Type: Bug`
- Feature/model/API additions: `Type: Feature`
- Updates to docs or samples: `Type: Documentation`
- Dependencies/code cleanup: `Type: Maintenance`

----

PR #1542, opened 2023-02-10
feat(team_members): allow team id lookup from slug

dennislapchenko:
Duplicated approach taken in: https://github.com/integrations/terraform-provider-github/pull/1502

----

### Before the change?

* resource `github_team_members` allowed only team id in `team_id` field.

### After the change?

*  resource `github_team_members` now allows team id as well as team slug into `team_id` field

### Other information

* 

----

### Pull request checklist

- [x] No

If `Yes`, what's the impact:  

* N/A

### Pull request type

 
 

----


github/resource_github_team_members.go
```diff
@@ -64,9 +63,9 @@ func resourceGithubTeamMembersCreate(d *schema.ResourceData, meta interface{}) e
 	orgId := meta.(*Owner).id
 
 	teamIdString := d.Get("team_id").(string)
-	teamId, err := strconv.ParseInt(teamIdString, 10, 64)
+	teamId, err := getTeamID(teamIdString, meta)
```
ilmax, line 66: I have the same concern as @csainty [here](https://github.com/integrations/terraform-provider-github/pull/1502#issuecomment-1418282281) This is making additional call and I'm already hitting api limits. 
Maybe this can try to parse the teamId and if it fails look it up by slug, to avoid adding additional api calls

nickfloyd: @ilmax Agreed, that seems like the right thing to do here.  @dennislapchenko is that something you'd be up for changing in the 4 calls here?

dennislapchenko: Sure, @nickfloyd ! I thought y'all agreed it is not a real issue since that PR got merged.
I have committed and update to `getTeamID` func. It does make sense to remove API call when an integer is passed, since github would rightfully return an error if the team does not exist.

dennislapchenko:
Sorry for the request review train, took a bit to understand only 1 person could be re-requested, rather than issues with my internet 😅

dennislapchenko:
@ilmax @nickfloyd hey guys, any traction on reviewing this please?

ilmax:
@dennislapchenko sorry for the late reply, this LGTM. There's a small change though that now `getTeamById` is not validating the passed in id, but I'm fine with that since 99% of the time we are not making this id up but use the attribute of some other resources.
I will leave the approval to @nickfloyd since I'm fairly new here and may not have the full picture.

kfcampbell:
Thanks for the contribution! I'd like to sit on this a little bit longer before getting it in to allow more time for potential overhead issues described [here](https://github.com/integrations/terraform-provider-github/pull/1502#issuecomment-1418282281) to crop up. I'm somewhat hesitant to replicate the pattern elsewhere just yet.

dennislapchenko:
@kfcampbell Cheers! The pattern is not replicated though as int parsing will not be an API call anymore. It seems that it does make sense it that way, because in 'invalid team id' case githup api would return a sensible error, right?

kfcampbell:
Ahh I had missed that, thank you for correcting me. I'll get this merged and released soon.

kfcampbell (approved):

Issue #1532, opened 2023-02-05
[FEAT]: all org members(users) data source


mac2000:
### Describe the need

Will be nice to have GitHub org members data source so me may retrieve all members and their public email if it is available

Use case: make sure that github org does not contain deactivated active directory members

Current impl: we ask teammates to make sure that GitHub account they are using have publicly available work email, so script that sync members can rely on it (e.g. if by mistake you will make your email private or personal, script will just kick you out of org)

Yes, I know, there are enterprise'ish plans everywhere, so everything may be in sync out of the box, it is just an use case example

I was able to retrieve [azuread users](https://registry.terraform.io/providers/hashicorp/azuread/latest/docs/data-sources/users) list successfully, but it seems that in github provider both [users](https://registry.terraform.io/providers/integrations/github/latest/docs/data-sources/users) and [members](https://registry.terraform.io/providers/integrations/github/latest/docs/data-sources/membership) are requiring user list to be retrieved

Wondering if that option may be set as optional, so if not set, whole list is returned, something like:

```graphql
query getAllMembers($org:String!,$after:String) {
  organization(login:$org) {
    membersWithRole(first:100,after:$after) {
      nodes {
        id
        login
        email
      }
    }
  }
}
```

somewhere [here](https://github.com/integrations/terraform-provider-github/blob/main/github/data_source_github_users.go#L79) as an else statement:

<img width="498" alt="image" src="https://user-images.githubusercontent.com/88868/216819487-caef8af1-987c-45bd-bed7-3c4a027cf048.png">

which technically should solve this feature request and will be backward compatible

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

kfcampbell:
@mac2000 do you have any interest in opening a PR to implement this behavior?

mac2000:
while we are waiting for an review and approve for emails, here is how we can receive member role from an graphql side:

```gql
query  {
  organization(login:"your_org_name") {
    membersWithRole(first:100) {
      edges {
        role # ADMIN, MEMBER
        node { id login }
      }
    }
  }
}
```

not sure yet if this will be as easy as emails addition, but i bed that when it will come to part two with reading all users it definitely will become clear

PR #1557, opened 2023-02-19
retrieve publicly available emails as well

mac2000:
This is an very first baby step going to solve #1532 feature request

Technically all we are doing here is adding one more field to GraphQL query and map it to response, aka:

```gql
query getAllMembers($org:String!,$after:String) {
  organization(login:$org) {
    membersWithRole(first:100,after:$after) {
      nodes {
        id
        login
        email # <- added
      }
    }
  }
}
```

This should not brake anything because we are not touching any existing fields, only adding new one, the idea was to introduce this first and as a next step try to figure out how can we rewrite code in a such way so it will retrieve all users if no inputs given, but to me it seems better to do it step by step, otherwise it may become to many changes to review

Here is how sample output will look like:

image

To be fair I'm not sure if passing such data in different lists is a good idea but cannot touch it without braking something else so at first iteration we may keep it simple

----

### After the change?

* Email field added to response

### Other information

* @kfcampbell I'm not sure if you asked  about something like this, but I did tried to do my best

----

### Pull request checklist
- [+] Tests for the changes have been added (for bug fixes / features)
- [+] Docs have been reviewed and added / updated if needed (for bug fixes / features)
- [+] Added the appropriate label for the given change

- [+] No

### Pull request type

Type: Feature

----


kfcampbell:
@mac2000 thanks! Could you provide appropriate documentation for this field in the website/docs/d/users.html.markdown file so our docs stay up-to-date?

nickfloyd (approved):
Thanks, @mac2000 for this change! ❤️

PR #1588, opened 2023-03-09
data_source_github_organization: expand members

### LINKED ISSUE #1532: [FEAT]: all org members(users) data source

mac2000:
Fixes #1532

At moment github organization datasource reads all members but exposes only their logins

```go
for index := range allMembers {
	memberList = append(memberList, *allMembers[index].Login)
}
```

Technically we are receiving `*github.User` and may use it, aka something like:

```go
for _, member := range allMembers {
  memberList = append(memberList, map[string]interface{}{
    "login": *member.Login,
    "id":    fmt.Sprintf("%d", *member.ID),
    "email": *member.Email,
    "role":  *member.RoleName,
  })
}
```

But somehow rest api does not return wanted emails and roles, that's why we are switching to graphql and running query like this:

```gql
query { 
  organization(login:"your_org_name") {
    membersWithRole(first:100, after:null) {
      edges {
        role
        node {
          id
          login
          email
        }
      }
    }
  }
}
```

Following example:

```tf
  name = "you_org"
}
output "example" {
  value = data.github_organization.example.members
}
```

before the change:

```log
Changes to Outputs:
  + example = [
      + "user1",
      + "user2",
    ]
```

after the change:

```log
Changes to Outputs:
  + example = [
      + {
          + "id"    = "1"
          + "login" = "user1"
          + "email" = "user1@contoso.com"
          + "role" = "ADMIN"
        },
      + {
          + "id"    = "2"
          + "login" = "user2"
          + "email" = "user2@contoso.com"
          + "role" = "MEMBER"
        },
    ]
```

Things that may be improved:

- at moment schema defines members as list of string maps, aka `[]map[string]string` wondering if thats possible to have something strongly typed

Why this one may be even better:

- we do get only wanted data
- it will be easier to add new fields in future (e.g. if someone will ask for user avatar or something like this)

Considerations:

- to make it fully backward compatible - we may leave members field as is and add second field with expanded members - it will guarantee that we wont brake anything, but not sure if that's a good idea to have such duplicated fields

Notes:

- @kfcampbell here is another attempt that solves whole issue at once, so I even not sure if we will need #1557 anymore
- @pascal-hofmann it should solve your issue with roles as well


mac2000:
@pascal-hofmann yep, confirming, while I have playground did tested your guess and indeed

```diff
- github_organization.this.members
+ github_organization.this.members[*].login
```

```tf
  name = "rabotaua"
}

output "example" {
  value = data.github_organization.example.members[*].login
}
```

gives us the same output that was previously

```log
Changes to Outputs:
  + example = [
      + "user1",
      + "user2",
    ]
```

kfcampbell:
Hmm...I like the idea here, though I'm a little torn as [v5.0.0](https://github.com/integrations/terraform-provider-github/releases/tag/v5.0.0) was released in September. I was reading some [Terraform documentation](https://developer.hashicorp.com/terraform/plugin/sdkv2/best-practices/versioning#versioning-specification) recently which recommends:

> Increasing the major number indicates that significant breaking changes have been made, and thus extra care or attention is required during an upgrade. To allow practitioners sufficient time and opportunity to upgrade to the latest version of the provider, we recommend releasing major versions no more than once per year. Releasing major versions more frequently could present a barrier to adoption due to the effort required to upgrade.

Using another data source would allow this to be shipped right away, though it isn't the most elegant way. I don't love the idea of duplicating the information in the map to keep backwards compatibility. :thinking:

pascal-hofmann:
In the past this provider had frequent breaking changes in minor versions. Maybe just release without major version bump? 🙈😬

mac2000:
@kfcampbell, @pascal-hofmann guys I have an idea which you may like

In a GraphQL world it is totally fine to add new field and mark old one as deprecated, then with help of statistics we may see when it become unusable and removing it - we are using such approach heavily to make sure we wont broke clients, even more - all this happens automatically

From what I see so far - [Terraform has exactly such approach](https://developer.hashicorp.com/terraform/plugin/sdkv2/best-practices/deprecations#provider-data-source-or-resource-removal), and fields in schema have deprecated flag

So the proposition is:

1. add new field `users` (or whatever name you will like)
2. mark `members` as deprecated and in description point users that they should switch to new field

As a result:

- no new datasources - there is already to many ways to receive users and adding one more will just blown everyones minds
- we definitely wont brake anything
- we are covering @kfcampbell complain about having multiple similar fields in map by marking old one as deprecated
- we gently notifying users of whats happening so they will have time to switch
- whenever next major version will arrive we will just cleanup deprecated fields
- profit

Just my thoughts, not saying that we should do exactly like this - but so far it sounds like a good compromise where everyone will be happy

BR
Alex

kfcampbell (approved):
Thank you for the update! I like the deprecation approach.

PR #1617, opened 2023-03-23
feat: Split out Pages resource

### LINKED ISSUE #782: Proposal to decouple GitHub pages functionality from github_repository resource
Hi!

I would like to be able to create a repo and enable the GitHub Pages functionality for a branch using Terraform. Currently I'm unable to do this without multiple Terraform runs and would like to propose decoupling the GitHub Pages functionality from the `github_repository` resource (in a similar way `github_branch_protection` is it's own resource).

If the branch to be used by GitHub pages does not exist, the API returns the following:

```json
{
    "message": "The gh-pages branch must exist before GitHub Pages can be built.",
    "documentation_url": "https://docs.github.com/rest/reference/repos#create-a-github-pages-site"
}
```

With the current implementation it's not possible to create a repository, create a branch and enable Pages for that branch in a single run.

I'm happy pick this up but wanted to share the idea before investing time into it.

Thanks!

isaacsanders:
Resolves #782 

----

### Before the change?

* When configuring a repository for the first time, one can not specify pages, since a branch needs to exist before pages can be configured.

### After the change?

* Decoupling these two concepts allows for more fine control to be used to manage the dependencies, using tools Terraform already gives us.

### Other information

* I'm not a Go developer. This code builds, but I'm eager to take more feedback, and to get help from people who know what they are doing more than I do.

----

### Pull request checklist

- [x] Yes (Please add the `Type: Breaking change` label)
- [ ] No

If `Yes`, what's the impact:  

* The `pages` property has been removed from the repository resource
* I think this could be added back and deprecated, but it was helpful for me to understand how the code changes if the feature is removed


isaacsanders:
@nickfloyd What can I do to move this along?

kfcampbell:
@isaacsanders I think the next best step would be to handle migrations for this breaking change gracefully. We've got some prior art on state migration [here](https://github.com/integrations/terraform-provider-github/blob/main/github/resource_github_repository.go#L31). I'm sure there's Hashicorp docs about this somewhere, though I can't seem to find them at the moment.

highb:
Are [these the docs](https://developer.hashicorp.com/terraform/plugin/sdkv2/resources/state-migration) that you're thinking of @kfcampbell ?

kfcampbell:
Looks like it, thanks!

kfcampbell:
@isaacsanders we're planning on cutting a new major version in the near future. Do you have the interest/inclination to continue this work so we can get it merged? If not, perhaps somebody from the community sees value in resuming this feature.

isaacsanders:
I’d like to make the split happen, but it isn’t something I have a lot of time for right now. The current API is broken, and I think I’ve done enough work in the PR that GitHub could get it over the line.

laurenty:
Would flagging the `pages` argument under `repository` as deprecated instead of removing it make it easier to get this change out since it wouldn't be a breaking change? Something similar was done with `github_repository.default_branch` and `github_branch_default`.

kfcampbell:
@laurenty Ideally we'd do that and at the same time provide a separate resource for managing pages for a repository. I'd like to avoid deprecation without an alternative in place.

laurenty:
@kfcampbell that's exactly what I was suggesting 🙂 This PR adds a new dedicated resource `github_repository_pages`, but removes support for GH Pages from `github_repository`. I think it would be useful to keep the `github_repository` resource as-is, flag the `pages` property as deprecated and update the docs to warn user that using both `github_repository.pages` and `github_repository_pages` will result in unexpected results (just like `github_branch_default` does https://registry.terraform.io/providers/integrations/github/latest/docs/resources/branch_default)

kfcampbell:
Ahh, my mistake! Thank you for correcting me; that's a great course of action.

isaacsanders:
I’m getting back into coding outside of work, so I should have time to get this in order.

tpreviero:
@isaacsanders, if I can help in any way to resume the work on this PR I'd be happy to!

I was trying to resume splitting the GitHub pages from repository resource by [adding deprecations to the docs](https://github.com/tpreviero/terraform-provider-github/commit/b5424528ee841fed69344fbff20a6d3c9dcd8005) based on the work in this PR, but I don't want to create another stream of changes if this one's still going 🙃

PR #1639, opened 2023-04-06
Add datasource to get app JWT token

phillebaba:
----

### Before the change?

* It was not possible to get application JWT token.

### After the change?

* The new datasource exposes the application JWT token.

### Other information

This change introduces a datasource that is useful for specific edge cases where the application JWT token is needed outside of the provider. I do not expect a lot of people to require this feature but it is very useful for those who do, as the process of generating the JWT is complex and not possible otherwise in Terraform.

----

### Pull request checklist

- [x] No

If `Yes`, what's the impact:  

* N/A

### Pull request type

 
 

- Bugfix: `Type: Bug`

----


kfcampbell:
Hmm...in what use cases would this be needed? Is there an issue tracking this addition?

phillebaba:
@kfcampbell I have not created an issue for this, as I thought it was easier to discuss with the implementation present, and implementing the datasource was quick.

So one of the things I work on is the [Flux Terraform Provider](https://registry.terraform.io/providers/fluxcd/flux/latest) which interacts with Git to bootstrap Flux in a Kubernetes cluster. I have worked with different authentication schemes with GitHub and found issues with a lot of them. PATs are linked to users unless you create bot accounts and the same could be said about deploy keys. A bigger issue is that using a PAT to create a deploy key and then removing the PAT will result in the deploy key being removed. My reasoning has been that GitHub apps has been the be compromise between creating a bot account and using PATs from end users.

The use case for exposing the JWT created by the provider would be to be able to use it in the Flux provider to make commits.

kfcampbell:
Oh interesting, thanks for sharing. 

I agree that PATs aren't the most optimal solution; their long-lived nature and broad permissions make me uncomfortable. Bot accounts also aren't elegant and I don't like having to maintain the public-facing parts of them.

> My reasoning has been that GitHub apps has been the be compromise between creating a bot account and using PATs from end users.

That seems reasonable to me.

kfcampbell:
Before merging, we'll need to reference the function you've created in `provider.go` where the other data sources are referenced, and add appropriate documentation and reference it in the `github.erb` file.

phillebaba:
@kfcampbell I just thought about this feature a bit more. It may be better to create a datasource that just takes the application credentials and outputs a JWT. Instead of outputting the JWT from the provider. I will close this PR and create an issue instead to discuss how that could be implemented.

PR #1640, opened 2023-04-06
Allow parallel requests to take place when non using github.com

### LINKED ISSUE #1597: [MAINT]: Enable parallelism to increase performance for self hosted Enterprise servers
### Describe the need

As a self hosted  Github Enterprise user I am not bound to github.com rate limits so I would like to discuss if it is possible to unlock the possibility to have parallel requests.

I understand [your concerns](https://github.com/integrations/terraform-provider-github/pull/235) related to rate limiting and best practices but, in my opinion, it should not limit everyone.

### SDK Version

_No response_

### API Version

_No response_

### Relevant log output

_No response_

### Code of Conduct

Roviluca:
Resolves #1597
----

### Before the change?
* A locking system was enforcing the client to make api calls one at a time.

### After the change?
* Api call are made in a serial way by default but a new provider setting parallel_requests to let users decide to allow parallel requests
* Serial requests are enforced in case the api call is made to https://api.github.com

### Other information
-

----

### Pull request checklist

no

- [x] No

If `Yes`, what's the impact:  

* N/A

### Pull request type
Feature


github/transport.go
```diff
@@ -59,7 +60,7 @@ func (rlt *RateLimitTransport) RoundTrip(req *http.Request) (*http.Response, err
 	// Make requests for a single user or client ID serially
 	// This is also necessary for safely saving
 	// and restoring bodies between retries below
-	rlt.lock(req)
+	rlt.smartLock(true)
```
kfcampbell, line 63: Can you describe what this change is doing?

github/provider.go
```diff
@@ -308,13 +318,21 @@ func providerConfigure(p *schema.Provider) schema.ConfigureFunc {
 		}
 		log.Printf("[DEBUG] Setting read_delay_ms to %d", readDelay)
 
+		parallelRequests := d.Get("parallel_requests").(bool)
+		isGithubDotCom, _ := regexp.MatchString("^https://api.github.com", baseURL)
+		if parallelRequests && isGithubDotCom {
+			return nil, fmt.Errorf("parallel_requests cannot be true when connecting to public github")
+		}
+		log.Printf("[DEBUG] Setting parallel_requests to %t", parallelRequests)
```
kfcampbell, line 326: What does this log print for a case of using github.com? It looks like it would print "Setting parallel_requests to false".

github/provider.go
```diff
@@ -56,6 +57,12 @@ func Provider() terraform.ResourceProvider {
 				Default:     0,
 				Description: descriptions["read_delay_ms"],
 			},
+			"parallel_requests": {
```
kfcampbell, line 60: By default, does this make all requests parallel? Is there a limit to the number of requests that can be made in parallel?

github/provider.go
```diff
@@ -226,6 +233,9 @@ func init() {
 			"Defaults to 1000ms or 1s if not set.",
 		"read_delay_ms": "Amount of time in milliseconds to sleep in between non-write requests to GitHub API. " +
 			"Defaults to 0ms if not set.",
+		"parallel_requests": "Allow the provider to make parallel api calls to the Github server." +
+			"It can't be enabled for github.com in order to comply on rate_limits and best practices." +
+			"Defaults to false if not set",
```
kfcampbell, line 238:
```suggestion
		"parallel_requests": "Allow the provider to make parallel API calls to GitHub. " +
			"It can't be enabled for github.com in order to avoid triggering abuse rate limits. " +
			"Defaults to false if not set",
```

github/provider.go
```diff
@@ -308,13 +318,21 @@ func providerConfigure(p *schema.Provider) schema.ConfigureFunc {
 		}
 		log.Printf("[DEBUG] Setting read_delay_ms to %d", readDelay)
 
+		parallelRequests := d.Get("parallel_requests").(bool)
+		isGithubDotCom, _ := regexp.MatchString("^https://api.github.com", baseURL)
```
kfcampbell, line 322: What's the reason we're ignoring this error here?

github/provider.go
Roviluca: The default for the flag is to keep request serial for backward consistency and to be safe when calling github.com without changing it.
In regards of the upper limit, I am allowing terraform to work "normally" and use the go routines chosen by the user through the `--parallelism` flag that is documented [here](https://developer.hashicorp.com/terraform/cli/commands/apply#parallelism-n); default is 10. The locking system was preventing that to comply with github.com rate limits.

github/provider.go
Roviluca: From the doc was triggered  only when setting a malformed regexp so I skipped it.I am going to add the error check.

github/provider.go
Roviluca: yes, it is going to print that. Isn't it correct? I tried to keep the format consistent with other settings i.e.: `log.Printf("[DEBUG] Setting read_delay_ms to %d", readDelay)`

github/provider.go
kfcampbell: Understood, thank you!

github/provider.go
kfcampbell: Thank you!

github/provider.go
```diff
@@ -309,13 +319,24 @@
 		}
 		log.Printf("[DEBUG] Setting read_delay_ms to %d", readDelay)
 
+		parallelRequests := d.Get("parallel_requests").(bool)
+		isGithubDotCom, err := regexp.MatchString("^https://api.github.com", baseURL)
```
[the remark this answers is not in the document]
Roviluca: FIxed this in [d66449e](https://github.com/integrations/terraform-provider-github/pull/1640/commits/d66449eb972ec4caa5b26543c9e44e9d765ead59)

Roviluca:
Fixed the failed check for "Incomplete regular expression for hostnames"

github/transport.go
Roviluca: Updated the comment before the first smartLock occurrence and the function.

github/provider.go
kfcampbell: You're right, I think I misread something, sorry for the confusion.

kfcampbell:
Can you talk to me more about the `smartLock` concept? Is the idea that the previously used locking functions are going away and all calls of lock/unlock should switch to the new locking function?

Roviluca:
Yes, that is the idea. I replaced  all occurrences of `rlt.lock()` and `rlt.unlock()` with a new function `smartLock()` that accepts a boolean as an input. When True tries to lock, when False unlocks. By default it keeps the same behaviour as before.

The reasons that made me create the new function are:
- make evident with the name that is not  a simple lock but there is some logic inside
- skip locking/unlocking when `rlt.parallelRequests` is set to true ( the purpose of the PR ) 
- avoided me to add the `if rlt.parallelRequests` in two places ( the old lock/unlock functions )  or more if I decided to put the if in all invocations

kfcampbell (commented):
Thank you for working with me here. I have two testing and one comment requests for you and then I'll be 👍 on the change: can you please

- add unit test(s) that interrogate the regex for checking if we're on dotcom
- add unit test(s) that interrogate the smartLock function
- make a provider.go change to the description of parallel_requests that makes it clear when you'd want to use it

Roviluca:
Added unit tests for smart_lock and fixed description for parallel_requests.

Regarding `add unit test(s) that interrogate the regex for checking if we're on dotcom`  i have the feeling that it is over testing. The reason is that what I am doing there is just input validation with a very simple regexp.
On the other hand, it may be useful to add an acceptance test for that but, I am having some difficulties wrapping my head around how the acceptance are working. It seems like setting the [config var](https://github.com/integrations/terraform-provider-github/blob/main/github/provider_test.go#L145) is not changing the behaviour so they seems like not testing anything 🤔.

kfcampbell:
> Added unit tests for smart_lock and fixed description for parallel_requests.

Thank you!

>  i have the feeling that it is over testing. The reason is that what I am doing there is just input validation with a very simple regexp

That's a reasonable argument, we can leave that test out. 

> I am having some difficulties wrapping my head around how the acceptance are working. It seems like setting the [config var](https://github.com/integrations/terraform-provider-github/blob/main/github/provider_test.go#L145) is not changing the behaviour so they seems like not testing anything thinking.

At the momnet, acceptance tests are currently run manually and are not triggered on automated PR checks. I am hopeful that this will change in the future.

Roviluca:
Fixed a linting error 👆

kfcampbell (approved):
Yep, thanks for working with me here. I'll get this merged and released shortly.

PR #1663, opened 2023-04-24
feat: Add build_type to github_repository (pages)

### LINKED ISSUE #1267: github_repository pages source should allow for github actions source
### Affected Resource(s)
- github_repository

### Expected Behavior
I should be allowed to specify the "source" of my pages enabled github_repository resource to be GitHub Actions, much like I can in the GitHub UI.

### Actual Behavior
I am only allowed to specify my pages source as branch/path.

0x46616c6b:
Resolves #1267

----

### Before the change?

* It is not possible to configure a GitHub Pages with GitHub Actions

### After the change?

* The `build_type` option is added to `pages` property

----

### Pull request checklist

- [x] No

## Open Questions

- How can I ensure if the `build_type = "legacy"` that `source` is not optional anymore? 
- This PR is related with #1617, can it be introduced before the resource is split out?


lukas-tide:
@0x46616c6b to validate the field you could do the checks [here](https://github.com/0x46616c6b/terraform-provider-github/blob/add-build-type-to-repository-pages/github/resource_github_repository.go#L725)

Validation functions are only supported for basic types, so the only option I have found is to do the checks on creation and update and in this case creation reuses update, so only update would work fine I guess. Maybe @kfcampbell could clarify if that is the case and if this is needed for the PR to get merged

website/docs/r/repository.html.markdown
```diff
@@ -124,7 +124,9 @@ initial repository creation and create the target branch inside of the repositor
 
 The `pages` block supports the following:
 
-* `source` - (Required) The source branch and directory for the rendered Pages site. See [GitHub Pages Source](#github-pages-source) below for details.
+* `source` - (Optional) The source branch and directory for the rendered Pages site. See [GitHub Pages Source](#github-pages-source) below for details.
+
+* `build_type` - (Optional) The type of GitHub Pages site to build. Can be `legacy` or `workflow`. If you use `legacy` as build type you need to set the option `source`.
```
kfcampbell, line 129: What happens if both source and build_type are present, which one takes precedence? What happens if legacy is set for build_type but source isn't set?

kfcampbell:
I'm okay with the validation approach; not running checks on creation/update should result in an error at the API level that will be passed back to the user in any case. 

@0x46616c6b does this result in a breaking change for users? I'm not sure if providing "legacy" by default on all build_types results in any API-level changes or not.

0x46616c6b:
I understand that upgrading the provider and applying the changes will cause a difference (in state only), but it won't affect the repository configuration. If this change is considered a breaking change (due to the new key changing the state), we should reflect that too.

website/docs/r/repository.html.markdown
0x46616c6b: **Allowed Options**

```terraform
  name = "example"

  pages {
    source {
      branch = "main"
      path = "/"
    }
    build_type = "legacy"
  }
}
```

```terraform
  name = "example"

  pages {
    source {
      branch = "main"
      path = "/"
    }
  }
}
```

```terraform
  name = "example"

  pages {
    build_type = "workflow"
  }
}
```

**Problematic / wrong Configuration**

```terraform
  name = "example"

  pages {
    source {
      branch = "main"
      path = "/"
    }
    build_type = "workflow"
  }
}
```

kfcampbell (approved):
I'm okay with state changing as a non-breaking change as long as the API/GitHub state remains consistent. Thanks!

benlangfeld:
This appears to have broken some use cases: https://github.com/integrations/terraform-provider-github/issues/1724

PR #1671, opened 2023-05-02
Add new datasource to get GitHub App token

### LINKED ISSUE #1656: [FEAT]: Github App JWT Datasource
### Describe the need

Generating a Github app JWT from client credentials requires multiple steps that is not possible in pure Terraform. Currently it is possible to pass these credentials to the provider config so that the provider can use the application credentials to authenticate with the API.

There are however use cases where the JWT could be used outside of the provider. One example for where it could be used in the [Flux Terraform Provider](https://registry.terraform.io/providers/fluxcd/flux/latest) which needs credentials to commit files to GitHub.

For this reason it would be useful to have a new datasource called `github_app_jwt` which takes the same credentials as the provider but outputs the JWT that is generated.

```
data "github_app_jwt" "this" {
    id                      = var.app_id             
    installation_id = var.app_installation_id 
    pem_file           = var.app_pem_file       
}
```

### SDK Version

_No response_

### API Version

_No response_

### Relevant log output

_No response_

### Code of Conduct

phillebaba:
Resolves #1656

----

### Before the change?

It was not possible to generate a GitHub App JWT to use outside of the GitHub provider.

### After the change?

The new datasource enables Terraform users to pass a GitHub App JWT to other provider resources so that they are able to authenticate with the GitHub API.

### Other information

N/A

----

### Pull request checklist

- [x] No

If `Yes`, what's the impact:  

* N/A

### Pull request type

 
 

- Bugfix: `Type: Bug`

----


examples/app_token/main.tf
```diff
@@ -0,0 +1,15 @@
...
+  }
+}
+
+provider "github" {}
+
+data "github_app_token" "this" {
+  app_id          = var.app_id
+  installation_id = var.installation_id
+  pem_file        = file(var.pem_file_path)
+}
```
kfcampbell, line 15: Is the empty file examples/app_token/outputs.tf below necesary?

github/data_source_github_app_token.go
```diff
@@ -0,0 +1,67 @@
...
+			"token": {
+				Type:      schema.TypeString,
+				Computed:  true,
+				Sensitive: true,
+			},
+		},
+	}
+}
+
+func dataSourceGithubAppTokenRead(d *schema.ResourceData, meta interface{}) error {
```
kfcampbell, line 44: Can you create appropriate testing for this please?

website/docs/d/app_token.html.markdown
```diff
@@ -0,0 +1,39 @@
+---
```
kfcampbell, line 1: In order to be accessible, this file needs to be linked appropriately in the github.erb file.

bateller (approved):

github/data_source_github_app_token.go
```diff
@@ -0,0 +1,67 @@
...
+				Type:        schema.TypeString,
+				Required:    true,
+				Description: descriptions["app_auth.installation_id"],
+			},
+			"pem_file": {
+				Type:        schema.TypeString,
+				Required:    true,
+				Description: descriptions["app_auth.pem_file"],
+			},
+			"base_url": {
```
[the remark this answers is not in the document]
phillebaba: Changed it to use the base url from the client instead.

website/docs/d/app_token.html.markdown
phillebaba: Created a link in the file. Do I need to generate a proper HTML file?

github/data_source_github_app_token.go
phillebaba: To test this I need a proper GitHub App. Would you say its a good idea to use the same setup that already exists to test the providers app auth?

https://github.com/integrations/terraform-provider-github/blob/main/github/apps_test.go

kfcampbell: I'm :+1: on reusing the same testing configuration.

website/docs/d/app_token.html.markdown
kfcampbell: Nope! You're all good; that's done on Hashicorp's side.

github/data_source_github_app_token.go
phillebaba: I found some issues with how to test this. Either we do this with real credentials which is not present in the current test scope. Or we solve this like the provider tests and mock the client response. The issue is that using httptest would require changing the base url, which is set in the provider scope. 

Do you have any other good idea for testing this datasource?

phillebaba: @kfcampbell any thoughts?

kfcampbell: I like the idea of mocking the client response. 

> The issue is that using httptest would require changing the base url, which is set in the provider scope.

I'm not quite sure changing the base URL is an issue in tests. We have several tests that change the base URL, like [this one](https://github.com/integrations/terraform-provider-github/blob/f280fb53922abeb72cbd3ccc29beea53374b8522/github/data_source_github_repository_file_test.go#L277).

phillebaba: @kfcampbell thanks for pointing me in the right direction. I have now implemented a test that validates the token creation with the mocked API endpoint. I removed any JWT validation because it would be impossible to actually have a static JWT. Hopefully this test is good enough.

kfcampbell (approved):
@phillebaba thanks for working with me here!

PR #1675, opened 2023-05-06
feat: Expose detailed repositories in github_team data_source

### LINKED ISSUE #1670: [FEAT]: Expose repository_ids in github_team data_source
### Describe the need

Currently the [`github_team`](https://registry.terraform.io/providers/integrations/github/latest/docs/data-sources/team) data_source exposes the attribute [`repositories`](https://registry.terraform.io/providers/integrations/github/latest/docs/data-sources/team#repositories), which is a list of repository names. 

My use-case is that I want to create a `github_actions_organization_secret` with the `selected_repository_ids` set to the repositories which a specific `github_team` has access to. The `github_actions_organization_secret` requires a list of  repository ids, but the `github_teams` data_source only provides a list of repository names. While this list of names can be chained with the [`github_repository`](https://registry.terraform.io/providers/integrations/github/latest/docs/resources/repository) resource to fetch the id:s, [the API call for fetching this information is already done](https://github.com/integrations/terraform-provider-github/blob/e9bb788f4a640bba0d0545880af8c2d63e8d47b5/github/data_source_github_team.go#L154-L168) within the `github_team` data_source, so it might as well be exposed there directly. 

## Suggested Change
A new attribute can be added to the `github_team` data_source exposing these id:s, `repository_ids` for example. It might be valuable to expose even more attributes, such as the `role_name` the team has for the repository. In my mind the `repositories` attribute should probably be a list of structs exposing these 3 fields:
* `repository_name`
* `repository_id`
* `role_name` (`admin`/`maintain`/`write`/...)

However, Implementing this would break backwards compatibility, so I think one of the 2 following solutions are the best options:
1. Adding a new attribute called `repository_detailed`, which contains each connected repository's `name`, `id` and `role_name`
    * This would not break any compatibility, but still expose all fields one would want 
    * I'm not that used to Terraform Provider development, but I think a list of structs could be extended in the future if more than these three attributes are wanted if the future
    * ```go
	  "repositories_detailed": {
		  Type:     schema.TypeList,
		  Computed: true,
		  Elem: map[string]*schema.Schema{
			  "repo_name": {
				  Type:     schema.TypeString,
				  Required: true,
			  },
			  "repo_id": {
				  Type:     schema.TypeInt,
				  Computed: true,
			  },
			  "role_name": {
				  Type:     schema.TypeString,
				  Computed: true,
			  },
		  },
	  } 
	  ```
2. Adding 2 new attributes, one for the `repository name`, one for the repository `role_name`
    * Not as elegant as 1, but it works

### SDK Version

_No response_

### API Version

_No response_

### Relevant log output

_No response_

### Code of Conduct

felixlut:
Resolves #1670 

----

### Before the change?

The `github_team` datasource only exposed the names of the connected repositories in the past. 

### After the change?

The `github_team` datasource now has an additional field called `repositories_detailed` that includes the `repo_id` and `role_name` of each connected repository. 

----

### Pull request checklist
- [ ] Added the appropriate label for the given change (No permissions to add labels)

It does not introduce a breaking change, but as discussed [in the issue](https://github.com/integrations/terraform-provider-github/issues/1670#issuecomment-1536614401), a new issue could be raised for creating one for the future. 

- [x] No

### Pull request type
* Feature/model/API additions `Type: Feature`

----


website/docs/d/team.html.markdown
```diff
@@ -34,3 +34,4 @@ data "github_team" "example" {
 * `permission` - the team's permission level.
 * `members` - List of team members (list of GitHub usernames). Not returned if `summary_only = true`
 * `repositories` - List of team repositories (list of repo names). Not returned if `summary_only = true`
+* `repositories_detailed` - List of team repositories (list of `repo_id` and [`role_name`](https://registry.terraform.io/providers/integrations/github/latest/docs/resources/team_repository#permission)). Not returned if `summary_only = true`
```
kfcampbell, line 37: Are there future commits incoming for the specified `summary_only` property? I don't see an implementation.

felixlut: No need, it's already there. The entire [L109-L192 block](https://github.com/integrations/terraform-provider-github/blob/ec307a35525e744f579fa64399f4f1bd2f8ab1a5/github/data_source_github_team.go#L109-L192) only happens if the `summary_only` is set to true. 

My implementation is more or less just a copy-paste of the `repositories` field, which already had the `summary_only` field implemented 😅

kfcampbell: Oh thank you, I missed that piece somehow!

kfcampbell (approved):

PR #1722, opened 2023-06-12
Improve github team resource

bpaquet:
Fix various issue with the `github_team` resource

* During import, set `create_default_maintainer` to false to avoid a diff in the resource. This parameter is not used outside of the `Create` method
* Improve the handling of `parent_team_id`. This parameter is complicated to manage because it can be an ID or a slug, and the current behaviour is not correct (no cast from Int to String when the parameter is Set in the `Read` function). A cleaner way to do things is to have two parameters `parent_team_id` and `parent_team_slug` for example. But I do not want to introduce compatibility issues, so that's why this PR
  * Cast the ID to string when the `Read` function set it
  * Create two new computed params, which store the slug name and the ID in two different parameters
  * Ignore diff on `parent_team_id` if this paremeter match one of the two low level params.

With these modifications, I'm able to import a team, with a parent id, and having no diff in a following plan.

Integration tests are still green.


bitcoinfinancier (approved):

kfcampbell (approved):
Thank you @bpaquet! Confirming all integration tests are passing for me as well.

Issue #679, opened 2021-01-28
`allow_merge_commit`, `allow_rebase_merge`, `allow_squash_merge` fields ignored


Anton-Shutik:
Hi there,

### Terraform Version
```
Terraform v0.14.5
provider registry.terraform.io/integrations/github v4.3.1
```

### Affected Resource(s)
- github_repository

### Terraform Configuration Files
```
resource "github_repository" "rebrand" {
  name                = "rebrand"
  visibility          = "public"
  has_issues          = true
  has_projects        = true
  has_wiki            = true
  has_downloads       = true
  allow_merge_commit  = true
  allow_squash_merge  = true
  allow_rebase_merge  = true
}
```

### Expected Behavior
>terraform state show github_repository.rebrand
```
resource "github_repository" "rebrand" {
    allow_merge_commit     = true
    allow_rebase_merge     = true
    allow_squash_merge     = true
# all the rest fields
}

```

### Actual Behavior
```
resource "github_repository" "rebrand" {
    allow_merge_commit     = false
    allow_rebase_merge     = false
    allow_squash_merge     = false
# all the rest fields
}

```

### Steps to Reproduce
Please list the steps required to reproduce the issue, for example:
1. `terraform import github_repository.rebrand rebrand`

### Important Factoids
I'm trying to import github resources that already exist on GH and the issue is that it ignores `allow_merge_commit`, `allow_rebase_merge`, `allow_squash_merge` fields and sets them with `false`, while they explicitly set to `true` in the config

tibbes:
Hi @Anton-Shutik, I'm having trouble reproducing this problem. Are you running against github.com or GitHub Enterprise? If GitHub Enterprise, which version?

Also, what level of access does the user have to the repository? (the user associated with the GITHUB_TOKEN you are using with terraform, that is)

Anton-Shutik:
Hi, 
I'm running against github.com.
The user has read/write permissions.

majormoses:
I suspect you need to run this as a repository admin at the minimum here and in most cases it really needs to be an org admin. At my org we do scope down the oauth token to prevent certain actions via automation that we deem undesirable from a bug perspective (such as deleting repositories) and require a manual process around that to ensure that we make sure its intended, not in use (checking references) and a backup/archive before we actually clean it up.

[...]

PR #1727, opened 2023-06-15
Add note documenting weird permissions required

kfcampbell:
See #679. The API requires write permissions to properly hydrate GET requests in some github_repository cases. While this is the case in the API, we should document it here to avoid confusion if possible.

### DIFF
--- a/website/docs/r/repository.html.markdown
+++ b/website/docs/r/repository.html.markdown
@@ -11,7 +11,7 @@ This resource allows you to create and manage repositories within your
 GitHub organization or personal account.
 
 ~> Note: When used with GitHub App authentication, even GET requests must have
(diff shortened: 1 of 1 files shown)


website/docs/r/repository.html.markdown
```diff
@@ -10,6 +10,10 @@ description: |-
 This resource allows you to create and manage repositories within your
 GitHub organization or personal account.
 
+~> Note: When used with GitHub App authentication, even GET requests must have
+the `contents:write` permission or else `allow_merge_commit`, `allow_rebase_merge`,
```
kfcampbell, line 14:
```suggestion
the `contents:write` permission or else the `allow_merge_commit`, `allow_rebase_merge`,
```

PR #1750, opened 2023-06-22
Fixes accidental deletion of Repository resource from state

t0yv0:
Resolves #ISSUE_NUMBER

----

### Before the change?

Currently when the user does not authenticate the provider, refreshing a GithubRepository resource drops it from the state, which is unexpected and very confusing.

### After the change?

The suggested fix consults the resource data to infer the appropriate owner to use instead of the empty owner. For pubic repositories this makes the Read method proceed correctly.

### Other information

The suspected root cause of this is: when AnonymousHTTPClient is used, owner == "" and this causes resourceGithubRepositoryRead to issue requests to non-existent URLs such as https://github.com//myrepo and subsequently interpret 404 as a reason to drop the resource from the state.

----

### Pull request checklist

- [x] No

If `Yes`, what's the impact:  

* N/A

### Pull request type

 
 

- Bugfix: `Type: Bug`

----

### DIFF
--- a/github/resource_github_repository.go
+++ b/github/resource_github_repository.go
@@ -593,8 +593,15 @@ func resourceGithubRepositoryCreate(d *schema.ResourceData, meta interface{}) er
 
 func resourceGithubRepositoryRead(d *schema.ResourceData, meta interface{}) error {
 	client := meta.(*Owner).v3client
+
 	owner := meta.(*Owner).name
 	repoName := d.Id()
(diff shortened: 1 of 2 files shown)


kfcampbell (approved):
In a perfect world, we would have something like this for every resource/data source. Right now using the provider unauthenticated is not a good experience. 

Thank you for contributing!

Issue #1746, opened 2023-06-21
[BUG]: Regression with Dependabot Organization secret since v5.27.0


frankywahl:
Hi there,

There seems to be a regression introduced by version 5.28. My best understanding would be something relating to https://github.com/integrations/terraform-provider-github/pull/1718/files#diff-f746541448646d1732c2951600cb385e74f53f48d4d182d878c26d330ef04e0cR54 and how the go-github client moved over to using int64 as ids in v53.0.0

### Terraform Version
Terraform version: 1.5.0
Github Provider Version: v5.28.0 (Also happens in v5.27.0)

### Affected Resource(s)
- github_dependabot_organization_secret

### Panic Output
<details>
<summary> Stack Trace Output </summary>

```bash
panic: interface conversion: interface {} is int, not int64

goroutine 88 [running]:
github.com/integrations/terraform-provider-github/v5/github.resourceGithubDependabotOrganizationSecretCreateOrUpdate(0x10e1be0?, {0x104dbe0?, 0xc00015c4c0?})
	github.com/integrations/terraform-provider-github/v5/github/resource_github_dependabot_organization_secret.go:104 +0x5ad
github.com/hashicorp/terraform-plugin-sdk/helper/schema.(*Resource).Apply(0xc000408500, 0xc000398be0, 0xc00078d0e0, {0x104dbe0, 0xc00015c4c0})
	github.com/hashicorp/terraform-plugin-sdk@v1.17.2/helper/schema/resource.go:326 +0x3e3
github.com/hashicorp/terraform-plugin-sdk/helper/schema.(*Provider).Apply(0xc00004bf80, 0xc0008498c8, 0x12a0d2c?, 0xf?)
	github.com/hashicorp/terraform-plugin-sdk@v1.17.2/helper/schema/provider.go:294 +0x70
github.com/hashicorp/terraform-plugin-sdk/internal/helper/plugin.(*GRPCProviderServer).ApplyResourceChange(0xc00000f248, {0xc0004795e0?, 0x4b8786?}, 0xc0004795e0)
	github.com/hashicorp/terraform-plugin-sdk@v1.17.2/internal/helper/plugin/grpc_provider.go:895 +0x7c5
github.com/hashicorp/terraform-plugin-sdk/internal/tfplugin5._Provider_ApplyResourceChange_Handler({0x1248b80?, 0xc00000f248}, {0x15efd60, 0xc000780180}, 0xc000479570, 0x0)
[...]

PR #1759, opened 2023-06-27
Fix `resourceGithubDependabotOrganizationSecretCreateOrUpdate`

frankywahl:
Fixes a regression used when trying to create/update a dependabot organization secret. List is an `int` so cannot be cast to a different type. We therefore double convert it.

There was a regression introduced by version 5.27. This happened as the go-github client moved over to using int64 as ids in v53.0.0

```hcl
terraform {
  required_providers {
    github = {
      version = "5.27.0"
    }
  }
}

provider "github" {
  owner = var.owner
}

variable "owner" {
  type = string
}

  name        = "example"
  visibility = "public"
}

resource "github_dependabot_organization_secret" "example" {
  secret_name     = "example"
  visibility      = "selected"
  plaintext_value = "anything"
  selected_repository_ids = [
    github_repository.example.repo_id,
  ]
}
```

Resolves https://github.com/integrations/terraform-provider-github/issues/1746

### Pull request checklist

It should fix an issue after the provider get upgraded. 

- [x] No

Notes: 
  * Unsure how to write a test for this change. Guidance would be nice. I ran an integration test and it all worked using the sample code and a local build of the provider pointing the client that had the change.
  * Depends on https://github.com/google/go-github/pull/2817 being released (and client updated)

### DIFF
--- a/github/resource_github_dependabot_organization_secret.go
+++ b/github/resource_github_dependabot_organization_secret.go
@@ -4,11 +4,12 @@ import (
 	"context"
 	"encoding/base64"
 	"fmt"
+	"log"
+	"net/http"
+
 	"github.com/google/go-github/v53/github"
 	"github.com/hashicorp/terraform-plugin-sdk/helper/schema"
 	"github.com/hashicorp/terraform-plugin-sdk/helper/validation"
-	"log"
-	"net/http"
 )
 
 func resourceGithubDependabotOrganizationSecret() *schema.Resource {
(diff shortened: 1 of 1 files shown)


github/resource_github_dependabot_organization_secret.go
```diff
@@ -4,11 +4,12 @@ import (
 	"context"
 	"encoding/base64"
 	"fmt"
+	"log"
+	"net/http"
+
```
frankywahl, line 9: These were auto-fixed by `fmt`

kfcampbell (approved):
Thank you!

Issue #1701, opened 2023-05-25
[BUG] github_branch_protection_v3 tries to delete the contexts every time


julioagos:
### Problem
I created some branch protection rules for some branches. These rules included required checks. After I applied them, it created matching contexts and a new check that is just one of the required checks with an app_id that always changes (First 4 digits remain the same, the rest changes). Now, every time I ran plan, it says that it is going to delete them again. The branch protections are working just as intended but, running plan always generates meaningless noise that get's in the way especially since we have a lot of branch protection rules once we fully migrate it over to Terraform. The applies are all run by `Atlantis` on PRs. This is what an example output of what `terraform state show` looks like for one of the branches:
```
    required_status_checks {
        checks         = [
            "test 1",
            "test 2",
            "test 1:<random-numbers>",
        ]
        contexts       = [
            "test 1",
            "test 2",
        ]
    }
```
Running `terraform plan`  returns something like this
```
      ~ required_status_checks {
          ~ checks         = [
              - "test 1:<random-numbers>",
                # (2 unchanged elements hidden)
            ]
          ~ contexts       = [
              - "test 1",
              - "test 2",
            ]
            # (2 unchanged attributes hidden)
        }
```
There are no error messages at any point and the apply works as intended. It states that all plans have been applied.

### Versions
Terraform: v1.4.4
GitHub Provider: Originally 5.23.0 but also happens in 5.25.1

svalevka:
We have a very similiar issue with 5.25.1 provider.


```
  # module.github_repository.github_branch_protection_v3.branch["redacted_name.master"] will be updated in-place
  ~ resource "github_branch_protection_v3" "branch" {
        id                              = "redacted_name:master"
        # (6 unchanged attributes hidden)

      ~ required_status_checks {
          ~ checks         = [
              - "Naming conventions / Pull request:0",
                # (1 unchanged element hidden)
            ]
          ~ contexts       = [
              - "Naming conventions / Pull request",
            ]
            # (2 unchanged attributes hidden)
        }
```

0x46616c6b:
I have the same problem. Since #1774 (v5.31.0) the `checks` diffs are gone but every plan produce a diff for `contexts` (regardless if they are set or not).

0x46616c6b:
[...]

PR #1774, opened 2023-07-06
Attempted fix for #1657

### LINKED ISSUE #1657: [BUG]: github_branch_protection_v3.required-status-checks.checks - app_id changes all the time
### Describe the need

Hej all,

We are in the middle of moving from the hashicorp/github provider to integrations/github provider and are also doing a major jump from v4.xx to latest 5.xx (as of writing/testing it was 5.22).

We are now facing an issue due to changing from contexts to checks in the required-status-checks block, because the new structure of the status checks is "context:app_id", but app_id changes with every run of our pipeline?

This is of course a bad thing, because it would lead to "unnecessary" changes with every terraform plan/apply.

### SDK Version

_No response_

### API Version

_No response_

### Relevant log output

```shell
Result of first plan:
---------------------
~ resource "github_branch_protection_v3" "main" {
        id                              = "repo-name:main"

      ~ required_status_checks {
          ~ checks         = [
              - "Terraform:824642352016",
                # (1 unchanged element hidden)
            ]
          ~ contexts       = [
              - "Terraform",
            ]
        }

        # (1 unchanged block hidden)
    }

Result of second plan, without any changes in between:
------------------------------------------------------
  ~ resource "github_branch_protection_v3" "main" {
        id                              = "repo-name:main"

      ~ required_status_checks {
          ~ checks         = [
              - "Terraform:824635284400",
                # (1 unchanged element hidden)
            ]
          ~ contexts       = [
              - "Terraform",
            ]
        }

        # (1 unchanged block hidden)
    }

As one can see, the app_id changed from 824642352016 to 824635284400
Another plan then delivered 824642245200.

So the app_id is kind of generic and this makes it hard for us to use this new version.
```

### Code of Conduct

yaakov-h:
Resolves #1657

----

## Behavior

Given the following resource:

```hcl
resource "github_branch_protection_v3" "myrepo-master" {
  repository     = "MyRepo"
  branch         = "master"

  # irrelevant fields omitted

  required_status_checks {
    checks = [
      # format: "<name>:<GitHub App ID>"
      "MyStatusCheck:123456"
    ]
  }
}
```

### Before the change?

* Generating the plan would always show too many checks
* If a GitHub App ID was specified it would not be seen as correct:

```plain
  ~ resource "github_branch_protection_v3" "myrepo-master" {
        id                              = "MyRepo:master"

      ~ required_status_checks {
          ~ checks         = [
              - "MyStatusCheck",
              - "MyStatusCheck:824642865624",
              + "MyStatusCheck:123456",
            ]
          ~ contexts       = [
              - "MyStatusCheck",
            ]
        }
    }

```

### After the change?

* Status checks with GitHub App IDs are seen as valid and do not get recreated.
* There are no extra "checks" elements
* There are extra "contexts" elements, but this is due to server behaviour - see below.

```plain
  ~ resource "github_branch_protection_v3" "myrepo-master" {
        id                              = "MyRepo:master"

      ~ required_status_checks {
          ~ contexts       = [
              - "MyStatusCheck",
            ]
        }
    }

```

The GitHub API returns the `contexts` value, so users will either need to include this deprecated field in their HCL or set `ignore_changes` for `required_status_checks.contexts`, and then ignore the deprecation warning that the field they are ignoring is deprecated.

### Other information
I'm not very experienced in Go so this can probably be done better.

I've also ignored the AppID entirely if it is omitted from JSON or is a JSON `null`, which can happen if a user changes the branch protection policy from requiring a specific app to allowing any GitHub app.

This will probably need some unit tests added too, though #1415 didn't have any when it added checks in the first place.

----

### Pull request checklist

- [X] No

If `Yes`, what's the impact:  

* N/A

### Pull request type

 
 

- Bugfix: `Type: Bug`

----


kfcampbell:
@yaakov-h can you describe how you've tested this manually so I can replicate it? Thank you for contributing!

domsleee:
I believe this is a valid repro: https://github.com/integrations/terraform-provider-github/issues/1657#issuecomment-1522390303

The current behaviour is that this line is printing a pointer to the app ID as an integer:
[> checks = append(contexts, fmt.Sprintf("%s:%d", chk.Context, chk.AppID))](https://github.com/integrations/terraform-provider-github/blob/1acafed6d79e871f146bcd183f2a5cfa36444758/github/resource_github_branch_protection_v3_utils.go#L62C23-L62C23)

From the end-user perspective, the App ID is "changing all the time" described in the issue. Because we are printing the pointer and not the actual value. The suggested fix here looks reasonable to me 👍

yaakov-h:
@kfcampbell I used the existing config I was working on (the resource I posted at the top of the PR) and used Terraform `dev_overrides` to use a custom build with these changes. Running `terraform plan` produced the output shown in "After the change".

kfcampbell (approved):

entropitor:
> The GitHub API returns the contexts value, so users will either need to include this deprecated field in their HCL or set ignore_changes for required_status_checks.contexts, and then ignore the deprecation warning that the field they are ignoring is deprecated.

@yaakov-h The problem is that if you set both `contexts` and `checks` it's trying to apply both and that fails as you can only specify one. Should we instead only look at `checks` if both are set while applying?

yaakov-h:
@entropitor I think if you set both then that would be consumer error, but tbh I'd be in favour of removing contexts entirely as mentioned in #1701 as that also gets rid of the problem of deprecation warnings.

PR #1783, opened 2023-07-09
feat: add ability to downgrade membership when `github_membership` is destroyed

### LINKED ISSUE #1782: [FEAT]: Ability to downgrade org membership to 'member' on `github_membership` destroy
### Describe the need

I manage an org that primarily adds people to GitHub using an IDP. So we don't use the `github_membership` resource primarily to manage org membership. But I'm interested in using `github_membership` to assign some specific members to `admin` and have that managed by Terraform.

`github_membership` allows for a "create or update" resource creation, so if the membership exists before the resource is managed by Terraform then Terraform will just update the membership. But when the resource is destroyed, membership from the org is revoked. In my case, I'd prefer if "resource destroy" just downgraded the member back to `member`.

The resource definition could look like this:
```
locals {
  member_permissions = {
    "jsifuentes": "admin"
  }
}

resource "github_membership" "member_role" {
  for_each = local.member_permissions

  username             = each.key
  role                 = each.value
  downgrade_on_destroy = true
}

```

### SDK Version

_No response_

### API Version

_No response_

### Relevant log output

_No response_

### Code of Conduct

jsifuentes:
Resolves #1782

----

### Before the change?

* github_membership allows for a "create or update" resource creation, so if the membership exists before the resource is managed by Terraform then Terraform will just update the membership. When the resource is destroyed, membership from the org is always revoked.

### After the change?

* You can now add `downgrade_on_destroy = true` to a `github_membership` resource. Instead of revoking org membership, the membership role will be set to 'member' when the resource is destroyed.

----

## Additional info

* I added a new test case specifically to test this downgrade scenario. The function `testAccGithubMembershipConfigDowngradable` sets the role to `admin` and sets the `downgrade_on_destroy` to true. Since the Destroy no longer completely removes the membership during this test case, the membership is cleaned up during the `CheckDestroy`.

### Pull request checklist

- [x] No

If `Yes`, what's the impact:  

* N/A

### Pull request type

----


github/resource_github_membership.go
```diff
@@ -38,6 +38,12 @@ func resourceGithubMembership() *schema.Resource {
 				Type:     schema.TypeString,
 				Computed: true,
 			},
+			"downgrade_on_destroy": {
+				Type:        schema.TypeBool,
+				Optional:    true,
+				Default:     false,
+				Description: "Instead of removing the member from the org, you can choose to downgrade their membership to 'member' when this resource is destroyed.",
```
kfcampbell, line 45:
```suggestion
				Description: "Instead of removing the member from the org, you can choose to downgrade their membership to 'member' when this resource is destroyed. This is useful when wanting to downgrade admins while keeping them in the organization",
```

github/resource_github_membership.go
```diff
@@ -126,8 +132,19 @@ func resourceGithubMembershipDelete(d *schema.ResourceData, meta interface{}) er
...
 
-	_, err = client.Organizations.RemoveOrgMembership(ctx,
-		d.Get("username").(string), orgName)
+	username := d.Get("username").(string)
+	downgradeOnDestroy := d.Get("downgrade_on_destroy").(bool)
+	downgradeTo := "member"
+
+	if downgradeOnDestroy {
+		log.Printf("[INFO] Downgrading '%s' membership for '%s' to '%s'", orgName, username, downgradeTo)
+		_, _, err = client.Organizations.EditOrgMembership(ctx, username, orgName, &github.Membership{
```
kfcampbell, line 141: What happens if a "member" is destroyed while `downgrade_on_destroy` is true?

jsifuentes: If a membership was revoked outside of Terraform and then afterward removed from Terraform, the code at the time you commented would have re-invited the user back to the organization.

Now I load the membership status and check if the user is still in the org before I call `EditOrgMembership`.

Thanks.

kfcampbell (approved):

Issue #922, opened 2021-10-04
Unable to set branch names/patterns when `custom_branch_policies` is `true`.


Omicron7:
### Affected Resource(s)
- github_repository_environment

### Terraform Configuration Files
```hcl
resource "github_repository_environment" "environment" {
  repository  = "repository_name"
  environment = "testing"

  deployment_branch_policy {
    protected_branches     = false
    custom_branch_policies = true
  }
}
```
In the GitHub console, when `custom_branch_policies` is set to `true`, it allows setting one or more Branch names/patterns to match against. It doesn't seem like this is possible in Terraform. I'm unable to find any other GitHub resource that would allow managing that list per environment. Am I missing something?

bishwash-delphix:
I was looking for the same feature... is it possible to "Add deployment branch rule"?

gilfthde:
Same for me. I also need to set a branch name/pattern and don't now how to do it.

figadore:
Agree, seems like there should be a way to specify the patterns that can deploy to the environment. Maybe something like

```hcl
resource "github_repository_environment" "default" {
  repository  = var.repo
  environment = "${var.env_id}"
  reviewers {
    teams = var.reviewers_team
    users = []
  }
  deployment_branch_policy {
    protected_branches          = false
    custom_branch_policies = true
    custom_branch_regex = ["main"]
  }
}
```
?

It doesn't look like the go-github package supports this at the moment 
https://github.com/google/go-github/blob/7add3a89d10dbdcbbfab71036aa9c12e5ee997c8/github/repos_environments.go#L42

StephenWithPH:
I added a comment in GitHub's product discussions to try to elevate this to their PM team: https://github.com/github-community/community/discussions/14190#discussioncomment-3065590

StephenWithPH:
https://github.com/orgs/community/discussions/14190#discussioncomment-3931426 points to https://docs.github.com/en/rest/deployments/branch-policies#create-a-deployment-branch-policy, which suggests that there is now an API endpoint for this.

kfcampbell:
If that's true, that's great news. I'd be happy to review/test/merge a PR to cover this behavior!

mBlomsterberg:
Would really like to see this implemented. Having a custom_branch_policies attribute feels a bit misleading since it's only a bool.

romikoops:
Really disappointing that we still do not have it. If we automate something, it should have 100% automation, not 99 and etc.

[...]

PR #1530, opened 2023-02-02
Add deployment policy resource

ilmax:
- Add the initial code to manage the resource
- Add sample configuration used to test it
- Add test

Resolves #922 

----

### Before the change?
Not possible to set environment deployment branch policies

### After the change?

You can to set environment deployment branch policies

### Other information

N/A

----

### Pull request checklist
- [x] Tests for the changes have been added (for bug fixes / features)
- [x] Docs have been reviewed and added / updated if needed (for bug fixes / features)
- [x] Added the appropriate label for the given change

### Does this introduce a breaking change?

Please see our docs on [breaking changes](https://github.com/octokit/.github/blob/master/community/breaking_changes.md) to help!

- [ ] Yes (Please add the `Type: Breaking change` label)
- [x] No

If `Yes`, what's the impact:  

* N/A

### Pull request type

 
 

Please add the corresponding label for change this PR introduces:
- Bugfix: `Type: Bug`
- Feature/model/API additions: `Type: Feature`
- Updates to docs or samples: `Type: Documentation`
- Dependencies/code cleanup: `Type: Maintenance`

----

PR #1799, opened 2023-07-19
fix: Github Environments Policy feature causing the provider to produce inconsistent result

edheliel:
Resolves https://github.com/integrations/terraform-provider-github/issues/922
Stale initial PR: https://github.com/integrations/terraform-provider-github/pull/1530
----

## Behavior
This PR is based on the now appearing to be stale PR https://github.com/integrations/terraform-provider-github/pull/1530. 
- Fix the issue with the provider becoming inconsistent
  - It was cause by double encoding which made the environment name that contains a `/` wrong therefore not existing
  - use `url.PathEscape` since the github API operations use path parameters rather than query

### Before the change?
- `envName` was encoded 2 time which cause the Environment Branch Policy resource Read operation to not be able to find the file that was just created. That caused the provider to produce an inconsistent result
- Usage of `url.QueryEscape` instead of `url.PathEscape` 

### After the change?
- delete and read operation don't try to encode the `envName` a second time as they are receiving it already encoded and a second encoding breaks the name and url path
- use `url.PathEscape`

## Additional info

Most of the work was done on the original PR which seem to have become stale. 
My motivation for creating this one is to unblock a feature that will allow for improvement in security when it comes to Github Actions Workflows and etc

### Pull request checklist

No

- [x] No

### Pull request type
Bug fix

----


kfcampbell (approved):
Thank you @edheliel and @ilmax for digging in here! I'll get this merged and released shortly.

PR #1806, opened 2023-07-21
Enforce valid chars in repo name

### LINKED ISSUE #1215: `github_repository` with a `name` attribute that contains spaces or special characters fails
Hi,

As per title—
When the repository name has spaces or special characters, only the original creation operation in the provider code will work, subsequent edits will fail.

_Context:_ `github_repository` does a `POST` request (e.g.`POST /orgs/<org>/repos`) to create the repo followed by one or more `PATCH` requests (`PATCH /repos/<org>/<repo>`) to set other options.

The `POST` method will take the name as-is but will return a repo with the spaces and special characters in the name replaced by dashes (`-`) (e.g. `Repo Test` -> `Repo-Test`).

The subsequent `PATCH` request(s) will send the original `name` in its payload (e.g. `Repo Test`), which the API will reject with a `422 Unprocessable Entity`:

```json
{
  "message": "Validation Failed",
  "errors": [],
  "documentation_url": "https://docs.github.com/rest/reference/repos#update-a-repository"
 }
```

### Terraform Version

```
Terraform v1.2.4
on linux_amd64
+ provider registry.terraform.io/integrations/github v4.26.1
```

### Affected Resource(s)

`github_repository`

```hcl
resource "github_repository" "repo-test-space" {
  name        = "Repo Test"
  description = "Repo Test Description"
}
```

### Debug Output

https://gist.github.com/pndurette/b33d98cddd97892f98fb02738e61c788

### Panic Output

N/A

### Expected Behavior

It should create the `github_repository` resource and it should not be tainted.

Either:
* The provider should limit the repository `name` to what the API does with it (`^[a-zA-Z0-9_.]+$` afaik)
* The provider should "transform" the given name (like the API would) and use that in all repository operations (i.e. `client.Repositories.CreateFromTemplate()`, `client.Repositories.Create()`, `client.Repositories.Edit()`).
* The provider should get (read) the repository from the API before the update that comes just after create?
* ...?

### Actual Behavior

A Terraform _create_ or _update_ on a `github_repository` resource when the `name` contains spaces or special characters fails.

```hcl
Error: PATCH https://api.github.com/repos/<org>/<repo>: 422 Validation Failed []

  with github_repository.repo-test-space,
  on github_repos.tf line 93, in resource "github_repository" "repo-test-space":
  93: resource "github_repository" "repo-test-space" {
```

Which comes from this [`PATCH` request](https://docs.github.com/en/rest/repos/repos#update-a-repository):
```
PATCH /repos/<org>/Repo-Test HTTP/1.1
Host: api.github.com
User-Agent: go-github
Content-Length: 387
Accept: application/vnd.github.baptiste-preview+json, application/vnd.github.nebula-preview+json
Content-Type: application/json
Accept-Encoding: gzip

{
 "name": "Repo Test",
 "description": "Repo Test Description",
 "homepage": "",
 "auto_init": false,
 "allow_rebase_merge": true,
 "allow_squash_merge": true,
 "allow_merge_commit": true,
 "allow_auto_merge": false,
 "delete_branch_on_merge": false,
 "archived": false,
 "has_issues": false,
 "has_wiki": false,
 "has_projects": false,
 "has_downloads": false,
 "is_template": false,
 "license_template": "",
 "gitignore_template": ""
}
```

Notice `Repo-Test` in the URL and `Repo Test` in the payload.

### Steps to Reproduce

1. Have an auth method configured to the right GitHub owner, having the `repo` scope
2. `terraform apply`

### Important Factoids

* Workaround is to set `github_repository` with a `name` that matches `^[a-zA-Z0-9_.]+$`
* After the `apply` failure, changing the `name` to another one with illegal characters will _work_ for some reason, but will fail again on the next `apply`.

### References

Are there any other GitHub issues (open or closed) or Pull Requests that should be linked here?

- #850 
- #1001

btlogy:
Resolves #1215

----

### Before the change?

* The name of the repository can be a string long from 1 to 100 containing any type of character.

### After the change?

* The name of the repository can be a string long from 1 to 100 containing only valid characters (as per API).

### Pull request checklist

- [x] Yes: Adding/altering a validation rule to an existing parameter
- [ ] No

----


btlogy:
I guess I should try to add a few test for this validation...
Too bad I can not find any for the current validation (name length between 1 and 100) :-(

btlogy:
I wish I could have linked this regexp with some API documentation, but I've not found any!?
Also: sorry I've mentioned "dash" in place of "hyphen" in my commit (I do not think it is worth a force push, right?).

btlogy:
About this being a "breaking change":
According to #1215, it looks to me like one could have ever successfully created and manage repositories containing space in their name using this provider.
Maybe not breaking anything after all...

kfcampbell (approved):

PR #1820, opened 2023-08-03
Documentation enhancements for `github_repository_deployment_branch_policy` resource

calebplum:
### Before the change?
1. The documentation example is missing the required `depends_on` meta-argument. Without this, Terraform will often try to deploy the `github_repository_deployment_branch_policy` before the corresponding `github_repository_environment` which results in an ambiguous 404 error being returned from the API.

2. A separate (but equally ambiguous) 404 is returned from the API when the `github_repository_environment` referenced by the `github_repository_deployment_branch_policy` resource has `custom_branch_policies` set to `false`. I believe the design pattern is to catch this error in the Terraform provider rather than rely on the API returning a descriptive error (which it doesn't do). This can catch one off-guard since the documentation says the `custom_branch_policies` must be set to `true`, but doesn't mention the somewhat unusual behaviour of the API returning a 404 (rather than Terraform complaining about the attribute) if this attribute is set incorrectly. This personally caused me a few hours of frustration, so it would be good to make others aware of this behaviour.

3. The Terraform code in the example hadn't been run through `terraform fmt`.

### After the change?
1. The required `depends_on` attribute has been added to the documentation example to prevent the branch policy being created before the corresponding environment.

2. The `environment_name` argument's description has been enhanced to describe the behaviour that occurs if `deployment_branch_policy.custom_branch_policies` is not set to `true`.

3. The examples have been run through `terraform fmt`.

### Pull request checklist

- [ ] Yes
- [x] No

----

### DIFF
--- a/website/docs/r/repository_deployment_branch_policy.html.markdown
+++ b/website/docs/r/repository_deployment_branch_policy.html.markdown
@@ -23,9 +23,11 @@ resource "github_repository_environment" "env" {
 }
 
 resource "github_repository_deployment_branch_policy" "foo" {
-  repository = "my_repo"
+  depends_on = [github_repository_environment.env]
+
+  repository       = "my_repo"
   environment_name = "my_env"
-  name = "foo"
+  name             = "foo"
 }
 ```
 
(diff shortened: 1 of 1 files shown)


nickfloyd (approved):
Thanks for the contributions! ❤️

PR #1832, opened 2023-08-08
Add owner attribute for repository resource

### LINKED ISSUE #1702: [FEAT]: github_repository select organization to use in Enterprise
### Describe the need

We are leveraging GitHub Enterprise Managed Users and have multiple organizations in the enterprise.

We are trying to create repositories for specific organizations, but currently do not see an argument in github_repository for setting the organization.

Potentially something like setting the owner field below:
```
  name        = "example"
  owner         = "my-org"

  visibility = "internal"
}
```

I believe a workaround is we could create a separate Terraform pipeline for each organization and set the owner in the provider, but it would require a maintaining the terraform state for each organization.

Would the workaround be the recommended approach? Thoughts on this as a feature?

### SDK Version

_No response_

### API Version

_No response_

### Relevant log output

_No response_

### Code of Conduct

tdabasinskas:
An attempt to fix #1702.

The lack of ability to specify `owner` for an individual repository resource makes this provider pretty much unusable in a GHES environment containing over 300 organizations – having a separate provider for every single organization is probably not a great idea.

I'm not sure my solution is ideal – the current code is quite biased on the owner being resolved automatically, so it's a bit challenging to add support for an explicit owner without introducing a breaking change.

I've tested this quite extensively manually – seems to be working as expected. If someone could help with adding some tests to confirm this, I would appreciate. Also, these changes should be applied to all `repository_` resources, but prior to doing that, I wanted to get some feedback.

----

### Before the change?
* `repository` resource uses `owner` specified at the provider level.

### After the change?
* `repository` resource supports an optional `owner` argument.
* If the argument is provided, the resource is managed under that specific owner.
* If the argument is not provided, the owner defined at the provider level is used.

### Pull request checklist

- [ ] Yes
- [X] No

----


website/docs/r/repository.html.markdown
```diff
@@ -55,6 +55,9 @@ The following arguments are supported:
 
 * `name` - (Required) The name of the repository.
 
+* `owner` - (Optional) The GitHub organization or user the repository is owned by.
+ Defaults to the owner/organization specified in the provider configuration.
+
```
kfcampbell, line 60:
```suggestion
* `owner` - (Optional) The GitHub organization or user the repository is owned by.
Defaults to the owner/organization specified in the provider configuration.
If neither are given, this field defaults to the authenticated user. 
```

kfcampbell:
This seems like a good approach to the change to me, though I also feel somewhat uncomfortable without integration tests.

nickfloyd:
Hey @tdabasinskas, just a quick follow up.  Is this something that you are interested in wrapping up? 
I think the remaining items are: 
* @kfcampbell left a quick suggestion on the docs
* Some integration tests if you get the chance

Thanks again for making things better for the community here! ❤️

tdabasinskas:
Hi @nickfloyd,

Sorry for dropping a ball on this.

I've updated the docs as per @kfcampbell proposal. However, I'm not sure when I could do the integration tests, and, more importantly, add this to all the rest of repository-related resources. Being able to set `owner` only for `repository` resource but still requiring dedicated providers for all other resources, is not ideal 😕

kfcampbell:
> Being able to set owner only for repository resource but still requiring dedicated providers for all other resources, is not ideal 😕

I definitely agree with this take, and this is something the provider can stand to do a lot better on. We'd definitely be receptive to anybody wanting to continue this work.

avidspartan1:
Hey every, just a heads up - I merged this branch into my own and am continuing this work there. Feedback welcome!

PR #1780, opened 2023-07-07
Upgrade to Terraform Plugin SDK v2

kfcampbell:
The [most recent release failed](https://github.com/integrations/terraform-provider-github/actions/runs/5469879925/jobs/9959316344) due to some lingering `go mod tidy` issues we've seen. This can be more permanently resolved by making a concerted effort to update Terraform provider SDK versions. 

The upgrade has been performed by following the steps [here](https://developer.hashicorp.com/terraform/plugin/sdkv2/guides/v2-upgrade-guide) and using the `tf-sdk-migrator` tooling provided by Hashicorp.

PR #1833, opened 2023-08-11
GitHub org ignore archived repos

### LINKED ISSUE #1826: [FEAT]: allow filtering of archived repos in github_organization
### Describe the need

Currently the `github_organization` datasource exposes all the repos of the org in the [`repositories`](https://registry.terraform.io/providers/integrations/github/latest/docs/data-sources/organization#repositories) attribute. I would like to make it possible to filter these repos based on whether they are ``archived`` or not. While this is possible to do afterwards, you'd have to make a second API call via a [``github_repository``](https://registry.terraform.io/providers/integrations/github/latest/docs/data-sources/repository#archived) datasource, which I want to avoid. 

My initial idea would be to add a `ignore_archived_repos` argument (defaulted to ``false`` if not explicitly set to not break current usage), which simply would simply filter the list returned in the [`repositories`](https://registry.terraform.io/providers/integrations/github/latest/docs/data-sources/organization#repositories) attribute to only return repos that aren't ``archived``. I could also add a separate attribute called `non-archived-repos` or similar, returning both the filtered and unfiltered lists. I'm leaning towards the former approach, but I could be convinced to go the other way

And I could work on this.  

### SDK Version

_No response_

### API Version

_No response_

### Relevant log output

_No response_

### Code of Conduct

felixlut:
Resolves #1826

----

### Before the change?

* The `repositories` attribute of a `github_organization` always included all repos in the entire Organziation, including archived ones. 

### After the change?

* The `repositories` attribute of a `github_organization` can now selectively ignore archived repos for the `repositories` attribute. 

### Pull request checklist

No

- [ ] Yes
- [x] No

----


github/data_source_github_organization_test.go
```diff
@@ -54,4 +55,69 @@ func TestAccGithubOrganizationDataSource(t *testing.T) {
...
+			output "should_be_false" {
+				value = contains(data.github_organization.skip_archived.repositories, github_repository.archived.full_name)
+			}
+			output "should_be_true" {
+				value = contains(data.github_organization.all_repos.repositories, github_repository.archived.full_name)
+			}
+		`, randomID, testOrganization, testOrganization)
+
+		check := resource.ComposeTestCheckFunc(
+			resource.TestCheckOutput("should_be_false", "false"),
```
felixlut, line 92: @kfcampbell this test is based on the ones I found [here](https://github.com/ovh/terraform-provider-ovh/blob/95e83bb4df06b18b8cc5b92dcb42981e153535be/ovh/data_cloud_project_users_test.go#L44), where it appears that ``TestCheckOutput()`` handles the conversions of bools and strings, i.e., `false` --> `"false"`. That code uses `terraform-plugin-sdk/v2`, while the GitHub provider uses `v1`. Is there a reason for not using v2 here?

I think I read somewhere in the docs that you guys don't want the developer community to not mess with bumping dependencies ourselves, which is why I'm asking 😄

kfcampbell: The reason is due to lack of prioritization of #1780, not anything super intentional. I'm fine adding v2 as a dependency since we'll need to cut over to it anyways.

felixlut: So how do we move forward here? Should I wait until #1780 is merged, or import and use `v2` in `data_source_github_organization_test.go` (if that is even possible, haven't dealt much with golang dependencies and vendoring before)? I don't care about getting this in fast, so if it takes another couple of weeks for #1780, I'm fine to wait

kfcampbell: Let's move forward with your second option. The approach you describe is indeed possible and I think it's unlikely #1780 is going to be merged in the near future.

felixlut: I played around a bit with option 2, but I think using both v1 and v2 of the sdk is not compatible, at least not without doing some real weird stuff. 

First of you need to use a version of the sdk before `v2.10.1` (which released in dec 2021). Using a later version requires a bump of `terraform-exec`, to a version above `v0.16.0`, [which deprecated the `tfinstall` package](https://github.com/hashicorp/terraform-exec/blob/f8daf3dc8e047bd082f80a0918488f06e3ce233b/CHANGELOG.md?plain=1#L79C1-L99) that v1 of the sdk depends on

I just go stuck with the tests using the [`testAccProviders` (defined in `github/provider_test.go`)](https://github.com/integrations/terraform-provider-github/blob/52489e144754805407365053e0c15f620ce8f39a/github/provider_test.go#L12-L14), which uses v1. They cannot be used for the v2 tests in `github/data_source_github_organization_test.go`.

kfcampbell: Oh gross...thanks for doing the legwork on that! I'm :+1: on waiting for now.

kfcampbell: If you're interested in picking up #1780 and fixing the build errors, please feel free! You should be able to create a new PR off of my branch if you desire.

felixlut: @kfcampbell that thought did cross my mind, especially since the time I spent working around it could have been enough to just make it work properly. I learnt a lot about go dependencies at least 😅 

I make take a look later

kfcampbell: Haha, I'm glad you were able to take something positive out of it! Should you end up deciding to do so, I'd be happy to look at it.

felixlut: With the new release this actually works, I've tested it locally.

kfcampbell (approved):
I'm glad the SDK bump unblocked this work! Confirming that the added integration test is passing for me.

PR #1846, opened 2023-08-17
[FEAT]: Add github_repository_topics resource

### LINKED ISSUE #1845: [FEAT]: Add github_repository_topics resource
### Describe the need

In my org we currently don't have all repos inside of Terraform, but we want to be able to control the repository topics on them to enforce certain standards. Currently setting the topics via Terraform is only possible when using the `github_repository` resource, which is too intrusive for my usecase (unless I make some magic with the [`ignore_changes`](https://developer.hashicorp.com/terraform/language/meta-arguments/lifecycle#ignore_changes) argument). 

Another use-case is also that you can use this resource to set some topics from a platform prospective, while still allowing your users the freedom to add their own topics on top. See example below

```terraform
data "github_repository" "test" {
  name = "my-cool-repo"
}

resource "github_repository_topics" "test" {
    repository    = data.github_repository.test.name
    # this result in all topics that was manually put on the repo + the standard ones the platform team wants
    topics        = concat(data.github_repository.test.topics, [lower(data.github_repository.test.primary_language), "owner-team", "product-name"])
}

output "name" {
  value = github_repository_topics.test.topics
}

```

There are API endpoints for doing this: https://docs.github.com/en/enterprise-server@3.10/rest/repos/repos?apiVersion=2022-11-28

I can work on it

### SDK Version

_No response_

### API Version

_No response_

### Relevant log output

_No response_

### Code of Conduct

felixlut:
Resolves #1845

----

### Before the change?

* Repository topics can only be set via the `github_repository` resource

### After the change?

* Now it's also possible to work directly on the repository topics without touching anything else about the repository

### Pull request checklist

This is not a breaking change, but usage of the `topics` attribute in `github_repository` and `github_repository_topics` cannot be used simultaneously

- [ ] Yes
- [x] No

----


website/docs/r/repository.html.markdown
```diff
@@ -116,6 +116,8 @@ initial repository creation and create the target branch inside of the repositor
 
 * `topics` - (Optional) The list of topics of the repository.
 
+~> Note: This attribute is not compatible with the ``github_repository_topics`` resource. Use one of them
```
kfcampbell, line 119: Can you please add a note here that describes when you would want to use one resource vs. the other?

website/docs/r/repository_topics.html.markdown
```diff
@@ -0,0 +1,43 @@
+---
```
kfcampbell, line 1: Do you mind linking to this doc in the `github.erb` file so it's discoverable?

website/docs/r/repository.html.markdown
felixlut: Done! I've expanded the docs a bit, and linked to #1845, which shows how I will use it!

website/docs/r/repository_topics.html.markdown
felixlut: Done! Sorry for missing that!

felixlut:
@kfcampbell ping

website/docs/r/repository.html.markdown
```diff
@@ -116,6 +116,8 @@ initial repository creation and create the target branch inside of the repositor
 
 * `topics` - (Optional) The list of topics of the repository.
 
+~> Note: This attribute is not compatible with the ``github_repository_topics`` resource. Use one of them. `github_repository_topics` is only meant to be used if the repository itself is not handled via terraform, for example if it's only read as a datasource (see [issue #1845](https://github.com/integrations/terraform-provider-github/issues/1845)).
```
kfcampbell, line 119:
```suggestion
~> Note: This attribute is not compatible with the `github_repository_topics` resource. Use one of them. `github_repository_topics` is only meant to be used if the repository itself is not handled via terraform, for example if it's only read as a datasource (see [issue #1845](https://github.com/integrations/terraform-provider-github/issues/1845)).
```

kfcampbell (approved):
Thank you for contributing! You've been on fire recently.

PR #1873, opened 2023-09-01
chore(docs): use consistent spacing in github_organization docs

garnertb:
### Before the change?

* Some GitHub org attributes are being rendered as sub-attributes of the users list.

image

### After the change?

* I have updated the spacing in the markdown to be consistent with the other attributes.

### Pull request checklist

No

- [ ] Yes
- [ ] No

----


kfcampbell (approved):
Good catch, thank you! 

It's weird we use that extra space before the list items; I wonder why that is.

PR #1875, opened 2023-09-05
Check repository_deployment_branch_policy for 404 in read

### LINKED ISSUE #1853: [Bug] Terraform refresh fails if deployment environment's branch policy has been deleted
### Describe the need

If a repository deployment environment has a branch policy rule that's tracked in Terraform state and that branch policy is deleted outside of Terraform, subsequent Terraform refresh operations will fail as the Github API returns a 404.

Rather than shitting the bed, the Github provider should handle this 404 gracefully and re-create the deleted branch policy.

Re-creating the branch policy manually doesn't fix the issue. The only way I've found to resolve it is to manually remove the `github_repository_deployment_branch_policy` resource from the Terraform state with `terraform state rm` and re-run the refresh operation.

### Steps to reproduce:
1. Deploy a `github_repository_deployment_branch_policy` resource
2. Manually delete the branch policy in the Github console
3. Run any of a `terraform plan`, `terraform apply` or `terraform refresh` operation. The operation will fail with an error similar to:
```
│ Error: GET https://api.github.com/repos/redacted/redacted/environments/redacted/deployment-branch-policies/redacted: 404 Not Found []
│
│   with github_repository_deployment_branch_policy.main,
│   on main.tf line 1, in resource "github_repository_deployment_branch_policy" "main":
│  1: resource "github_repository_deployment_branch_policy" "main" {
│
```

This completely breaks any project that implements branch policies and unnecessarily requires manual remediation to fix it.

### SDK Version

5.32

### API Version

_No response_

### Relevant log output

```shell
As above
```

### Code of Conduct

o-sama:
Resolves #1853

----

### Before the change?

* Deleting this resource outside of TF currently breaks the configuration.

### After the change?

* The provider should be able to recognize when this resource no longer exists.

### Pull request checklist

Neither of the above is required, there's no functionality change outside of checking for a 404 status when reading the resource.

- [ ] Yes
- [x] No

----

### DIFF
--- a/github/resource_github_repository_deployment_branch_policy.go
+++ b/github/resource_github_repository_deployment_branch_policy.go
@@ -2,6 +2,8 @@ package github
 
 import (
 	"context"
+	"log"
+	"net/http"
 	"strconv"
 
 	"github.com/google/go-github/v54/github"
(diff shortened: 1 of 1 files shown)


kfcampbell:
Is this the desired behavior? Perhaps when deleted outside of GitHub, the provider should reinstate the resource as described in the HCL. Can you talk to me a little more about what prompted this PR and the scenarios you're hoping to avoid?

o-sama:
That is the intended behaviour, previously the read would just return the error if it isn't a `304 not found` and so nothing would happen. The goal here is to remove the resource from state if it's deleted outside of Terraform, which would then recreate on the next apply to match the code. 

I did forget to remove the old error checks though which should not be hit anymore so I'll do that in a commit in a few mins.

kfcampbell (approved):
@o-sama thank you for the context and explanation!

PR #1884, opened 2023-09-11
Fix typo in org rulesets.repository_id

o-sama:
----

### Before the change?

A typo was introduced when this feature was added in #1808.

* Resource `repository_organization_ruleset.conditions.repository_id` has field `inlcude`.
* Docs have the same typo.

### After the change?

* Typo fixed in both places.

### Pull request checklist

- [x] Yes
- [ ] No

----

The breaking change is in the organization ruleset resource, a field name is being changed so users must reference the new name to move to the new version of the provider once this is released. @kfcampbell let me know if you'd like me to instead support two versions of the field, it's doable but in my opinion there hasn't been much time to adopt this feature and so we should release a patch instead. I'll be ready to add support for both either way.

### DIFF
--- a/github/resource_github_organization_ruleset.go
+++ b/github/resource_github_organization_ruleset.go
@@ -118,7 +118,7 @@ func resourceGithubOrganizationRuleset() *schema.Resource {
 							ConflictsWith: []string{"conditions.0.repository_id"},
 							Elem: &schema.Resource{
 								Schema: map[string]*schema.Schema{
-									"inlcude": {
+									"include": {
 										Type:        schema.TypeList,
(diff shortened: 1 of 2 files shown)


timja (approved):

o-sama:
BTW, @kfcampbell, if we do release a patch as is and break the use of `inlcude`, would it be feasible to do so as quick as possible to avoid more people using the version with the typo? Again, my bad on this one, I missed it when working on the feature.

nickfloyd (approved):
Thank you for the correction here.  We'll drop a major version.

PR #1885, opened 2023-09-11
Revert "Fix typo in org rulesets.repository_id"

nickfloyd:
Reverts integrations/terraform-provider-github#1884

We'll need to drop a major release and use MigrationFuncs [pattern](https://github.com/integrations/terraform-provider-github/blob/a361b158a645282a238cdefa5c40ae950556a4a7/github/migrate_github_repository.go#L20)  between the schema changes so that state doesn't get messed up for consumers.

I pulled the trigger on this one too quickly.  Apologies for the hassle here. I'll reopen #1884 and put a vNext label on it.

cc: @kfcampbell, @o-sama

### DIFF
--- a/github/resource_github_organization_ruleset.go
+++ b/github/resource_github_organization_ruleset.go
@@ -118,7 +118,7 @@ func resourceGithubOrganizationRuleset() *schema.Resource {
 							ConflictsWith: []string{"conditions.0.repository_id"},
 							Elem: &schema.Resource{
 								Schema: map[string]*schema.Schema{
-									"include": {
+									"inlcude": {
 										Type:        schema.TypeList,
 										Required:    true,
 										Description: "Array of repository names or patterns to include. One of these patterns must match for the condition to pass. Also accepts `~ALL` to include all repositories.",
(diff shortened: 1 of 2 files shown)


o-sama:
@nickfloyd no problem, the mistake's mine. I'll get the code down for it tonight or tomorrow morning.

kfcampbell (approved):
Thank you @nickfloyd! Apologies for my poor communication about the next version release. 

@o-sama when we make breaking changes like this, first we should introduce a new schema element that's spelled correctly and mark the misspelled property as `Deprecated` with a message about switching to the new element. 

When that's released and has a chance to "bake" for a while, then we can add a state migration function like Nick mentioned and officially cut a major version. I know this seems like a really small thing we can easily fix, but it can cause a big hassle with consumers who have existing state that this will disrupt. Hashicorp [recommends](https://developer.hashicorp.com/terraform/plugin/best-practices/versioning#versioning-specification) we don't cut breaking changes more than once per year as a result of the state hassles. 

@o-sama I love the enthusiasm and we definitely appreciate all the help so far. If you'd like to push this along, please feel free to start making changes as described above!

o-sama:
@kfcampbell I totally agree and definitely don't want to be breaking people's existing work using what's been released. I was just hoping since the release was just a few days ago that we could get away with it 😅 

I'll get the ball rolling on this with the deprecation of the typo. Thanks for the tips also!

PR #1887, opened 2023-09-12
Fixes for org and repo rulesets

### LINKED ISSUE #1886: [BUG]: import github_organization_ruleset crashed provider
### Expected Behavior

should work expected to import github_organization_ruleset

### Actual Behavior

Stack trace from the terraform-provider-github_v5.36.0 plugin:

[signal SIGSEGV: segmentation violation code=0x2 addr=0x0 pc=0x102f0d764]

goroutine 83 [running]:
github.com/integrations/terraform-provider-github/v5/github.flattenConditions(0x140009daa38, 0x1)
        github.com/integrations/terraform-provider-github/v5/github/respository_rules_utils.go:180 +0x314
github.com/integrations/terraform-provider-github/v5/github.resourceGithubOrganizationRulesetRead(0x14000c80bd0, {0x10331a260?, 0x14000224f00?})
        github.com/integrations/terraform-provider-github/v5/github/resource_github_organization_ruleset.go:492 +0x5bc
github.com/hashicorp/terraform-plugin-sdk/helper/schema.(*Resource).RefreshWithoutUpgrade(0x14000492be0, 0x14000120500, {0x10331a260, 0x14000224f00})
        github.com/hashicorp/terraform-plugin-sdk@v1.17.2/helper/schema/resource.go:470 +0x1ac
github.com/hashicorp/terraform-plugin-sdk/internal/helper/plugin.(*GRPCProviderServer).ReadResource(0x1400011f638, {0x1400010d020?, 0x0?}, 0x1400010d020)
        github.com/hashicorp/terraform-plugin-sdk@v1.17.2/internal/helper/plugin/grpc_provider.go:535 +0x338
github.com/hashicorp/terraform-plugin-sdk/internal/tfplugin5._Provider_ReadResource_Handler({0x1035460e0?, 0x1400011f638}, {0x1035a8f08, 0x14000909c80}, 0x14000c80700, 0x0)
        github.com/hashicorp/terraform-plugin-sdk@v1.17.2/internal/tfplugin5/tfplugin5.pb.go:3269 +0x174
google.golang.org/grpc.(*Server).processUnaryRPC(0x140003a8000, {0x1035ad108, 0x14000592d00}, 0x14000a117a0, 0x14000881770, 0x103e3c190, 0x0)
        google.golang.org/grpc@v1.55.0/server.go:1337 +0xc7c
google.golang.org/grpc.(*Server).handleStream(0x140003a8000, {0x1035ad108, 0x14000592d00}, 0x14000a117a0, 0x0)
        google.golang.org/grpc@v1.55.0/server.go:1714 +0x840
google.golang.org/grpc.(*Server).serveStreams.func1.1()
        google.golang.org/grpc@v1.55.0/server.go:959 +0x88
created by google.golang.org/grpc.(*Server).serveStreams.func1
        google.golang.org/grpc@v1.55.0/server.go:957 +0x170

Error: The terraform-provider-github_v5.36.0 plugin crashed!

This is always indicative of a bug within the plugin. It would be immensely
helpful if you could report the crash with the plugin's maintainers so that it
can be fixed. The output above should help diagnose the issue.

### Terraform Version

terraform core version 1.5.7
github terraform version: terraform-provider-github_v5.36.0

### Affected Resource(s)

github_organization_ruleset

```hcl
resource "github_organization_ruleset" "sre-general-rule" {
  name        = "sre-general-rule"
  target      = "branch"
  enforcement = "active"

  conditions {
    ref_name {
      include = ["~ALL"]
      exclude = []
    }
  }

  rules {
    creation                = true
    update                  = true
    deletion                = true
    required_linear_history = true
    required_signatures     = true

    branch_name_pattern {
      name     = "example"
      negate   = false
      operator = "starts_with"
      pattern  = "ex"
    }
  }
}
```

### Steps to Reproduce

import existent org repository ruleset

### Debug Output

_No response_

### Panic Output

_No response_

### Code of Conduct

o-sama:
Resolves #1886 

Thank you to @csainty and @zoezhangmattr for finding the issues I'm addressing here!

----

### Before the change?

* `bypass_actors.bypass_mode` was optional, causing failures if not set since it is required on the GitHub API.
* Typo in `github_organization_ruleset.conditions.repository_name`, `include` spelled as `inlcude`.
* Importing rulesets didn't set `bypass_actors` correctly in state, only setting `[]`.
* Importing organization rulesets with `conditions.repository_name` set crashed the provider.

### After the change?

* Added default to `github_organization_ruleset.conditions.repository_name.protected`. 
* Fixed importing `bypass_actors` state when importing rulesets.
  * Bypass actors are correctly set now.
  * Added a sort when flattening bypass actors to sort descending by ID, when TF reads the input it seems to sort by descending ID and this mimics that. I opted out of `DiffSuppressFunc` as that makes an individual call to the function on each field of each instance, making a comparison of the input frustrating.
* Added nil checks for pointers when flattening conditions and rules to avoid nil pointer dereferencing.
* Fixed the typo in `github_organization_ruleset.conditions.repository_name`. More on why this is not a breaking change in the comments.

### Pull request checklist

- [ ] Yes
- [x] No

----
@kfcampbell @nickfloyd I realize we discussed this yesterday but while looking into these other issues I kind of realized why it isn't breaking any existing use of this resource to fix the typo. I'll explain more in the comments on the PR.


github/resource_github_repository_ruleset.go
```diff
@@ -67,7 +67,7 @@ func resourceGithubRepositoryRuleset() *schema.Resource {
 						},
 						"bypass_mode": {
 							Type:         schema.TypeString,
-							Optional:     true,
+							Required:     true,
```
o-sama, line 70: Same as the reasoning in organization ruleset.

github/resource_github_organization_ruleset.go
```diff
@@ -63,11 +62,12 @@ func resourceGithubOrganizationRuleset() *schema.Resource {
 						},
 						"bypass_mode": {
 							Type:         schema.TypeString,
-							Optional:     true,
+							Required:     true,
```
o-sama, line 65: This is required in the GitHub API and so previous usage would just end in an error when the request is made, which is why it's a patch and not a breaking change.

o-sama (commented):
I hope the comments I left here clarify my reasoning for choices I made in the PR and make a good case for why this doesn't need to be a major version change 🙂 

Also to note: I found a bug which is caused by `go-github` which it seems has been fixed in v55. This would cause using `conditions.repository_name` to be broken as is right now even with these bug fixes until the provider is updated to use that version.

The issue happens when either `include` or `exclude` is empty, since in the type defined in the client, `omitempty` is used here. This doesn't affect `conditions.ref_name` because `omitempty` isn't used there.

csainty:
I ran this against my local config I was working towards. The plan look right now 🙌 
Thanks for your work!

nickfloyd:
Wow @o-sama such good sluthing here!  Thank you so much for making these change but also being super clear and thoughtful as to how the were approached and the whys behind them. I completely agree with your assessments. 

Also, thank you @csainty for validating these changes with your local! 

Y'all are what makes our community incredible! Thank you for all that you do!

PR #1906, opened 2023-09-18
Fixed invalid default hostname for gh auth token

### LINKED ISSUE #1905: [BUG]: Incorrect default hostname for tokenFromGhCli() function
### Expected Behavior

The expected default hostname for `gh auth login` and its associated `gh auth token` cli command is github.com. When the provider initializes and there is no base_url specified in its configuration or in the environment, the auth token fetched from the GitHub CLI should use github.com as its hostname.

### Actual Behavior

The default hostname is currently set to `api.github.com` which is incorrect and prevents the provider from correctly authenticating with GitHub via the GitHub CLI

### Terraform Version

Terraform version: 1.5.7

### Affected Resource(s)

- provider.go

_No response_

### Steps to Reproduce

_No response_

### Debug Output

_No response_

### Panic Output

_No response_

### Code of Conduct

jkstpierre:
Resolves #1905

----

### Before the change?

The current default hostname for getting the authentication token from GitHub CLI is "api.github.com"

### After the change?

The default hostname is set to "github.com" which is the correct default hostname for `gh auth login`.

There should be no breaking changes contained in this PR.

----

### DIFF
--- a/github/provider.go
+++ b/github/provider.go
@@ -398,7 +398,7 @@ func tokenFromGhCli(baseURL string) (string, error) {
 	}
 	hostname := ""
 	if baseURL == "" {
-		hostname = "api.github.com"
+		hostname = "github.com"
 	} else {
 		parsedURL, err := url.Parse(baseURL)
 		if err != nil {


nickfloyd (approved):
Great catch @jkstpierre, thank you for this contrib ❤️

Issue #1909, opened 2023-09-20
[BUG]: Organization ruleset invalid property condition


jcshadhara:
### Expected Behavior

The organization ruleset specified in the terraform file applies without errors

### Actual Behavior

The API throws an error when trying to apply an organization ruleset with a condition field:
```
│ Error: POST https://api.github.com/orgs/MyOrganization/rulesets: 422 Invalid request.
│ 
│ Invalid property /conditions: data matches no possible input. See `documentation_url`. []
```

### Terraform Version

Terraform version >= 1.3.6
on linux_amd64
provider registry.terraform.io/integrations/github v5.37.0

### Affected Resource(s)

github_organization_ruleset

### Terraform Configuration Files

```hcl
Ruleset configuration as follows:

resource "github_organization_ruleset" "master_main_branch_protection" {
  name        = "Master and main branches protection"
  target      = "branch"
  enforcement = "active"

  conditions {
    ref_name {
      include = [
        "refs/heads/master",
        "refs/heads/main"
      ]
      exclude = []
    }
  }

  rules {
    deletion         = true
    non_fast_forward = true
    pull_request {
      dismiss_stale_reviews_on_push   = true
      require_code_owner_review       = true
      required_approving_review_count = 1
      require_last_push_approval      = true
    }
  }
}

```
```


### Steps to Reproduce

_No response_

### Debug Output

_No response_

### Panic Output

_No response_

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

**Comment by o-sama** [comment]
Hi @jcshadhara, i think I know what's happening. Can you try setting either `repository_name` or `repository_id` and reporting back please?

The API expects one of them to be defined even if it's empty.

I'll add that to the documentation and potentially change the schema to reflect that, but it could potentially be a major version change so I'll look into that before saying anything for sure.

Hope this helps!

**Comment by jcshadhara** [comment]
I confirm that adding the following code into the `conditions` property made it work:
```
repository_name {
  include = []
  exclude = []
}
```
Before adding the aforementioned piece of code, I tried with the property `repository_id` but threw the same error posted in the main post.

o-sama:
Interesting, let me play around with that. The API should accept the `repository_id` and `ref_name` combination as per the docs

tj-smith47:
This still does not reflect in the documentaiton: https://registry.terraform.io/providers/integrations/github/latest/docs/resources/organization_ruleset

[...]

PR #1911, opened 2023-09-21
Fix organization ruleset conditions schema validation, update docs

### LINKED ISSUE #1899: Organization Ruleset missing support for OrganizationAdmin bypass
### Expected Behavior

`actor_id` should not be required because it is not applicable to OrganizationAdmin role type 

### Actual Behavior

`actor_id` causes apply to fail if it is not set or set to null. Setting to 0 doesn't work either. Creating a ruleset in the UI with org admin bypass and fetching the payload shows that the actor id should be null. 

Curious on what @o-sama thinks and if a change is needed. 

### Terraform Version

V1.5.5

### Affected Resource(s)

- github_organization_ruleset

_No response_

### Steps to Reproduce

_No response_

### Debug Output

_No response_

### Panic Output

_No response_

### Code of Conduct

o-sama:
Resolves #1899 & #1909

----

### Before the change?

* Validation on `organization_ruleset.conditions.0` was partially broken.
  * It correctly prevented setting both `repository_name` and `repository_id`.
  * It incorrectly allowed setting neither `repository_name` nor `repository_id`.
* Documentation for `bypass_actors` didn't specify anything related to `actor_id` when setting type to `OrganizationAdmin`, the API docs also don't mention anything but the expected value is `1` as tested by @YElyousfi.

### After the change?

* At least one of `repository_name` and `repository_id` is required for organization rulesets conditions, previously this would fail on apply since the API requires at least one to be set alongside `ref_name`. This shouldn't require a major version change because requests sent using the previous iteration weren't actually succeeding, so there shouldn't be a state change or breakage of existing use.
* Updated docs for both of the issues mentioned in the before section.

### Pull request checklist

- [ ] Yes
- [x] No

----

### DIFF
--- a/github/resource_github_organization_ruleset.go
+++ b/github/resource_github_organization_ruleset.go
@@ -52,7 +52,7 @@ func resourceGithubOrganizationRuleset() *schema.Resource {
 						"actor_id": {
 							Type:        schema.TypeInt,
 							Required:    true,
-							Description: "The ID of the actor that can bypass a ruleset",
+							Description: "The ID of the actor that can bypass a ruleset. When `actor_type` is `OrganizationAdmin`, this should be set to `1`.",
 						},
 						"actor_type": {
 							Type:         schema.TypeString,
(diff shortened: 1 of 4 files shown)


lewismiddleton:
I [left a comment](https://github.com/integrations/terraform-provider-github/issues/1899#issuecomment-1730072504) on #1899 with some more bypass actor_id quirkiness. Could expand the scope of this PR to address that too. 

I've also opened #1915 which may be of interest to you.

nickfloyd (approved):
Thanks for the commits here! ❤️

PR #1939, opened 2023-10-05
[fix]: Fixes parsing for checks with `:` in the name.

### LINKED ISSUE #1531: [BUG] github_branch_protection_v3 `checks` does not allow for colons
If my status check context includes a colon then it fails.
Our CircleCi jobs for example `ci/circleci: unit_test` produces the following error:
```
Error: Could not parse unit_test:-1 as valid app_id
```

Potential solutions are:
- Use a different different character to split the string
- Separate out the context and app_id as separate input variables (e.g. `checks = [ { context = "ci/circleci: unit_test", app_id = 1234 } ]`

[Link to original "checks" feature PR](https://github.com/integrations/terraform-provider-github/pull/1415)

yorik:
Resolves #1531

----

### Before the change?

*  Checks with `:` in the name were not possible to use because the string was split by the first `:`.

### After the change?

* The string is split by the last `:`, so it's possible to have check names with `:` charater.
* Founds the last `:` in a check string and uses all after it as app_id.

### Pull request checklist

- [ ] Yes
- [x] No

----


nickfloyd (approved):
@yorik Thank you for the contributions! ❤️

Issue #1938, opened 2023-10-05
[BUG]: github_branch data source constructs incorrect API URL for public repo


NickFitz:
### Expected Behavior

Given the provider configuration:
```
provider "github" {
  owner = "someorg"
}
```
the data source:
```
data "github_branch" "main" {
  repository = "some_public_repository"
  branch = "main"
}
```
is expected to result in `data.github_branch.main` having the attributes documented at https://registry.terraform.io/providers/integrations/github/latest/docs/data-sources/branch relating to the GitHub repository at `someorg/some_public_repository`.

Examining the `DEBUG` logs shows that the path being requested from the GitHub API is
```
/repos//some_public_repository/git/ref/heads/main
```
when it should be
```
/repos/someorg/some_public_repository/git/ref/heads/main
```

Setting the `GITHUB_OWNER` environment variable does not resolve the issue. It appears from the debug output that the absence of a token (which is not required as the repository is public) causes the data source to configure an anonymous owner. But an anonymous owner does not seem to have a name, and therefore the interpolation of the owner name as the organisation name in URL construction fails, resulting in the `//` seen in the URL.

This appears to be the same, or very similar, behaviour to that identified in issue #942 with the `github_repository_file` data source. A fix for that issue is in #1494, and from comparing that PR and the corresponding code in `github_branch`, I believe the fix would be the same or very similar. I also suspect the same issue may be present in other data sources.

### Actual Behavior

The attributes of `data.github_branch.main` are all `null` due to the malformed path in the API request, which results in a `404 Not Found` response.

### Terraform Version

```
Terraform v1.5.7
on darwin_arm64
+ provider registry.terraform.io/hashicorp/aws v5.18.1
+ provider registry.terraform.io/hashicorp/local v2.4.0
+ provider registry.terraform.io/integrations/github v5.39.0
```

### Affected Resource(s)

- github_branch

### Steps to Reproduce

```
terraform apply
```

### Debug Output

```shell
2023-10-05T11:35:49.128+0100 [WARN]  ValidateProviderConfig from "provider[\"registry.terraform.io/integrations/github\"]" changed the config value, but that value is unused
2023-10-05T11:35:49.129+0100 [DEBUG] provider.terraform-provider-github_v5.39.0: 2023/10/05 11:35:49 [INFO] Selecting owner  from GITHUB_OWNER environment variable
[...]

PR #1029, opened 2022-01-08
Allow setting owner in anonymous mode

kfcampbell:
Fixes #647. 

The provider currently doesn't respect any ownership configuration when running in anonymous mode. This is annoying because it blocks all kinds of uses, for example data sources retrieving public repos. 

This PR makes a small change to respect ownership in anonymous mode, and now those data sources work!

PR #1943, opened 2023-10-07
Add owner attribute for reading public repos

Nmishin:
Resolves [#1938](https://github.com/integrations/terraform-provider-github/issues/1938)

The main idea here that we need to calculate owner parameter no matter this is anonymous user or user with token. 

### Before the change?

* No owner added if no token specified. Neither via GITHUB_OWNER env variable, neither via specifying owner in the provider configuration.
* Because of this path to repos calculated wrongly, for example like: `/repos//some_public_repository/git/ref/heads/main`

### After the change?

* Owner can be added if no token used. Works via GITHUB_OWNER env variable and also in the provider configuration.
* All configuration precedence works as described in [provider configuration docks](https://registry.terraform.io/providers/integrations/github/latest/docs).

### Pull request checklist

- [ ] Yes
- [x] No

----


kfcampbell:
@Nmishin have you looked at #1029? What do you think about that approach vs. yours?

Nmishin:
@kfcampbell I didn't saw it, honestly :)
My idea was exactly the same, I think we need to merge your MR

kfcampbell:
@Nmishin I've done so now and will release it with the next version of the provider. Thank you for your help and contributions!

PR #1950, opened 2023-10-11
Fix ruleset bypass actors diff issues

### LINKED ISSUE #1952: [BUG]: github_organization_ruleset unable to remove bypassers
### Expected Behavior

When you set [bypass_actors](https://registry.terraform.io/providers/integrations/github/latest/docs/resources/organization_ruleset#bypass_actors) to null remove all bypassers.

### Actual Behavior

It does not remove them. It also gets confused on every plan and it tries to swap existing bypassers for some reason!

### Terraform Version

Terraform v1.6.1
on linux_amd64
+ provider registry.terraform.io/integrations/github v5.39.0

### Affected Resource(s)

- github_organization_ruleset

_No response_

### Steps to Reproduce

_No response_

### Debug Output

```shell
# module.organization.github_organization_ruleset.this["pr_default"] will be updated in-place
  ~ resource "github_organization_ruleset" "this" {
        id          = "118285"
        name        = "Pull Request Default Ruleset"
        # (5 unchanged attributes hidden)

      ~ bypass_actors {
          ~ actor_id    = 5 -> 1
          ~ actor_type  = "RepositoryRole" -> "OrganizationAdmin"
        }
      ~ bypass_actors {
          ~ actor_id    = 1 -> 5
          ~ actor_type  = "OrganizationAdmin" -> "RepositoryRole"
        }

        # (4 unchanged blocks hidden)
    }
```

### Panic Output

_No response_

### Code of Conduct

o-sama:
Resolves #1952 (Partial)

----

### Before the change?

* When updating a ruleset you can get the following:
```
# github_repository_ruleset.test2 will be updated in-place
  ~ resource "github_repository_ruleset" "test2" {
        id          = "xxxxx"
        name        = "test-12345"

      ~ bypass_actors {
          ~ actor_id    = 5 -> 1
          ~ actor_type  = "RepositoryRole" -> "OrganizationAdmin"
        }
      ~ bypass_actors {
          ~ actor_id    = 1 -> 5
          ~ actor_type  = "OrganizationAdmin" -> "RepositoryRole"
        }

        # (1 unchanged block hidden)
    }
```

I've seen this happen when the order in which `bypass_actor` blocks happen to be defined changes. Whether this is done manually or through an automation (I've noticed this happen with something I'm working on which uses a yaml file for defining things).

A while back I made a small change which tries to fix this by sorting when flattening the block, but that doesn't cover this case (and potentially others) and isn't technically the way to do it.

### After the change?

* Now we have a DiffSuppressFunc defined which checks every bypass actor and makes sure a diff only shows up when something does change. The caveat here is it's done several times because each key is passed to the function once. Personally though I would still take this over having a diff every time TF runs, and logically it doesn't make sense to have _many_ bypass actors on a ruleset, so we shouldn't expect a real impact on performance.

The following is the same change as above:

```
github_repository_ruleset.test2: Refreshing state... [id=xxxxx]
github_organization_ruleset.test: Refreshing state... [id=xxxxx]

No changes. Your infrastructure matches the configuration.
```

### Pull request checklist

- [ ] Yes
- [x] No

----


kfcampbell:
> I will test out my theory and if needed, make a PR to go-github to get this addressed in the next release of the client.

That sounds great, thank you @o-sama! We appreciate all your contributions.

felixlut:
I randomly stumbled upon this issue myself last week in https://github.com/jdamata/terraform-provider-sonarqube/pull/211.

I started out with defining the `selected_projects` attribute as a [`schema.TypeList`](https://developer.hashicorp.com/terraform/plugin/sdkv2/schemas/schema-types#typelist). While testing the provider I noticed that terraform plan always tried to alter the ordering of the projects, which is just noise. The way I solved it was to use a [`schema.TypeSet`](https://developer.hashicorp.com/terraform/plugin/sdkv2/schemas/schema-types#typeset) instead. From the Terraform documentation the latter is meant to be used when the order matters, while the former is unordered. This solved the issue because Terraform in the `plan` stage simply didn't care about the ordering of the underlying data, just that they included the same objects. 

I'm not sure what other implications there are now that `schema.TypeList` is out in the wild already, but it might be valuable information to know in the future. From most of the Terraform providers I've looked into so far (including this one), it seems that `schema.TypeSet` is highly underused in favor of `schema.TypeList`, especially for nested attributes. I might be overlooking a reason not to use it, but I think an unordered set usually is a more accurate representation than an ordered list.

o-sama:
I remember when we worked on this initially I saw `TypeList` used elsewhere and so I used it for most of the nested schemas.

Similarly I wasn't aware this would be an issue to be honest and so it seems like it would be fine.

In hindsight it'd have been better to use `TypeSet`, but like you pointed out it has potential implications if we change it, even with a state migration function it can still be a breaking change and so a custom diff function is the less intrusive way of avoiding a diff every time this is run.

I am interested though in why it is underused compared to `TypeList`, although things should be better as the provider is slowly migrated away from the old sdk.

felixlut:
Ye I'm in the same camp. I'd never used the set for a nested schema before last week, and I'm confused as to why it's not used more. Maybe I'll find out in a month when I curse myself for using it in the SonarQube provider 😅

kfcampbell:
@o-sama where did you land on the go-github PR? What does that mean for this PR? I'd like to check just to make sure this doesn't get forgetten.

stevehipwell:
What needs to happen before we can get this addressed? We're currently seeing a constant churn on both the `github_organization_ruleset` & `github_repository_ruleset`.

o-sama:
@kfcampbell I was tied up between work and a different feature in go-github and haven't implemented the change I mentioned yet. I've worked on it locally but I just have to add test cases and push my changes for review.

I think the PR as is should fix the diff issue but it wouldn't add the ability to remove bypass actors since as far as I'm aware the client lacks that ability still.

I've been abroad for the last week or so but should have some time to work on that in the next couple of days. My bad on the delay!

usmonster:
Thanks for your effort on this, @o-sama! Please let me know if there's anything you need help with. 🙏

o-sama:
@kfcampbell  Apologies for the delay here, It's been a hectic couple of months for me. I'm going to be creating the PR on `go-github` in the next couple of hours, but in the meantime is it possible for us to get this out and possibly get a release for it when that's appropriate? I think this will still help a lot of folks out with the perpetual diff they're seeing in their rulesets which costs a lot of time since each ruleset edit is at least 1 second with the write delay to account for rate limits.

I'll have to revisit my notes but IIRC once I get the `go-github` PR merged and released, it'll simply be a matter of updating the client's version and the issue with removing bypass actors would be fixed, so it'd be more of a general update than a code change specific to rulesets.

kfcampbell:
@o-sama Sounds good! When that's merged, we can also clean up the diff suppress functions introduced here.

kfcampbell (approved):

o-sama:
@kfcampbell The diff suppress func will likely need to stay since this issue comes from the default diff func comparing the objects as a string one line at a time, so you'd likely always have a diff. I'll try changing that to a set at some point and having a state migration  func to handle the change without needing anything on the users' end, how does that sound?

By the way thank you for the release! Always appreciate your timely responses!

kfcampbell:
Ahh, thanks for clarifying. That sounds good to me! I'm grateful for your contributions to this project.

PR #1965, opened 2023-10-17
Team reviewers exclusions

### LINKED ISSUE #1972: [FEAT]: Exclude list for Team Code Review settings
### Describe the need

This is a continuation of #1249 which I think should be closed already by #1356

The current version of [`github_team_settings`](https://registry.terraform.io/providers/integrations/github/latest/docs/resources/team_settings) is missing one field that should be available which makes it possible to exclude certain members of the team from being assigned to review PR:s (see image below). 

image

The GraphQL API seems to be the only way to programmatically alter these fields, and they are [currently in preview](https://docs.github.com/en/graphql/overview/schema-previews#team-review-assignments-preview). The [`excludedTeamMemberIds`](https://docs.github.com/en/graphql/reference/input-objects#updateteamreviewassignmentinput) list is available in the [`updateTeamReviewAssignment`](https://docs.github.com/en/graphql/reference/mutations#updateteamreviewassignment) mutation call, so the functionality should be possible. 

### SDK Version

_No response_

### API Version

_No response_

### Relevant log output

_No response_

### Code of Conduct

felixlut:
Resolves #1972

----

### Before the change?

* It's not possible to alter the exclusion list for team members that shouldn't be added as reviews to pull requests via the terraform provider

### After the change?

* It's now possible to alter the exclusion list for team members that shouldn't be added as reviews to pull requests via the terraform provider

### Pull request checklist

- [ ] Yes
- [x] No

----


felixlut:
@kfcampbell I started working on adding functionality for this issue and quickly realized that it's currently limited by the API, [which is currently in preview](https://docs.github.com/en/enterprise-server@3.10/graphql/overview/schema-previews#team-review-assignments-preview). While it's possible to to SET the exclusion list, it's not possible to read it. This leads to terraform never being able to make proper plans, since the exclusion list always will be empty from the providers perspective: 

image

My question now is whether you're able to ask around to see what the maturity plan is for the GitHub team GraphQL API is? I could spend some time trying to work around the issue, but the "proper" solution is just for the read of the exclusion list to be added to the API. If that's upcoming in the pipeline, I don't think whatever hack I come up with is the way to solve it. 

It seems like it's been in preview since at least `2020-01-14`, which is when the Team preview was added to the docs of [Enterprise Server 2.20](https://docs.github.com/en/enterprise-server@2.20/graphql/overview/schema-previews#team-review-assignments-preview) (it's nowhere to be found on [the same page for 2.19](https://docs.github.com/en/enterprise-server@2.19/graphql/overview/schema-previews#team-review-assignments-preview)). Would be nice to know what the status of the preview is 😅

kfcampbell:
> the "proper" solution is just for the read of the exclusion list to be added to the API

I totally agree with you. I don't think it's worth hacking around until it's available in the API. Unfortunately I don't have any insight as to the plan there. Logging a public request [here](https://github.com/orgs/community/discussions/categories/api-and-webhooks) is probably the best bet.

felixlut:
I've added a discussion about it here: https://github.com/orgs/community/discussions/72860

Rowern:
Hello, just saw some activity on the discussion page, do you know if this is mergeable @felixlut ?
This would be very helpful to be able to manage those settings via the provider.

felixlut:
@Rowern no, sadly it's not in a mergeable state. A month or so ago I got back to this and checked again if the API needed for this feature was available now, as [the reply to my discussion suggested](https://github.com/orgs/community/discussions/72860#discussioncomment-7491027). It was not. It isn't now. The API is no longer in preview though, which is something.

I've tried reaching out again [here](https://github.com/orgs/community/discussions/72860#discussioncomment-9988382), but haven't seen an update since.

janeklb:
Another attempt at this: https://github.com/integrations/terraform-provider-github/pull/3095

Issue #1975, opened 2023-10-23
[BUG]: Odd behaviour when running provider tests


scott-doyland-burrows:
### Expected Behavior

The provider test should have the same result when setting the ```TF_LOG``` level to different values.

I would assume the ```TF_LOG``` setting would have no bearing on the outcome of the test results.

I was going to attempt to fix a bug I reported in another issue, but I now don't know whether the issue reported here needs fixing first - I will look into it.

### Actual Behavior

```TF_LOG=WARN TF_ACC=1 go test -v ./... -run ^TestAccGithubTeam```
tests fail

```
    testing.go:705: Step 0 error: Check failed: 3 errors occurred:
        	 Check 2/4 error: data.github_team.test: Attribute 'repositories_detailed.#' expected "1", got "0"
        	 Check 3/4 error: data.github_team.test: Attribute "repositories_detailed.0.repo_id" not set, but "repo_id" is set in github_repository.test as "708930695"
        	 Check 4/4 error: data.github_team.test: Attribute "repositories_detailed.0.role_name" not set, but "permission" is set in github_team_repository.test as "admin"
```

```TF_LOG=DEBUG TF_ACC=1 go test -v ./... -run ^TestAccGithubTeam```
tests pass

### Terraform Version

Terraform v1.6.2
on linux_amd64

### Affected Resource(s)

Provider tests

### Terraform Configuration Files

```hcl
N/A
```

### Steps to Reproduce

Follow the instructions here:

https://github.com/integrations/terraform-provider-github/blob/main/CONTRIBUTING.md#local-development-setup

and use the ```TestAccGithubTeam``` test.

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

Issue #1668, opened 2023-04-25
[FEAT]: Remove parent team when parent_team_id == 0


Milkshak3s:
## Currently:
If a team has a parent and the `parent_team_id` is updated to `0` or `null` in the associated Terraform resource, the TF state will update the `parent_team_id` to `0`, but the GitHub team will remain under its original parent.

This brings the TF state out of line with what is true in GitHub, and is unexpected behavior.

## Suggested change
If the new `parent_team_id` is `0` and the field `parent_team_id` is not currently `0` or `null`, pass `true` as the `removeParent` argument to `EditTeamByID` (rather than hardcoding it to `false`, as is current`:
https://github.com/integrations/terraform-provider-github/blob/e83bb8d8ffd505d63d068527df25efcdccdfc7a0/github/resource_github_team.go#L233

https://github.com/integrations/terraform-provider-github/blob/e83bb8d8ffd505d63d068527df25efcdccdfc7a0/vendor/github.com/google/go-github/v51/github/teams.go#L224

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

kfcampbell:
@Milkshak3s that seems like a reasonable approach! Do you have the inclination to open a PR for this behavior?

Milkshak3s:
@kfcampbell I would love to! A bit swamped this week, but I should have some cycles next week to get something up.

scott-doyland-burrows:
It is worth noting that there are some other issues with parent teams - almost the reverse issue of what was reported above.

I create two teams in terraform and apply, and then alter one in terraform to have a parent team and apply.

I then go to GitHub and update the child team to not have the parent team.

I run a terraform apply and it does not report a diff, even though the terraform code still has the parent team defined.
The statefile shows the parent team (for the child team) as set to ```""```.

Terraform should show a diff.

Looks like the logic for parent teams is broken - and I wonder if using a new resource would be preferable, as per my suggestion here:

https://github.com/integrations/terraform-provider-github/issues/1954

scott-doyland-burrows:
Just to add to the above, if I delete the parent team via terraform (it remains in place in GitHub which is the wrong behaviour), then I see the statefile alters from this:

```
            "name": "scott-team02",
            "node_id": "T_kwDOBQjlXc4AhnuY",
            "parent_team_id": "8813444",
            "parent_team_read_id": "8813444",
            "parent_team_read_slug": "scott-team01",
            "privacy": "closed",
[...]

PR #1979, opened 2023-10-25
Fixes parent team removal, fixes the teams data source tests, adds tests for the parent team fix

scott-doyland-burrows:
Resolves https://github.com/integrations/terraform-provider-github/issues/1668
Resolves https://github.com/integrations/terraform-provider-github/issues/1975

I have closed https://github.com/integrations/terraform-provider-github/pull/1976 (which I initially setup) as this also resolves https://github.com/integrations/terraform-provider-github/issues/1975 but is no longer needed as this PR will now fix it.

----

### Before the change?
https://github.com/integrations/terraform-provider-github/issues/1668
Parent teams were created in terraform.  However if a team was removed from the parent team then the statefile was updated, but the team still remained under a parent team in GitHub.

Also, if a team was removed from under a parent team in GitHub but still was defined in terraform, then terraform would not show a diff.

Overall, nested teams were not correctly managed by Terraform.

https://github.com/integrations/terraform-provider-github/issues/1975
Before the change there was a failure reported in a test in the ```data_source_github_team_test.go``` file. The reason is the terraform data resource in the ```queries an existing team with connected repositories``` test block is only waiting on the ```github_team``` resource, and the data resource actually needs to wait on the ```github_team_repository``` resource.

Running this test:
```TF_LOG=WARN TF_ACC=1 go test -v ./... -run ^TestAccGithubTeamDataSource```
would result in a failure:
```
    testing.go:705: Step 0 error: Check failed: 3 errors occurred:
        	 Check 2/4 error: data.github_team.test: Attribute 'repositories_detailed.#' expected "1", got "0"
        	 Check 3/4 error: data.github_team.test: Attribute "repositories_detailed.0.repo_id" not set, but "repo_id" is set in github_repository.test as "708930695"
        	 Check 4/4 error: data.github_team.test: Attribute "repositories_detailed.0.role_name" not set, but "permission" is set in github_team_repository.test as "admin"
```

However, if using ```TF_LOG=DEBUG``` it would finish successfully - I believe this is just a timing issue.  ```TF_LOG=DEBUG``` takes longer to run each resource, so by the time the data resource is actioned, the previous resources have all (just by chance) also completed.

It appears that the test for ```TestAccGithubTeamDataSource``` was defined incorrectly.  I have refactored to bring the test in-line with how other tests are defined and this appears to have resolved the issue.

### Formatting
The ```.tf``` files are now formatted correctly.

* 

### After the change?
Parent teams are correctly managed by Terraform.  Tests have been defined and pass.

The data source tests for teams work.

* 

### Pull request checklist

- [ ] Yes
- [x] No

----


kfcampbell (approved):
This is great! Thank you for the thoughtful PR. All updated integration tests are passing for me as well.

Issue #1955, opened 2023-10-12
[BUG]: github_repository_environment - no terraform diff when resetting deployment branch policies via GitHub UI


scott-doyland-burrows:
### Expected Behavior

I set my code as follows and apply - this works.

``` 
resource "github_repository_environment" "example" {
  environment  = "example"
  repository   = "test"

  deployment_branch_policy {
    protected_branches     = true
    custom_branch_policies = false
  }
}
```

If I update the deployment branch policy in GitHub UI to ```no restriction```, then re-apply I expect terraform to show a diff and reset the deployment branch policy to that of the terraform code.

### Actual Behavior

Terraform does not reset the deployment branch policy.

Instead it reports no diff.

The statefile still shows the  deployment branch policy as ```"protected_branches": true```

The same scenario occurs with the ```custom_branch_policies```

Looks like there is various bad logic in this resource with regards to the deployment branch policies.

### Terraform Version

Terraform v1.6.0
on linux_amd64
+ provider registry.terraform.io/integrations/github v5.39.0

### Affected Resource(s)

github_repository_environment

### Terraform Configuration Files

``` 
resource "github_repository_environment" "example" {
  environment  = "example"
  repository   = "test"

  deployment_branch_policy {
    protected_branches     = true
    custom_branch_policies = false
  }
}
```

### Steps to Reproduce

```
terraform apply
```

Reset policy as shown in GitHub UI:

![image](https://github.com/integrations/terraform-provider-github/assets/61512176/6223c95c-cd8d-4330-821d-59981954e13c)

```terraform apply``` and there will be no diff shown.
State file still shows ```"protected_branches": true```

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

PR #1981, opened 2023-10-26
Fixed repo envs so terraform will add back in deployment branch protection if it is removed manually from GitHub

scott-doyland-burrows:
Resolves https://github.com/integrations/terraform-provider-github/issues/1955

----

### Before the change?
Manually removing an environment's deployment branch protection in GitHub (setting it back to no restriction) would not show a diff when terraform was applied.  This meant terraform and GitHub were out of line.

* 

### After the change?
Terraform brings the environment's deployment branch protection back inline with terraform.

The tests pass - albeit they don't give full coverage, and never did (I have not updated the tests).  They need some extra work to make them better.

There are also two minor changes (as vscode was complaining) which just set lowercase for an error message and remove a full stop for another error message in two unrelated files.

* 

### Pull request checklist

- [ ] Yes
- [X] No

----

### DIFF
--- a/github/resource_github_branch.go
+++ b/github/resource_github_branch.go
@@ -191,7 +191,7 @@ func resourceGithubBranchImport(d *schema.ResourceData, meta interface{}) ([]*sc
 
 	// resourceGithubBranchRead calls d.SetId("") if the branch does not exist
 	if d.Id() == "" {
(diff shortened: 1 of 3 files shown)


kfcampbell (approved):
Thank you @scott-doyland-burrows!

Issue #1964, opened 2023-10-17
[BUG]: github_team_members cannot import team member by name


scott-doyland-burrows:
### Expected Behavior

Team membership should be imported successfully using the team slug or team ID.

### Actual Behavior

Terraform can only import the team membership using the team ID if the original team members were added manually (as would normally be expected if we are importing them to terraform).

For example:

If the team membership was added to the team via the UI, then when terraform attempts to import the team membership via the team slug, it will want to destroy and recreate the team membership.  This is because terraform wants to set the ```id``` and ```team_id``` for the ```github_team_members``` resource in the statefile to that of the team slug on an apply.

Whereas terraform initially pulls these values into the statefile as the team ID during the initial import (this is what I assume it does anyway from the way it is behaving).

It seems the main issue is that ```id``` and ```team_id``` for the ```github_team_members``` in the statefile can either be the team ID or the team slug - so there is no consistency and so it gets confused at import time.

### Terraform Version

Terraform v1.6.1
on linux_amd64
+ provider registry.terraform.io/integrations/github v5.40.0

### Affected Resource(s)

```github_team_members```

### Terraform Config

The following will destroy/recreate the imported resource as the ```team_id``` is using ```github_team.team.slug```
If ```github_team.team.id``` was used then it would work as expected.

```
resource "github_team" "team" {
  name        = "scott-team01"
  description = "my team again"
  privacy     = "closed"
}

resource "github_team_members" "members" {
  team_id = github_team.team.slug

  members {
    username = "scott-doyland-burrows"
    role     = "maintainer"
  }
}

import {
  id = "scott-team01"
  to = github_team_members.members
}
```
```
Terraform used the selected providers to generate the following execution plan. Resource actions are indicated with the following symbols:
-/+ destroy and then create replacement

Terraform will perform the following actions:

  # github_team_members.members must be replaced
  # (imported from "scott-team01")
  # Warning: this will destroy the imported resource
-/+ resource "github_team_members" "members" {
      ~ id      = "8868843" -> (known after apply)
      ~ team_id = "8868843" -> "scott-team01" # forces replacement

        members {
            role     = "maintainer"
            username = "scott-doyland-burrows"
        }
    }

Plan: 1 to import, 1 to add, 0 to change, 1 to destroy.
```

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

PR #1991, opened 2023-10-31
update documentation for github_team_members to recommended team id

scott-doyland-burrows:
Resolves https://github.com/integrations/terraform-provider-github/issues/1964

The issue seemed difficult to correct via code. It seems better to still allow the current behaviour but recommend the desired approach.

----

### Before the change?
Documentation did not describe the pitfalls of using the team slug when creating or importing the ```github_team_members``` resource.

* 

### After the change?
Documentation recommends that the team id, rather than team slug, is used for the  ```github_team_members``` resource and gives the reasons why.

* 

### Pull request checklist

- [ ] Yes
- [x] No

----

### DIFF
--- a/website/docs/r/team_members.html.markdown
+++ b/website/docs/r/team_members.html.markdown
@@ -59,7 +59,10 @@ resource "github_team_members" "some_team_members" {
 
 The following arguments are supported:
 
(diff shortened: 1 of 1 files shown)


kfcampbell (approved):
This seems like a reasonable approach. Thank you for the contribution!

PR #1994, opened 2023-11-02
fix: Use ETag when reading `github_branch_default` resources.

### LINKED ISSUE #1993: [BUG]: `github_branch_default` does not use ETags for reads.
### Expected Behavior

Request quota should not be consumed by plans if there are no changes and no drift.

### Actual Behavior

Quota is consumed by `github_branch_default` resources.

### Terraform Version

Terraform v1.5.4
on linux_amd64
+ provider registry.terraform.io/integrations/github v5.40.0

### Affected Resource(s)

- github_branch_default

_No response_

### Steps to Reproduce

_No response_

### Debug Output

_No response_

### Panic Output

_No response_

### Code of Conduct

nairb774:
This resource was missing saving and using the underlying resource's ETag. This was discovered when we were hitting over-quota errors via this resource.

The etag related implementation was copied from the `github_repository` implementation.

Aside: There could be some refactoring to have these resources share some common code. Such a refactoring seemed outside the scope of this change.

Resolves #1993

----

### Before the change?

* Every plan containing a `github_branch_default` would use one request per resource read.

### After the change?

* `github_branch_default` should only consume quota when there is drift in the plan.

### Pull request checklist
   **Note:** It didn't look like there were any existing tests validating the etag behaviors. If tests are desired for this feature, a pointer to an example would be appreciated, or a rough sketch if none exists. Any guidance would be appreciated.
   **Note:** Not all resources which implement the etag feature have the field documented. The field feels like an internal implementation detail which wouldn't normally be documented. Given the repo is inconsistent, guidance on if the field should be documented or not would be appreciated.

- [ ] Yes
- [X] No

----

### DIFF
--- a/github/resource_github_branch_default.go
+++ b/github/resource_github_branch_default.go
@@ -2,6 +2,8 @@ package github
 
 import (
 	"context"
+	"log"
+	"net/http"
 
 	"github.com/google/go-github/v55/github"
 	"github.com/hashicorp/terraform-plugin-sdk/helper/schema"
@@ -34,6 +36,10 @@ func resourceGithubBranchDefault() *schema.Resource {
 				Optional:    true,
 				Default:     false,
 				Description: "Indicate if it should rename the branch rather than use an existing branch. Defaults to 'false'.",
+			},
+			"etag": {
+				Type:     schema.TypeString,
+				Computed: true,
 			},
 		},
 	}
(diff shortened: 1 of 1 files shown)


kfcampbell (approved):
Thanks for doing this!

kfcampbell:
PRs to refactor common ETag behavior are of course very welcome should you feel so inclined!

Issue #1804, opened 2023-07-20
[BUG]: Fix CanAdminsBypass to Environment


nirav-chotai:
### Describe the need

While using `github_repository_environment`, it is automatically taking `can_admins_bypass` and making invalid PUT request. 

```
"message": "Invalid request.\n\n\"can_admins_bypass\" is not a permitted key.",
```

Example code:
```
resource "github_repository_environment" "dev" {
  for_each = toset(
    [
      "dev-demo"
    ]
  )
  repository  = each.key
  environment = "dev-demo"
  reviewers {
    teams = ["123"]
  }
  deployment_branch_policy {
    protected_branches     = true
    custom_branch_policies = false
  }
}
```

There is no field for `can_admins_bypass` here, plan is successful, but when apply runs, and making PUT request, below is DEBUG error

```2023-07-20T13:00:28.754-0700 [DEBUG] provider.terraform-provider-github_v5.31.0: 2023/07/20 13:00:28 [DEBUG] GitHub API Request Details:
2023-07-20T13:00:28.754-0700 [DEBUG] provider.terraform-provider-github_v5.31.0: ---[ REQUEST ]---------------------------------------
2023-07-20T13:00:28.754-0700 [DEBUG] provider.terraform-provider-github_v5.31.0: PUT /api/v3/repos/my/dev-demo/environments/dev-demo HTTP/1.1
2023-07-20T13:00:28.754-0700 [DEBUG] provider.terraform-provider-github_v5.31.0: Host: abc.com
2023-07-20T13:00:28.754-0700 [DEBUG] provider.terraform-provider-github_v5.31.0: User-Agent: go-github/v53.2.0
2023-07-20T13:00:28.754-0700 [DEBUG] provider.terraform-provider-github_v5.31.0: Content-Length: 167
2023-07-20T13:00:28.754-0700 [DEBUG] provider.terraform-provider-github_v5.31.0: Accept: application/vnd.github.v3+json,application/vnd.github.stone-crop-preview+json
2023-07-20T13:00:28.754-0700 [DEBUG] provider.terraform-provider-github_v5.31.0: Content-Type: application/json
2023-07-20T13:00:28.754-0700 [DEBUG] provider.terraform-provider-github_v5.31.0: X-Github-Api-Version: 2022-11-28
2023-07-20T13:00:28.754-0700 [DEBUG] provider.terraform-provider-github_v5.31.0: Accept-Encoding: gzip
[...]

PR #2001, opened 2023-11-04
Add feature can_admins_bypass for repo environments | fixed wait_timer issue for repo environments

scott-doyland-burrows:
For resource ```github_repository_environment```

Resolves https://github.com/integrations/terraform-provider-github/issues/1804
```can_admins_bypass``` implemented

Resolves https://github.com/integrations/terraform-provider-github/issues/1996
```wait_timer``` issue fixed

----

### Before the change?
```can_admins_bypass``` was not implemented for repository environments.

```wait_timer``` would not show a diff if updated outside of terraform.

* 

### After the change?
```can_admins_bypass``` is implemented.

```wait_timer``` correctly shows a diff if updated outside of terraform.
* 

### Pull request checklist

- [ ] Yes
- [x] No

----

### CODE REVIEW

File: github/resource_github_repository_environment.go [hunk 1]
```diff
@@ -134,6 +140,8 @@ func resourceGithubRepositoryEnvironmentRead(d *schema.ResourceData, meta interf
 
 	d.Set("repository", repoName)
 	d.Set("environment", envName)
+	d.Set("wait_timer", nil)
```
kfcampbell, line 143: Do you mind explaining why setting the wait_timer to nil here solves the issue? It's not quite clear to me.

scott-doyland-burrows: When no wait_timer is removed then the API does not show the wait_timer setting at all when listing the environment.

So the following happens in the current code:

1. wait_timer is set in terraform

2. wait_timer is removed outside of terraform

3. We run a terraform apply and the code runs through lines 138-166 in resource_github_repository_environment.go (in the current main branch).

4. This code iterates over the protection rules BUT it never finds the wait_timer as it is just not shown at all in the API (if it were shown as "wait_timer = 0" it would find it, but its not there at all).

5. Therefore terraform never gets sent the current (ie 0) value of the wait_timer, so as far as terraform is concerned it has not changed.



Setting the wait_timer to "nil" means terraform now sees there is a diff at that point in time (note this is BEFORE terraform reports the diff to the user).

It then loops over the protection rules code.

If it does NOT find the wait_timer (ie it is removed/0 outside of terraform) then it already has it set to "nil" so it has a final diff to fix.


AND...


If it DOES find it in the protection rules (ie it has NOT been removed outside of terraform) is it compares that value with the value in the statefile (and we get a diff or not as normal, ie depending on whether the GitHUb value matches that of the statefile.

PR #2011, opened 2023-11-10
Update to google/go-github/v56

kfcampbell:
There's one [breaking change](https://github.com/google/go-github/pull/2939) that affects us here. I've chosen to follow two redirects based on the discussion in that PR, though this can be easily adjusted as necessary.


kfcampbell:
Closing in favor of #2081.

PR #2042, opened 2023-12-04
Fix typo (admiin -> admin)

yaakov-h:
At least, I assume this was a typo.


kfcampbell (approved):
It is indeed a typo. Thank you!

PR #2044, opened 2023-12-05
Make resource repository_deploy_key read_only a bool, and expand the example

rwblokzijl:
I updated the example. The read_only was implied to be a string, while it should be a bool. And i added some extra code to reflect a common usecase.


website/docs/r/repository_deploy_key.html.markdown
```diff
@@ -21,12 +21,17 @@ Further documentation on GitHub repository deploy keys:
 ## Example Usage
 
 ```hcl
-# Add a deploy key
+# Generate an ssh key
+resource "tls_private_key" "example_repository_deploy_key" {
```
kfcampbell, line 25: I'm using assuming you're using the [hashicorp/tls](https://registry.terraform.io/providers/hashicorp/tls/latest/docs/resources/private_key) provider here. Can you provide a small comment noting the provider? 

Otherwise the changes look good!

rwblokzijl: Done

kfcampbell (approved):

PR #2052, opened 2023-12-09
Ikegentz/1840

### LINKED ISSUE #1840: [MAINT]: issue_label doesn't work for Organizations
### Describe the need

There is currently no way to create an issue_label for a repo owned by an organization, rather than the account on which the PAT is used for authentication in the provider.

Specifically, the `repository` attribute in the `github_issue_label` resource is hard-coded to this format: `https://api.github.com/repos/<PAT user>/`, meaning if you try to pass `<my organization>/<repo>`, you'll end up with `https://api.github.com/repos/<PAT user>/<my organization>/<repo>`. 

[The API](https://docs.github.com/en/rest/issues/labels?apiVersion=2022-11-28#list-labels-for-a-repository) already supports an `owner` argument, so it seems like that could be exposed in the resource so that this would work.

### SDK Version

_No response_

### API Version

_No response_

### Relevant log output

```shell
│ Error: POST https://api.github.com/repos/<PAT user>/<my org>/<repo>/labels: 404 Not Found []
```

### Code of Conduct

ikegentz:
Resolves #1840

----

### Before the change?

* There was no way with the `github_issue_label` resource to specify an Organization in which the repo you wanted to create the label for, lived

### After the change?

* This adds an `org` attribute to the `github_issue_label` resource that lets the end-user explicitly specify where the repo lives they want to add the label to

### Pull request checklist

Nothing I'm aware of!

- [ ] Yes
- [X] No

----

### DIFF
--- a/examples/issue_label/main.tf
+++ b/examples/issue_label/main.tf
@@ -1,0 +1,14 @@
+resource "github_issue_label" "individual_repo_label" {
+  repository  = var.individual_repo
+  name        = "foo"
+  color       = "000000"
+  description = "Test issue"
+}
+
+  resource "github_issue_label" "org_repo_label" {
+  repository  = var.org_repo
+  name        = "bar"
+  color       = "000000"
+  description = "Test issue"
+  org         = var.org
+}
--- a/examples/issue_label/providers.tf
+++ b/examples/issue_label/providers.tf
@@ -1,0 +1,4 @@
+provider "github" {
+  owner = var.owner
+  token = var.github_token
+}
--- a/examples/issue_label/variables.tf
+++ b/examples/issue_label/variables.tf
@@ -1,0 +1,24 @@
+variable "owner" {
+  description = "GitHub owner used to configure the provider"
+  type        = string
+}
+
+variable "github_token" {
+  description = "GitHub access token used to configure the provider"
+  type        = string
+}
+
+variable "individual_repo" {
+  description = "Name of repo owned by an individual to create the issue in"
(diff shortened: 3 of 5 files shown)


kfcampbell (commented):
Thanks for the contribution! I'm a little torn as to whether we want to standardize on `owner` instead of `org` here. I think it would be better to be consistent with the keywords we use at the provider level...what do you think?

ikegentz:
> Thanks for the contribution! I'm a little torn as to whether we want to standardize on `owner` instead of `org` here. I think it would be better to be consistent with the keywords we use at the provider level...what do you think?

Wow I totally forgot about this PR, sorry for the delay!

Hm, IMO the current version of the provider is confusing because an owner of a repo doesn't _have_ to be an organization, it could be an individual too.

So on https://registry.terraform.io/providers/integrations/github/latest/docs#authentication, we have this example:
```
provider "github" {
  owner = var.github_organization
  app_auth {
    id              = var.app_id              # or `GITHUB_APP_ID`
    installation_id = var.app_installation_id # or `GITHUB_APP_INSTALLATION_ID`
    pem_file        = var.app_pem_file        # or `GITHUB_APP_PEM_FILE`
  }
}
```
specifically, `owner = var.github_organization`. IMO, this is currently confusing, so I suppose I like how I've got it here in this PR.

That said, perhaps there are some other implications to doing this I am not considering outside of the current context of this PR. Look forward to your thoughts on this @kfcampbell

PR #2072, opened 2023-12-23
[Fix] Correctly handle environments with names containing spaces

### LINKED ISSUE #1545: [BUG]: Using spaces in environment names does not work
## Context

When creating an environment with spaces using "github_repository_environment" "enviroment", the environment name gets URL encoded, rather than the spaces being parsed as spaces, they are changed to +

**Example**:
```tf
resource "github_repository_environment" "enviroment" {
  repository = "testrepo"
  environment = "Test - Env"
}
```

## Expected Outcome

The env will be created with the name "Test - Env"

## Actual Outcome

The env is created as "Test+-+Env"

## Related

Looks like this was related to #1392 where a change was made to support "/" characters inside environment names

nint8835:
Resolves #1545 

----

### Before the change?

* Creating an environment with a space in the name will result with an environment with a plus in the name instead
* Attempting to import an environment with a name containing a space will result in a not found error

### After the change?

* Creating an environment with a space in the name will correctly create an environment with a space in the name
* Importing an environment with a name containing a space will succeed

### Pull request checklist

- [ ] Yes
- [x] No

----


kfcampbell (approved):
Thank you!

Issue #2014, opened 2023-11-13
[BUG]: Unable to enter anonymous mode as such fail with 401


duxbuse:
### Expected Behavior

In v5.32.0 I can easily provide no token and it correctly enters anonymous mode and fetches files from a public repository.

I expected the same in later versions.

#### WORKING
``` log
[DEBUG] provider: using plugin: version=5
[DEBUG] No provider meta schema returned
[WARN]  ValidateProviderConfig from "provider[\"registry.terraform.io/integrations/github\"]" changed the config value, but that value is unused
[DEBUG] provider.terraform-provider-github_v5.32.0: 2023/11/13 12:26:16 [INFO] Selecting owner  from GITHUB_OWNER environment variable
[DEBUG] provider.terraform-provider-github_v5.32.0: 2023/11/13 12:26:16 [INFO] Setting write_delay_ms to 1000
[DEBUG] provider.terraform-provider-github_v5.32.0: 2023/11/13 12:26:16 [DEBUG] Setting read_delay_ms to 0
[DEBUG] provider.terraform-provider-github_v5.32.0: 2023/11/13 12:26:16 [DEBUG] Setting parallel_requests to false
[DEBUG] provider.terraform-provider-github_v5.32.0: 2023/11/13 12:26:16 [INFO] No token present; configuring anonymous owner.
```

### Actual Behavior

In V5.42.0 it doesnt enter anonymous mode and then tries to authenticate to the `v3/user` API and fails with a http 401 response.

#### BROKEN

``` log
[DEBUG] provider: using plugin: version=5
[DEBUG] No provider meta schema returned
[WARN]  ValidateProviderConfig from "provider[\"registry.terraform.io/integrations/github\"]" changed the config value, but that value is unused
[DEBUG] provider.terraform-provider-github_v5.42.0: 2023/11/13 12:35:45 [INFO] Selecting owner  from GITHUB_OWNER environment variable
[DEBUG] provider.terraform-provider-github_v5.42.0: 2023/11/13 12:35:45 [INFO] Setting write_delay_ms to 1000
[DEBUG] provider.terraform-provider-github_v5.42.0: 2023/11/13 12:35:45 [DEBUG] Setting read_delay_ms to 0
[DEBUG] provider.terraform-provider-github_v5.42.0: 2023/11/13 12:35:45 [DEBUG] Setting parallel_requests to false
[...]

Issue #2008, opened 2023-11-07
[BUG]: Empty or no provider information leads to error


schlichtanders:
### Expected Behavior

the github integration should work anonymously without providing any provider information

### Actual Behavior

with no `provider "github" {}` I get the following error on the most recent version
```
│ Error: Invalid provider configuration
│ 
│ Provider "registry.terraform.io/integrations/github" requires explicit configuration. Add a provider block to the root module and configure the provider's required
│ arguments as described in the provider documentation.
│ 
╵
╷
│ Error: GET https://api.github.com/user: 401 Requires authentication []
│ 
│   with provider["registry.terraform.io/integrations/github"],
│   on <empty> line 0:
│   (source code not available)
│ 
```

with setting the empty `provider "github" {}`, I still get 
```
╷
│ Error: GET https://api.github.com/user: 401 Requires authentication []
│ 
│   with provider["registry.terraform.io/integrations/github"],
│   on kube.tf line 754, in provider "github":
│  754: provider "github" {}
│ 
```

### Terraform Version

Terraform v1.5.3
on linux_amd64

Your version of Terraform is out of date! The latest version
is 1.6.3. You can update by downloading from https://www.terraform.io/downloads.html

### Affected Resource(s)

- [kube-hetzner](https://github.com/kube-hetzner/terraform-hcloud-kube-hetzner)
- probably others as well

### Steps to Reproduce

you could create a kubernetes cluster using https://github.com/kube-hetzner/terraform-hcloud-kube-hetzner and see that it fails, but I guess that is not yet minimal enough.

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

rcambrj:
This appears to have been introduced in [v5.41.0..v5.42.0](https://github.com/integrations/terraform-provider-github/compare/v5.41.0...v5.42.0)

mysticaltech:
Waiting for the fix too.

mnencia:
Solved in release 5.44.0.

@kfcampbell please close this issue.

mabilinab:
Hey! I'm getting this error again. Any tips?

PR #2074, opened 2023-12-27
Avoid getting user when anonymous

mnencia:
Resolves regressions #2008 and #2014 introduced in #1029

### Before the change?

* When anonymous, you always obtain the following initialization error
   ```
   Error: GET https://api.github.com/user: 401 Requires authentication []
   ```

### After the change?

* When anonymous, it works without errors.

### Pull request checklist

- [ ] Yes
- [X] No


kfcampbell (approved):
Thanks for catching this!

PR #2085, opened 2024-01-09
docs: Update branch protection documentation

LiamMacP:
### Before the change?

* Contexts documentation for branch protection documentation was not clear on the formatting of the strings that should be provided to correctly require action completion.

### After the change?

* Contexts documentation now includes a note that indicates the as yet 3 identified patterns that can be specified and used. 

### Pull request checklist

- [ ] Yes
- [X] No

----

The contexts attribute can contain multiple variations of strings that are not documented clearly. These variations differ greatly depending on the usecase, including:
- Typical usage with name/id
- Appending matrix value(s)
- Utilising reusable workflows.


nickfloyd (approved):
Thanks for the contributions here @LiamMacP

PR #2100, opened 2024-01-15
feat: automatic branch creation for resource 'github_repository_file'

### LINKED ISSUE #2101: [FEAT]: github_repository_file should support automatic branch creation if the ref does not exist
### Describe the need

The resource github_repository_file does not currently support automatic branch creation. For example, a potential use case is rather than directly writing changes to a long lived branch such as master/main writing to a short lived ephemeral branch to ultimately be reviewed and merged via a pull request.
It's not possible to do this gracefully without supporting automatic branch creation given if we leveraged the github_branch resource, terraform would continue to recreate the branch after it's deleted in Github (either on deletion of stale branches, or immediate deletion of the branch following a merged pull request).

### SDK Version

_No response_

### API Version

_No response_

### Relevant log output

_No response_

### Code of Conduct

wparr-circle:
Resolves #2101

----

### Before the change?

* Currently [repository_file](https://registry.terraform.io/providers/integrations/github/latest/docs/resources/repository_file) does not support automatic creation of branches. This means in order manage files via pull requests (rather than directly pushing to the branch ie. main), a [branch](https://registry.terraform.io/providers/integrations/github/latest/docs/resources/branch) would need created. This has impacts on terraform lifecycle post-creation ie. a pull request merging the branch and deleting it, in which the branch would attempt to be re-created causing a long-lived branch to exist in a repo.

### After the change?

* When the flag `autocreate_branch` is set true, if the branch does not already exist then it shall be created based on `autocreate_branch_source_sha` or `autocreate_branch_source_branch` which defaults to main during create, update or destroy. Given the branch could be ephemeral; on reads, if the branch does not exist then content shall be read from autocreate_branch_source_branch as it may been merged.

### Pull request checklist

- [ ] Yes
- [x] No

----


github/resource_github_repository_file.go
```diff
@@ -115,6 +115,24 @@ func resourceGithubRepositoryFile() *schema.Resource {
...
+			"autocreate_branch": {
+				Type:        schema.TypeBool,
+				Optional:    true,
+				Description: "Automatically create the branch if it could not be found. Subsequent reads if the branch is deleted will occur from 'autocreate_branch_source_branch'",
+				Default:     false,
+			},
+			"autocreate_branch_source_branch": {
+				Type:        schema.TypeString,
+				Default:     "main",
+				Optional:    true,
```
kfcampbell, line 127: For both this attribute and the below one, Terraform offers a `RequiredWith` attribute that we can set to ensure that `autocreate_branch_source_sha` can't be used without `autocreate_branch`. Would you mind making that change?

wparr-circle: Added the `RequiredWith` option to the schema. Thanks @kfcampbell

kfcampbell:
@wparr-circle there's a pretty quick `go fmt` issue to resolve here, then I can get this merged and released.

nickfloyd (approved):
Hey thanks for the contributions here! ❤️

PR #2102, opened 2024-01-15
feat: support signed commits for resource 'github_repository_file'

wparr-circle:
Resolves #ISSUE_NUMBER

----

### Before the change?

* 

### After the change?

* 

### Pull request checklist

- [ ] Yes
- [ ] No

----


github/resource_github_repository_file.go
```diff
@@ -10,6 +11,7 @@ import (
 
 	"github.com/google/go-github/v57/github"
 	"github.com/hashicorp/terraform-plugin-sdk/helper/schema"
+	"golang.org/x/crypto/openpgp"
```
wparr-circle, line 14: https://github.com/golang/go/issues/44226

Need to replace this library, and add some tests for the actually using the sign commit opt.

nickfloyd:
Hey @wparr-circle Thanks for the contributions here.  Please run lint when you get the chance! It looks like CI is getting hung up on that. Thanks.

wparr-circle:
Ran against linters now @nickfloyd! Thanks :)

wparr-circle:
@kfcampbell Sure no problem! Sorry if I wasn't clear.
Current implementation of this resource is utilising the [GitHub Contents API](https://docs.github.com/en/rest/repos/contents?apiVersion=2022-11-28).
We get some verified signature support using this like auto sign for bots/github actions.
However, for the use case of GPG based signing - we can't leverage the contents API. Rather we need to manipulate the git tree directly.

Does that help explain?

I left the old contents API way of working as the default behaviour, because of the size of change creeping up.

marek-karwacki-rdx:
Hi, is there a timeline on this feature? Thanks

M0NsTeRRR:
Hello, is something missing @kfcampbell to get this merged ?

ahanafy:
Landed at this PR after realizing the resource doesn't support signing. @kfcampbell do you have any direction or feedback on this PR to get it completed? Trying to get an idea on whether this feature is planned for this resource or if its not achievable?

nickfloyd (dismissed):
Hey thanks for the contributions here! ❤️

kfcampbell:
@wparr-circle we just merged #2100 and we're ready to go with this before we cut a release with your new features in it. I didn't anticipate that there might be a merge conflict between your two PRs though, and I'm wondering if you'd feel comfortable resolving the conflict. If you'd prefer that Nick and I do it, please let me know and we'll get to it!

Thanks for the contributions, and I'm looking forward to seeing commit signing in the wild.

ahanafy:
@wparr-circle / @nickfloyd / @kfcampbell I think this PR will need to have conflicts resolved and review occur pretty tightly together in order to get it past the goal. Thoughts?

rwblokzijl:
@nickfloyd / @kfcampbell Thank you for all the work on this provider! At the risk of sounding too pushy, we would love to see this PR merged. Would it be possible to review again?

wparr-circle:
@nickfloyd sorry I haven't had time to follow up on this until now. I've just resolved merge conflicts - please review when you have a chance 🙏

rwblokzijl:
I would love it if this resource supported using environment variables to set `pgp_signing_key` and `pgp_signing_key_passphrase`. That way they don't have to be stored in the state. Perhaps the specific env var to use could be passed to the resource. Eg:
```terraform
resource "github_repository_file" "file" {
  ...
  pgp_signing_key_env = "SECRET_PGP_SIGNING_KEY"
  ...
}
```

rwblokzijl:
Actually, terraform just released [Write-only attributes](https://developer.hashicorp.com/terraform/plugin/sdkv2/resources/write-only-arguments) and i think this is the perfect usecase. By marking the key as write-only it won't be persisted in the state.

deiga:
@wparr-circle Hey,

Would you be available to refactor this to use write-only attributes?

deiga:
Hey there,

Thanks for the contribution and sorry that you've had to endure such a long wait time.

We've been discussing this internally and we've come to the conclusion that we don't want to add `git` CLI interactions into the provider.

I think you might be able to achieve something similar already with the `local` provider instead

rwblokzijl:
This is a real shame. I understand this is a workaround and the "real" fix should be in the API. Is there any chance of getting that done?

deiga:
@rwblokzijl If you want changes to the API then you'll need to ask GitHub directly for that

PR #2106, opened 2024-01-16
Bump version to v6

kfcampbell:
This small change ensures our major version is set to v6.


andersthorbeck:
What was the backwards incompatible breaking change that prompted a major version bump?

kfcampbell:
#1918 is technically a breaking change, but the main reasons I'm considering this a major bump is #1780 (errors and warnings may change) and #2045 (once updated).

PR #2107, opened 2024-01-16
[WIP] Custom properties

### LINKED ISSUE #1956: [FEAT]: Support GitHub repository custom properties
### Describe the need

GitHub just announced the feature of GitHub repository custom properties. We can use the information of custom properties to apply rules to repository in GitHub organization(s). It would be good to support the feature in the terraform provider as well. By doing that, we will be able to manage custom properties and rule sets in terraform.

https://github.blog/changelog/2023-10-12-github-repository-custom-properties-beta/

### SDK Version

_No response_

### API Version

_No response_

### Relevant log output

_No response_

### Code of Conduct

alileza:
Resolves #1956

----

### Before the change?

* 

### After the change?

* 

### Pull request checklist

- [ ] Yes
- [ ] No

----


Simon-Boyer:
If I'm ready this correctly, this is only adding support to create/read custom properties definitions, would it also be possible to add support for defining custom properties values (by repo)?

github/resource_github_organisation_custom_properties.go
```diff
@@ -0,0 +1,141 @@
...
+	d.Set("value_type", customProperty.ValueType)
+
+	return nil
+}
+
+func resourceGithubCustomPropertiesUpdate(d *schema.ResourceData, meta interface{}) error {
+	if err := resourceGithubCustomPropertiesCreate(d, meta); err != nil {
+		return err
+	}
+	return resourceGithubTeamRead(d, meta)
```
roeishuster, line 123: you probably want to return the resourceGithubCustomPropertiesRead(d, meta) value, I guess this is a typo

Simon-Boyer:
> If I'm ready this correctly, this is only adding support to create/read custom properties definitions, would it also be possible to add support for defining custom properties values (by repo)?

I've looked a bit into it, and it should be possible once PR #2188 is merged to update go-github, since there was bug relating to that api that was fixed. I'm likely to try to build a working version of that feature on my side, i'll follow up there once it's done.

Pirikara:
Looks like go-github has been updated to v62 with PR https://github.com/integrations/terraform-provider-github/pull/2304 merged in place of PR https://github.com/integrations/terraform-provider-github/issues/2187 !

github/resource_github_organisation_custom_properties.go
```diff
@@ -0,0 +1,141 @@
...
+				Type:        schema.TypeBool,
+				Optional:    true,
+				Description: "Whether the custom property is required",
+			},
+			"default_value": {
+				Type:        schema.TypeString,
+				Description: "The default value of the custom property",
+				Optional:    true,
+				Computed:    true,
+			},
```
felixlut, line 48: I'm working in the repository level equivalents of these resources (i.e., `github_repository_custom_properties` as opposed to `github_organization_custom_properties`) in https://github.com/integrations/terraform-provider-github/pull/2316. The value of a custom property, and by extension the `default_value` of it as well, can either be a string or a list of strings depending on the type of custom property (`multi_select` --> list of strings). [See `Response schema` in the API docs](https://docs.github.com/en/enterprise-cloud@latest/rest/orgs/custom-properties?apiVersion=2022-11-28#get-a-custom-property-for-an-organization). 

I've opted to represent list values as a comma separated string for simplicity, and would suggest doing so here as well. It wouldn't change the schema at all, only the logic when creating a custom property of type `multi_select`

isometry: Please don't use comma-separated strings for this! We have a use-case for explicitly using comma-separated strings as the values within our multi-select custom properties :-)

felixlut: Back to the drawing board then 😅 

I guess it's possible to use something other than a comma as a separator, but that has the same problem if someone actually had a usecase for it. 

Another alternative is having two different attributes depending on the property type. ` valueList`  for ` multi_select`  and ` valueString`  for the rest. And have some logic for not allowing both to be set at the same time

isometry: What about an explicitly [dynamic](https://developer.hashicorp.com/terraform/plugin/framework/handling-data/types/dynamic) type with validation?

felixlut: Oh cool, I wasn't aware that existed! I've restricted myself to the [Terraform SDKv2 docs](https://developer.hashicorp.com/terraform/plugin/sdkv2) so far, as that is what this project is using (it upgraded from SDKv1 in Januari #1780). Dynamic values are only available if you use the [Terraform Plugin Framework](https://developer.hashicorp.com/terraform/plugin/framework). Looking at the Terraform Framework, it looks like it has some [major advantages compared to SDKv2](https://developer.hashicorp.com/terraform/plugin/framework-benefits). 

With that said, going down that path requires buy in from the maintainers of the provider itself, since it's no longer a simple feature request. There seems to be [support for using both the Framework and SDKv2 at the same time](https://developer.hashicorp.com/terraform/plugin/mux), but I don't think it's worth adding the framework just for using [Dynamic Types](https://developer.hashicorp.com/terraform/plugin/framework/handling-data/types/dynamic) for one resource. In that case it should be for the other [benefits](https://developer.hashicorp.com/terraform/plugin/framework-benefits) it brings. I think @kfcampbell is the right person to ask for opinions here

In either case, I think we should be able to figure out an interface for this that doesn't require structural changes to the provider itself,. I can see the following options: 

* String with some separator. Bad for the reasons mentioned previously in the thread
* `valueAsString` and `valueAsStringList`
* Treat all values as a list of strings. I.e., the `value` for a non-`multi_select` custom property is a list of strings of length 1, and `mutli_select` properties are a list of length x

My vote is probably on the latter. It would loke something like:

```HCL
resource "github_repository_custom_properties" "test" {
  repository = "test-custom-properties"
  properties = {
    true_false    = ["true"]
    text          = ["asd123"]
    single_select = ["option1"]
    multi_select  = ["option2", "option3"]
  }
}
```

isometry: Bother, indeed. I've only been working with the newer framework recently.
Bear in mind that whilst they're likely not yet exposed through go-github, upstream custom properties already support additional types, including boolean and multi-select, so argument per-type likely doesn't scale well into the future.

felixlut: Mhm, looks like my idea of treating everything as lists of string isn't supported by SDKv2 either...

felixlut: Hmm, actually the above is only true as long as the schema is using a Map structure, which is how I've currently implemented the [`github_repository_custom_properties` datasource](https://github.com/integrations/terraform-provider-github/pull/2316/files#diff-5c462458978462a7ab2813fa6eec63eba029f913f31edbe8519b60be89f2fd2cR23-R28), since it reads all custom properties of the repo. Nothing stops the `default_value` in this PR to be either a TypeList/TypeString

muru: I'm not familiar with the internals of Terraform providers, but would it make sense to use a `property` block instead?

```hcl
resource "github_repository_custom_properties" "test" {
  repository = "test-custom-properties"
  property {
    name  = "true_false"
    value = ["true"]
  }
  property {
    name  = "text"
    value = ["asd123"]
  }
  property {
    name  = single_select
    value = ["option1"]
  }
  property {
    name  = multi_select
    value = ["option2", "option3"]
  }
}
```

isometry: Internally, blocks are implemented more-or-less the same as object lists. Sense idiomatic to me.
Whether implemented with blocks or not, what about encoding the values as JSON to support the different property types. This is  pretty common pattern, and HCL lends itself to easy marshal/unmarshal.

I would also personally vote for this being introduced as a block on the existing github_repository resource rather than introducing another (with associated per-resource billing models from most commercial state backends)

```terraform
resource "github_repository" {
  ...

  custom_property {
    name  = "owner"
    value = jsonencode("bob")
  }

  custom_property {
    name  = "PoC"
    value = jsonencode(true)
  }

  custom_property {
    name  = "promotion_flow"
    value = jsonencode(["main","staging","production"])
  }
}
```

stevehipwell:
Given that #2316 has been merged and released this should really be a priority. @alileza are you still actively working on this?

graham1228:
➕ 1️⃣  on getting this merged. Would love to migrate our bash scripts to TF, but need this functionality.

nickfloyd:
CI is failing because it's looking for an older version of go-github @ v57 whereas the version in branch is v66.

nickfloyd:
Additionally, we will need to [merge in docs](https://github.com/integrations/terraform-provider-github/pull/2785) for this change after this PR get's merged.

PR #2129, opened 2024-02-07
fix doc for #2054: Clarify how allows_force_pushes is working

### LINKED ISSUE #2054: [BUG]: github_branch_protection allow_force_pushes permanent diff when force_push_bypassers is specified
### Expected Behavior

Value for allows_force_pushes should remain constant after apply even if force_push_bypassers are set

### Actual Behavior

allows_force_pushes sticks to false even though it is shows enabled in the GitHub console

image

image

### Terraform Version

Terraform v1.3.8
on linux_amd64
+ provider registry.terraform.io/integrations/github v5.29.0

I tested this with terraform 1.6 and the github provider vs 5.52.0 as well.

### Affected Resource(s)

github_branch_protection 

```hcl
resource "github_branch_protection" "this" {
  repository_id = github_repository.this.node_id

  pattern = "main"
  allows_force_pushes             = true
  force_push_bypassers            = [ "org/team1" ]
}
```

### Steps to Reproduce

Run terraform apply

Then run terraform plan and you should see terraform trying to change allows_force_pushes from false to true

### Debug Output

_No response_

### Panic Output

_No response_

### Code of Conduct

bpaquet:
Resolves #2054

#2054 is not really a bug, it's more a documentation issue.
At graphql level, `allowsForcePushes` will be `true` only if **everyone** is allowed to bypass. If a list of bypasser is provided, 
 `allowsForcePushes` will be `false`.

So in the terraform config, when `force_push_bypassers` is used, `allowsForcePushes` should be kept to `false.

This PR improves the doc and add a unit test.

### DIFF
--- a/github/resource_github_branch_protection_test.go
+++ b/github/resource_github_branch_protection_test.go
@@ -623,6 +623,7 @@ func TestAccGithubBranchProtection(t *testing.T) {
 			  name      = "tf-acc-test-%s"
 			  auto_init = true
 			}
+
 			resource "github_team" "test" {
 				name = "tf-acc-test-%s"
 			}
@@ -637,7 +638,7 @@ func TestAccGithubBranchProtection(t *testing.T) {
 			  repository_id  = github_repository.test.node_id
 			  pattern        = "main"
 
-				force_push_bypassers = [
+			  force_push_bypassers = [
 					"%s/${github_team.test.slug}"
 				]
 			}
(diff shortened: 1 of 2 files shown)


meliamzz (approved):

nickfloyd (approved):
Hey thanks @bpaquet  for the contributions here! ❤️

PR #2154, opened 2024-02-17
fix(data_source_github_rest_api): only allow for 404 on err

### LINKED ISSUE #2156: [BUG]: rest api swallows all errors (in order to allow for 404's)
### Expected Behavior

currently decode of json fails for certain response types, but since all exceptions that come out of the rest client are ignored there's just `null` stored in state.

### Actual Behavior

check explicitly for the response status (in case you want allow 404 as expected behaviour).

### Terraform Version

v6.0.0

### Affected Resource(s)

- data_source_github_rest_api

_No response_

### Steps to Reproduce

_No response_

### Debug Output

_No response_

### Panic Output

_No response_

### Code of Conduct

riezebosch:
Resolves #2156 

----

### Before the change?

* all exceptions are swallowed, even the json decode issue

### After the change?

* 404 is allowed as status code (keeping current behaviour), other errors are raised

### Pull request checklist

- [ ] Yes
- [x] No

----

### DIFF
--- a/github/data_source_github_rest_api.go
+++ b/github/data_source_github_rest_api.go
@@ -50,7 +50,10 @@ func dataSourceGithubRestApiRead(d *schema.ResourceData, meta interface{}) error
 		return err
 	}
 
-	resp, _ := client.Do(ctx, req, &body)
+	resp, err := client.Do(ctx, req, &body)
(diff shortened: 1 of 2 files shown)


riezebosch:
Came up with a test by using an endpoint that exceeds the maximum limit for url length.

PR #2050, opened 2023-12-06
Add support for tag-based environment deployment branch policy

mcevoypeter:
Resolves #1974

----

### Before the change?

* The `github_repository_environment_deployment_policy` only supported branch-based policies even though the [GitHub API](https://docs.github.com/en/rest/deployments/branch-policies?apiVersion=2022-11-28#create-a-deployment-branch-policy) has support for both branch-based and tag-based policies.

### After the change?

* A `tag_pattern` attribute has been added to the `github_repository_environment_deployment_policy` resource.

### Pull request checklist
- [x] Tests for the changes have been added (for bug fixes / features)
- [x] Docs have been reviewed and added / updated if needed (for bug fixes / features)

### Does this introduce a breaking change?

Please see our docs on [breaking changes](https://github.com/octokit/.github/blob/master/community/breaking_changes.md) to help!

- [ ] Yes
- [x] No

----

PR #2165, opened 2024-02-20
Add support for tag-based environment deployment branch policy

### LINKED ISSUE #1974: [FEAT]: Add support for tag pattern to `github_repository_environment_deployment_policy` resource
### Describe the need

The resource `github_repository_environment_deployment_policy` has `branch_pattern` to add branch based protection so could you also implement the same for tag based pattern. The github API already supports this as `type=tag`.

https://docs.github.com/en/enterprise-cloud@latest/rest/deployments/branch-policies?apiVersion=2022-11-28#create-a-deployment-branch-policy

### SDK Version

_No response_

### API Version

_No response_

### Relevant log output

_No response_

### Code of Conduct

sumnerwarren:
Resolves #1974. Supersedes #2050.

----

### Before the change?

* The `github_repository_environment_deployment_policy` only supported branch-based policies even though the [GitHub API](https://docs.github.com/en/rest/deployments/branch-policies?apiVersion=2022-11-28#create-a-deployment-branch-policy) has support for both branch-based and tag-based policies.

### After the change?

* A `tag_pattern` attribute has been added to the `github_repository_environment_deployment_policy` resource.

### Pull request checklist

- [ ] Yes
- [x] No

----
Because the [update API](https://docs.github.com/en/rest/deployments/branch-policies?apiVersion=2022-11-28#update-a-deployment-branch-policy) does not support pattern type (only name), I attempted to make switching between `branch_pattern` and `tag_pattern` require a new resource. Not entirely sure I did that or tested it well, so would appreciate particular attention on that aspect of this PR.


kfcampbell:
Tests are all passing for me. The chosen `branch_pattern` vs. `tag_pattern` choice seems reasonable to me. Can you explain more about 

> I attempted to make switching between branch_pattern and tag_pattern require a new resource

It doesn't seem to me like a new resource in Terraform is created here; the same `github_repository_deployment_branch_policy` is used. 

One other reasonable option might be to create a new resource, something like `github_repository_deployment_tag_policy` rather than use the branch policy one for tags.

sumnerwarren:
> Can you explain more about

The [update endpoint for deployment branch policies](https://docs.github.com/en/rest/deployments/branch-policies?apiVersion=2022-11-28#update-a-deployment-branch-policy) does not support changing from a branch pattern to a tag pattern or vice versa; it only supports pattern changes within the pattern type that was chosen at creation time. So if terraform config changes like this:

```diff
 resource "github_repository_environment_deployment_policy" "test" {
 	repository     = github_repository.test.name
 	environment    = github_repository_environment.test.environment
-	branch_pattern = "release/*"
+	tag_pattern    = "release/*"
 }
```

then this plugin marks the resource with the `ForceNew` attribute so it will be recreated.

These two tests: ([1](https://github.com/integrations/terraform-provider-github/pull/2165/files#diff-5e7e8ce91ab71e0f9ef85c58806ba9ff57b95cffe2096d7658f89cb4654f813eR462), [2](https://github.com/integrations/terraform-provider-github/pull/2165/files#diff-5e7e8ce91ab71e0f9ef85c58806ba9ff57b95cffe2096d7658f89cb4654f813eR604])) were designed to ensure that a new resource is created by comparing the unique ids of the created and updated policies. This is in contrast to these two tests ([1](https://github.com/integrations/terraform-provider-github/pull/2165/files#diff-5e7e8ce91ab71e0f9ef85c58806ba9ff57b95cffe2096d7658f89cb4654f813eR95), [2](https://github.com/integrations/terraform-provider-github/pull/2165/files#diff-5e7e8ce91ab71e0f9ef85c58806ba9ff57b95cffe2096d7658f89cb4654f813eR320)) which confirm the policy id does not change when only the pattern changes.

A simpler alternative is to always, overeagerly recreate environment deployment policies when their configuration changes.


> One other reasonable option might be to create a new resource, something like github_repository_deployment_tag_policy rather than use the branch policy one for tags.

GitHub only uses one term (deployment branch policy) for both branch-based and tag-based policies and I think there's some advantage in matching what they're doing. That said, I don't think the naming used for these resources is perfect. Right now we have both: `github_repository_deployment_branch_policy` and `github_repository_environment_deployment_policy`. This PR mostly updates the latter. I'm not sure what the history is around having two resource types that seem to represent the same thing in GitHub; maybe the former was not removed simply for backward compatibility? I would support combining and renaming these to better match what GitHub calls them, but that seems like a bigger conversation.

thulasirajkomminar:
Any update on this?

the-smooth-operator:
bump

av-andrii-ievtukhov:
Do you know if there's any progress on this? ))

aatzer:
Is there any progress on this?

avivek:
This is something that we are waiting for.
Would be great to get this in.

CarlosEspejo:
In the mean time I am using this work around that leverages the [GitHub CLI](https://cli.github.com). 
In my module I created `bin/create_tag_policy.sh` script with 

```
#!/bin/bash

# Variables
ORG=[replace-me-with-your-org]
REPO=$1
ENVIRONMENT=$2
TAG_PATTERN=$3

# Create or update the environment with the tag pattern
gh api \
  -X POST \
  -H "Accept: application/vnd.github.v3+json" \
  "/repos/$ORG/$REPO/environments/$ENVIRONMENT/deployment-branch-policies" \
  -f "name=$TAG_PATTERN" \
  -f "type=tag"

echo "Environment '$ENVIRONMENT' in repository '$REPO' has been configured with tag pattern '$TAG_PATTERN'."
```

Next I call this script using the local provisioner 

```
resource "github_repository_environment" "production" {
  count               = var.enable_production ? 1 : 0
  environment         = var.production_env_name
  repository          = var.repository_name
  prevent_self_review = true
  can_admins_bypass   = false

  reviewers {
    teams = [
      data.github_team.devops.id,
    ]
  }

  deployment_branch_policy {
    protected_branches     = false
    custom_branch_policies = true
  }

  ## Work around until Tag based rules are merged into the GitHub provider
  ## https://github.com/integrations/terraform-provider-github/pull/2165
  provisioner "local-exec" {
    command = "./bin/create_tag_policy.sh ${var.repository_name} ${var.production_env_name} ${local.production_tag_pattern}"
  }
}
```

Once it executes you will end up with:


Screenshot 2024-09-22 at 3 18 30 AM

kfcampbell (approved):
There's an obnoxious issue with the repository resource's vulnerability alerts that was breaking the tests. Adding the line `ignore_vulnerability_alerts_during_read = true` to the repository's test schema fixes it. Now the old and newly-added tests are all passing. Thank you for the contributions!

Issue #1537, opened 2023-02-09
[FEAT]: Support app auth and PAT via two provider instances


MajorBreakfast:
### Describe the need

I'd like to be able to instantiate the GitHub provider twice. One instance uses app auth. The other instance uses personal access token (PAT) auth. 

The code that I'd like to write looks like this:

```tf
provider "github" { # Uses GITHUB_APP_... env vars
  app_auth {}
}

provider "github" {  # Uses GITHUB_TOKEN env var
  alias = "pat"
}
```

The first provider block errors because, although I'm not explicitly setting the `token` argument, it is set implicitly due to the existence of the env var `GITHUB_TOKEN`.

Desired behavior:
- `app_auth {}` block set => Provider uses the `GITHUB_APP_...` env vars (currently the case*), ignores `GITHUB_TOKEN` env var (new)
- `app_auth {}` block not set => Provider uses the `GITHUB_TOKEN` env var (currently the case*), ignores `GITHUB_APP_...` env vars (currently the case*)

`*` I manually confirmed these statements about current behavior. In particular, the `GITHUB_APP_...` env vars are really ignored unless the `app_auth {}` block is set.

### Relevant log output

```shell
│ Error: "app_auth": conflicts with token
│ 
│   with provider["registry.terraform.io/hashicorp/github"],
│   on main.tf line 1, in provider "github":
│    1: provider "github" {
│
```

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

PR #2174, opened 2024-03-04
feat: app-based authentication via env vars without app_auth block

### LINKED ISSUE #1877: [FEAT]: Switching Between PAT and GitHub App Authentication Without Modifying Terraform Code
### Describe the need

Hello,

I'm trying to differentiate between authentication methods: using PAT (Personal Access Token) in my local environment and the GitHub App in the CI environment. However, after adding the app_auth block for the CI setup, I receive an error in the local environment indicating that id, installation_id, and pem_file are not set. Is there a way to toggle between authentication methods without changing the Terraform code?

Thank you for your assistance.

(Note: This message was translated with the assistance of a machine translation tool.)

### SDK Version

_No response_

### API Version

_No response_

### Relevant log output

_No response_

### Code of Conduct

laughedelic:
Resolves #1877

This PR adds new provider configuration parameters that mirror those in the `app_auth` block and make it possible to switch between token-based and app-based authentication via environment variables _without altering existing provider configuration code_. This allows flexibility of using a GitHub app for provider authentication when running in CI (or another automated environment like Atlantis), and using a personal access token when developing locally.

Related:
- https://github.com/integrations/terraform-provider-github/issues/1537 should still work as described (since `app_auth` is not going anywhere)

----

### Before the change?

* `app_auth {}` block is required for app-based provider authentication
* it's not possible to switch between token-based and app-based authentication without modifying provider configuration code

### After the change?

* it's possible to use app-based authentication by setting `GITHUB_APP_*` env vars without an `app_auth` block in the code
* same code can be used in different environments either with `GITHUB_TOKEN` or `GITHUB_APP_*` env vars

#### Example:

```hcl
provider "github" {
  owner = var.github_owner
}
```

- if `GITHUB_TOKEN` is set, provider will pick it up
- otherwise, if `GITHUB_APP_*` are set, provider will use app-based auth and generate an app installation token

### Pull request checklist
  - _Some_ tests added, but need more work. See below for details ⬇️ 

- [x] No

This was intended as a non-breaking change, so the `app_auth` block is kept and only new (redundant) parameters are added. Existing behavior is preserved.

----

### Provider auth matrix

Here I want to show different configuration scenarios and outcomes before/after. The only new case is in the first line when the `GITHUB_APP_*` env vars are set but there's no `app_auth` block in the code: before it would be an error (app vars would be ignored), but now it works as an app-based configuration.

| `GITHUB_APP_*` | `GITHUB_TOKEN` | `app_auth {}` | Before  | After   |                                       |
| :------------: | :------------: | :-----------: | ------- | ------- | ------------------------------------- |
|       ✔        |                |               | ❌ error | 🤖 app   | **new**: no `app_auth` block needed   |
|                |       ✔️        |               | 🔑 token | 🔑 token | just token auth                       |
|       ✔        |       ✔        |               | 🔑 token | 🔑 token | prioritize token for compatibility    |
|       ✔        |       ✔        |       ✔       | 🤖 app   | 🤖 app   | prioritize app auth for compatibility |
|       ✔        |                |       ✔       | 🤖 app   | 🤖 app   | `app_auth {}` is redundant            |
|                |                |       ✔       | ❌ error | ❌ error | only `app_auth {}` with no values     |

### Tests

I'm new to Go, so I need some help to write proper tests for this. I tried manual testing in `examples/app_authentication` and it worked. 

I also added some tests in `provider_test.go` following the pattern of existing test cases. But in those tests parameters are set explicitly and I'm not sure how to test the behavior of picking up parameters from the environment variables (with an empty provider configuration).

I would love to add tests for all of the cases in the matrix above, but I don't know how to approach it code-wise. Guidance would be highly appreciated 🙏 

### Code review

The main code change in `provider.go` nested existing code in an if block, so it's much easier to see the actual change if you review it with [whitespace changes ignored](https://github.com/integrations/terraform-provider-github/pull/2174/files?diff=unified&w=1).


github/provider_test.go
```diff
@@ -93,6 +93,55 @@ func TestAccProviderConfigure(t *testing.T) {
 
 	})
 
+	t.Run("can be configured with an app_auth block", func(t *testing.T) {
```
laughedelic, line 96: I didn't have luck running those tests locally:

```shell
$ GITHUB_APP_ID='xxx' GITHUB_APP_INSTALLATION_ID='xxx' GITHUB_APP_PEM_FILE='xxx' GITHUB_OWNER='xxx' TF_ACC=1 go test -v ./... -run ^TestAccProviderConfigure

...
=== RUN   TestAccProviderConfigure/can_be_configured_to_run_anonymously
    provider_test.go:60: Step 1/1 error: Error running pre-apply refresh: exit status 1
        
        Error: Duplicate provider configuration
        
          on terraform_plugin_test.tf line 3:
           3:                   provider "github" {}
        
        A default (non-aliased) provider configuration for "github" was already given
        at terraform_plugin_test.tf:1,1-18. If multiple configurations are required,
        set the "alias" argument for alternative configurations.
```

I don't quite understand where this is coming from and how to fix it.

kfcampbell: It's complaining because on line 111 below, you're already setting a provider, which is the provider defined [here](https://github.com/integrations/terraform-provider-github/blob/main/github/provider_test.go#L17-L19). 

If it makes you feel better, these tests are broken as-is, it's not you. I'm not sure if they ever even worked. 

I'm looking at the [Azure Terraform provider tests](https://github.com/hashicorp/terraform-provider-azurerm/blob/main/internal/provider/provider.go) as an example, and while they do have some configuration to tweak the provider, it's about tweaking the schema, not what's provided to it. 

I'm not sure what the best way of validating this behavior would be. [Hashicorp plugin development testing documentation](https://developer.hashicorp.com/terraform/plugin/sdkv2/testing) doesn't appear to address this case. Probably in a best case scenario, we'd add another Actions workflow with a separate configuration to run tests as a whole. Thoughts?

nicky-lenaers-phoenicks:
This would be a very nice feature indeed. Thanks for putting up this PR!

github/provider_test.go
laughedelic: Thanks for the pointers!

> these tests are broken as-is, it's not you. I'm not sure if they ever even worked

I see that all acceptance tests are failing in CI. Do these tests run at all? Or are they just ignored?

Reading the docs, I think the `Config` field in the `TestStep`s is misused for the provider configuration (which as you pointed out is already initialized). Maybe it's possible to refactor this, but I wouldn't brave it (at least not in the scope of this PR).

> Probably in a best case scenario, we'd add another Actions workflow with a separate configuration to run tests as a whole. Thoughts?

I see that in the current CI setup there is a job per authentication method: anonymous, individual and organization. Are you suggesting adding another job for the app-based authentication? I think it would generally make sense. But that would require fixing current issues with CI and setting up an app for testing.

ei-grad:
@jcudit @kfcampbell Just a quick, friendly reminder in case you have some time to review and help move this PR forward. Your feedback would be greatly appreciated!

valorl:
This would be a big quality-of-life improvement. Perhaps we can get someone else to have a look if you're busy @kfcampbell ?

briceburg:
agree -- this  is more than a QOL improvement, it focuses on sensible _environment driven_ configuration. 

there is a clear case to support auth via GITHUB token or via GITHUB_APP_* envars without having to enforce which one via provider configuration, and this PR went the extra mile (:cake:) to allow being explicit as to which auth method is prefered in case both envars are present; 
* if `app_auth {}` is set, it will prefer GITHUB_APP_ envars
* if `app_auth {}` is not set, it will prefer GITHUB_TOKEN before falling back to GITHUB_APP_ envars.

sensible indeed. is there pushback/feedback?

github/provider.go
```diff
@@ -122,6 +122,25 @@ func Provider() *schema.Provider {
...
+				Optional:    true,
+				DefaultFunc: schema.EnvDefaultFunc("GITHUB_APP_INSTALLATION_ID", nil),
+				Description: descriptions["app_auth.installation_id"],
+			},
+			"app_pem_file": {
+				Type:        schema.TypeString,
+				Optional:    true,
+				Sensitive:   true,
+				DefaultFunc: schema.EnvDefaultFunc("GITHUB_APP_PEM_FILE", nil),
+				Description: descriptions["app_auth.pem_file"],
```
deiga, line 142: If we want to add these, let's try to add some additional validation.
```suggestion
			"app_id": {
				Type:        schema.TypeString,
				Optional:    true,
				DefaultFunc: schema.EnvDefaultFunc("GITHUB_APP_ID", nil),
				Description: descriptions["app_auth.id"],
				RequiredWith: ["app_installation_id", "app_pem_file"],
				ConflictsWith: ["app_auth.0.app_id"]
			},
			"app_installation_id": {
				Type:        schema.TypeString,
				Optional:    true,
				DefaultFunc: schema.EnvDefaultFunc("GITHUB_APP_INSTALLATION_ID", nil),
				Description: descriptions["app_auth.installation_id"],
				RequiredWith: ["app_id", "app_pem_file"],
				ConflictsWith: ["app_auth.0.app_installation_id"]
			},
			"app_pem_file": {
				Type:        schema.TypeString,
				Optional:    true,
				Sensitive:   true,
				DefaultFunc: schema.EnvDefaultFunc("GITHUB_APP_PEM_FILE", nil),
				Description: descriptions["app_auth.pem_file"],
				RequiredWith: ["app_id", "app_installation_id"],
				ConflictsWith: ["app_auth.0.app_pem_file"]
```

deiga (changes requested):
Hey @laughedelic 

I'm sorry this has been sitting this long. The provider code has changed quite a bit since you suggested his PR, would you be available to rebase (or maybe just start the branch from scratch and re-do the changes)

We're really interested in getting this feature in soon (would love it into v6.11.0 if possible)

I would also challenge the idea you had here. 
While we can't be removing anything in a Minor version, together with this I think we should introduce something like `auth_mode` where a user can explicitly set `token`, `anonymous` or `app_auth`.
And together with that we should deprecate the `app_auth` block entirely.

`auth_mode` should be settable via ENV variable as well and have a default state of anonymous, so that `provider {}` still works (mostly) as before.

laughedelic:
@deiga Hi! Thanks for reviving this. I'll try to spend some time on it this weekend

stevehipwell:
This should have been resolved by #3448.

PR #2193, opened 2024-03-12
chore: Refactored tooling for best practices

stevehipwell:
This PR refactors the provider tooling to align it to GitHub Actions best practices and [TF provider best practices](https://developer.hashicorp.com/terraform/registry/providers/publishing#creating-a-github-release). The following things have changed, but there shouldn't be a material change in how the project works.

- Fixed all GitHub Actions to a SHA with the version in a comment so Dependabot can keep them up to date but the runs are idempotent (to the checked in code)
- Added Dependabot group support to make GitHub Actions & Go module dependency changes easier to manage when they're not major versions
- Updated the setup Go GH action to use the Go version in _go.mod_ instead of needing to copy the current version throughout the codebase.
- Added a _terraform-registry-manifest.json_ file as per best practices
- Updated the `golangci-lint` config to the recommended pattern
- Commented out failing linting rules in _.golangci.yml_
- Updated the `goreleaser` config


.github/workflows/release.yml
```diff
@@ -1,47 +1,45 @@
...
-      -
-        name: Set up Go
-        uses: actions/setup-go@v5
+      - uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11 # v4.1.1
         with:
-          go-version: 1.21
-      -
-        name: Import GPG key
+          # Allow goreleaser to access older tag information.
+          fetch-depth: 0
```
kfcampbell, line 23: Why is this necessary?

.github/workflows/release.yml
```diff
@@ -1,47 +1,45 @@
...
+
+      - uses: actions/setup-go@0c52d547c9bc32b1aa3301fd7a9cb496313a4491 # v5.0.0
+        with:
+          go-version-file: 'go.mod'
+          cache: true
+
+      - name: Import GPG key
+        uses: crazy-max/ghaction-import-gpg@01dd5d3ca463c7f10f7f4f7b4f177225ac661ee4 # v6.1.0
         id: import_gpg
-        uses: paultyng/ghaction-import-gpg@v2.1.0
```
kfcampbell, line 35: Do you mind checking the blame that introduced this fork in the first place? I don't want to remove it without knowing why it was introduced.

.github/workflows/release.yml
```diff
@@ -1,47 +1,45 @@
...
-        name: Run GoReleaser
-        uses: goreleaser/goreleaser-action@v5
         with:
+          gpg_private_key: ${{ secrets.GPG_PRIVATE_KEY }}
+          passphrase: ${{ secrets.PASSPHRASE }}
+
+      - name: Run GoReleaser
+        uses: goreleaser/goreleaser-action@7ec5c2b0c6cdda6e8bbb49444bc797dd33d74dd8 # v5.0.0
+        with:
+          args: release --clean
```
kfcampbell, line 40: Why add the `--clean` arg?

.goreleaser.yml
```diff
@@ -3,39 +3,40 @@
 before:
```
kfcampbell, line 3: What's the justification for the changes in this file? Where can I read about the "recommended pattern"?

kfcampbell (commented):
I appreciate the attention paid to the details here. I had a first pass of questions/comments.

.github/workflows/release.yml
stevehipwell: It's necessary for the reason in the comment. As this wasn't an option in the original checkout `v2` you had the removed fetch command below to do the same thing.

.github/workflows/release.yml
stevehipwell: I was added to support signing the release, it looks like this action is unmaintained and the TF provider template has switched to the version I've added.

.goreleaser.yml
stevehipwell: I linked to the TF provider docs in the PR description with a link to the template repo. Basically it adds the manifest and switches the YAML to standard formatting (double quotes and indentation for readability).

I've actually spotted a typo which I'm about to fix, when enabling `mod tidy`. I've re-enabled this as the criteria for taking it out looks to have passed based on the comment.

stevehipwell:
@kfcampbell I've replied to your questions.

.github/workflows/release.yml
kfcampbell: Thanks!

.github/dependabot.yml
```diff
@@ -1,11 +1,25 @@
...
-    directory: "/"
+  - package-ecosystem: github-actions
+    directory: /
+    groups:
+      github-actions:
+        patterns:
+          - "*"
+        update-types:
+          - minor
+          - patch
```
kfcampbell, line 11: Why not include major version updates here and below? Shouldn't the PR fire so we're aware the update exists even if we don't want to take it immediately?

kfcampbell (commented):
I had one final question, thank you for your patience here.

stevehipwell:
@kfcampbell I've rebased this ready to be merged.

PR #2204, opened 2024-03-21
Allow to pick a specific user from `gh` CLI to work in GH EMU scenarios

nfx:
Invoke `gh auth token ... --user ...` if the special parameter is added to provider conf. Otherwise `gh auth token` will give the token for the user that did last `gh auth login` command

### DIFF
--- a/github/provider.go
+++ b/github/provider.go
@@ -122,6 +122,12 @@ func Provider() *schema.Provider {
 					},
 				},
 			},
+			"gh_user": {
+				Type:        schema.TypeString,
+				Optional:    true,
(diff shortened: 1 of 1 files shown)


nfx:
@nickfloyd what should be the name for the provider attribute here and the default env var? will fix the linters after that

kfcampbell:
Hmm...I'm somewhat partial to `gh_cli_user` as it makes it clear the authentication is coming from the `gh` CLI in the normal scenario. Thoughts?

nfx:
@kfcampbell this functionality is required when you have more than one user authenticated, eg EMU

PR #2217, opened 2024-04-01
fix(github_repository_tag_protection): Handle non-existent tag protections in state refresh

fixe:
This commit updates the `resourceGithubRepositoryTagProtectionRead` function to remove tag protections from Terraform's state if they no longer exist in GitHub. This ensures that the state accurately reflects the actual infrastructure and that Terraform will recreate the tag protections if necessary.

### DIFF
--- a/github/resource_github_repository_tag_protection.go
+++ b/github/resource_github_repository_tag_protection.go
@@ -98,6 +98,13 @@ func resourceGithubRepositoryTagProtectionRead(d *schema.ResourceData, meta inte
 		}
 		return err
 	}
+
+	if len(tag_protection) == 0 {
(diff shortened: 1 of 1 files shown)


github/resource_github_repository_tag_protection.go
```diff
@@ -98,6 +98,13 @@ func resourceGithubRepositoryTagProtectionRead(d *schema.ResourceData, meta inte
 		}
 		return err
 	}
+
```
kfcampbell, line 101: Wouldn't we also want to do this if the call to `client.Repositories.ListTagProtection` 404s? What happened when you tested this?

fixe: Returns [], worked as expected.

PR #2232, opened 2024-04-16
fix(branch_protection_v3): ignore context when value exists in checks

### LINKED ISSUE #1701: [BUG] github_branch_protection_v3 tries to delete the contexts every time

jporzucek:
Resolves #1701 
----

### Before the change?
`github_branch_protection_v3` flattens all contexts even if they're only configured in `checks` and if `contexts` field is empty in `required_status_checks`.

### After the change?
If the context is already in checks, it's not parsed into context field.

### Pull request checklist

- [ ] Yes
- [X] No

It doesn't break anything since contexts in `checks` are still applied and they'll be applied by GitHub to `contexts`.

----


bahag-hammera:
@jporzucek Is there an Update on this PR we are waiting heavily on this fix :) @kfcampbell

rsi-mrobinson:
would be super duper awesome if someone could put eyes on this PR that potentially fixes a super annoying bug. Pretty disappointing that this has been sitting open since april.

stevehipwell:
@jporzucek I've been able to fix this issue in the acceptance tests by setting `Computed: true` for both `contexts` & `checks` which AFAIK is the idiomatic solution. So I don't think you need this additional code.

PR #2247, opened 2024-05-01
feat: data organization secret repositories

moutonjeremy:
Resolves no issue

This PR add a new data for retrieve the list of repositories of a secret of an organisation.

----

### Before the change?
* It's not possible to have the repositories list for a secret.

### After the change?

* Data action for get the repositories list for a secret

### Pull request checklist

- [ ] Yes
- [X] No

----


kfcampbell:
Can you talk to me a little more about the use case for this and why you want to add it to the provider?

moutonjeremy:
For my use case, we have a process to init/create (automatically) services (and repos) via a template but we must give access to a secret for each repo.
The service creation process is independent for each service, so we cannot manage access to the secret in centralized mode, so we must know the previous state before making a modification.
This is not an ideal situation.

moutonjeremy:
Hello, will the PR be reviewed or will it be refused?

PR #2155, opened 2024-02-17
feat: Add `github_actions_enterprise_permissions`

ErikElkins:
Resolves part of #1199 

This adds a new resource `github_actions_enterprise_permissions` to control GitHub Actions policy on an enterprise.

```tf
data "github_enterprise" "example-enterprise" {
  slug = "my-enterprise"
}

data "github_organization" "example-org" {
  name = "my-org"
}

resource "github_actions_enterprise_permissions" "test" {
  enterprise_id = data.github_enterprise.example-enterprise.slug
  allowed_actions = "selected"
  enabled_organizations = "selected"
  allowed_actions_config {
    github_owned_allowed = true 
    patterns_allowed     = ["actions/cache@*", "actions/checkout@*"]
    verified_allowed     = true
  }
  enabled_organizations_config {
    organization_ids = [data.github_organization.example-org.id]
  }
}
```

### Pull request checklist
- [x] Tests for the changes have been added (for bug fixes / features)
- [x] Docs have been reviewed and added / updated if needed (for bug fixes / features)

### Does this introduce a breaking change?

Please see our docs on [breaking changes](https://github.com/octokit/.github/blob/master/community/breaking_changes.md) to help!

- [ ] Yes
- [x] No

----

PR #2274, opened 2024-05-28
fix: Fixing `enterprise_id` import on `github_enterprise_actions_permissions`

ErikElkins:
Resolves issue where `enterprise_id` wasn't getting imported as mentioned here: https://github.com/integrations/terraform-provider-github/pull/2155#issuecomment-1993599613

----

### Before the change?

* The `enterprise_id` field wasn't getting set in state as a part of the import process of the `github_enterprise_actions_permissions` resource. Also, github/resource_github_enterprise_actions_runner_group_test.go was misspelled.

### After the change?

* The `github_enterprise_actions_permissions` now uses the `enterprise_slug` field to align with `resource_github_enterprise_actions_runner_group` and other enterprise resources, and sets it as a part of the import process.

### Pull request checklist

- [X] Yes
- [ ] No

----


kfcampbell:
@ErikElkins thanks for the contribution! I'm torn about the breaking change here -- since the behavior was broken in the first place, I wonder if it's okay to release the schema change without a corresponding major version bump, since [Hashicorp is pretty clear that releasing major versions too frequently is a bad idea](https://developer.hashicorp.com/terraform/plugin/best-practices/versioning): "To allow practitioners sufficient time and opportunity to upgrade to the latest version of the provider, we recommend releasing major versions no more than once per year. Releasing major versions more frequently could present a barrier to adoption due to the effort required to upgrade."

What do you think?

ErikElkins:
Yeah I think it's fair to avoid the major version change. It's probably only me and the one other commenter using this resource right now.

kfcampbell (approved):

PR #2278, opened 2024-05-31
docs: mark enterprise runner visibility required

gustavlasko:
If the `visibility` property is not specified on the `github_enterprise_actions_runner_group`, an error is thrown.

### Pull request checklist

Nope!

- [ ] Yes
- [X] No

----


kfcampbell (approved):
Our [API docs](https://docs.github.com/en/enterprise-cloud@latest/rest/actions/self-hosted-runner-groups?apiVersion=2022-11-28#create-a-self-hosted-runner-group-for-an-enterprise) show that this is actually optional, so the required piece is an implementation detail from this repository that could be relaxed. The smart thing to do would be to make that optional in our code. 

However, this is a good step in the meantime that corrects the docs error. Thanks!

PR #2298, opened 2024-06-26
RateLimitRoundtripper: Fix mutex leak and not respecting context cancellation

pete-woods:
…

Resolves #ISSUE_NUMBER

----

### Before the change?

* 

### After the change?

* 

### Pull request checklist

- [ ] Yes
- [ ] No

----

### DIFF
--- a/github/transport.go
+++ b/github/transport.go
@@ -2,6 +2,7 @@ package github
 
 import (
 	"bytes"
+	"context"
 	"io"
 	"log"
 	"net/http"
@@ -65,7 +66,7 @@ func (rlt *RateLimitTransport) RoundTrip(req *http.Request) (*http.Response, err
 	// for read and write requests. See isWriteMethod for the distinction between them.
 	if rlt.nextRequestDelay > 0 {
 		log.Printf("[DEBUG] Sleeping %s between operations", rlt.nextRequestDelay)
-		time.Sleep(rlt.nextRequestDelay)
+		sleep(req.Context(), rlt.nextRequestDelay)
 	}
 
 	rlt.nextRequestDelay = rlt.calculateNextDelay(req.Method)
@@ -81,6 +82,7 @@ func (rlt *RateLimitTransport) RoundTrip(req *http.Request) (*http.Response, err
 	// See https://github.com/google/go-github/pull/986
 	r1, r2, err := drainBody(resp.Body)
 	if err != nil {
+		rlt.smartLock(false)
 		return nil, err
 	}
 	resp.Body = r1
@@ -93,7 +95,7 @@ func (rlt *RateLimitTransport) RoundTrip(req *http.Request) (*http.Response, err
 		retryAfter := arlErr.GetRetryAfter()
 		log.Printf("[DEBUG] Abuse detection mechanism triggered, sleeping for %s before retrying",
 			retryAfter)
-		time.Sleep(retryAfter)
+		sleep(req.Context(), retryAfter)
 		rlt.smartLock(false)
 		return rlt.RoundTrip(req)
 	}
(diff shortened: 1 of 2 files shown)


deiga:
Hey @pete-woods 👋

Thanks for the contribution!

Would you be available to resolve the merge conflict?

Would it also be possible to explain in more detail how the changes work? 
I'm not a golang expert and for such a project having more descriptive code/comments is just a good practice ☺️

pete-woods:
Absolutely - I'll rebase when I get a moment

github/transport.go
```diff
@@ -116,6 +118,16 @@ func (rlt *RateLimitTransport) RoundTrip(req *http.Request) (*http.Response, err
 	return resp, nil
 }
 
+func sleep(ctx context.Context, dur time.Duration) {
```
pete-woods, line 122: Compared to `time.Sleep()`, this sleep function will stop sleeping as soon as the context is cancelled (e.g. if the user hits control-C or calls SIGTERM on the terraform job)

github/transport.go
pete-woods, line 77: You can see this lock is released in other exit pathways in this function.

github/transport.go
```diff
@@ -82,6 +83,7 @@ func (rlt *RateLimitTransport) RoundTrip(req *http.Request) (*http.Response, err
 	// See https://github.com/google/go-github/pull/986
 	r1, r2, err := drainBody(resp.Body)
 	if err != nil {
+		rlt.smartLock(false)
```
pete-woods, line 86: It was missing here

github/transport.go
```diff
@@ -95,7 +97,7 @@ func (rlt *RateLimitTransport) RoundTrip(req *http.Request) (*http.Response, err
 		retryAfter := arlErr.GetRetryAfter()
 		log.Printf("[WARN] Abuse detection mechanism triggered, sleeping for %s before retrying",
 			retryAfter)
-		time.Sleep(retryAfter)
+		sleep(req.Context(), retryAfter)
 		rlt.smartLock(false)
```
pete-woods, line 101: It was also called here in this exit path

pete-woods:
Okay, rebased and resolved the conflict now

pete-woods:
@nickfloyd sorry for direct @ - is there any chance you have time to look at this PR that tries to fix a deadlock/cancellation issue? (You're the #1 contributor to the repo)

deiga:
@pete-woods It's Thanksgiving, so most US-based folks won't be around. We've raised this PR as something to look at next week :)

github/transport.go
deiga: question: Instead of needing to add `rlt.smartLock(false)` to each exit pathway, could we achieve the same by adding `defer rlt.smartLock(false)` to the beginning of the function?

pete-woods: I'd have done that, and it's usually the right approach, but doing it in that way will require some refactoring.

There was some exit paths that unlock themselves IIRC, and I didn't want to get pulled into a larger refactoring.

stevehipwell:
@pete-woods could you please rebase and I'll see if we can get it merged.

pete-woods:
> @pete-woods could you please rebase and I'll see if we can get it merged.

Done

github/transport_test.go
```diff
@@ -160,6 +160,43 @@ func TestRateLimitTransport_abuseLimit_get(t *testing.T) {
...
+	defer ts.Close()
+
+	httpClient := http.DefaultClient
+	httpClient.Transport = NewRateLimitTransport(http.DefaultTransport)
+
+	client := github.NewClient(httpClient)
+	u, _ := url.Parse(ts.URL + "/")
+	client.BaseURL = u
+
+	ctx := context.WithValue(context.Background(), ctxId, t.Name())
```
stevehipwell:
```suggestion
	ctx := context.WithValue(t.Context(), ctxId, t.Name())
```
Could you use this pattern here?

pete-woods: Done

pete-woods:
> Thanks for this PR @pete-woods.
> 
> FYI I'm not sure this is going to have much of an immediate impact given the general lack usage of the context based provider methods (most call to the client use `context.Background()`). That said it will provide a noticeable benefit to resources where we adopt the context based patterns.
> 
> LGTM

It's been quite some time since I made this PR, but some brain cells are telling me it was getting stuck in the sleep between retries, at least as best as I can remember.

stevehipwell:
@pete-woods yes it does, but I suspect that it will still do so in most cases as there is no context in the legacy methods to propagate down. Your fix looks a lot like the pattern in the library I use when writing GitHub apps, which coincidentally is where I'd like to move this implementation. I think we'll also need to look at the default context timeout and add provider scoped timeout configuration for retries as the default is 20 mins.

PR #2188, opened 2024-03-09
Bump go-gitub to v60

nayuta:
Resolves #2187

----

### Before the change?

* go-github v57.0.0

### After the change?

* go-github v60.0.0
  * Fix code for
    * https://github.com/google/go-github/pull/3070
    * https://github.com/google/go-github/pull/3073

### Pull request checklist
- [ ] Tests for the changes have been added (for bug fixes / features)
- [ ] Docs have been reviewed and added / updated if needed (for bug fixes / features)

### Does this introduce a breaking change?

Please see our docs on [breaking changes](https://github.com/octokit/.github/blob/master/community/breaking_changes.md) to help!

- [ ] Yes
- [x] No

----

PR #2304, opened 2024-07-01
Bump go-github to v62 + fixes

EttoreFoti:
This is updating the core go-github package from v57 to v62 and adjusting all the resources impacted to be backwards compatible.
This is also fixing tests that were previously broken, covering most of the resources, this would be a new activity to plan IMHO.

----

### Before the change?

* go-github@v57

### After the change?

* go-github@v62

### Pull request checklist
- [X ] Tests for the changes have been added (for bug fixes / features)

- [ ] Yes
- [ X] No

----


github/resource_github_branch_protection_v3_utils.go
```diff
@@ -241,6 +241,7 @@ func flattenAndSetRestrictions(d *schema.ResourceData, protection *github.Protec
 	return d.Set("restrictions", []interface{}{})
 }
 
+// TODO: review pointer structure
```
kfcampbell, line 244: Do you mind explaining what's going on here?

github/resource_github_organization_webhook.go
```diff
@@ -219,3 +219,49 @@ func resourceGithubOrganizationWebhookDelete(d *schema.ResourceData, meta interf
 	_, err = client.Organizations.DeleteHook(ctx, orgName, hookID)
 	return err
 }
+
+func webhookConfigFromInterface(config map[string]interface{}) *github.HookConfig {
```
kfcampbell, line 223: Ahh, is this workaround for fixing the bug I was complaining about [here](https://github.com/integrations/terraform-provider-github/pull/2188#issuecomment-2187446020)? Nice!

github/resource_github_release.go
```diff
@@ -82,6 +82,56 @@ func resourceGithubRelease() *schema.Resource {
 				Type:     schema.TypeString,
 				Computed: true,
 			},
+			"release_id": {
```
kfcampbell, line 85: How come this additional schema is added in this PR? Shouldn't there be a corresponding docs change for this as well?

github/resource_github_repository_webhook.go
```diff
@@ -115,13 +109,13 @@ func resourceGithubRepositoryWebhookCreate(d *schema.ResourceData, meta interfac
...
 	// We would prefer to store the real secret in state, so we'll
 	// write the configuration secret in state from our request to GitHub
-	if hook.Config["secret"] != nil {
-		hook.Config["secret"] = hk.Config["secret"]
+	if hook.Config.Secret != nil {
+		hook.Config.Secret = hk.Config.Secret
 	}
 
-	hook.Config = insecureSslStringToBool(hook.Config)
+	// hook.Config = insecureSslStringToBool(hook.Config)
```
kfcampbell, line 116: This comment should be removed entirely rather than left in the PR; same for the similar line on 176.

kfcampbell (commented):
Thank you for your effort here! I'm looking forward to getting this in.

github/resource_github_branch_protection_v3_utils.go
EttoreFoti: This was just a note added while testing and resolved, removing in last commit

github/resource_github_organization_webhook.go
EttoreFoti: Yes, this was the nicest way to not break backwards compatibility

github/resource_github_release.go
EttoreFoti: They are already in the docs as attribute reference, they were missing from the schema I guess because it was allowed in older tf sdk versions but as of today the resource is breaking and this is fixing it. The attributes are all computed and so only attribute reference, not parameters.

kfcampbell: Ahh got it, thanks!

kfcampbell (approved):
This is awesome, thank you for updating!

PR #2316, opened 2024-07-14
feat: support repository level custom_property resource and custom_properties datasource

### LINKED ISSUE #1956: [FEAT]: Support GitHub repository custom properties

felixlut:
Contributes to #1956

Custom properties are defined on an organizational level, and then applied on a repository level. This PR aims to solve the latter of those two, while the former already has a PR for it in #2107. 

Edit: Leaving the original comment below. Update can be found below in https://github.com/integrations/terraform-provider-github/pull/2316#pullrequestreview-2456483783

This is a work in progress (WIP) since the underlying sdk:s are lacking for the [REST endpoints pertaining to the repository level custom properties](https://docs.github.com/en/enterprise-cloud@latest/rest/repos/custom-properties?apiVersion=2022-11-28#about-custom-properties):

* Regular `go-github` lacks support for the `multi_select` custom property type, see https://github.com/google/go-github/issues/3198. There is a PR close to being merged for solving this https://github.com/google/go-github/pull/3200
    * The PR is merged, but no new release of `go-github` has been released since
* As @kfcampbell mentioned [here](https://github.com/integrations/terraform-provider-github/issues/1956#issuecomment-1986141803), the GitHub SDK team has created a SDK generated from GitHub's OpenAPI spec. I've added and used it to implement a `github_repository_custom_properties` datasource with it due to the lacking functionality of `go-github` in this area currently. I'm currently trying to get it to write the custom properties, which I've been unsuccessful in so far. See https://github.com/octokit/go-sdk/issues/90

If there's any progress on either of these I'll update this PR

----

### Before the change?

* It was not possible to add a custom property to a repo before this change (or read them)

### After the change?

* It is now possible to add a custom property to a repo before this change (or read them)

### Pull request checklist

- [ ] Yes
- [x] No

----


ViktorLindgren95:
@felixlut  It seems that v64 https://github.com/google/go-github/pull/3240 added support for Multi select finally

felixlut:
> @felixlut It seems that v64 [google/go-github#3240](https://github.com/google/go-github/pull/3240) added support for Multi select finally. Only issue is that the terraform provider codebase is still at v63

Seems like #2359 will deal with bumping the dep, so I'll revisit this once it's merged

felixlut:
I started looking into this again over the weekend but hit a bottleneck again. I made a fix in https://github.com/google/go-github/pull/3309, but that requires another bump of the `go-github` provider. Hopefully that is the last blocker before we can use that provider fully for this feature. Reason I'd rather use that than the [`go-sdk`](https://github.com/octokit/go-sdk) is that the PR simply becomes huge, and anytime another PR is merged I have to resolve conflicts (or at least run the risk of it). Dealing with that on a long-living PR like this is pretty painful. Using the `go-github` will skip that pain 😅

github/data_source_github_repository_custom_properties.go
```diff
@@ -0,0 +1,103 @@
...
+		Read: dataSourceGithubOrgaRepositoryCustomProperties,
+
+		Schema: map[string]*schema.Schema{
+			"repository": {
+				Type:        schema.TypeString,
+				Required:    true,
+				Description: "Name of the repository which the custom properties should be on.",
+			},
+			"property": {
+				Type:        schema.TypeSet,
```
felixlut, line 22: The way I see most people using this resource is to:

1. Read the custom properties of a repo
2. For each specific property do something based on it's value. Eg, do X to the repo if the `owner` property equals Y, do Z to the repo if the `environment` property equals ...

This means that it should be as easy as possible to get the value of a specific property. IMO that would be best achieved by using a [`TypeMap`](https://developer.hashicorp.com/terraform/plugin/sdkv2/schemas/schema-types#typemap). Unfortunately that is not supported by v2 of the terraform-plugin-sdk ([see thread](https://github.com/integrations/terraform-provider-github/pull/2107#discussion_r1677557505)). 

The choice is then between [``TypeList``](https://developer.hashicorp.com/terraform/plugin/sdkv2/schemas/schema-types#typelist) or [``TypeSet``](https://developer.hashicorp.com/terraform/plugin/sdkv2/schemas/schema-types#typeset). Reading the docs the latter sounds more correct to me, as the order of the custom properties does not matter. We've seen issues regarding this in this provider in the past, eg in #1950, that should be avoided by using `TypeSet`.

It's worth noting that there are some other implications of using `TypeSet` though, namely that they are [non-indexable](https://github.com/hashicorp/terraform-website/issues/1495#issuecomment-887841699). [This common pattern](https://stackoverflow.com/a/59437020) will for example not work, and throw the error below:

```HCL
data "github_repository_custom_properties" "test" {
  repository    = "test-custom-properties"
}
locals {
  # Assuming the repo have a property called "string" set already
  test = data.github_repository_custom_properties.test.property[index(data.github_repository_custom_properties.test.property.*.property_name, "string")]
}
```

image

That does not mean that it's impossible to fetch the value of a custom property though. The snippet below does work, and is IMO more readable. But I think I'd be amiss if I didn't mention this. 

```HCL
data "github_repository_custom_properties" "test" {
  repository    = "test-custom-properties"
}
locals {
  test = one([ for prop in data.github_repository_custom_properties.test.property: prop if prop.property_name == "string" ])
}
```

I personally lean towards using a `TypeSet` to avoid possible weird diffs, but I don't feel strongly about it. If anyone have any opinions here I'm all ears 😄

felixlut (commented):
Aight, long time coming but I finally think this PR is in a reviewable state! I've made some comments on certain parts where I'm not 100% sure on the design. Of the bat these 2 things are noteworthy as well:

1. I opted for a non-authoriative custom_propert**Y** resource as opposed to an authoriative custom_propert**IES** resource. Reason being that fulfills the use-case I have. I think an authoriative resource should exist, as it has its uses similarily to [github_repository_collaborators](https://registry.terraform.io/providers/integrations/github/latest/docs/resources/repository_collaborators). Nothing stops later PR:s doing just that 
    * On a similar note I opted opted out of adding `custom_property` to `github_repository`, as proposed [here](https://github.com/integrations/terraform-provider-github/pull/2107#discussion_r1811119876) for the same reason. Nothing stops future PR:s here either. Docs just needs to reflect that both shouldn't be used as the same time
2. For the data_source I opted for custom_propert**IES** over reading each property individually. The reason for this is that the [API endpoint for reading custom properties on a repo](https://docs.github.com/en/enterprise-cloud@latest/rest/repos/custom-properties?apiVersion=2022-11-28#get-all-custom-property-values-for-a-repository) returns ALL properties. Filtering that full list and returning just 1 property seemed redundant to me. With that said creating a custom_propert**Y** data source should be trivial with a future PR 😄

github/data_source_github_repository_custom_properties.go
stevehipwell: @felixlut have you considered also adding a `github_repository_custom_property` data source to access a singular property? It would allow for more direct access to properties at the cost of increased API usage.

felixlut: [I have](https://github.com/integrations/terraform-provider-github/pull/2316#pullrequestreview-2456483783), and nothing stops future PR:s for doing just that. The actual implementation would be to simply filter it after [this call](https://github.com/integrations/terraform-provider-github/blob/15b61e7ed0a568dd8b0cb0b61492d8896e5c20e5/github/data_source_github_repository_custom_properties.go#L56-L59)

ChaosCypher:
🚀

hashildy:
Any update on getting this PR merged? This would be a really useful enhancement to support custom properties!

hashildy:
@stevehipwell - is this PR able to be reviewed/merged?

stevehipwell:
@hashildy I'm just a contributor, I have 9 PRs that have been ready to review merge in the last 3 months that haven't had a single comment by a maintainer.

tayven-bigelow:
@kfcampbell - Any chance of getting a review / approval of this PR? 
Custom properties would be great to have support around in terraform.

kfcampbell (approved):

PR #2345, opened 2024-08-11
Rate limit

### LINKED ISSUE #2344: [BUG]: Repository files deleted on rate limiting
### Expected Behavior

When the token in use hits rate limiting the file being read should not be removed from the state

### Actual Behavior

The file is removed from state resulting in the file needing to be imported back to state.

### Terraform Version

Tested on multiple versions of provider and terraform, I don't believe this is version specific.

### Affected Resource(s)

- github_repository_file 

_No response_

### Steps to Reproduce

_No response_

### Debug Output

_No response_

### Panic Output

_No response_

### Code of Conduct

grahamhar:
Resolves #2344

----

### Before the change?

This change stops the file from being removed from the state when the operation is rate limited, I am unsure if we should also handle 5xx here too. 

### After the change?

The operation fails with an error due to rate limiting leaving the file in the state.

### Pull request checklist
- [NA] Tests for the changes have been added (for bug fixes / features)
- [NA] Docs have been reviewed and added / updated if needed (for bug fixes / features)

I don't believe I can add an acceptance test for this, there don't appear to be any unit level tests so have not added anything. Happy to look at this on your guidance.

- [ ] Yes
- [X] No

----


kfcampbell (approved):
Thank you for the attention to the integration tests. I agree it'd be difficult to write a test with our current setup to force a 429. I'll accept your manual testing for now.

PR #2366, opened 2024-08-26
feat: add github_user_ssh_signing_key

### LINKED ISSUE #1917: [FEAT]: Support SSH signing keys
### Describe the need

For quite some time, GitHub supports signing commits with an SSH key. You can add an SSH key in the UI for authentication or signing. Currently, this provider only supports SSH keys for authentication, not for signing.

### SDK Version

_No response_

### API Version

_No response_

### Relevant log output

_No response_

### Code of Conduct

### LINKED ISSUE #2365: add difference of auth and signing keys to github_user_ssh_key
### Describe the need

Currently the `github_user_ssh_key` resource can only be used to manage "ssh keys". There is no difference between the authentication and signing ssh keys.

I would like to manage both key types via terraform. 

### SDK Version

_No response_

### API Version

_No response_

### Relevant log output

_No response_

### Code of Conduct

simonostendorf:
Resolves #2365

Resolves #1917

----

### Before the change?

* managing ssh signing keys was not possible

### After the change?

* managing ssh signing keys is possible with `github_user_ssh_signing_key` resource
* the resource is equal to the existing `github_user_ssh_key` resource that manages ssh auth keys

### Pull request checklist

- [ ] Yes
- [x] No

----

### DIFF
--- a/github/resource_github_user_ssh_key_migration.go
+++ b/github/resource_github_user_ssh_key_migration.go
@@ -1,0 +1,60 @@
+package github
+
+import (
+	"context"
+	"fmt"
+	"strconv"
+	"strings"
+
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+)
+
+func resourceGithubUserSshKeyV0() *schema.Resource {
+	return &schema.Resource{
+		SchemaVersion: 0,
+		Schema: map[string]*schema.Schema{
+			"title": {
+				Type:        schema.TypeString,
+				Required:    true,
+				ForceNew:    true,
+				Description: "A descriptive name for the new key.",
+			},
+			"key": {
+				Type:        schema.TypeString,
+				Required:    true,
+				ForceNew:    true,
+				Description: "The public SSH key to add to your GitHub account.",
+				DiffSuppressFunc: func(k, oldV, newV string, d *schema.ResourceData) bool {
+					newTrimmed := strings.TrimSpace(newV)
+					return oldV == newTrimmed
+				},
+			},
+			"url": {
+				Type:        schema.TypeString,
+				Computed:    true,
+				Description: "The URL of the SSH key.",
+			},
+			"etag": {
+				Type:     schema.TypeString,
+				Computed: true,
+			},
+		},
+	}
+}
+
+func resourceGithubUserSshKeyStateUpgradeV0(ctx context.Context, rawState map[string]any, m any) (map[string]any, error) {
+	if rawState == nil {
+		return nil, fmt.Errorf("resource state upgrade failed, state is nil")
+	}
+
+	// copy d.Id() into key_id
+	if id, ok := rawState["id"].(string); ok {
+		keyID, err := strconv.ParseInt(id, 10, 64)
+		if err != nil {
+			return nil, fmt.Errorf("resource state upgrade failed, invalid SSH key ID format: %w", err)
+		}
+		rawState["key_id"] = keyID
+	}
+
+	return rawState, nil
+}
--- a/github/resource_github_user_ssh_signing_key.go
+++ b/github/resource_github_user_ssh_signing_key.go
@@ -12,18 +12,17 @@ import (
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
 )
 
-func resourceGithubUserSshKey() *schema.Resource {
+func resourceGithubUserSshSigningKey() *schema.Resource {
 	return &schema.Resource{
-		CreateContext: resourceGithubUserSshKeyCreate,
-		ReadContext:   resourceGithubUserSshKeyRead,
-		DeleteContext: resourceGithubUserSshKeyDelete,
+		CreateContext: resourceGithubUserSshSigningKeyCreate,
+		ReadContext:   resourceGithubUserSshSigningKeyRead,
+		DeleteContext: resourceGithubUserSshSigningKeyDelete,
 		Importer: &schema.ResourceImporter{
-			StateContext: resourceGithubUserSshKeyImport,
+			StateContext: resourceGithubUserSshSigningKeyImport,
 		},
 
-		Description: "Manages a SSH key for the authenticated user.",
+		Description: "Manages a SSH signing key for the authenticated user.",
 
-		SchemaVersion: 1,
 		Schema: map[string]*schema.Schema{
 			"title": {
 				Type:        schema.TypeString,
@@ -35,41 +34,28 @@ func resourceGithubUserSshKey() *schema.Resource {
 				Type:        schema.TypeString,
 				Required:    true,
 				ForceNew:    true,
-				Description: "The public SSH key to add to your GitHub account.",
-			},
-			"url": {
-				Type:        schema.TypeString,
-				Computed:    true,
-				Description: "The URL of the SSH key.",
+				Description: "The public SSH signing key to add to your GitHub account.",
 			},
 			"key_id": {
 				Type:        schema.TypeInt,
 				Computed:    true,
-				Description: "The unique identifier of the SSH key.",
+				Description: "The unique identifier of the SSH signing key.",
 			},
 			"etag": {
 				Type:     schema.TypeString,
 				Computed: true,
 			},
 		},
-
-		StateUpgraders: []schema.StateUpgrader{
-			{
-				Version: 0,
-				Type:    resourceGithubUserSshKeyV0().CoreConfigSchema().ImpliedType(),
-				Upgrade: resourceGithubUserSshKeyStateUpgradeV0,
-			},
-		},
 	}
 }
 
-func resourceGithubUserSshKeyCreate(ctx context.Context, d *schema.ResourceData, meta any) diag.Diagnostics {
+func resourceGithubUserSshSigningKeyCreate(ctx context.Context, d *schema.ResourceData, meta any) diag.Diagnostics {
 	client := meta.(*Owner).v3client
 
 	title := d.Get("title").(string)
 	key := d.Get("key").(string)
 
-	userKey, resp, err := client.Users.CreateKey(ctx, &github.Key{
+	userKey, resp, err := client.Users.CreateSSHSigningKey(ctx, &github.Key{
 		Title: new(title),
 		Key:   new(key),
 	})
@@ -85,9 +71,6 @@ func resourceGithubUserSshKeyCreate(ctx context.Context, d *schema.ResourceData,
 	if err = d.Set("etag", resp.Header.Get("ETag")); err != nil {
 		return diag.FromErr(err)
 	}
-	if err = d.Set("url", userKey.GetURL()); err != nil {
-		return diag.FromErr(err)
-	}
 	if err = d.Set("title", userKey.GetTitle()); err != nil {
 		return diag.FromErr(err)
 	}
@@ -95,40 +78,28 @@ func resourceGithubUserSshKeyCreate(ctx context.Context, d *schema.ResourceData,
 	return nil
 }
 
-func resourceGithubUserSshKeyRead(ctx context.Context, d *schema.ResourceData, meta any) diag.Diagnostics {
+func resourceGithubUserSshSigningKeyRead(ctx context.Context, d *schema.ResourceData, meta any) diag.Diagnostics {
 	client := meta.(*Owner).v3client
 
 	keyID := int64(d.Get("key_id").(int))
-	userKey, resp, err := client.Users.GetKey(ctx, keyID)
+	_, resp, err := client.Users.GetSSHSigningKey(ctx, keyID)
 	if err != nil {
-		return diag.FromErr(deleteResourceOn404AndSwallow304OtherwiseReturnError(err, d, "user SSH key (%d)", keyID))
+		return diag.FromErr(deleteResourceOn404AndSwallow304OtherwiseReturnError(err, d, "user SSH signing key (%d)", keyID))
 	}
 
 	// set computed fields
 	if err = d.Set("etag", resp.Header.Get("ETag")); err != nil {
 		return diag.FromErr(err)
 	}
-	if err = d.Set("url", userKey.GetURL()); err != nil {
-		return diag.FromErr(err)
-	}
 
 	return nil
 }
 
-func resourceGithubUserSshKeyDelete(ctx context.Context, d *schema.ResourceData, meta any) diag.Diagnostics {
+func resourceGithubUserSshSigningKeyDelete(ctx context.Context, d *schema.ResourceData, meta any) diag.Diagnostics {
 	client := meta.(*Owner).v3client
 
 	keyID := int64(d.Get("key_id").(int))
-	// fallback to d.Id() for backward compatibility when key_id is not set
-	if keyID == 0 {
-		var err error
-		keyID, err = strconv.ParseInt(d.Id(), 10, 64)
-		if err != nil {
-			return diag.FromErr(fmt.Errorf("invalid SSH key ID format: %w", err))
-		}
-	}
-
-	resp, err := client.Users.DeleteKey(ctx, keyID)
+	resp, err := client.Users.DeleteSSHSigningKey(ctx, keyID)
 	if err != nil {
 		if resp != nil && resp.StatusCode == http.StatusNotFound {
 			return nil
@@ -138,20 +109,20 @@ func resourceGithubUserSshKeyDelete(ctx context.Context, d *schema.ResourceData,
 	return nil
 }
 
-func resourceGithubUserSshKeyImport(ctx context.Context, d *schema.ResourceData, meta any) ([]*schema.ResourceData, error) {
+func resourceGithubUserSshSigningKeyImport(ctx context.Context, d *schema.ResourceData, meta any) ([]*schema.ResourceData, error) {
 	client := meta.(*Owner).v3client
 
 	keyID, err := strconv.ParseInt(d.Id(), 10, 64)
 	if err != nil {
-		return nil, fmt.Errorf("invalid SSH key ID format: %w", err)
+		return nil, fmt.Errorf("invalid SSH signing key ID format: %w", err)
 	}
 
-	key, _, err := client.Users.GetKey(ctx, keyID)
+	key, _, err := client.Users.GetSSHSigningKey(ctx, keyID)
 	if err != nil {
 		var ghErr *github.ErrorResponse
 		if errors.As(err, &ghErr) {
 			if ghErr.Response.StatusCode == http.StatusNotFound {
-				return nil, fmt.Errorf("SSH key with ID %d not found", keyID)
+				return nil, fmt.Errorf("SSH signing key with ID %d not found", keyID)
 			}
 		}
 		return nil, err
--- a/github/resource_github_user_ssh_signing_key_test.go
+++ b/github/resource_github_user_ssh_signing_key_test.go
@@ -15,65 +15,59 @@ import (
 	"golang.org/x/crypto/ssh"
 )
 
-func TestAccGithubUserSshKey(t *testing.T) {
-	t.Parallel()
-
-	t.Run("creates and destroys a user SSH key without error", func(t *testing.T) {
-		t.Parallel()
-
+func TestAccGithubUserSshSigningKey(t *testing.T) {
+	t.Run("creates and destroys a user SSH signing key without error", func(t *testing.T) {
 		randomID := acctest.RandStringFromCharSet(5, acctest.CharSetAlphaNum)
 		name := fmt.Sprintf(`%s-%s`, testResourcePrefix, randomID)
-		testKey := newTestKey()
+		testKey := newTestSigningKey()
 
 		config := fmt.Sprintf(`
-			resource "github_user_ssh_key" "test" {
+			resource "github_user_ssh_signing_key" "test" {
 				title = "%[1]s"
 				key   = "%[2]s"
 			}
 		`, name, testKey)
 
 		resource.Test(t, resource.TestCase{
-			PreCheck:          func() { skipUnlessMode(t, individual) },
+			PreCheck:          func() { skipUnauthenticated(t) },
 			ProviderFactories: providerFactories,
 			Steps: []resource.TestStep{
 				{
 					Config: config,
 					ConfigStateChecks: []statecheck.StateCheck{
-						statecheck.ExpectKnownValue("github_user_ssh_key.test", tfjsonpath.New("title"), knownvalue.StringExact(name)),
-						statecheck.ExpectKnownValue("github_user_ssh_key.test", tfjsonpath.New("key"), knownvalue.StringExact(testKey)),
+						statecheck.ExpectKnownValue("github_user_ssh_signing_key.test", tfjsonpath.New("title"), knownvalue.StringExact(name)),
+						statecheck.ExpectKnownValue("github_user_ssh_signing_key.test", tfjsonpath.New("key"), knownvalue.StringExact(testKey)),
 					},
 				},
 			},
 		})
 	})
 
-	t.Run("imports an individual account SSH key without error", func(t *testing.T) {
-		t.Parallel()
-
+	t.Run("imports an individual account SSH signing key without error", func(t *testing.T) {
 		randomID := acctest.RandStringFromCharSet(5, acctest.CharSetAlphaNum)
 		name := fmt.Sprintf(`%s-%s`, testResourcePrefix, randomID)
-		testKey := newTestKey()
+		testKey := newTestSigningKey()
 
 		config := fmt.Sprintf(`
-			resource "github_user_ssh_key" "test" {
+			resource "github_user_ssh_signing_key" "test" {
 				title = "%[1]s"
 				key   = "%[2]s"
 			}
 		`, name, testKey)
 
 		resource.Test(t, resource.TestCase{
-			PreCheck:          func() { skipUnlessMode(t, individual) },
+			PreCheck:          func() { skipUnauthenticated(t) },
 			ProviderFactories: providerFactories,
 			Steps: []resource.TestStep{
 				{
 					Config: config,
 					ConfigStateChecks: []statecheck.StateCheck{
-						statecheck.ExpectKnownValue("github_user_ssh_key.test", tfjsonpath.New("title"), knownvalue.StringExact(name)),
-						statecheck.ExpectKnownValue("github_user_ssh_key.test", tfjsonpath.New("key"), knownvalue.StringExact(testKey)),
+						statecheck.ExpectKnownValue("github_user_ssh_signing_key.test", tfjsonpath.New("title"), knownvalue.StringExact(name)),
+						statecheck.ExpectKnownValue("github_user_ssh_signing_key.test", tfjsonpath.New("key"), knownvalue.StringExact(testKey)),
 					},
 				},
 				{
-					ResourceName:      "github_user_ssh_key.test",
+					ResourceName:      "github_user_ssh_signing_key.test",
 					ImportState:       true,
 					ImportStateVerify: true,
 				},
@@ -82,7 +76,7 @@ func TestAccGithubUserSshKey(t *testing.T) {
 	})
 }
 
-func newTestKey() string {
+func newTestSigningKey() string {
 	privateKey, _ := rsa.GenerateKey(rand.Reader, 1024)
 	publicKey, _ := ssh.NewPublicKey(&privateKey.PublicKey)
 	return strings.TrimRight(string(ssh.MarshalAuthorizedKey(publicKey)), "\n")
--- a/templates/resources/user_ssh_signing_key.md.tmpl
+++ b/templates/resources/user_ssh_signing_key.md.tmpl
@@ -1,0 +1,38 @@
+---
+page_title: "{{.Name}} ({{.Type}}) - {{.RenderedProviderName}}"
+description: |-
+  Provides a GitHub user's SSH signing key resource.
+---
+
+# {{.Name}} ({{.Type}})
+
+Provides a GitHub user's SSH signing key resource.
+
+This resource allows you to add/remove SSH signing keys from your user account.
+
+## Example Usage
+
+{{ tffile "examples/resources/user_ssh_signing_key/example_1.tf" }}
+
+## Argument Reference
+
+The following arguments are supported:
+
+- `title` - (Required) A descriptive name for the new key.
+- `key` - (Required) The public SSH signing key to add to your GitHub account.
+
+## Attributes Reference
+
+The following attributes are exported:
+
+- `id` - The ID of the SSH signing key
+- `key_id` - The unique identifier of the SSH signing key.
+- `etag`
+
+## Import
+
+SSH signing keys can be imported using their ID e.g.
+
+```shell
+terraform import github_user_ssh_signing_key.example 1234567
+```
--- a/docs/resources/user_ssh_key.md
+++ b/docs/resources/user_ssh_key.md
@@ -32,6 +32,8 @@ The following attributes are exported:
 
 - `id` - The ID of the SSH key
 - `url` - The URL of the SSH key
+- `key_id` - The unique identifier of the SSH key.
+- `etag`
 
 ## Import
 
--- a/docs/resources/user_ssh_signing_key.md
+++ b/docs/resources/user_ssh_signing_key.md
@@ -1,0 +1,43 @@
+---
+page_title: "github_user_ssh_signing_key (Resource) - GitHub"
+description: |-
+  Provides a GitHub user's SSH signing key resource.
+---
+
+# github_user_ssh_signing_key (Resource)
+
+Provides a GitHub user's SSH signing key resource.
+
+This resource allows you to add/remove SSH signing keys from your user account.
+
+## Example Usage
+
+```terraform
+resource "github_user_ssh_signing_key" "example" {
+  title = "example title"
+  key   = file("~/.ssh/id_rsa.pub")
+}
+```
+
+## Argument Reference
+
+The following arguments are supported:
+
+- `title` - (Required) A descriptive name for the new key.
+- `key` - (Required) The public SSH signing key to add to your GitHub account.
+
+## Attributes Reference
+
+The following attributes are exported:
+
+- `id` - The ID of the SSH signing key
+- `key_id` - The unique identifier of the SSH signing key.
+- `etag`
+
+## Import
+
+SSH signing keys can be imported using their ID e.g.
+
+```shell
+terraform import github_user_ssh_signing_key.example 1234567
+```
--- a/examples/resources/user_ssh_signing_key/example_1.tf
+++ b/examples/resources/user_ssh_signing_key/example_1.tf
@@ -1,0 +1,4 @@
+resource "github_user_ssh_signing_key" "example" {
+  title = "example title"
+  key   = file("~/.ssh/id_rsa.pub")
+}
--- a/github/acc_test.go
+++ b/github/acc_test.go
@@ -270,6 +270,16 @@ func configureSweepers() {
 		Name: "teams",
 		F:    sweepTeams,
 	})
+
+	resource.AddTestSweepers("user_ssh_keys", &resource.Sweeper{
+		Name: "user_ssh_keys",
+		F:    sweepUserSSHKeys,
+	})
+
+	resource.AddTestSweepers("user_ssh_signing_keys", &resource.Sweeper{
+		Name: "user_ssh_signing_keys",
+		F:    sweepUserSSHSigningKeys,
+	})
 }
 
 func sweepTeams(_ string) error {
@@ -324,6 +334,56 @@ func sweepRepositories(_ string) error {
 			fmt.Printf("destroying repository %s\n", name)
 
 			if _, err := client.Repositories.Delete(ctx, owner, name); err != nil {
+				return err
+			}
+		}
+	}
+
+	return nil
+}
+
+func sweepUserSSHKeys(_ string) error {
+	fmt.Println("sweeping user SSH keys")
+
+	client := testAccConf.meta.v3client
+	owner := testAccConf.meta.name
+	ctx := context.Background()
+
+	keys, _, err := client.Users.ListKeys(ctx, owner, nil)
+	if err != nil {
+		return err
+	}
+
+	for _, k := range keys {
+		if title := k.GetTitle(); strings.HasPrefix(title, testResourcePrefix) {
+			fmt.Printf("destroying user SSH key %s\n", title)
+
+			if _, err := client.Users.DeleteKey(ctx, k.GetID()); err != nil {
+				return err
+			}
+		}
+	}
+
+	return nil
+}
+
+func sweepUserSSHSigningKeys(_ string) error {
+	fmt.Println("sweeping user SSH signing keys")
+
+	client := testAccConf.meta.v3client
+	owner := testAccConf.meta.name
+	ctx := context.Background()
+
+	keys, _, err := client.Users.ListSSHSigningKeys(ctx, owner, nil)
+	if err != nil {
+		return err
+	}
+
+	for _, k := range keys {
+		if title := k.GetTitle(); strings.HasPrefix(title, testResourcePrefix) {
+			fmt.Printf("destroying user SSH signing key %s\n", title)
+
+			if _, err := client.Users.DeleteSSHSigningKey(ctx, k.GetID()); err != nil {
 				return err
 			}
 		}
--- a/github/provider.go
+++ b/github/provider.go
@@ -240,6 +240,7 @@ func NewProvider(version, commit string) func() *schema.Provider {
 				"github_user_gpg_key":                                                   resourceGithubUserGpgKey(),
 				"github_user_invitation_accepter":                                       resourceGithubUserInvitationAccepter(),
 				"github_user_ssh_key":                                                   resourceGithubUserSshKey(),
+				"github_user_ssh_signing_key":                                           resourceGithubUserSshSigningKey(),
 				"github_enterprise_organization":                                        resourceGithubEnterpriseOrganization(),
 				"github_enterprise_actions_runner_group":                                resourceGithubActionsEnterpriseRunnerGroup(),
 				"github_enterprise_ip_allow_list_entry":                                 resourceGithubEnterpriseIpAllowListEntry(),
--- a/github/resource_github_user_ssh_key_migration_test.go
+++ b/github/resource_github_user_ssh_key_migration_test.go
@@ -1,0 +1,49 @@
+package github
+
+import (
+	"context"
+	"reflect"
+	"testing"
+)
+
+func Test_resourceGithubUserSshKeyStateUpgradeV0(t *testing.T) {
+	t.Parallel()
+
+	for _, d := range []struct {
+		testName    string
+		rawState    map[string]any
+		want        map[string]any
+		shouldError bool
+	}{
+		{
+			testName: "migrates_v0_to_v1",
+			rawState: map[string]any{
+				"id":    "123",
+				"title": "test-key",
+				"key":   "test-key-data",
+				"url":   "test-url",
+			},
+			want: map[string]any{
+				"id":     "123",
+				"key_id": int64(123),
+				"title":  "test-key",
+				"key":    "test-key-data",
+				"url":    "test-url",
+			},
+			shouldError: false,
+		},
+	} {
+		t.Run(d.testName, func(t *testing.T) {
+			t.Parallel()
+
+			got, err := resourceGithubUserSshKeyStateUpgradeV0(context.Background(), d.rawState, nil)
+			if (err != nil) != d.shouldError {
+				t.Fatalf("unexpected error state")
+			}
+
+			if !d.shouldError && !reflect.DeepEqual(got, d.want) {
+				t.Fatalf("got %+v, want %+v", got, d.want)
+			}
+		})
+	}
+}


kfcampbell:
@simonostendorf can you please link to the new doc you created in the `github.erb` file so it's discoverable?

simonostendorf:
> @simonostendorf can you please link to the new doc you created in the `github.erb` file so it's discoverable?

@kfcampbell sorry forgot that. I added it now.

quotidian-ennui:
#1917 should perhaps be formally linked to this PR as well since it appears equivalent to #2365

simonostendorf:
@kfcampbell hey, can we get this merged?

kilianpaquier:
Would love to see this merged

github/resource_github_user_ssh_signing_key.go
```diff
@@ -0,0 +1,119 @@
...
+	client := meta.(*Owner).v3client
+
+	id, err := strconv.ParseInt(d.Id(), 10, 64)
+	if err != nil {
+		return unconvertibleIdErr(d.Id(), err)
+	}
+	ctx := context.WithValue(context.Background(), ctxId, d.Id())
+	if !d.IsNewResource() {
+		ctx = context.WithValue(ctx, ctxEtag, d.Get("etag").(string))
+	}
```
deiga:
```suggestion
	if !d.IsNewResource() {
		ctx = context.WithValue(ctx, ctxEtag, d.Get("etag").(string))
	}
```

deiga (changes requested):

deiga:
Hey @simonostendorf 
Sorry that you have needed to wait for so long. Could you rebase your branch and adjust your code to fit our newer standards?

- Use Contex-aware functions: https://github.com/integrations/terraform-provider-github/issues/2996
- Use `tflog` for logging: https://github.com/integrations/terraform-provider-github/issues/3070
- Adhere to new testing structure

simonostendorf:
> Hey @simonostendorf Sorry that you have needed to wait for so long. Could you rebase your branch and adjust your code to fit our newer standards?
> 
> * Use Contex-aware functions: [[MAINT] Migrate all resources and data sources to Context-aware CRUD functions #2996](https://github.com/integrations/terraform-provider-github/issues/2996)
> * Use `tflog` for logging: [[MAINT] Tracking issue for refactoring logging to use `tflog` instead of `log` #3070](https://github.com/integrations/terraform-provider-github/issues/3070)
> * Adhere to new testing structure

Should be implemented now

github/resource_github_user_ssh_signing_key.go
```diff
@@ -0,0 +1,120 @@
...
+		Title: github.Ptr(title),
+		Key:   github.Ptr(key),
+	})
+	if err != nil {
+		return diag.FromErr(err)
+	}
+
+	d.SetId(strconv.FormatInt(*userKey.ID, 10))
+
+	return resourceGithubUserSshSigningKeyRead(ctx, d, meta)
```
deiga: Could you remove the call to `Read` and directly set any computed fields based on create response?

github/resource_github_user_ssh_signing_key.go
```diff
@@ -0,0 +1,120 @@
...
+func resourceGithubUserSshSigningKeyDelete(ctx context.Context, d *schema.ResourceData, meta interface{}) diag.Diagnostics {
+	client := meta.(*Owner).v3client
+
+	id, err := strconv.ParseInt(d.Id(), 10, 64)
+	if err != nil {
+		return diag.Errorf("failed to convert ID %s: %v", d.Id(), err)
+	}
+	ctx = context.WithValue(ctx, ctxId, d.Id())
+
+	_, err = client.Users.DeleteSSHSigningKey(ctx, id)
```
deiga: Could you add handling for 404 here to ignore the error? We use this in most resources already

github/resource_github_user_ssh_signing_key_test.go
```diff
@@ -0,0 +1,91 @@
...
+)
+
+func TestAccGithubUserSshSigningKey(t *testing.T) {
+	t.Run("creates and destroys a user SSH signing key without error", func(t *testing.T) {
+		randomID := acctest.RandStringFromCharSet(5, acctest.CharSetAlphaNum)
+		testKey := newTestSigningKey()
+
+		config := fmt.Sprintf(`
+			resource "github_user_ssh_signing_key" "test" {
+				title = "tf-acc-test-%s"
```
deiga: Could you change the title to use `testResourcePrefix` for generating it?

And could you add Sweeper in acc_test.go which can remove any test ash keys?

deiga (changes requested):

github/resource_github_user_ssh_signing_key.go
simonostendorf: I checked some files but did not found anything, I hope my solution is also okay :D

github/resource_github_user_ssh_signing_key_test.go
simonostendorf: Should be implemented, thanks.

github/resource_github_user_ssh_signing_key.go
simonostendorf: Should be implemented, thanks.

github/resource_github_user_ssh_signing_key.go
```diff
@@ -0,0 +1,133 @@
...
+			},
+			"etag": {
+				Type:     schema.TypeString,
+				Computed: true,
+			},
+		},
+	}
+}
+
+func resourceGithubUserSshSigningKeyCreate(ctx context.Context, d *schema.ResourceData, meta interface{}) diag.Diagnostics {
```
deiga:
```suggestion
func resourceGithubUserSshSigningKeyCreate(ctx context.Context, d *schema.ResourceData, meta any) diag.Diagnostics {
```

github/resource_github_user_ssh_signing_key.go
```diff
@@ -0,0 +1,133 @@
...
+func resourceGithubUserSshSigningKeyRead(ctx context.Context, d *schema.ResourceData, meta interface{}) diag.Diagnostics {
+	client := meta.(*Owner).v3client
+
+	id, err := strconv.ParseInt(d.Id(), 10, 64)
+	if err != nil {
+		return diag.Errorf("failed to convert ID %s: %v", d.Id(), err)
+	}
+	if !d.IsNewResource() {
+		ctx = context.WithValue(ctx, ctxEtag, d.Get("etag").(string))
+	}
```
deiga: This isn't needed anymore
```suggestion
```

github/resource_github_user_ssh_signing_key.go
```diff
@@ -0,0 +1,133 @@
...
+}
+
+func resourceGithubUserSshSigningKeyDelete(ctx context.Context, d *schema.ResourceData, meta interface{}) diag.Diagnostics {
+	client := meta.(*Owner).v3client
+
+	id, err := strconv.ParseInt(d.Id(), 10, 64)
+	if err != nil {
+		return diag.Errorf("failed to convert ID %s: %v", d.Id(), err)
+	}
+	ctx = context.WithValue(ctx, ctxId, d.Id())
```
deiga:
```suggestion
```

github/resource_github_user_ssh_signing_key_test.go
```diff
@@ -0,0 +1,93 @@
...
+				"github_user_ssh_signing_key.test",
+				"title",
+				regexp.MustCompile(randomID),
+			),
+			resource.TestMatchResourceAttr(
+				"github_user_ssh_signing_key.test",
+				"key",
+				regexp.MustCompile("^ssh-rsa "),
+			),
+		)
```
deiga: please inline this

deiga (changes requested):

github/resource_github_user_ssh_signing_key.go
```diff
@@ -0,0 +1,129 @@
...
+
+	userKey, resp, err := client.Users.CreateSSHSigningKey(ctx, &github.Key{
+		Title: github.Ptr(title),
+		Key:   github.Ptr(key),
+	})
+	if err != nil {
+		return diag.FromErr(err)
+	}
+
+	d.SetId(strconv.FormatInt(*userKey.ID, 10))
```
deiga: question: does `userKey` have a `GetID` accessor we could use?

deiga (approved):

github/resource_github_user_ssh_signing_key.go
simonostendorf: fixed, using GetID now :)

github/resource_github_user_ssh_signing_key.go
```diff
@@ -0,0 +1,129 @@
...
+			},
+			"key": {
+				Type:        schema.TypeString,
+				Required:    true,
+				ForceNew:    true,
+				Description: "The public SSH key to add to your GitHub account.",
+				DiffSuppressFunc: func(k, oldV, newV string, d *schema.ResourceData) bool {
+					newTrimmed := strings.TrimSpace(newV)
+					return oldV == newTrimmed
+				},
```
stevehipwell:
```suggestion
```
This isn't required.

github/resource_github_user_ssh_signing_key.go
```diff
@@ -0,0 +1,129 @@
...
+
+	if err = d.Set("etag", resp.Header.Get("ETag")); err != nil {
+		return diag.FromErr(err)
+	}
+	if err = d.Set("title", userKey.GetTitle()); err != nil {
+		return diag.FromErr(err)
+	}
+	if err = d.Set("key", userKey.GetKey()); err != nil {
+		return diag.FromErr(err)
+	}
```
stevehipwell, line 80:
```suggestion
```
These shouldn't be modified here.

github/resource_github_user_ssh_signing_key.go
```diff
@@ -0,0 +1,129 @@
...
+func resourceGithubUserSshSigningKey() *schema.Resource {
+	return &schema.Resource{
+		CreateContext: resourceGithubUserSshSigningKeyCreate,
+		ReadContext:   resourceGithubUserSshSigningKeyRead,
+		DeleteContext: resourceGithubUserSshSigningKeyDelete,
+		Importer: &schema.ResourceImporter{
+			StateContext: schema.ImportStatePassthroughContext,
+		},
+
+		Schema: map[string]*schema.Schema{
```
stevehipwell, line 28: Please add `key_id` to the schema as computed.

github/resource_github_user_ssh_signing_key.go
```diff
@@ -0,0 +1,129 @@
...
+	return nil
+}
+
+func resourceGithubUserSshSigningKeyRead(ctx context.Context, d *schema.ResourceData, meta interface{}) diag.Diagnostics {
+	client := meta.(*Owner).v3client
+
+	id, err := strconv.ParseInt(d.Id(), 10, 64)
+	if err != nil {
+		return diag.Errorf("failed to convert ID %s: %v", d.Id(), err)
+	}
```
stevehipwell: This should use `key_id`.

github/resource_github_user_ssh_signing_key.go
```diff
@@ -0,0 +1,129 @@
...
+
+	if err = d.Set("etag", resp.Header.Get("ETag")); err != nil {
+		return diag.FromErr(err)
+	}
+	if err = d.Set("title", key.GetTitle()); err != nil {
+		return diag.FromErr(err)
+	}
+	if err = d.Set("key", key.GetKey()); err != nil {
+		return diag.FromErr(err)
+	}
```
stevehipwell:
```suggestion
```
These are immutable.

github/resource_github_user_ssh_signing_key.go
```diff
@@ -0,0 +1,129 @@
...
+				})
+				d.SetId("")
+				return nil
+			}
+		}
+	}
+
+	if err = d.Set("etag", resp.Header.Get("ETag")); err != nil {
+		return diag.FromErr(err)
+	}
```
stevehipwell, line 100: I'm not sure if there is any point modifying this, given that the whole resource is immutable and we're only checking if it's been removed.

github/resource_github_user_ssh_signing_key.go
```diff
@@ -0,0 +1,129 @@
...
+	return nil
+}
+
+func resourceGithubUserSshSigningKeyDelete(ctx context.Context, d *schema.ResourceData, meta interface{}) diag.Diagnostics {
+	client := meta.(*Owner).v3client
+
+	id, err := strconv.ParseInt(d.Id(), 10, 64)
+	if err != nil {
+		return diag.Errorf("failed to convert ID %s: %v", d.Id(), err)
+	}
```
stevehipwell: This should use `key_id`.

stevehipwell (changes requested):
Thanks for the PR @simonostendorf. I've added some review comments. Could you also add an import method. And ideally the existing SSH key resource should be aligned with this implementation (off the top of my head - migration to add `key_id`, code changes, and deprecation of `url`).

github/resource_github_user_ssh_signing_key.go
simonostendorf: What do you mean with your suggestion?

github/resource_github_user_ssh_signing_key.go
simonostendorf: Added in latest commit.

github/resource_github_user_ssh_signing_key.go
simonostendorf: Removed in latest commit.

github/resource_github_user_ssh_signing_key.go
simonostendorf: Adjusted in latest commit.

github/resource_github_user_ssh_signing_key.go
simonostendorf: Removed update of etag (and dont added update of key_id) because it will not change.

github/resource_github_user_ssh_signing_key.go
simonostendorf: Also removed update of key and title because they are immutable.

github/resource_github_user_ssh_signing_key.go
simonostendorf: Adjusted in latest commit.

simonostendorf:
How can I update the website? When I run `make website` terraform-website downloads and fails because a unknown makefile target is called.

github/resource_github_user_ssh_signing_key.go
stevehipwell: The key is immutable, so it shouldn't be modified from what the user passed in.

stevehipwell:
> How can I update the website? When I run make website terraform-website downloads and fails because a unknown makefile target is called.

@simonostendorf you just manually modify the website files, check some previous PRs to see how it was done.

github/resource_github_user_ssh_signing_key.go
simonostendorf: Should be resolved.

simonostendorf:
Everything should be fixed now (hopefully) :D

stevehipwell:
@simonostendorf it looks like you've got a conflict, could you please fix this before I give it a review?

simonostendorf:
> @simonostendorf it looks like you've got a conflict, could you please fix this before I give it a review?

@stevehipwell fixed :)

stevehipwell:
@simonostendorf could you please fix the lint errors.

simonostendorf:
> @simonostendorf could you please fix the lint errors.

Should be fixed now :)

stevehipwell:
@simonostendorf please take a look at the sweep failure.

simonostendorf:
> @simonostendorf please take a look at the sweep failure.

@stevehipwell Got the wrong package as import in my test file and therefore the sweep redefined error occured. Should be fixed now "make test" runs fine locally.

stevehipwell (dismissed):
LGTM

github/resource_github_user_ssh_key.go
```diff
@@ -33,15 +34,11 @@ func resourceGithubUserSshKey() *schema.Resource {
 				Required:    true,
 				ForceNew:    true,
 				Description: "The public SSH key to add to your GitHub account.",
-				DiffSuppressFunc: func(k, oldV, newV string, d *schema.ResourceData) bool {
-					newTrimmed := strings.TrimSpace(newV)
-					return oldV == newTrimmed
-				},
 			},
-			"url": {
```
robert-crandall, line 43: This is a breaking schema change. Existing Terraform state will not have `key_id` set, so `d.Get("key_id").(int64)` in Read and Delete will return 0, causing API calls to fail or target the wrong key. There is no state migration function to backfill `key_id` from the existing resource ID.

Suggested fix:
- Keep the `url` attribute with Deprecated: "Use key_id instead" for at least one major version.
- Add a `StateUpgraders` / `SchemaVersion` migration that copies `d.Id()` into `key_id`. 
- In Read, fall back to parsing `d.Id()` as the key ID when `key_id` is zero.

github/resource_github_user_ssh_key.go
```diff
@@ -51,80 +48,102 @@ func resourceGithubUserSshKey() *schema.Resource {
...
-	if err != nil {
-		return unconvertibleIdErr(d.Id(), err)
-	}
-	ctx := context.WithValue(context.Background(), ctxId, d.Id())
-	if !d.IsNewResource() {
-		ctx = context.WithValue(ctx, ctxEtag, d.Get("etag").(string))
-	}
-
-	key, resp, err := client.Users.GetKey(ctx, id)
+	keyID := d.Get("key_id").(int64)
```
robert-crandall: `key_id` is declared as `schema.TypeInt` but called with `d.Get("key_id").(int64)` on read. Terraform SDK stores TypeInt as int, not int64. This type assertion will panic at runtime when Read or Delete is called.

github/resource_github_user_ssh_key.go
```diff
@@ -51,80 +48,102 @@ func resourceGithubUserSshKey() *schema.Resource {
...
-	if err = d.Set("url", key.GetURL()); err != nil {
-		return err
-	}
+func resourceGithubUserSshKeyDelete(ctx context.Context, d *schema.ResourceData, meta any) diag.Diagnostics {
+	client := meta.(*Owner).v3client
 
-	return nil
+	keyID := d.Get("key_id").(int64)
+	resp, err := client.Users.DeleteKey(ctx, keyID)
+	if resp.StatusCode == http.StatusNotFound {
```
robert-crandall: `resp.StatusCode` is accessed before checking whether `err != nil`. If the HTTP call fails with a network error (no response), `resp` could be nil, causing a nil pointer dereference panic.


```suggestion
resp, err := client.Users.DeleteKey(ctx, keyID)
if err != nil {
    if resp != nil && resp.StatusCode == http.StatusNotFound {
        return nil
    }
    return diag.FromErr(err)
}
```

github/resource_github_user_ssh_signing_key.go
```diff
@@ -0,0 +1,149 @@
...
+		}
+	}
+	return nil
+}
+
+func resourceGithubUserSshSigningKeyDelete(ctx context.Context, d *schema.ResourceData, meta any) diag.Diagnostics {
+	client := meta.(*Owner).v3client
+
+	keyID := d.Get("key_id").(int64)
+	resp, err := client.Users.DeleteSSHSigningKey(ctx, keyID)
```
robert-crandall: Same comment as earlier.


```suggestion
resp, err := client.Users.DeleteSSHSigningKey(ctx, keyID)
if err != nil {
    if resp != nil && resp.StatusCode == http.StatusNotFound {
        return nil
    }
    return diag.FromErr(err)
}
```

github/resource_github_user_ssh_key.go
simonostendorf: So the solution is to convert them back and forth because the github api wants int64?

e.g. `keyID := int64(d.Get("key_id").(int))`

github/resource_github_user_ssh_key.go
simonostendorf: Because I never written one of these before, please have a look at my commit and I am very happy about feedback.

simonostendorf: Implemented in https://github.com/integrations/terraform-provider-github/pull/2366/commits/4db92ca6a7e33ca193ef9bf5005d1728e2e0c035, please resolved if code is okay.

github/resource_github_user_ssh_key.go
simonostendorf: Implemented in https://github.com/integrations/terraform-provider-github/pull/2366/commits/4db92ca6a7e33ca193ef9bf5005d1728e2e0c035, please resolved if code is okay.

github/resource_github_user_ssh_key.go
```diff
@@ -33,98 +36,174 @@ func resourceGithubUserSshKey() *schema.Resource {
 				Required:    true,
 				ForceNew:    true,
 				Description: "The public SSH key to add to your GitHub account.",
-				DiffSuppressFunc: func(k, oldV, newV string, d *schema.ResourceData) bool {
-					newTrimmed := strings.TrimSpace(newV)
-					return oldV == newTrimmed
-				},
```
deiga, line 41: question: was there a reason why this needed to be removed?

github/resource_github_user_ssh_key.go
```diff
@@ -33,98 +36,174 @@ func resourceGithubUserSshKey() *schema.Resource {
...
+						}
+						rawState["key_id"] = keyID
+					} else {
+						return nil, fmt.Errorf("resource state upgrade failed, missing or invalid 'id' field in state")
+					}
+
+					return rawState, nil
+				},
+			},
+		},
```
deiga: issue: Please don't inline these, instead follow the established pattern for migrations

github/resource_github_user_ssh_key.go
```diff
@@ -33,98 +36,174 @@ func resourceGithubUserSshKey() *schema.Resource {
...
-	if !d.IsNewResource() {
-		ctx = context.WithValue(ctx, ctxEtag, d.Get("etag").(string))
+	keyID := int64(d.Get("key_id").(int))
+	// fallback to d.Id() for backward compatibility when key_id is not set
+	if keyID == 0 {
+		var err error
+		keyID, err = strconv.ParseInt(d.Id(), 10, 64)
+		if err != nil {
+			return diag.FromErr(fmt.Errorf("invalid SSH key ID format: %w", err))
+		}
```
deiga: Issue: this is not necessary. The Schema Migration will set the keyId

github/resource_github_user_ssh_key.go
```diff
@@ -33,98 +36,174 @@ func resourceGithubUserSshKey() *schema.Resource {
...
-					d.Id())
+				tflog.Info(ctx, fmt.Sprintf("Removing user SSH key %s from state because it no longer exists in GitHub", d.Id()), map[string]any{
+					"ssh_key_id": d.Id(),
+				})
 				d.SetId("")
 				return nil
 			}
 		}
-		return err
 	}
```
deiga, line 113: issue: you should still be setting computed fields in Read

github/resource_github_user_ssh_key.go
```diff
@@ -33,98 +36,174 @@ func resourceGithubUserSshKey() *schema.Resource {
...
+	}
+	if err = d.Set("title", key.GetTitle()); err != nil {
+		return nil, err
+	}
+	if err = d.Set("key", key.GetKey()); err != nil {
+		return nil, err
 	}
-	ctx := context.WithValue(context.Background(), ctxId, d.Id())
 
-	_, err = client.Users.DeleteKey(ctx, id)
```
deiga: Please set only required fields in Import. Read will be automatically called afterwards

github/resource_github_user_ssh_key.go
```diff
@@ -33,98 +36,174 @@ func resourceGithubUserSshKey() *schema.Resource {
...
-				DiffSuppressFunc: func(k, oldV, newV string, d *schema.ResourceData) bool {
-					newTrimmed := strings.TrimSpace(newV)
-					return oldV == newTrimmed
-				},
 			},
 			"url": {
 				Type:        schema.TypeString,
 				Computed:    true,
 				Description: "The URL of the SSH key.",
+				Deprecated:  "Use key_id instead.",
```
deiga: question: why does this field need to be deprecated?

deiga (changes requested):

github/resource_github_user_ssh_key.go
simonostendorf: See https://github.com/integrations/terraform-provider-github/pull/2366#discussion_r2816272867

github/resource_github_user_ssh_key.go
simonostendorf: I took this from https://developer.hashicorp.com/terraform/plugin/framework/migrating/resources/state-upgrade#examples.

So just moving this to a separarte function is fine?

github/resource_github_user_ssh_key.go
simonostendorf: See https://github.com/integrations/terraform-provider-github/pull/2366#discussion_r2934263631

github/resource_github_user_ssh_key.go
deiga: Please look at other files with migrations in the repo

github/resource_github_user_ssh_key.go
deiga: But that was a comment since you removed the field. Why does the field need to be removed at all?

github/resource_github_user_ssh_key.go
robert-crandall: There's still some places this occurs, [here](https://github.com/simonostendorf/integrations-terraform-provider-github/blob/4db92ca6a7e33ca193ef9bf5005d1728e2e0c035/github/resource_github_user_ssh_signing_key.go#L83) and [here](https://github.com/simonostendorf/integrations-terraform-provider-github/blob/4db92ca6a7e33ca193ef9bf5005d1728e2e0c035/github/resource_github_user_ssh_signing_key.go#L106).

simonostendorf:
I only wanted to say that I havent got much time this and the next week and will fix the comments here later.

github/resource_github_user_ssh_key.go
simonostendorf: Readded url. I think I removed it because the field doesnt exist on the signing key api resource and the comment was that I should align both resources.

github/resource_github_user_ssh_key.go
simonostendorf: removed all field sets that arent required.

github/resource_github_user_ssh_key.go
simonostendorf: readded setting computed fields.

github/resource_github_user_ssh_key.go
simonostendorf: removed.

github/resource_github_user_ssh_key.go
simonostendorf: aligned with other migration structure.

github/resource_github_user_ssh_key.go
simonostendorf: fixed.

github/resource_github_user_ssh_key.go
```diff
@@ -33,98 +35,150 @@ func resourceGithubUserSshKey() *schema.Resource {
...
 	if err != nil {
 		var ghErr *github.ErrorResponse
 		if errors.As(err, &ghErr) {
 			if ghErr.Response.StatusCode == http.StatusNotModified {
 				return nil
 			}
 			if ghErr.Response.StatusCode == http.StatusNotFound {
-				log.Printf("[INFO] Removing user SSH key %s from state because it no longer exists in GitHub",
-					d.Id())
+				tflog.Info(ctx, fmt.Sprintf("Removing user SSH key %s from state because it no longer exists in GitHub", d.Id()), map[string]any{
```
deiga: suggestion: we prefer not to use fmt.Sprintf in tflog calls

github/resource_github_user_ssh_key_migration.go
```diff
@@ -0,0 +1,62 @@
...
+	// copy d.Id() into key_id
+	if id, ok := rawState["id"].(string); ok {
+		keyID, err := strconv.ParseInt(id, 10, 64)
+		if err != nil {
+			return nil, fmt.Errorf("resource state upgrade failed, invalid SSH key ID format: %w", err)
+		}
+		rawState["key_id"] = keyID
+	} else {
+		return nil, fmt.Errorf("resource state upgrade failed, missing or invalid 'id' field in state")
+	}
```
deiga: issue: this else is unnecessary. Migrations should only be run on existing resources and existing resources have an id populated

deiga: If you want to keep the extra defense, then move it to an if!ok block with early return to not add nesting to the logic

github/resource_github_user_ssh_signing_key.go
```diff
@@ -0,0 +1,155 @@
...
+
+func resourceGithubUserSshSigningKey() *schema.Resource {
+	return &schema.Resource{
+		CreateContext: resourceGithubUserSshSigningKeyCreate,
+		ReadContext:   resourceGithubUserSshSigningKeyRead,
+		DeleteContext: resourceGithubUserSshSigningKeyDelete,
+		Importer: &schema.ResourceImporter{
+			StateContext: resourceGithubUserSshSigningKeyImport,
+		},
+
```
deiga, line 23: issue: Please add top-level Description

github/resource_github_user_ssh_signing_key.go
```diff
@@ -0,0 +1,155 @@
...
+			if ghErr.Response.StatusCode == http.StatusNotModified {
+				return nil
+			}
+			if ghErr.Response.StatusCode == http.StatusNotFound {
+				tflog.Info(ctx, fmt.Sprintf("Removing user SSH signing key %s from state because it no longer exists in GitHub", d.Id()), map[string]any{
+					"ssh_signing_key_id": d.Id(),
+				})
+				d.SetId("")
+				return nil
+			}
```
deiga: suggestion: this could be replaced with the handle404 function that exists already

github/resource_github_user_ssh_signing_key_test.go
```diff
@@ -0,0 +1,85 @@
...
+			resource "github_user_ssh_signing_key" "test" {
+				title = "%[1]s"
+				key   = "%[2]s"
+			}
+		`, name, testKey)
+
+		check := resource.ComposeTestCheckFunc(
+			resource.TestMatchResourceAttr("github_user_ssh_signing_key.test", "title", regexp.MustCompile(randomID)),
+			resource.TestMatchResourceAttr("github_user_ssh_signing_key.test", "key", regexp.MustCompile("^ssh-rsa ")),
+		)
```
deiga: Please use ConfigStsteChecks and don't use a variable for these

website/docs/r/user_ssh_key.html.markdown
```diff
@@ -24,15 +24,15 @@ resource "github_user_ssh_key" "example" {
...
-* `title` - (Required) A descriptive name for the new key. e.g. `Personal MacBook Air`
+* `title` - (Required) A descriptive name for the new key.
 * `key` - (Required) The public SSH key to add to your GitHub account.
 
 ## Attributes Reference
 
 The following attributes are exported:
 
-* `id` - The ID of the SSH key
-* `url` - The URL of the SSH key
```
deiga: question: why are these removed?

deiga (changes requested):

github/resource_github_user_ssh_signing_key.go
simonostendorf: I cant find the handle404 function

github/resource_github_user_ssh_key.go
simonostendorf: removed.

github/resource_github_user_ssh_key_migration.go
simonostendorf: removed unneeded else block.

github/resource_github_user_ssh_signing_key.go
simonostendorf: added.

github/resource_github_user_ssh_signing_key_test.go
simonostendorf: using ConfigStateChecks now.

website/docs/r/user_ssh_key.html.markdown
simonostendorf: readded them, website has to be manually created (see discussion above) and I forgot to readd them.

github/resource_github_user_ssh_signing_key.go
deiga: deleteResourceOn404AndSwallow304OtherwiseReturnError

simonostendorf: using function now.

website/docs/r/user_ssh_signing_key.html.markdown
```diff
@@ -0,0 +1,44 @@
...
+  title = "example title"
+  key   = file("~/.ssh/id_rsa.pub")
+}
+```
+
+## Argument Reference
+
+The following arguments are supported:
+
+* `id` - The ID of the SSH signing key
```
deiga: issue: This is not an argument, it's an attribute

website/docs/r/user_ssh_key.html.markdown
```diff
@@ -24,15 +24,17 @@ resource "github_user_ssh_key" "example" {
 
 The following arguments are supported:
 
-* `title` - (Required) A descriptive name for the new key. e.g. `Personal MacBook Air`
+* `id` - The ID of the SSH key
+* `url` - The URL of the SSH key
```
deiga: issue: These are not arguments, they are attributes

github/resource_github_user_ssh_signing_key.go
```diff
@@ -0,0 +1,144 @@
...
+		var ghErr *github.ErrorResponse
+		if errors.As(err, &ghErr) {
+			if ghErr.Response.StatusCode == http.StatusNotFound {
+				return nil, fmt.Errorf("SSH signing key with ID %d not found", keyID)
+			}
+		}
+		return nil, err
+	}
+
+	d.SetId(strconv.FormatInt(key.GetID(), 10))
```
deiga: question: Isn't this unnecessary? `d.Id()` is already set with `keyID` and AFAIK `keyID == key.GetID()`?

github/resource_github_user_ssh_signing_key.go
```diff
@@ -0,0 +1,144 @@
...
+
+	if err = d.Set("title", key.GetTitle()); err != nil {
+		return nil, err
+	}
+	if err = d.Set("key", key.GetKey()); err != nil {
```
deiga, line 145: issue: Read uses `key_id` but Import doesn't set it

github/resource_github_user_ssh_signing_key.go
```diff
@@ -0,0 +1,144 @@
...
+	keyID := int64(d.Get("key_id").(int))
+	userKey, resp, err := client.Users.GetSSHSigningKey(ctx, keyID)
+	if err != nil {
+		return diag.FromErr(deleteResourceOn404AndSwallow304OtherwiseReturnError(err, d, "user SSH signing key (%d)", keyID))
+	}
+
+	// set computed fields
+	if err = d.Set("key_id", userKey.GetID()); err != nil {
+		return diag.FromErr(err)
+	}
```
deiga: This seems redundant to set, since it should never change

github/resource_github_user_ssh_key_migration.go
```diff
@@ -0,0 +1,60 @@
...
+	"fmt"
+	"strconv"
+	"strings"
+
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+)
+
+func resourceGithubUserSshKeyV0() *schema.Resource {
+	return &schema.Resource{
+		SchemaVersion: 1,
```
deiga:
```suggestion
		SchemaVersion: 0,
```

deiga (changes requested):

github/resource_github_user_ssh_signing_key.go
simonostendorf: setting key_id in import.

github/resource_github_user_ssh_signing_key.go
simonostendorf: removed because I also think its unnecessary.

github/resource_github_user_ssh_signing_key.go
simonostendorf: removed setting key_id in read.

website/docs/r/user_ssh_key.html.markdown
simonostendorf: fixed.

github/resource_github_user_ssh_key_migration.go
simonostendorf: fixed.

website/docs/r/user_ssh_signing_key.html.markdown
simonostendorf: fixed.

simonostendorf:
@deiga saw you assigned the 6.13.0 milestone, nice that it gets released. Can you ping me here when I have to resolve the merge conflicts again? I dont want to do it every time a conflict pops up because that happens every week because go imports change

templates/resources/user_ssh_signing_key.md.tmpl
deiga, line 1: issue: Please remove the custom template so that it uses the default template instead. This doesn't have any need for template customization

github/resource_github_user_ssh_key.go
```diff
@@ -33,98 +36,136 @@ func resourceGithubUserSshKey() *schema.Resource {
...
 				Description: "The URL of the SSH key.",
 			},
+			"key_id": {
+				Type:        schema.TypeInt,
+				Computed:    true,
+				Description: "The unique identifier of the SSH key.",
+			},
 			"etag": {
 				Type:     schema.TypeString,
 				Computed: true,
```
deiga: SInce you're modifying this resource, please ensure all attributes have a `Description` field

github/resource_github_user_ssh_key.go
```diff
@@ -33,98 +36,136 @@ func resourceGithubUserSshKey() *schema.Resource {
...
-	id, err := strconv.ParseInt(d.Id(), 10, 64)
+	keyID := int64(d.Get("key_id").(int))
+	// fallback to d.Id() for backward compatibility when key_id is not set
+	if keyID == 0 {
+		var err error
+		keyID, err = strconv.ParseInt(d.Id(), 10, 64)
+		if err != nil {
+			return diag.FromErr(fmt.Errorf("invalid SSH key ID format: %w", err))
+		}
+	}
```
deiga: issue: There should be no situation where `d.Id()` has a value, but `key_id` doesn't.

Can you give an example when that would not be the case?

github/resource_github_user_ssh_key.go
```diff
@@ -33,98 +36,136 @@ func resourceGithubUserSshKey() *schema.Resource {
...
+	keyID, err := strconv.ParseInt(d.Id(), 10, 64)
+	if err != nil {
+		return nil, fmt.Errorf("invalid SSH key ID format: %w", err)
 	}
 
-	key, resp, err := client.Users.GetKey(ctx, id)
+	key, _, err := client.Users.GetKey(ctx, keyID)
 	if err != nil {
 		var ghErr *github.ErrorResponse
 		if errors.As(err, &ghErr) {
```
deiga: suggestion: Please update this to use `errors.AsType`

github/resource_github_user_ssh_key.go
```diff
@@ -33,98 +36,136 @@ func resourceGithubUserSshKey() *schema.Resource {
...
+			return diag.FromErr(fmt.Errorf("invalid SSH key ID format: %w", err))
+		}
+	}
+
+	resp, err := client.Users.DeleteKey(ctx, keyID)
 	if err != nil {
-		return unconvertibleIdErr(d.Id(), err)
+		if resp != nil && resp.StatusCode == http.StatusNotFound {
+			return nil
+		}
```
deiga: issue: Please use the `errors.AsType` convention here instead

github/resource_github_user_ssh_key_test.go
```diff
@@ -20,36 +22,26 @@ func TestAccGithubUserSshKey(t *testing.T) {
 		t.Parallel()
 
 		randomID := acctest.RandStringFromCharSet(5, acctest.CharSetAlphaNum)
+		name := fmt.Sprintf(`%s-%s`, testResourcePrefix, randomID)
```
deiga: `testResourcePrefix` already contains a `-` suffix
```suggestion
		name := fmt.Sprintf(`%s%s`, testResourcePrefix, randomID)
```

github/resource_github_user_ssh_signing_key.go
```diff
@@ -0,0 +1,142 @@
...
+
+func resourceGithubUserSshSigningKeyDelete(ctx context.Context, d *schema.ResourceData, meta any) diag.Diagnostics {
+	client := meta.(*Owner).v3client
+
+	keyID := int64(d.Get("key_id").(int))
+	resp, err := client.Users.DeleteSSHSigningKey(ctx, keyID)
+	if err != nil {
+		if resp != nil && resp.StatusCode == http.StatusNotFound {
+			return nil
+		}
```
deiga: issue: Please use the `errors.AsType` convention instead

deiga (changes requested):

github/resource_github_user_ssh_key.go
simonostendorf: I took that from the long discussion before here so I think it can be removed.

github/resource_github_user_ssh_key.go
```diff
@@ -3,24 +3,27 @@ package github
...
-		Delete: resourceGithubUserSshKeyDelete,
+		CreateContext: resourceGithubUserSshKeyCreate,
+		ReadContext:   resourceGithubUserSshKeyRead,
+		DeleteContext: resourceGithubUserSshKeyDelete,
 		Importer: &schema.ResourceImporter{
-			StateContext: schema.ImportStatePassthroughContext,
+			StateContext: resourceGithubUserSshKeyImport,
 		},
 
+		Description: "Manages a SSH key for the authenticated user.",
```
stevehipwell:
```suggestion
		Description: "Resource to manage a SSH key for the authenticated user.",
```

github/resource_github_user_ssh_key.go
```diff
@@ -33,98 +36,126 @@ func resourceGithubUserSshKey() *schema.Resource {
...
 
-	d.SetId(strconv.FormatInt(*userKey.ID, 10))
+	d.SetId(strconv.FormatInt(userKey.GetID(), 10))
+
+	if err = d.Set("key_id", userKey.GetID()); err != nil {
+		return diag.FromErr(err)
+	}
+	if err = d.Set("etag", resp.Header.Get("ETag")); err != nil {
+		return diag.FromErr(err)
+	}
```
stevehipwell: Can we set the etag last.

github/resource_github_user_ssh_signing_key.go
```diff
@@ -0,0 +1,141 @@
...
+func resourceGithubUserSshSigningKey() *schema.Resource {
+	return &schema.Resource{
+		CreateContext: resourceGithubUserSshSigningKeyCreate,
+		ReadContext:   resourceGithubUserSshSigningKeyRead,
+		DeleteContext: resourceGithubUserSshSigningKeyDelete,
+		Importer: &schema.ResourceImporter{
+			StateContext: resourceGithubUserSshSigningKeyImport,
+		},
+
+		Description: "Manages a SSH signing key for the authenticated user.",
```
stevehipwell:
```suggestion
		Description: "Resource to manage a SSH signing key for the authenticated user.",
```

github/resource_github_user_ssh_signing_key.go
```diff
@@ -0,0 +1,141 @@
...
+				Type:        schema.TypeString,
+				Computed:    true,
+				Description: "The ETag of the SSH signing key.",
+			},
+		},
+	}
+}
+
+func resourceGithubUserSshSigningKeyCreate(ctx context.Context, d *schema.ResourceData, meta any) diag.Diagnostics {
+	client := meta.(*Owner).v3client
```
stevehipwell:
```suggestion
func resourceGithubUserSshSigningKeyCreate(ctx context.Context, d *schema.ResourceData, m any) diag.Diagnostics {
	meta, _ := m.(*Owner)
	client := meta.v3client
```

github/resource_github_user_ssh_signing_key.go
```diff
@@ -0,0 +1,141 @@
...
+
+	title := d.Get("title").(string)
+	key := d.Get("key").(string)
```
stevehipwell:
```suggestion
	title, _ := d.Get("title").(string)
	key, _ := d.Get("key").(string)
```

stevehipwell (commented):
@simonostendorf it doesn't look like you're using `hashicorp/terraform-plugin-docs` to generate the docs page, this is required and CI will fail if it's not been implemented. Off the top of my head you need to add a template, the example naming needs fixing and you need to add the import examples. It also looks like you haven't followed other parts of the contribution guide such as the required linting?

simonostendorf:
Thanks @stevehipwell, that was helpful, especially the pointer to tfplugindocs. Rebased onto main and pushed.

The docs page was not generated as I was told in feburary. Added the missing template and the import examples, and regenerated.

The rebase pulled in #3474, which made `etag` `Optional + Computed` with diff suppression across 28 resources. That conflicted with my commits from before. I resolved in favour of #3474 for `github_user_ssh_key` and applied the same block to `github_user_ssh_signing_key`, so the new resource follows the convention instead of the plain `Computed` etag I originally wrote.

Two things I did beyond your comments, happy to split either out if you prefer:

- Converted `github_user_ssh_key` to the generated docs form as well. This PR already changes that resources schema, and I was patching its hand maintained attribute list by hand.
- `tools/tools.go` now passes `--provider-name github`. tfplugindocs derives the provider short name from the provider directory name, so `make generatedocs` wipes `docs/` and then fails in any clone not named `terraform-provider-github`. Unrelated to this feature, but it is a one line fix and it cost me a while to work out.

I also switched the `github_user_ssh_key` acceptance test from `skipUnlessMode(t, individual)` to `skipUnauthenticated(t)`. The test is otherwise already modernised in this PR, and under the old precheck it never runs in the org test mode that the sweepers I added assume.

github/resource_github_user_ssh_signing_key_test.go
```diff
@@ -0,0 +1,89 @@
...
+			},
+		})
+	})
+}
+
+func newTestSigningKey() string {
+	privateKey, _ := rsa.GenerateKey(rand.Reader, 1024)
+	publicKey, _ := ssh.NewPublicKey(&privateKey.PublicKey)
+	return strings.TrimRight(string(ssh.MarshalAuthorizedKey(publicKey)), "\n")
+}
```
stevehipwell, line 89: Please add this to a test helpers file and use the `must` pattern with `t.Fatal()` in the error handlers.

github/resource_github_user_ssh_signing_key_test.go
```diff
@@ -0,0 +1,89 @@
+package github
```
stevehipwell, line 1: Please look at `RESOURCES.md` for a resource with the tests fully refactored and copy the pattern there. I'd expect to have the import as a step in the happy path test.

github/resource_github_user_ssh_key_test.go
```diff
@@ -20,36 +22,26 @@ func TestAccGithubUserSshKey(t *testing.T) {
...
-			resource.TestMatchResourceAttr(
-				"github_user_ssh_key.test", "url",
-				regexp.MustCompile("^https://api.github.com/[a-z0-9]+/keys/"),
-			),
-		)
+		`, name, testKey)
 
 		resource.Test(t, resource.TestCase{
-			PreCheck:          func() { skipUnlessMode(t, individual) },
+			PreCheck:          func() { skipUnauthenticated(t) },
```
stevehipwell, line 36: I'm not sure I follow this change, isn't the resource intended to be used by the currently authenticated user? Also the latest testing pattern has resource/data-source scoped skips at the top of the function instead of here.

stevehipwell (changes requested):
Thanks for the quick response and detailed summary of the changes you made, it makes context shifting and re-reviewing much easier.

The main review point is could you please take a look at the tests for one of the updated resources and update the tests in this PR accordingly. Not the use of fewer test cases with more steps.

Thanks for pointing out what happened in #3474, that doesn't look right to me. But in a related point, could you make sure that you set the `etag` value to zero if the update fails (probably not needed for this resource, but it should be the pattern).

simonostendorf:
Thanks @stevehipwell, that all makes sense.

Rebased onto main and pushed.

**Tests.** Took `github_actions_variable` as the reference from RESOURCES.md and
rewrote both test files to match. The two near identical subtests are now one
`default` subtest with four steps: create, retitle, rekey, import. The two
rename steps assert `ResourceActionReplace` via `ConfigPlanChecks`, since both
attributes are `ForceNew`. Import is the last step of the happy path rather
than its own case, as you asked.

Also dropped `PreCheck` entirely and moved `skipUnauthenticated(t)` to the top
of the test function, matching the current pattern.

**Test helper.** `newTestKey` and `newTestSigningKey` are gone. Replaced by a
single `mustNewSshPublicKey(t)` in `github/helpers_test.go`, next to the
existing `mustNewPEM`, using `t.Helper()` and `t.Fatalf` on both error paths.
It now generates Ed25519 rather than RSA 1024 while I was in there. GitHub
rejects RSA keys below 2048 bits, so the old helper would not have survived a
real acceptance run.

**On `skipUnlessMode(t, individual)` to `skipUnauthenticated(t)`.** You are
right that this resource is for the authenticated user, and that is exactly
why I changed it. `individual` is one specific auth mode, so the old precheck
skipped the test in the org modes, including the mode the sweepers I added in
this PR run under. The resource does not care which mode it is, only that
there is an authenticated user, so `skipUnauthenticated` is the accurate
condition. Happy to revert if you would rather keep the narrower gate.

**Etag.** The rebase also pulled in #3637, which replaced the #3474
`Optional` + `DiffSuppressFunc` block with `CustomizeDiff: diffETag` and a
plain `Computed` etag. I had applied the #3474 form to
`github_user_ssh_signing_key` in an earlier commit, so I have moved it over to
`diffETag` and regenerated the docs for both resources. The etag attribute is
now Read-Only rather than Optional on both pages.

On zeroing the etag when an update fails: there is nowhere to put it here.
Neither resource has an update function, because the GitHub API has no update
endpoint for user SSH keys or signing keys, so `title` and `key` are both
`ForceNew`. `github_release` looks like the only resource doing this today
(`d.Set("etag", nil)` as the first statement of update).

PR #2378, opened 2024-09-04
Fix eventual consistency issue with github_actions_environment_variable

### LINKED ISSUE #2047: [BUG]: RANDOMLY github_actions_variable can't be created - root object was present, but now absent
### Expected Behavior

Variable for repository is created.

Howevever it randomly doesn't work and error occurs (described below)

### Actual Behavior

github_actions_variable.repo,
provider "provider[\"registry.terraform.io/integrations/github\"]" produced
an unexpected new value: Root object was present, but now absent.

### Terraform Version

Terraform 1.6.4
integrations/github v5.42.0

### Affected Resource(s)

- github_actions_variable

_No response_

### Steps to Reproduce

_No response_

### Debug Output

_No response_

### Panic Output

```shell
github_actions_variable.repo,
provider "provider[\"registry.terraform.io/integrations/github\"]" produced
an unexpected new value: Root object was present, but now absent.

```

### Code of Conduct

marcoths:
Resolves #2047

----

### Before the change?

* When creating `github_actions_environment_variable` resources, there is the chance on which the variable gets created, but the read operation fails - for example due to eventual consistency. Before this change, Terraform would issue the error 
```

When applying changes to
github_actions_environment_variable.my_variable, provider
"provider[\"registry.terraform.io/integrations/github\"]" produced an
unexpected new value: Root object was present, but now absent.

issue tracker.
```

### After the change?

* The retry package in the [terraform plugin sdk](https://github.com/hashicorp/terraform-plugin-sdk/blob/61766d8f9c10fa17a1e0b6f42fb2e02d633b82ae/helper/retry/state.go#L27) has a struct called `StateChangeConf` which implements a retry mechanism for these cases.  This is an example from [hashicorp/terraform-provider-aws](https://github.com/hashicorp/terraform-provider-aws/blob/84b687e8a78a82342569c93bc7eafce8dc3c95aa/internal/service/ecs/cluster.go#L206)

### Pull request checklist

No breaking change

- [ ] Yes
- [ ] No

----


acouvreur:
This should fix #2387 can we have an update on this ?

This fix is straight forward and we're looking for this to be merged!

Thanks @marcoths

PR #2382, opened 2024-09-06
feat: allow filtering on permission in repo collaborator datasource

### LINKED ISSUE #2381: [FEAT]: github_collaborators data_source should allow filtering based on permission
### Describe the need

Currently the [`github_collaborators`](https://registry.terraform.io/providers/integrations/github/latest/docs/data-sources/collaborators) data_source always returns all users that have access to the repo (unless you filter on `affiliation`, see last paragraph for limitation on that). This becomes a problem in large organizations with many users and some permission (like read) being [granted to all repos by default](https://docs.github.com/en/enterprise-cloud@latest/organizations/managing-user-access-to-your-organizations-repositories/managing-repository-roles/setting-base-permissions-for-an-organization). The result is that any call to this datasource have to list all the users in the entire organization, taking quite a long time. 

The underlying API endpoint [`/repos/OWNER/REPO/collaborators`](https://docs.github.com/en/rest/collaborators/collaborators?apiVersion=2022-11-28#list-repository-collaborators) supports filtering via `permission`, but the Terraform data source does not. This is what this issue is for!

The underlying [`go-github` SDK already support this](https://github.com/google/go-github/blob/173aa01e10b0260da38267a51f3b8d3141f76416/github/repos_collaborators.go#L13-L34), so the change in this repo would be pretty straight forward.

Note that it is currently possible to filter on [`affiliation`](https://registry.terraform.io/providers/integrations/github/latest/docs/data-sources/collaborators#affiliation) in the data source, but this is quite limited. You can only get:

* `outside` - All outside collaborators with access to the repo
* `direct` - All users which have direct access to the repo (notably this does not include users with access to the repo via a team assignment)
* `all` - Every user with any access to the repo

### SDK Version

_No response_

### API Version

_No response_

### Relevant log output

_No response_

### Code of Conduct

felixlut:
Resolves #2381

----

### Before the change?

* It was not possible to filter on `permission` for the [`github_collaborators`](https://registry.terraform.io/providers/integrations/github/latest/docs/data-sources/collaborators) datasource before this change

### After the change?

* It is now possible to filter on `permission` for the [`github_collaborators`](https://registry.terraform.io/providers/integrations/github/latest/docs/data-sources/collaborators) datasource

### Pull request checklist

- [ ] Yes
- [x] No
    - Worth noting is that I've changed the internal ID of the datasource to include the new `permission` argument if it is set [here](https://github.com/integrations/terraform-provider-github/pull/2382/files#diff-a177fd7eae43548088cdfa3133f154ec7fa07d4c2ed8786f7eaf01d9aa39ec63L127-R145). I don't think this counts as a breaking change since the ID only changes if the ID is the `permission` is actually set. Omitting it will default to the same ID as before this change
----

### DIFF
--- a/github/data_source_github_collaborators.go
+++ b/github/data_source_github_collaborators.go
@@ -32,6 +32,18 @@ func dataSourceGithubCollaborators() *schema.Resource {
 				Optional: true,
 				Default:  "all",
 			},
+			"permission": {
+				Type: schema.TypeString,
+				ValidateDiagFunc: toDiagFunc(validation.StringInSlice([]string{
+					"pull",
+					"triage",
+					"push",
+					"maintain",
+					"admin",
+				}, false), "permission"),
+				Optional: true,
+				Default:  "",
+			},
 			"collaborator": {
 				Type:     schema.TypeList,
 				Computed: true,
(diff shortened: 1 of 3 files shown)


kfcampbell (approved):
Nice, thanks!

Issue #2383, opened 2024-09-10
[BUG]: github_repository: 422 An enforced security configuration prevented modifying advanced security enablement.


eherde:
### Expected Behavior

We have been using the `github_repository` resource for some time. We recently signed up for GHAS, and enabled enforcement on our repos.

We expect that an `apply` of a plan like the following should succeed:
```
terraform apply -target module.repo.github_repository.repo
…
  # module.repo.github_repository.repo will be updated in-place
  ~ resource "github_repository" "repo" {
      ~ allow_rebase_merge          = false -> true
        id                          = "myrepo"
        name                        = "myrepo"
        # (33 unchanged attributes hidden)

        # (1 unchanged block hidden)
    }

Plan: 0 to add, 1 to change, 0 to destroy.
…
```

### Actual Behavior

The first time the terraform runs with an expected change, we get the following error:

```
│ Error: PATCH https://api.github.com/repos/groq/myrepo: 422 An enforced security configuration prevented modifying advanced security enablement. Contact your organization owner for details. []
```

After this run, manual inspection in the UI shows that the change is applied (in this case, allowing rebase and merge). The next run of the terraform rightfully detects that there are no changes:

```
No changes. Your infrastructure matches the configuration.
```

Note that if I disabled enforcement of GHAS and run the `terraform apply` again when there are pending changes (in this example, toggling allow_rebase_merge between true and false), then the apply succeed with no errors.

### Terraform Version

Terraform v1.9.3
on darwin_amd64
+ provider registry.terraform.io/buildkite/buildkite v1.10.2
+ provider registry.terraform.io/hashicorp/google v6.0.1
+ provider registry.terraform.io/integrations/github v6.2.3

### Affected Resource(s)

- github_repository

### Steps to Reproduce

```$ terraform apply```

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

sgsollie:
We've also ran into this where we use the terraform provider to create a new repo and add some branch protections. 

On adding a new repo with org wide GHAS enforcement on, Terraform exits with the `422` error above having partially done the create repo operation & more concerning, it tainted that particular `github_repository` resource. Worst case, that could lead to a repo being destroyed and re-created. In our case it was brand new so it was empty anyway, and we block deletes of repos, but this could really be dangerous for some.

jamestran201:
[...]

PR #2397, opened 2024-09-20
Update security_and_analysis settings only when there are changes

jamestran201:
The `resourceGithubRepositoryUpdate` function currently passes the current state of security_and_analysis to the PATCH /repo request. If the repo has an enforced security configuration, this can result in the API returning a 422 response because changing security settings is not allowed in this case. Even though the security_and_analysis settings remain the same, and only some other settings are changed, the request will still fail.

To avoid the request failing when security_and_analysis remains the same, we will only pass it in the request payload when changes have been made.

Resolves https://github.com/integrations/terraform-provider-github/issues/2383

----

### Before the change?

Currently, if a repository has Secret Scanning enabled. The payload in the PATCH /repository request will include this data, thus attempts to enable Secret Scanning on the repository again. This can happen even with this template that doesn't modify Secret Scanning
```terraform
resource "github_repository" "allow_rebase_merge" {
  name               = "test-tf-changes"
  allow_rebase_merge = false
}
```

### After the change?

The request payload for PATCH /repository will only contain the `security_and_analysis` object when security features are being modified. This means that the value for at least one of the features is being updated: `advanced_security`, `secret_scanning`, `secret_scanning_push_protection`, `secret_scanning_non_provider_patterns`, `secret_scanning_validity_checks`. By doing this, we will only attempt to modify these settings when a change is specified by the template.

### Pull request checklist

I didn't write an acceptance test for this change because the setup requires applying a security configuration to a repo. I tried the following manual test:
- Create `main.tf` that looks like
```terraform
provider "github" {
  token = "<your API token here>"
  owner = "<your org here>"
}

terraform {
  required_providers {
    github = {
    }
  }
}
resource "github_repository" "allow_rebase_merge" {
  name               = "manual-pr-test"
  allow_rebase_merge = false
  ignore_vulnerability_alerts_during_read = true
}
```

- Run `terraform apply`
- In the GitHub UI, create a security configuration that enforces all security features. Then apply it to the repo above
- Change `allow_rebase_merge` from `false` to `true` in the template
- Run `terraform apply` again
- The command should complete successfully

`ignore_vulnerability_alerts_during_read` is set for this test because the terraform state stored locally can be different from the repo's state after a security configuration is applied. This causes the update command to attempt changing Dependabot alerts enablement state, which will fail if the state is enforced by a security configuration.

- [ ] Yes
- [x] No

----


arkadiyt:
Would love to see this shipped, it is a very annoying issue for our org

kfcampbell (approved):

PR #2465, opened 2024-11-14
Return the error when deleting a file

### LINKED ISSUE #2464: [BUG]: Deleting github_repository_file resource might not remove the file
### Expected Behavior

When a successful apply of a destruction, I would not expect to see the file in the repository anymore

### Actual Behavior

Even if there is an HTTP error, the terraform process exits successfully. 

### Terraform Version

Terraform v1.9.8
on darwin_arm64
+ provider registry.terraform.io/integrations/github v6.3.1

### Affected Resource(s)

- github_repository_file

_No response_

### Steps to Reproduce

1. Create the resource 
2. Delete the resource after changing ownership preventing you from doing so. 

```hcl
resource "github_repository_file" "foo" {
  repository          = "tmp"
  branch              = "main"
  file                = ".gitignore"
  content             = "**/*.tfstate"
  commit_message      = "Managed by Terraform"
  commit_author       = "Terraform User"
  commit_email        = "terraform@example.com"
  overwrite_on_create = true
}
```

then 

```hcl
// resource "github_repository_file" "foo" {
//   repository          = "tmp"
//   branch              = "main"
//   file                = ".gitignore"
//   content             = "**/*.tfstate"
//   commit_message      = "Managed by Terraform"
//   commit_author       = "Terraform User"
//   commit_email        = "terraform@example.com"
//   overwrite_on_create = true
// }
```

### Debug Output

_No response_

### Panic Output

_No response_

### Code of Conduct

frankywahl:
There is currently an issue when destroying the github_repository_file resource. If the action fails because of permissions, or because there is a protected branch, or any of the above, the error gets silenced.

Resolves #2464

----

### Before the change?

* Delete a github_repository_file would be successful even if the API call failed. 

### After the change?

* A failed API call will not mark the apply as being successful 

- [x] No (unless people are expecting that bad behavior).

### DIFF
--- a/github/resource_github_repository_file.go
+++ b/github/resource_github_repository_file.go
@@ -504,11 +504,7 @@ func resourceGithubRepositoryFileDelete(d *schema.ResourceData, meta interface{}
 	}
 
 	_, _, err := client.Repositories.DeleteFile(ctx, owner, repo, file, opts)
-	if err != nil {
-		return nil
-	}
-
-	return nil
+	return err
 }
 
 func autoBranchDiffSuppressFunc(k, _, _ string, d *schema.ResourceData) bool {


frankywahl:
This was fixed in https://github.com/integrations/terraform-provider-github/pull/2837

PR #2501, opened 2024-12-05
fix: Fixed repository resource churn

### LINKED ISSUE #2489: [BUG]: The `github_repository` resource churns on `vulnerability_alerts` if not explicitly set
### Expected Behavior

When creating a new repository with the `github_repository` resource, this should not churn if `vulnerability_alerts` is not set.

### Actual Behavior

If `vulnerability_alerts` is not set it will constantly churn.

### Terraform Version

n/a

### Affected Resource(s)

- `github_repository`

```hcl
  name         = "test"
}
```

### Steps to Reproduce

Apply the above resource and then run a plan.

This issue is being caused by a missing `Computed: true` value on the schema. I'll open a PR to fix this, but long term this attribute should be deprecated and handled by it's own resource.

### Debug Output

_No response_

### Panic Output

_No response_

### Code of Conduct

### LINKED ISSUE #2495: [BUG]: The `github_repository` resource churns if pages are enabled and `build_type` is set to `workflow`
### Expected Behavior

When configuring a `github_repository` resource to enables pages and use the workflow build type I'd expect the resource to not churn.

### Actual Behavior

The `github_repository` resource constantly churns.

### Terraform Version

n/a

### Affected Resource(s)

- `github_repository`

_No response_

### Steps to Reproduce

_No response_

### Debug Output

_No response_

### Panic Output

_No response_

### Code of Conduct

stevehipwell:
Resolves #2489 
Resolves #2495

----

### Before the change?

* The `github_repository` resource churned on `vulnerability_alerts`
* The `github_repository` resource churned on pages if build type was workflow

### After the change?

* No churning on the `github_repository` resource

### Pull request checklist

- [ ] Yes
- [x] No

----


github/resource_github_repository.go
```diff
@@ -320,6 +320,7 @@ func resourceGithubRepository() *schema.Resource {
 			"vulnerability_alerts": {
 				Type:        schema.TypeBool,
 				Optional:    true,
+				Computed:    true,
```
Lekensteyn, line 323: According to the [documentation](https://developer.hashicorp.com/terraform/plugin/sdkv2/schemas/schema-behaviors#computed), "`Computed` is often used to represent values that are not user configurable or can not be known at time of `terraform plan` or `apply`, such as date of creation or a service specific UUID."

I suppose that `Computed` is appropriate here since it could be enabled by default at the org level.

github/resource_github_repository.go
```diff
@@ -619,6 +619,11 @@ func resourceGithubRepositoryCreate(d *schema.ResourceData, meta interface{}) er
 		}
 	}
 
+	err := updateVulnerabilityAlerts(d, client, ctx, owner, repoName)
```
Lekensteyn, line 622: This seems wrong, the `return resourceGithubRepositoryRead(d, meta)` line below can end up calling `updateVulnerabilityAlerts` again. Is this change actually needed?

github/resource_github_repository.go
```diff
@@ -1062,3 +1069,13 @@ func customDiffFunction(_ context.Context, diff *schema.ResourceDiff, v interfac
 	}
 	return nil
 }
+
+func updateVulnerabilityAlerts(d *schema.ResourceData, client *github.Client, ctx context.Context, owner, repoName string) error {
+	updateVulnerabilityAlerts := client.Repositories.DisableVulnerabilityAlerts
+	if vulnerabilityAlerts, ok := d.GetOk("vulnerability_alerts"); ok && vulnerabilityAlerts.(bool) {
```
Lekensteyn, line 1075: Not new behavior, but this treats the absense of `vulnerability_alerts = "true"` as an alias for `vulnerability_alerts = "false"`.

That is not true when a global org-level policy enables vulnerability alerts by default. Is this desirable behavior?

github/resource_github_repository.go
stevehipwell: It is also required in combination with `optional` if an optional value will be persisted even if it's not been explicitly set.

github/resource_github_repository.go
stevehipwell: This part of the TF provider is a bit of a mess, AFAIK 3 months later this is required.

github/resource_github_repository.go
stevehipwell: @Lekensteyn this is a pattern used throughout the provider, the `GetOk()` function is intended to only respond if the value has been set.

stevehipwell:
@nickfloyd this would be a useful quality of life improvement.

PR #2506, opened 2024-12-06
chore: Deprecated classic project resources

### LINKED ISSUE #2494: [BUG]: Resources using the removed legacy projects API are still present and not deprecated in the provider
### Expected Behavior

Resources which will not work should not be in the provider and should have been deprecated when the API deprecation announcement was issued.

### Actual Behavior

The resources using the removed legacy projects API are still present and not deprecated in the provider.

### Terraform Version

n/a

### Affected Resource(s)

- `github_project_card`
- `github_project_column`
- `github_repository_project`

_No response_

### Steps to Reproduce

_No response_

### Debug Output

_No response_

### Panic Output

_No response_

### Code of Conduct

stevehipwell:
Resolves #2494

----

### Before the change?

* Resources using the removed classic projects REST API would not work and were not deprecated

### After the change?

* Resources using the removed classic projects REST API are deprecated

### Pull request checklist

- [ ] Yes
- [x] No

----

### DIFF
--- a/github/resource_github_organization_project.go
+++ b/github/resource_github_organization_project.go
@@ -13,6 +13,8 @@ import (
 
 func resourceGithubOrganizationProject() *schema.Resource {
 	return &schema.Resource{
+		DeprecationMessage: "This resource is deprecated as the API endpoints for classic projects have been removed. This resource no longer works and will be removed in a future version.",
+
 		Create: resourceGithubOrganizationProjectCreate,
 		Read:   resourceGithubOrganizationProjectRead,
 		Update: resourceGithubOrganizationProjectUpdate,
--- a/github/resource_github_organization_project_test.go
+++ b/github/resource_github_organization_project_test.go
@@ -13,6 +13,8 @@ import (
 )
 
 func TestAccGithubOrganizationProject_basic(t *testing.T) {
+	t.Skip("Skipping test as the GitHub REST API no longer supports classic projects")
+
 	if err := testAccCheckOrganization(); err != nil {
 		t.Skipf("Skipping because %s.", err.Error())
 	}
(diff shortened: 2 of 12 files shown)


0x46616c6b:
In the meantime, the go-github SDK removed the (classic) projects API: https://github.com/google/go-github/releases/tag/v68.0.0

So maybe this could be removed entirely now?

stevehipwell:
@0x46616c6b the intention here is/was to deprecate before removal as should have happened when the API was originally deprecated. Technically they were broken before I opened #2494 and should just be removed as their presence will block upgrading the SDK; but if a release is going out before the SDK is upgraded this PR is still valid. IMHO this PR should have been prioritised as it could have gone out in `v6.5.0` or `v6.6.0`.

stevehipwell:
@nickfloyd IMHO this should go into the codebase before the SDK is updated even if the SDK is updated and the files are removed before the next release.

nickfloyd:
> @nickfloyd IMHO this should go into the codebase before the SDK is updated even if the SDK is updated and the files are removed before the next release.

Agreed.  I was sketching out what transitioning to the new project endpoints in GraphQL would look like and it's going to be a fairly significant change, let alone moving to the SDK @ v76.  This type of comms is way over due.  Thank you for making it happen ❤️

nickfloyd (approved):

stevehipwell:
@nickfloyd there is a REST API for projects now.

https://docs.github.com/en/rest/projects/projects?apiVersion=2022-11-28

nickfloyd:
> @nickfloyd there is a REST API for projects now.

🤦 You're right.  Thank you for reminding me. ❤️

Issue #2513, opened 2024-12-13
[FEAT]: Support github_release_asset data source


mdb:
### Describe the need

A `github_release_asset` data source would enable users to natively fetch and/or read GitHub release assets in Terraform.

For example, imagine a `owner/repo` GitHub release with various `*.json` file assets. To read those `*.json` files' contents via a `github_release_asset` data source, I'm imagining something akin to the following (although, it would be useful to support additional/alternative arguments beyond just `id` too, and perhaps even the option to download the asset):

```tf
data "github_release" "latest" {
  repository  = "owner"
  owner       = "repo"
  retrieve_by = "latest"
}

data "github_release_asset" "release_assets" {
  count = length(data.github_release.latest.assets)
  url   = data.github_release.latest.assets[count.index].url
}

output "release_assets" {
  value = data.github_release_asset.release_assets.*.body
}
```

However, currently, without a `github_release_asset` data source, it's necessary to utilize suboptimal workarounds, such as the use of the [http](https://registry.terraform.io/providers/hashicorp/http/latest/docs) provider and some not-ideal mechanism (such as a `var.github_token`) for providing a GitHub token:

```tf
data "github_release" "latest" {
  repository  = "owner"
  owner       = "repo"
  retrieve_by = "latest"
}

data "http" "release_assets" {
  count = length(data.github_release.latest.assets)
  url   = data.github_release.latest.assets[count.index].url

  request_headers = {
    Accept        = "application/octet-stream"
    Authorization = "Bearer ${var.github_token}"
  }
}

output "release_assets" {
  value = data.http.release_assets.*.body
}
```

Note that, in addition to requiring the use of an additional provider (i.e. the `http` provider), the use of the `var.github_token` creates a few complications:

1. It's necessary to ensure it's not printed in plain text in logs, Terraform state, etc. 
2. If its value is generated dynamically via a GitHub app -- and if `plan` and `apply` are divided across distinct phases run on different CI/CD compute infrastructure, each with their own distinct `var.github_token`-generation  -- then the token's value can change across each phase, resulting in `Error: Can't change variable when applying a saved plan` during apply.

So, a `github_release_asset` data source would be very helpful!

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

mdb:
[...]

PR #2514, opened 2024-12-14
feat: add github_release_asset data source

mdb:
This addresses issue #2513 and adds support for a `github_release_asset` data source.

Example of passing acceptance tests:

```
$ GITHUB_ORGANIZATION=mdb GITHUB_OWNER=mdb
GITHUB_TEMPLATE_REPOSITORY=terraputs
GITHUB_TEMPLATE_REPOSITORY_RELEASE_ASSET_ID=206493547
GITHUB_TEMPLATE_REPOSITORY_RELEASE_ASSET_NAME=checksums.txt TF_ACC=1 go
test -v ./... -run ^TestAccGithubReleaseAssetDataSource

?       github.com/integrations/terraform-provider-github/v6    [no test
files]
=== RUN   TestAccGithubReleaseAssetDataSource
=== RUN   TestAccGithubReleaseAssetDataSource/queries_specified_asset_ID
=== RUN
TestAccGithubReleaseAssetDataSource/queries_specified_asset_ID/with_an_anonymous_account
    provider_utils.go:51: GITHUB_TOKEN environment variable should be
empty
    provider_utils.go:74: Skipping
TestAccGithubReleaseAssetDataSource/queries_specified_asset_ID/with_an_anonymous_account
which requires anonymous mode
=== RUN
TestAccGithubReleaseAssetDataSource/queries_specified_asset_ID/with_an_individual_account
=== RUN
TestAccGithubReleaseAssetDataSource/queries_specified_asset_ID/with_an_organization_account
--- PASS: TestAccGithubReleaseAssetDataSource (5.20s)
    --- PASS:
TestAccGithubReleaseAssetDataSource/queries_specified_asset_ID (5.20s)
        --- SKIP:
TestAccGithubReleaseAssetDataSource/queries_specified_asset_ID/with_an_anonymous_account
(0.00s)
        --- PASS:
TestAccGithubReleaseAssetDataSource/queries_specified_asset_ID/with_an_individual_account
(2.57s)
        --- PASS:
TestAccGithubReleaseAssetDataSource/queries_specified_asset_ID/with_an_organization_account
(2.62s)
PASS
ok      github.com/integrations/terraform-provider-github/v6/github
5.688s
```

Resolves #ISSUE_NUMBER

----

### Before the change?

* 

### After the change?

* 

### Pull request checklist

- [ ] Yes
- [ ] No

----


mdb:
Hi @lafrenierejm - This has been waiting on reviews for ~6 months. Is there anything I can do to help steward this along to get it reviewed, merged, and released?

mdb:
@deiga Given your help reviewing PR #2515, would you mind reviewing this as well? Thanks!

deiga:
Hey @mdb 👋

Thanks for the contribution!

The changes look good for the most part. 
I have a few requests for you:
- Could you elaborate the headers change? The comment wasn't enough for me to understand what is happening.
- I wonder if it'd be possible to get a test for the headers change which ensures that it doesn't break any behaviour?
- We're moving towards Context-aware provider functions, could you refactor the Schemata to use ReadContext?
- We're also going to start using documentation generating from resource/data source Descriptions. Could you ensure that all necessary info is in the different Description fields of the Schemata?

mdb:
Thanks for the quick response @deiga ! 

> * Could you elaborate the headers change? The comment wasn't enough for me to understand what is happening.

I've added a comment linking to corresponding context over in https://github.com/google/go-github/pull/3392 (and note that this is fixed in [go-github/v68](https://github.com/google/go-github/releases/tag/v68.0.0) (Admittedly, this is all a bit foggy for me -- it's been over a year since I opened this PR as well as since I opened https://github.com/google/go-github/pull/3392)).

> * I wonder if it'd be possible to get a test for the headers change which ensures that it doesn't break any behaviour?

I added unit tests. Does that address your concern?

> * We're moving towards Context-aware provider functions, could you refactor the Schemata to use ReadContext?

Done!

> * We're also going to start using documentation generating from resource/data source Descriptions. Could you ensure that all necessary info is in the different Description fields of the Schemata?

Done!

mdb:
@deiga 👋 Gentle nudge! Have I addressed your concerns in the response above?

deiga:
@mdb Thanks for the changes! These look good!

I realized that this PR won't land at least before v6.10.0 and there is a major test refactor coming in there. Would you be available to rebase on top of #2986 and conform your `TestAcc*` with those changes?

mdb:
> @mdb Thanks for the changes! These look good!
> 
> I realized that this PR won't land at least before v6.10.0 and there is a major test refactor coming in there. Would you be available to rebase on top of #2986 and conform your `TestAcc*` with those changes?

@deiga Thanks! I pushed a commit doing so. How's that look?

github/data_source_github_release_asset_test.go
```diff
@@ -0,0 +1,72 @@
...
+
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/resource"
+)
+
+func TestAccGithubReleaseAssetDataSource(t *testing.T) {
+
+	testReleaseRepository := testAccConf.testPublicRepository
+
+	// NOTE: the default repository, owner, asset ID, asset name, and asset content
+	// values can be overridden with GITHUB_TEST* environment variables to exercise
```
deiga: The refactoring changes the Env variable format slightly, could you align with that?
```suggestion
	// values can be overridden with GH_TEST* environment variables to exercise
```

mdb: @deiga Thanks! Done!

github/data_source_github_release_asset.go
```diff
@@ -0,0 +1,161 @@
...
+}
+
+func dataSourceGithubReleaseAssetRead(ctx context.Context, d *schema.ResourceData, meta any) diag.Diagnostics {
+	repository := d.Get("repository").(string)
+	owner := d.Get("owner").(string)
+
+	client := meta.(*Owner).v3client
+
+	var err error
+	var asset *github.ReleaseAsset
```
stevehipwell: How come these are being defined here instead of in the idiomatic pattern with a `:=`?

github/data_source_github_release_asset.go
```diff
@@ -0,0 +1,161 @@
...
+	if err != nil {
+		return diag.FromErr(err)
+	}
+	defer respBody.Close()
+
+	buf := new(strings.Builder)
+	_, err = io.Copy(buf, respBody)
+	if err != nil {
+		return diag.FromErr(err)
+	}
```
stevehipwell:
```suggestion
	if _, err := io.Copy(buf, respBody); err != nil {
		return diag.FromErr(err)
	}
```
This would be easier to read.

github/data_source_github_release_asset.go
```diff
@@ -0,0 +1,161 @@
...
+
+	d.SetId(strconv.FormatInt(asset.GetID(), 10))
+	err = d.Set("body", buf.String())
+	if err != nil {
+		return diag.FromErr(err)
+	}
```
stevehipwell:
```suggestion
	if err := d.Set("body", buf.String()); err != nil {
		return diag.FromErr(err)
	}
```
This pattern would be easier to read (same applied to the ones below).

github/data_source_github_release_asset.go
```diff
@@ -0,0 +1,161 @@
...
+	var err error
+	var asset *github.ReleaseAsset
+
+	assetID := int64(d.Get("asset_id").(int))
+	asset, _, err = client.Repositories.GetReleaseAsset(ctx, owner, repository, assetID)
+	if err != nil {
+		return diag.FromErr(err)
+	}
+
+	var respBody io.ReadCloser
```
stevehipwell: Same as above.

github/data_source_github_release_asset.go
```diff
@@ -0,0 +1,161 @@
...
+	clientCopy := client.Client()
+	respBody, _, err = client.Repositories.DownloadReleaseAsset(ctx, owner, repository, assetID, clientCopy)
```
stevehipwell: Why are you passing in a redirect client here which is just a copy of the current client? I'd expect to see `nil` being passed in.

github/data_source_github_release_asset_test.go
```diff
@@ -0,0 +1,72 @@
...
+
+	testReleaseAssetName := "terraform-provider-github_6.4.0_manifest.json"
+	if os.Getenv("GH_TEST_REPOSITORY_RELEASE_ASSET_NAME") != "" {
+		testReleaseAssetName = os.Getenv("GH_TEST_REPOSITORY_RELEASE_ASSET_NAME")
+	}
+
+	testReleaseAssetContent := "{\n  \"version\": 1,\n  \"metadata\": {\n    \"protocol_versions\": [\n      \"5.0\"\n    ]\n  }\n}\n"
+	if os.Getenv("GH_TEST_REPOSITORY_RELEASE_ASSET_CONTENT") != "" {
+		testReleaseAssetContent = os.Getenv("GH_TEST_REPOSITORY_RELEASE_ASSET_CONTENT")
+	}
```
stevehipwell: Any env variable configuration should be in _acc_test.go_.

github/data_source_github_release_asset_test.go
```diff
@@ -0,0 +1,72 @@
...
+			resource.TestCheckResourceAttr(
+				"data.github_release_asset.test", "asset_id", testReleaseAssetID,
+			),
+			resource.TestCheckResourceAttr(
+				"data.github_release_asset.test", "name", testReleaseAssetName,
+			),
+			resource.TestCheckResourceAttr(
+				"data.github_release_asset.test", "body", testReleaseAssetContent,
+			),
+		)
```
stevehipwell: Please inline this instead of declaring it as a var, this pattern makes it harder to read the tests.

github/data_source_github_release_asset.go
mdb: @stevehipwell This is an emulation of the pattern from the `data_source_github_release.go`, though I'm happy to use `:=` if that's preferred:

https://github.com/integrations/terraform-provider-github/blob/main/github/data_source_github_release.go#L166-L167

github/data_source_github_release_asset_test.go
mdb: @stevehipwell For now, I removed the ability to configure the tests via `GH_TEST_*` env variables; that's arguably premature optimization. However, these tests do now source configuration from `testAccConfig` which I enhanced to support some new GitHub release asset-related fields. How's that look?

github/data_source_github_release_asset.go
mdb: Update: I went ahead and pushed a change using the `:=` pattern, despite what's done in `data_source_github_release.go`.

github/data_source_github_release_asset.go
mdb: @stevehipwell My understanding is that's necessary in order to download assets from a redirected location (like when a repository is renamed, for example), as [documented](https://github.com/integrations/terraform-provider-github/blob/main/vendor/github.com/google/go-github/v81/github/repos_releases.go#L345-L346):

> followRedirectsClient can be passed to download the asset from a redirected location

github/data_source_github_release_asset.go
```diff
@@ -0,0 +1,146 @@
...
+		return diag.FromErr(err)
+	}
+	defer respBody.Close()
+
+	buf := new(strings.Builder)
+	if _, err := io.Copy(buf, respBody); err != nil {
+		return diag.FromErr(err)
+	}
+
+	d.SetId(strconv.FormatInt(asset.GetID(), 10))
```
stevehipwell: Are the asset IDs globally unique? If not then you'd need to also use the repository in here.

github/data_source_github_release_asset.go
stevehipwell: Based on reading the SDK code my understanding is that this is only required if the primary client can't handle redirects. Given that you're just copying in the original client I can't see how this is required?

mdb: > I'd expect to see nil being passed in.
> ...
> I can't see how this is required?

@stevehipwell It's necessary to pass a redirect client in order to issue the release asset download request against `https://release-assets.githubusercontent.com`.

Otherwise, passing `nil` (i.e. `respBody, _, err = client.Repositories.DownloadReleaseAsset(ctx, owner, repository, assetID, nil)`) causes us to enter [this codepath](https://github.com/integrations/terraform-provider-github/blob/main/vendor/github.com/google/go-github/v81/github/repos_releases.go#L384) and return the `https://release-assets.githubusercontent.com` release asset URL (example: `https://release-assets.githubusercontent.com/github-production-release-asset/93446099/e3507a32-02f2-41be-81bb-c1d417157055?sp=r&sv=2018-11-09&sr=b&spr=https&se=2026-01-13T15%3A45%3A39Z&rscd=attachment%3B+filename%3Dterraform-provider-github_6.4.0_manifest.json&<REDATED>`). In turn, this causes acceptance test failures:

```
$ TF_ACC=1 \
  go test -v ./... -run ^TestAccGithubReleaseAssetDataSource
?       github.com/integrations/terraform-provider-github/v6    [no test files]
=== RUN   TestAccGithubReleaseAssetDataSource
=== RUN   TestAccGithubReleaseAssetDataSource/queries_specified_asset_ID
[signal SIGSEGV: segmentation violation code=0x2 addr=0x0 pc=0x103108548]

goroutine 94 [running]:
github.com/integrations/terraform-provider-github/v6/github.dataSourceGithubReleaseAssetRead({0x10392d9d0, 0x140003664d0}, 0x140005a3b00, {0x103669ae0, 0x1400001c4c0})
        /Users/mike.ball/go/src/github.com/integrations/terraform-provider-github/github/data_source_github_release_asset.go:105 +0x1b8
github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema.(*Resource).read(0x1400038f700, {0x10392d928, 0x1400049d590}, 0x140005a3b00, {0x103669ae0, 0x1400001c4c0})
        /Users/mike.ball/go/src/github.com/integrations/terraform-provider-github/vendor/github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema/resource.go:866 +0xe0
github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema.(*Resource).ReadDataApply(0x1400038f700, {0x10392d928, 0x1400049d590}, 0x0?, {0x103669ae0, 0x1400001c4c0})
        /Users/mike.ball/go/src/github.com/integrations/terraform-provider-github/vendor/github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema/resource.go:1088 +0x124
github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema.(*GRPCProviderServer).ReadDataSource(0x1400000c120, {0x10392d928?, 0x1400049d500?}, 0x1400049d4d0)
        /Users/mike.ball/go/src/github.com/integrations/terraform-provider-github/vendor/github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema/grpc_provider.go:1889 +0x544
github.com/hashicorp/terraform-plugin-go/tfprotov5/tf5server.(*server).ReadDataSource(0x1400040e3c0, {0x10392d928?, 0x1400049c480?}, 0x140000575e0)
        /Users/mike.ball/go/src/github.com/integrations/terraform-provider-github/vendor/github.com/hashicorp/terraform-plugin-go/tfprotov5/tf5server/server.go:736 +0x1b4
github.com/hashicorp/terraform-plugin-go/tfprotov5/internal/tfplugin5._Provider_ReadDataSource_Handler({0x1038e97c0, 0x1400040e3c0}, {0x10392d928, 0x1400049c480}, 0x14000203500, 0x0)
        /Users/mike.ball/go/src/github.com/integrations/terraform-provider-github/vendor/github.com/hashicorp/terraform-plugin-go/tfprotov5/internal/tfplugin5/tfplugin5_grpc.pb.go:843 +0x1b8
google.golang.org/grpc.(*Server).processUnaryRPC(0x140003f4200, {0x10392d928, 0x14000254f90}, 0x140005125a0, 0x1400051e780, 0x10435ca30, 0x0)
        /Users/mike.ball/go/src/github.com/integrations/terraform-provider-github/vendor/google.golang.org/grpc/server.go:1431 +0xbd0
google.golang.org/grpc.(*Server).handleStream(0x140003f4200, {0x10392e470, 0x140005a6340}, 0x140005125a0)
        /Users/mike.ball/go/src/github.com/integrations/terraform-provider-github/vendor/google.golang.org/grpc/server.go:1842 +0x858
google.golang.org/grpc.(*Server).serveStreams.func2.1()
        /Users/mike.ball/go/src/github.com/integrations/terraform-provider-github/vendor/google.golang.org/grpc/server.go:1061 +0x74
created by google.golang.org/grpc.(*Server).serveStreams.func2 in goroutine 116
        /Users/mike.ball/go/src/github.com/integrations/terraform-provider-github/vendor/google.golang.org/grpc/server.go:1072 +0x120
FAIL    github.com/integrations/terraform-provider-github/v6/github     1.640s
FAIL
```

github/data_source_github_release_asset.go
mdb: @stevehipwell My understanding is [they are unique identifiers](https://docs.github.com/en/rest/releases/assets?apiVersion=2022-11-28#get-a-release-asset), similar to the unique identifiers associated with releases (and this implementation emulates the [analogous pattern currently used by data_source_github_release](https://github.com/integrations/terraform-provider-github/blob/main/github/data_source_github_release.go#L194)), and so it's not necessary to build a composite ID that includes `OWNER/REPO`. Do you disagree?

github/data_source_github_release_asset.go
deiga: A valid quest here is: is the asset URL enough for this data source or is the asset contents something we want to store in the state?

mdb: @deiga Personally, I don't believe the asset URL is enough, at least not for the original use case I was imagining (and outlined in https://github.com/integrations/terraform-provider-github/issues/2513); I want to natively read the assets themselves in Terraform, then drive Terraform configuration based on their values.

stevehipwell: I'm trying to figure out why the SDK pattern is still like this (https://github.com/google/go-github/issues/3915).

github/data_source_github_release_asset.go
deiga: For safety sake I would opt for using `buildThreePartID(owner, repo, assetID)` here to ensure there are no clashes

github/data_source_github_release_asset_test.go
```diff
@@ -0,0 +1,49 @@
...
+			data "github_release_asset" "test" {
+				repository = "%s"
+				owner = "%s"
+				asset_id = "%s"
+			}
+		`, testReleaseRepository, testRepositoryOwner, testReleaseAssetID)
+
+		resource.Test(t, resource.TestCase{
+			ProviderFactories: providerFactories,
+			Providers:         testAccProviders,
```
deiga:
```suggestion
			ProviderFactories: providerFactories,
```

github/data_source_github_release_asset.go
mdb: @deiga Sounds reasonable; I pushed a commit doing so!

github/data_source_github_release_asset_test.go
mdb: @deiga Thanks; done!

github/data_source_github_release_asset.go
```diff
@@ -0,0 +1,146 @@
...
+	if err != nil {
+		return diag.FromErr(err)
+	}
+
+	clientCopy := client.Client()
+	respBody, _, err := client.Repositories.DownloadReleaseAsset(ctx, owner, repository, assetID, clientCopy)
+	if err != nil {
+		return diag.FromErr(err)
+	}
+	defer respBody.Close()
```
stevehipwell: Please don't use the `DownloadReleaseAsset` function as it modifies the GitHub client state which could cause unexpected behaviour in Terraform. You can use the client copy directly.

mdb: @stevehipwell I've pushed a commit using the client copy instead.

mdb:
> As a general comment I think that the binary download ought to be optional and off by default.

@stevehipwell Good call; I've pushed a change implementing this.

stevehipwell (changes requested):
Thanks @mdb, this PR is looking good. The only issue I have is that I don't like the term _body_ for this, I prefer _binary_ or _bytes_ but am open to suggestions?

mdb:
> Thanks @mdb, this PR is looking good. The only issue I have is that I don't like the term _body_ for this, I prefer _binary_ or _bytes_ but am open to suggestions?

@stevehipwell I don't love that either (nor do I love `download_body`). Although, on the other hand, it is the response _body_ to an asset download URL request. What about `asset_file_contents`?

Alternatively, could you elaborate a bit on your preference for _binary_ or _bytes_? I ask because, in my mind, it's not necessarily binary. For example, the acceptance tests fetch https://github.com/integrations/terraform-provider-github/releases/tag/v6.4.0 's `terraform-provider-github_6.4.0_manifest.json`'s contents, which is a JSON string. But perhaps I'm misunderstanding you?

stevehipwell:
@mdb I don't like body as a term for this, especially as the "body" of this API is determined by the accept type and effectively this data source carries 2 "bodies". What about `download_file` & `file`? Could you also expand on the fact that the `file` is base64 encoded bytes.

mdb:
> @mdb I don't like body as a term for this, especially as the "body" of this API is determined by the accept type and effectively this data source carries 2 "bodies". What about `download_file` & `file`? Could you also expand on the fact that the `file` is base64 encoded bytes.

@stevehipwell I think `download_file`/`file` sounds reasonable. I've pushed a commit implementing this.

website/docs/d/release_asset.html.markdown
```diff
@@ -0,0 +1,88 @@
...
+* `url` - URL of the asset
+* `node_id` - Node ID of the asset
+* `name` - The file name of the asset
+* `label` - Label for the asset
+* `content_type` - MIME type of the asset
+* `size` - Asset size in bytes
+* `created_at` - Date the asset was created
+* `updated_at` - Date the asset was last updated
+* `browser_download_url` - Browser URL from which the release asset can be downloaded
+* `file` - The release asset file contents (requires `download_file` to be `true`)
```
stevehipwell: Could you add some detail about the file format (e.g. base64 encoded bytes)?

mdb: @stevehipwell I'm confused about that statement, to be honest. Why do you believe `file`'s value to be base64-encoded bytes?

1. In the acceptance tests, it's a non-encoded string
2. I don't see any mention of base64-encoded in the [API endpoint documentation](https://docs.github.com/en/rest/releases/assets?apiVersion=2022-11-28#get-a-release-asset)

Am I confused and overlooking something?

github/data_source_github_release_asset.go
```diff
@@ -0,0 +1,170 @@
...
+	defer resp.Body.Close()
+
+	if resp.StatusCode != http.StatusOK {
+		return diag.Errorf("failed to get release asset (%s/%s %d): %s", owner, repository, assetID, resp.Status)
+	}
+
+	buf := new(strings.Builder)
+	if _, err := io.Copy(buf, resp.Body); err != nil {
+		return diag.FromErr(err)
+	}
```
stevehipwell, line 178:
```suggestion
	buf := new(strings.Builder)
	encoder := base64.NewEncoder(base64.StdEncoding, buf)
	defer encoder.Close()

	buffer := make([]byte, 4096)
	for {
		n, err := resp.Body.Read(buffer)
		if err != nil && err != io.EOF {
			return diag.FromErr(err)
		}
		if n > 0 {
			if _, err := encoder.Write(buffer[:n]); err != nil {
				return diag.FromErr(err)
			}
		}
		if err == io.EOF {
			break
		}
	}
```

Sorry, I missed this in my previous review.

The artifact content, as defined by the accept of `application/octet-stream`, is _"Generic binary data (or binary data whose true type is unknown) is application/octet-stream."_ ([ref](https://developer.mozilla.org/en-US/docs/Web/HTTP/Guides/MIME_types#application)). Therefore we need to treat the bytes as base64 instead of assuming that the content is utf8 and converting it into a string.

website/docs/d/release_asset.html.markdown
stevehipwell: Please see latest review.

github/data_source_github_release_asset.go
mdb: @stevehipwell Ah yes, right right - thanks for explaining. I've pushed a fix.

github/data_source_github_release_asset.go
```diff
@@ -0,0 +1,185 @@
...
+				Required:    true,
+				Description: "Name of the repository to retrieve the release asset from",
+			},
+			"download_file": {
+				Type:        schema.TypeBool,
+				Optional:    true,
+				Default:     false,
+				Description: "Whether to download the asset content into the file attribute",
+			},
+			"file": {
```
stevehipwell: Given that this is base64 encoded bytes I'm back to thinking that `contents` or `file_contents` would be a better fit here.

github/data_source_github_release_asset_test.go
```diff
@@ -0,0 +1,80 @@
...
+					Config: config,
+					Check: resource.ComposeTestCheckFunc(
+						resource.TestCheckResourceAttr(
+							"data.github_release_asset.test", "asset_id", testReleaseAssetID,
+						),
+						resource.TestCheckResourceAttr(
+							"data.github_release_asset.test", "name", testReleaseAssetName,
+						),
+						resource.TestCheckResourceAttr(
+							"data.github_release_asset.test", "file", base64EncodedAssetContent,
```
stevehipwell: Why not use a Terraform output and the [base64decode](https://developer.hashicorp.com/terraform/language/functions/base64decode) function for a more TF focused check?

github/data_source_github_release_asset.go
mdb: @stevehipwell Sounds good; I've pushed a commit updating to `download_file_contents`/`file_contents`.

github/data_source_github_release_asset_test.go
mdb: @stevehipwell Sounds good; I've pushed another commit doing so.

mdb:
@stevehipwell 👋 Hello! Gentle nudge: How do my latest round of changes look? I believe I've implemented all your suggestions :)

mdb:
> LGTM

@stevehipwell Thanks! It looks like [some static analysis checks failed CI](https://github.com/integrations/terraform-provider-github/actions/runs/21232219897/job/61621068985?pr=2514); I've pushed some commits that should fix these.

stevehipwell:
@mdb it looks like you've got failing tests.

mdb:
> @mdb it looks like you've got failing tests.

@stevehipwell Doh, thanks! I've pushed a fix.

nickfloyd (approved):

stevehipwell:
@mdb could you please rebase this so we can get it merged?

mdb:
> @mdb could you please rebase this so we can get it merged?

@stevehipwell Done!

github/data_source_github_release_asset.go
```diff
@@ -0,0 +1,185 @@
...
+
+	// Use a client copy to avoid possible mutation of shared GitHub client state
+	// by client.Repositories.DownloadReleaseAsset.
+	clientCopy := client.Client()
+	req, err := http.NewRequestWithContext(ctx, "GET", asset.GetBrowserDownloadURL(), nil)
+	if err != nil {
+		return diag.FromErr(err)
+	}
+
+	req.Header.Set("Accept", "application/octet-stream")
```
felipebraga, line 150: Set and `Accept` header causes the http request to return 406 Not Acceptable.

`req.Header.Set("Accept", "application/octet-stream")`

```suggestion
```

website/docs/d/release_asset.html.markdown
```diff
@@ -0,0 +1,88 @@
...
+
+To retrieve a specific release asset from a repository, and download the file
+into a `file` attribute on the data source:
+
+```hcl
+data "github_release_asset" "example" {
+    repository    = "example-repository"
+    owner         = "example-owner"
+    asset_id      = 12345
+    download_file = true
```
felipebraga, line 31:
```suggestion
    download_file_contents = true
```

Issue #2503, opened 2024-12-05
[FEAT]: Add `owner` attribute to `github_repository` data source and resource


gdavison:
### Describe the need

The `github_repository` data source works in two "modes": authenticated or anonymous.

When authenticated, the owner can be implied from the current provider instance, so all that is needed to look up the repository is the `name`. The `full_name` is populated in the `<organization>/<name>` format.

When anonymous, the owner cannot be implied, so the `full_name` must be used. `name` is returned as the name portion of the full name.

I propose adding an `Optional, Computed` `owner` field to the data source. This could be used in the anonymous mode along with `name` to specify the repository. Otherwise, it would be populated with the owner of the repository.

This would also have the benefit of keeping all information about the repository contained within the data source object so that it could be used in e.g. a module. Currently, either the owner must be passed as a separate variable or the `full_name` must be split. This would make the owner a "first-class" field of the data source alongside `name`, rather than needing to derive it.

For parity between the data source and the resource, also add `owner` as a `Computed` field to the resource.

For example, a module can take the resource or data source as an `object` value. The object will have `full_name` and `name` as attributes, but if the owner is needed, it must either be passed separately or parsed from `full_name`.

Currently your module must be like either

```terraform
variable "repository" {
  type = object({
    name = string
    ...
  })
}

variable "repository_owner" {
  type = string
}

# Do something with separate fields
...
  repo_owner = var.repository_owner
  repo_name  = var.repository.name
```

or

```terraform
variable "repository" {
  type = object({
    full_name = string
    name      = string
    ...
  })
}

locals {
  repository_owner = split("/", var.repository.full_name)[0]
}

# Do something with separate fields
...
  repo_owner = local. repository_owner
  repo_name  = var.repository.name
```

If `owner` is added as a field, we could do the following:

```terraform
variable "repository" {
  type = object({
    name  = string
    owner = string
    ...
  })
}

# Do something with separate fields
...
  repo_owner = var.repository.owner
  repo_name  = var.repository.name
```

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

stevehipwell:
@gdavison are you just asking for a computed R/O `owner` attribute on the `github_repository` resource and data source so you don't have to split the `full_name`?

gdavison:
[...]

PR #2528, opened 2025-01-03
[FEAT]: Add `owner` as parameter for `github_action_secret`

diofeher:
GitHub Actions secrets do not support repos from another organization. The only supported repos are the same as those from the owner's token.

----

### Before the change?

* You cannot set a secret for an organization's repo using a personal access token, because the owner will be the user.

### After the change?

* You can set the `owner` with the organization name

### Pull request checklist

- [ ] Yes
- [x] No

----


diofeher:
@nickfloyd any updates on this one?

diofeher:
Related to #2503

deiga:
@diofeher Thank you for this contribution!

What is the use-case for having a separate input field for owner? 
I mean, I understand that you might want to add a secret to repo that doesn't belong to the configured owner. But couldn't we use the repository resource/data source instead of adding a separate field?

For example:
If we'd use a `repo_full_name` field, we could pass in `github_repository.foo.full_name`
Or if we'd use a `repo_ref` field, we could pass in `github_repository.foo` and then parse whichever fields we wanted from it

Using a separate `owner` field feels clunky to me as I imagine we want to encourage the users to use resources/data sources and pass them "by reference", instead of "free text" fields.

What do you think?

diofeher:
@deiga Thanks for the response! I don't remember the specific use case for what I was trying to do. Still, I was probably using a personal classic token to access both an individual's and an organization's secrets simultaneously.

I enjoyed your idea about using the `repo_full_name`, let me rework on that!

But one thing must be noticed: we use the `owner` field in other resources, like `github_repository_milestone` and `github_repository_pull_request`.

diofeher:
The only caveat with my current implementation is that we would need to support creating repositories in other orgs using `github_repository` to have the full acceptance test implemented.

stevehipwell:
I think this behaviour change warrants wider discussion so I've opened #2940.

deiga:
> I think this behaviour change warrants wider discussion so I've opened #2940.

Due to this, I withdraw my original suggestion. 
@diofeher could you rebase and go back to using just an explicit string `owner` attribute?
Then this can get merged and we can think about reworking how `owner` is used in a different context

PR #2540, opened 2025-01-17
Add public field to repository pages

### LINKED ISSUE #1045: feature request: Add support for 'public' attribute on github pages configuration

BartolottiLuca:
Resolves #1045 

----

### Before the change?

* Pages availability is not configurable via terraform

### After the change?

* Pages availability can now be configured via terraform via the `public` field

### Pull request checklist

- [ ] Yes
- [x] No

----


website/docs/r/repository.html.markdown
```diff
@@ -136,6 +136,8 @@ The `pages` block supports the following:
 
 * `cname` - (Optional) The custom domain for the repository. This can only be set after the repository has been created.
 
+* `public` - (Optional) Whether the rendered GitHub Pages site is publicly availabe.
```
usmonster, line 139:
```suggestion
* `public` - (Optional) Whether the rendered GitHub Pages site is publicly available.
```

BartolottiLuca: thanks!

cyberkni:
I'd really like to have this feature available! Please!

github/resource_github_repository.go
```diff
@@ -364,9 +364,14 @@ func resourceGithubRepository() *schema.Resource {
...
 						},
+						"public": {
+							Type:        schema.TypeBool,
+							Computed:    true,
+							Description: "Whether the rendered GitHub Pages site is publicly availabe.",
+						},
 						"status": {
 							Type:        schema.TypeString,
-							Computed:    true,
+							Optional:    true,
```
deiga: issue: The `status` field AFAIK can't be set by the user so it needs to at least be kept as `Computed`. What did you encounter to require this change?

github/resource_github_repository.go
```diff
@@ -364,9 +364,14 @@ func resourceGithubRepository() *schema.Resource {
 							Computed:    true,
 							Description: "URL to the repository on the web.",
 						},
+						"public": {
+							Type:        schema.TypeBool,
+							Computed:    true,
+							Description: "Whether the rendered GitHub Pages site is publicly availabe.",
```
deiga: Let's use the more descriptive text to make sure users know what they are doing :)
```suggestion
							Description: "Whether the GitHub Pages site is publicly visible. If set to `true`, the site is accessible to anyone on the internet. If set to `false`, the site will only be accessible to users who have at least `read` access to the repository that published the site.",
```

github/resource_github_repository.go
```diff
@@ -1094,7 +1099,12 @@ func expandPages(input []any) *github.Pages {
 		buildType = github.String(v)
 	}
 
-	return &github.Pages{Source: source, BuildType: buildType}
+	var public *bool
+	// Only set the github.PagesUpdate public field if the value is a valid boolean.
+	if v, ok := pages["public"].(bool); ok {
+		public = github.Bool(v)
+	}
+	return &github.Pages{Source: source, BuildType: buildType, Public: public}
```
deiga: question: I'm wondering if this change is necessary/worthwhile. `expandPages` is only called in the `*Create` method before running `client.Repositories.EnablePages`, but `EnablePages` doesn't use the `Public` parameter at all, if I read the SDK code correctly.

It seems that visibility is only configured with `UpdatePages`

deiga (changes requested):

deiga:
@BartolottiLuca Please rebase your PR :)

github/resource_github_repository.go
BartolottiLuca: Unsure, it was a while ago and this is my first time working on a terraform provider. I trust you that it should be Computed

github/resource_github_repository.go
BartolottiLuca: Yes, you are correct, fixed it

BartolottiLuca:
@deiga Thanks for the review, I've applied your suggestions and rebased

github/resource_github_repository.go
```diff
@@ -1120,6 +1125,11 @@ func expandPagesUpdate(input []any) *github.PagesUpdate {
 		}
 		update.Source = &github.PagesSource{Branch: &sourceBranch, Path: &sourcePath}
 	}
+	
+	// Only set the github.PagesUpdate public field if the value is a valid boolean.
+	if v, ok := pages["public"].(bool); ok && v != "" {
+		update.Public = github.Bool(v)
```
deiga: `gitub.Bool` is deprecated
```suggestion
		update.Public = github.Ptr(v)
```

github/resource_github_repository_test.go
```diff
@@ -1312,6 +1312,53 @@ func Test_expandPages(t *testing.T) {
 			t.Errorf("got %q; want %q", pages.GetSource().GetPath(), "/docs")
 		}
 	})
+
+	t.Run("manages private pages feature for a repository", func(t *testing.T) {
+		config := fmt.Sprintf(`
+			resource "github_repository" "test" {
+				name         = "tf-acc-%s"
```
deiga: issue: Please use `testResourcePrefix` for all name attributes

github/resource_github_repository_test.go
```diff
@@ -1312,6 +1312,53 @@ func Test_expandPages(t *testing.T) {
...
+				}
+			}
+		`, randomID)
+
+		check := resource.ComposeTestCheckFunc(
+			resource.TestCheckResourceAttr(
+				"github_repository.test", "pages.0.public",
+				"false",
+			),
+		)
```
deiga, line 1445: Please inline this

github/resource_github_repository_test.go
```diff
@@ -1312,6 +1312,53 @@ func Test_expandPages(t *testing.T) {
...
+
+		testCase := func(t *testing.T, mode string) {
+			resource.Test(t, resource.TestCase{
+				PreCheck:  func() { skipUnlessMode(t, mode) },
+				Providers: testAccProviders,
```
deiga:
```suggestion
				ProviderFactories: providerFactories,
```

github/resource_github_repository_test.go
```diff
@@ -1312,6 +1312,53 @@ func Test_expandPages(t *testing.T) {
...
+			t.Skip("anonymous account not supported for this feature")
+		})
+
+		t.Run("with an individual account", func(t *testing.T) {
+			t.Skip("individual account not supported for this feature")
+		})
+
+		t.Run("with an organization account", func(t *testing.T) {
+			testCase(t, organization)
+		})
```
deiga: we don't use this structure anymore, please follow the new convention :)

github/resource_github_repository.go
```diff
@@ -372,6 +372,11 @@ func resourceGithubRepository() *schema.Resource {
 							Computed:    true,
 							Description: "URL to the repository on the web.",
 						},
+						"public": {
+							Type:        schema.TypeBool,
+							Computed:    true,
```
deiga, line 378: question: This doesn't seem to get set anywhere and in the docs you write that it's optional. Did you mean to use `Optional` here? Should it have `Default: false` as well to remove the need for type-checking this later?

deiga (changes requested):

github/resource_github_repository.go
BartolottiLuca: I'm unsure which is the best option here. From the [docs](https://docs.github.com/en/enterprise-cloud@latest/pages/getting-started-with-github-pages/changing-the-visibility-of-your-github-pages-site#about-access-control-for-github-pages-sites)

>To publish a GitHub Pages site privately, your organization must use GitHub Enterprise Cloud. For more information about how you can try GitHub Enterprise Cloud for free, see Setting up a trial of GitHub Enterprise Cloud.
If your enterprise uses Enterprise Managed Users, GitHub Pages sites can only be published as private, and all GitHub Pages sites are only accessible to other enterprise members. For more information about Enterprise Managed Users, see GitHub Pages limits.
If your organization uses GitHub Enterprise Cloud without Enterprise Managed Users, you can choose to publish your project sites privately or publicly to anyone on the internet.


So there's no clear default since:
1. anyone without github enterprise will be forced to have public pages
2. If you have enterprise and Enterprise Managed Users, the pages can only be private
3. only users that have enterprise and not Enterprise Managed Users can choose between public and private


I guess `Computed: true` is safer?

github/resource_github_repository.go
BartolottiLuca: applied

github/resource_github_repository_test.go
BartolottiLuca: done

github/resource_github_repository_test.go
BartolottiLuca: gave the new convention a go @deiga. let me know if it's correct

github/resource_github_repository.go
deiga: Yeah, that does look problematic. I think that having only `Computed: true` prevents users from setting the value, which would be unwelcome for GHEC non-EMU cases. 

Maybe you could check if the SDK sends the field or it's left blank, otherwise the Create and Update need to ensure it's not being sent unless explicitly set. 

You're right that Default isn't a good idea, but I guess it needs to be Computed & Optional

website/docs/r/repository.html.markdown
```diff
@@ -156,6 +156,8 @@ The `pages` block supports the following:
 
 * `cname` - (Optional) The custom domain for the repository. This can only be set after the repository has been created.
 
+* `public` - (Optional) Whether the GitHub Pages site is publicly visible. If set to `true`, the site is accessible to anyone on the internet. If set to `false`, the site will only be accessible to users who have at least `read` access to the repository that published the site.
```
deiga: Could you add a link to the section in the GH docs about the limitations for different Owner types? I think that would be valuable info for users

github/resource_github_repository_test.go
```diff
@@ -1339,6 +1339,38 @@ func Test_expandPages(t *testing.T) {
...
+				}
+			}
+		`, testRepoName)
+
+		check := resource.ComposeTestCheckFunc(
+			resource.TestCheckResourceAttr("github_repository.test", "pages.0.public", "false"),
+		)
+
+		resource.Test(t, resource.TestCase{
+			PreCheck:          func() { skipUnlessMode(t, organization) },
```
deiga: nitpick: we have a skipUnlessHasOrgs (or similar) function

deiga (changes requested):

BartolottiLuca:
@deiga sorry it took me so long, I've applied all the suggestions

deiga (changes requested):
Could you rebase your branch on top of [this PR](https://github.com/integrations/terraform-provider-github/pull/3168) and moving your changes to the pages resource? :)

We're in the process of deprecating the `pages` config in `github_repository`

deiga:
Superseded by #3168

PR #2544, opened 2025-01-22
Use the correct `update_allows_fetch_and_merge` parameter in the rules mapping

### LINKED ISSUE #2463: [BUG]: Property `update_allows_fetch_and_merge` indefinitely appear as a part of terraform plan/apply
### Expected Behavior

When the property `update_allows_fetch_and_merge` is changed to `false`, it should be applied once and no longer appear in the terraform plan.

### Actual Behavior

The terraform apply result looks fine:

`Apply complete! Resources: 0 added, 23 changed, 0 destroyed.`

But even after applying, it still appears in a terraform plan like below:

    ~ rules {
        ~ update_allows_fetch_and_merge = true -> false
      }

### Terraform Version

Terraform v1.8.4
on ubuntu-20.04
+ provider registry.terraform.io/integrations/github v6.3.0

### Affected Resource(s)

- github_repository_ruleset

_No response_

### Steps to Reproduce

_No response_

### Debug Output

_No response_

### Panic Output

_No response_

### Code of Conduct

yurii-kysel:
Previously the `update` parameter was used to set the `update_allows_fetch_and_merge` value.
Based on the documentation it should be `update_allows_fetch_and_merge` in case when `update` is set to true

Resolves #2463

----

### Before the change?

* When the `update` parameter is set to true - `update_allows_fetch_and_merge` also was set to true automatically, even if we have `update_allows_fetch_and_merge: false` in our rules

### After the change?

* When the `update` parameter is set to true -  `update_allows_fetch_and_merge` would be set based on the rule we defined in `update_allows_fetch_and_merge`

### Pull request checklist

- [ ] Yes
- [x] No

----

### DIFF
--- a/github/respository_rules_utils.go
+++ b/github/respository_rules_utils.go
@@ -209,7 +209,7 @@ func expandRules(input []interface{}, org bool) []*github.RepositoryRule {
 
 	if v, ok := rulesMap["update"].(bool); ok && v {
 		params := github.UpdateAllowsFetchAndMergeRuleParameters{}
-		if fetchAndMerge, ok := rulesMap["update"].(bool); ok && fetchAndMerge {
+		if fetchAndMerge, ok := rulesMap["update_allows_fetch_and_merge"].(bool); ok && fetchAndMerge {
 			params.UpdateAllowsFetchAndMerge = true
 		} else {
 			params.UpdateAllowsFetchAndMerge = false


yurii-kysel:
The tests were not added, as the use-case is only applicable to forked repositories (which is currently not possible to be created using the terraform). In case with the normal repo, the `update_allows_fetch_and_merge` would always be false in output

PR #2553, opened 2025-01-31
fix: Do not ignore file deletion errors

giner:
Resource: `github_repository_file`

When deleting a file errors (such as `409 Could not delete file: Changes must be made through a pull request`) should not be silently ignored leaving the file undeleted from the repo but deleted from the state

----

### Before the change?

Resource github_repository_file ignores any errors when deleting files which results in inconsistencies

### After the change?

Resource github_repository_file stops on errors if any when deleting files

### Pull request checklist

- [ ] Yes
- [ ] No

----


giner:
I've just found a similar change which has not been merged yet #2465

Arlington1985:
There should be functionality to be able to mute error as well. For example in our setup we only want to create file but not deleted with it.

stevehipwell:
This was resolved in #2837.

PR #2559, opened 2025-02-07
Remove ForceNew from GithubActionsOrganizationSecret

### LINKED ISSUE #2518: [BUG]: 'plaintext_value' forces delete and create instead of inplace update
### Expected Behavior

When changing 'plaintext_value' we want to replace the value inplace, as we can do in github UI.

### Actual Behavior

The force replace firstly deletes the value and then creates it which leads to secret downtime.  I think it would be better to update the value **INPLACE**

### Terraform Version

Terraform v1.8.4
on darwin_arm64
+ provider registry.terraform.io/hashicorp/azurerm v3.106.1
+ provider registry.terraform.io/hashicorp/github v6.2.1
+ provider registry.terraform.io/rewe-digital/sonarcloud v0.5.2

### Affected Resource(s)

- github_dependabot_secret
- github_actions_secret

maybe more that setup secrets

_No response_

### Steps to Reproduce

_No response_

### Debug Output

_No response_

### Panic Output

_No response_

### Code of Conduct

mjeffryes:
Resolves #2518

### Before the change?
The provider currently forces a replace on any change to the secret plain text, encrypted value or visibility. However, the Github API allows in-place updates to secrets so this is effectively just forcing an unnecessary delete. This causes performance issues as in the linked ticket and also correctness issues when `create_before_destroy` is used. (As it is by default, when this provider is used from Pulumi, see https://github.com/pulumi/pulumi-github/issues/250#issuecomment-2644205475)

### After the change?
Changes to the plaintext, encrypted value or visibility will cause an in place update of the secret. This results in faster and correct zero downtime updates to secrets.

### Pull request checklist

- [ ] Yes
- [x] No

----

### DIFF
--- a/github/resource_github_actions_organization_secret.go
+++ b/github/resource_github_actions_organization_secret.go
@@ -37,7 +37,6 @@ func resourceGithubActionsOrganizationSecret() *schema.Resource {
 			},
 			"encrypted_value": {
 				Type:             schema.TypeString,
-				ForceNew:         true,
 				Optional:         true,
 				Sensitive:        true,
 				ConflictsWith:    []string{"plaintext_value"},
@@ -46,7 +45,6 @@ func resourceGithubActionsOrganizationSecret() *schema.Resource {
 			},
 			"plaintext_value": {
 				Type:          schema.TypeString,
-				ForceNew:      true,
 				Optional:      true,
 				Sensitive:     true,
 				ConflictsWith: []string{"encrypted_value"},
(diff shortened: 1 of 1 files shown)


deiga:
Hey @mjeffryes 👋 

Thanks for your contribution! We're discussing what the original design decision was behind the `ForceNew` to determine if removing it is feasible.

We're currently trying to get v6.9.1 released and are not adding more changes to it. After that we're trying to get v7.0.0 released which has bigger fundamental changes. It's unclear if there will be another v6.x before we can get to v7

While you're PR doesn't currently require a rebase, would you anyway be available to rebase it?

PR #2637, opened 2025-04-25
feat: Add support for multi var/envvars in repos

### LINKED ISSUE #2526: [FEAT]: Reduce API throttling by reading all env variables per env at once.
### Describe the need

When running plans against 64 repos, we hit API throttling at about an order of magnitude earlier due to the reading of env variables one at a time rather than one API call to get all of them (the data source can read them all in one call).    WRT number of resources per repo, env variables currently dominate, since there may be 10 or 50 per environment.

We are re-planning repos and Azure resources (using a matrix per repo - so one plan per repo in parallel), and the process takes about 8 minutes on 64 repos, consuming about 60% of the budget, i.e., we cannot run this twice in the same hour.

I have identified 3 possible solutions:
- on a read, read all values for the environment and cache them in memory.  This is simple, but I'm unclear about how this might affect terraform.
- create a new resource that holds the env variable map for the entire environment, similar to the data source.
- Work around this with a provisioner and the data source to trigger the provisioner to run.  The provisioner would sync a per env map with the contents of the environment.

### SDK Version

_No response_

### API Version

_No response_

### Relevant log output

_No response_

### Code of Conduct

bradam12:
Resolves #2526

----

### Before the change?

N/A (new feature) 

### After the change?

* Able to manage repo variables and repo environment variables in a single resource.

### Pull request checklist

- [ ] Yes
- [x] No

----


deiga:
Hey @bradam12 👋 

Thank you for your contribution! Could you describe your need here in more detail?
Since each variable will need to have it's own API call, what is the benefit using this versus `github_actions_variable` with a `for_each` loop?

bradam12:
> Hey @bradam12 👋
> 
> Thank you for your contribution! Could you describe your need here in more detail? Since each variable will need to have it's own API call, what is the benefit using this versus `github_actions_variable` with a `for_each` loop?

It's fairly documented in https://github.com/integrations/terraform-provider-github/issues/2526. Reads can be done in one API call (per 100 variables), so future plans are much quicker and burn less API calls. Applys still write one at a time.

deiga:
Hmm, I see. 

I understand that the read can be quite consuming of the API rate. 

The implementation you proposed here has me worried that failures in one API call will leave the resource in an wonky state.

deiga:
While this might reduce the amount of API calls in some situations, the implementation is not something that the provider can support. 
Having an "unlimited" number of "sub-resources" that each require it's own API calls would make this resource incredibly brittle to weird state issues due to network flukes or rate limit throttling. 

Thanks for the contribution!

PR #2640, opened 2025-04-25
fix: imported custom property type

### LINKED ISSUE #2639: [BUG]: imported github_repository_custom_property type missing
### Expected Behavior

On importing a repository custom property, it should include the type, and the resource should not be replaced where the resource configuration is correct.

### Actual Behavior

The `property_type` is set to null, which causes the configuration to trigger a forced replacement of the resource.

### Terraform Version

Terraform v1.10.4
on darwin_arm64
+ provider registry.terraform.io/integrations/github v6.6.0

### Affected Resource(s)

- github_repository_custom_property

```hcl
resource "github_repository_custom_property" "test" {
  repository     = "test_repo"
  property_name  = "test_property"
  property_type  = "string"
  property_value = ["test_value"]
}

import {
  to = github_repository_custom_property.test
  id = "test_org:test_repo:test_property"
}
```

### Steps to Reproduce

Import an existing custom property with the config like the above, it will force recreation due to the difference in types.

### Debug Output

```shell
# Warning: this will destroy the imported resource
-/+ resource "github_repository_custom_property" "project" {
      ~ id             = "REDACTED" -> (known after apply)
        property_name  = "Project"
      + property_type  = "string" # forces replacement
        property_value = [
            "REDACTED",
        ]
        repository     = "REDACTED"
    }
```

### Panic Output

```shell

```

### Code of Conduct

imaspen:
Imported custom properties currently default the type to null. This causes the state to then update with the real type value, which triggers a forced replacement. This fetches the type from the organization property schema and adds it on import.

Resolves #2639

----

### Before the change?

* An imported custom property resource's `property_type` is set to null, which causes the configuration to trigger a forced replacement of the resource.

### After the change?

* The property type is fetched from the organisation schema and set correctly, preventing the resource replacement.

### Pull request checklist

- [ ] Yes
- [x] No

----

### DIFF
--- a/github/resource_github_repository_custom_property.go
+++ b/github/resource_github_repository_custom_property.go
@@ -22,7 +22,7 @@ func resourceGithubRepositoryCustomProperty() *schema.Resource {
 		Read:   resourceGithubRepositoryCustomPropertyRead,
 		Delete: resourceGithubRepositoryCustomPropertyDelete,
 		Importer: &schema.ResourceImporter{
-			StateContext: schema.ImportStatePassthroughContext,
+			StateContext: resourceGithubRepositoryCustomPropertyImport,
 		},
 
 		Schema: map[string]*schema.Schema{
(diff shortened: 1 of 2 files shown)


m0un10:
Thanks for the fix @imaspen , we are impacted by this as well. Hopefully GH brings it into one of the upcoming releases.

deiga:
Hey @imaspen 

Thanks for submitting this PR!
Would you be available to rebase this?

PR #2643, opened 2025-04-29
fix(github_repository_file): Commit detail changes should not do empty commits

### LINKED ISSUE #689: github_repository_file keeps sending empty commits with the same commit message despite no changes to file content
(I had already opened an issue about this in the old repo, but after it moved here, seems to have been lost...)

### Terraform Version

Terraform v0.14.5

### Affected Resource(s)

- github_repository_file

```hcl
resource "github_repository_file" "file" {
  for_each = {
    for x in flatten([
      for name, repo in var.repos : [
        for file in fileset("templates", "${repo.template}/**") : {
          repo     = name
          file     = trimprefix(file, "${repo.template}/")
          template = repo.template
        }
      ] if can(repo.template) && lookup(repo, "archived", false) == false
    ]) : "${x.repo}:${x.file}" => x
  }

  repository          = each.value.repo
  branch              = github_branch_default.default[each.value.repo].branch
  file                = each.value.file
  content             = file("templates/${each.value.template}/${each.value.file}")
  commit_message      = "chore(template): update ${each.value.file}"
  overwrite_on_create = true

  depends_on = [
    github_repository.repository
  ]
}
```

### Expected Behavior

Should not be sending empty commits with the file content not having changed!

### Actual Behavior

when running `terraform plan` everything looks correct, and there is no indication that any modifications to the referenced files will be triggered, but when running `terraform apply` a whole bunch of empty commits are sent with the same commit message template

see this repo for a recent example: https://github.com/ahmadnassri/node-glob-promise/commits/master

all the commits starting with `chore(template)` are an example of this

I tried adding 

```hcl
  lifecycle {
    ignore_changes = [
      commit_message
    ]
  }
```

but that simply re-uses whatever the previous commit message was for the file and commits anyways!

atorrescogollo:
When terraform changes any field related to how commits are done (email, commit message, etc), it pushes an unexpected empty commit to the repo.

This is specially relevant when importing an existing file into terraform, where the commit message you put in terraform will likely be different from the commit that pushed the file originally. In that sense, importing a file will always mean eventually doing empty commits for each file you imported.

The expected behaviour is just updating the state and avoid doing those empty commits. Otherwise, each empty commit can run any CI that might be configured, notifications, etc.

Resolves #689
Recreated from #2638

----

### Before the change?

* When any of  `commit_sha`, `commit_author`, `commit_email` or `commit_message` changed, it did a empty commit

### After the change?

* Now, it checks which fields have changes and only commits if there are changes in any other field

### Pull request checklist

- [ ] Yes
- [x] No

----


atorrescogollo:
@nickfloyd Hey, this PR was outdated due to being open for several months, so I just updated it with the latest main and resolved all conflicts.

Could you review it when you get a chance? Thanks

deiga:
Please rebase if you still want to get this merged

atorrescogollo:
@deiga Done already. It's ready for review

github/resource_github_repository_file_test.go
```diff
@@ -376,6 +377,65 @@ func TestAccGithubRepositoryFile(t *testing.T) {
...
+
+		randomID := acctest.RandStringFromCharSet(5, acctest.CharSetAlphaNum)
+		repoName := fmt.Sprintf("%srepo-file-%s", testResourcePrefix, randomID)
+		config := fmt.Sprintf(`
+
+			resource "github_repository" "test" {
+				name                 = "%s"
+				auto_init            = true
+				vulnerability_alerts = true
+			}
```
deiga: Please use the mustCreate methods to create resources that are not under test

github/resource_github_repository_file_test.go
```diff
@@ -376,6 +377,65 @@ func TestAccGithubRepositoryFile(t *testing.T) {
...
+		messageChanged := strings.Replace(config, "Managed by Terraform", "Changed by Terraform", 1)
+		contentChanged := strings.Replace(messageChanged, `"bar"`, `"baz"`, 1)
+
+		sameCommit := statecheck.CompareValue(compare.ValuesSame())
+		newCommit := statecheck.CompareValue(compare.ValuesDiffer())
+
+		resource.Test(t, resource.TestCase{
+			PreCheck:          func() { skipUnauthenticated(t) },
+			ProviderFactories: providerFactories,
+			Steps: []resource.TestStep{
```
deiga, line 406: Please add ConfigPlanChecks as well to determine when a plan action should happen

deiga (changes requested):
As you mention that importing will always cause this, would it make sense to update the import function to set the upstream values?

And I wonder if it would make sense to change the commit fields to be ForceNew instead of adding some "ignore changes" logic. 

@stevehipwell what do you think?

examples/repository_file/providers.tf
```diff
@@ -0,0 +1,12 @@
+terraform {
+  required_providers {
+    github = {
+      source = "integrations/github"
+    }
```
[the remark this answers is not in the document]
atorrescogollo: I guess this is irrelevant since most of the times you'll be testing the current local version that won't fetch any public version

atorrescogollo:
@deiga

>As you mention that importing will always cause this, would it make sense to update the import function to set the upstream values?

These fields describe the next commit, so seeding author/email from the last one stores a value that doesn't mean what the field means, and you'd just get a stale diff on first plan. `commit_message` is different since it's Computed. So leaving the others null on import seems fair to me.

>And I wonder if it would make sense to change the commit fields to be ForceNew instead of adding some "ignore changes" logic.

`ForceNew` would make Terraform replace the resource whenever a commit field changes. Replacing this resource means deleting the file and adding it again, which is the opposite of what this PR aimed for. There's also a moment where the file doesn't exist on the branch. i would rather not doing that to be honest.

deiga:
> These fields describe the next commit

I'm not sure I follow. Those fields describe the commit which created or modified the file, which would mean that on import we should be Computing the values instead of Configuring them.

> ForceNew would make Terraform replace the resource whenever a commit field changes. Replacing this resource means deleting the file and adding it again, which is the opposite of what this PR aimed for. There's also a moment where the file doesn't exist on the branch. i would rather not doing that to be honest.

The thing is, I think the premise of your PR is wrong. If you change a configured field value, then you should be expecting Terraform do changes upstream. If we'd disable that functionality, then we'd be breaking a fundamental contract a user has with Terraform.

Usually changes like that should be done on the users side with `ignore_changes`.

Maybe you could elaborate the use-case you're trying to solve?

atorrescogollo:
@deiga I can't reproduce this issue anymore in the latest version of the provider. The code changed a lot since April 2025 when I opened the PR, so I think at some point it was fixed part of some other work. So I'm closing this

PR #2649, opened 2025-05-02
Adding github_enterprise_ip_allow_list_entry resource

### LINKED ISSUE #2648: [FEAT]: IP Allow Lists for enterprises and organizations
### Describe the need

Both github enterprises and organizations can set an IP Allow List. Entries for theses lists could be controlled via terraform like so:

```hcl 
resource "github_enterprise_ip_allow_list_entry" "test" {
  enterprise_slug = "my-enterprise"
  ip                   = "192.168.1.0/20"
  name             = "My IP Range Name"
  is_active       = true
}
```

### SDK Version

_No response_

### API Version

_No response_

### Relevant log output

```shell

```

### Code of Conduct

ErikElkins:
Resolves #2648

----

### Before the change?

* None

### After the change?

* Adding the github_enterprise_ip_allow_list_entry resource.

### Pull request checklist

- [ ] Yes
- [X] No

----

### DIFF
--- a/github/resource_github_enterprise_ip_allow_list_entry.go
+++ b/github/resource_github_enterprise_ip_allow_list_entry.go
@@ -1,0 +1,218 @@
+package github
+
+import (
+	"context"
+	"log"
+	"strings"
+
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+	"github.com/shurcooL/githubv4"
+)
+
+func resourceGithubEnterpriseIpAllowListEntry() *schema.Resource {
+	return &schema.Resource{
+		Create: resourceGithubEnterpriseIpAllowListEntryCreate,
+		Read:   resourceGithubEnterpriseIpAllowListEntryRead,
+		Update: resourceGithubEnterpriseIpAllowListEntryUpdate,
+		Delete: resourceGithubEnterpriseIpAllowListEntryDelete,
+		Importer: &schema.ResourceImporter{
+			StateContext: schema.ImportStatePassthroughContext,
+		},
+
+		Schema: map[string]*schema.Schema{
+			"enterprise_slug": {
+				Type:        schema.TypeString,
+				Required:    true,
+				Description: "The slug of the enterprise to apply the IP allow list entry to.",
+			},
+			"ip": {
+				Type:        schema.TypeString,
+				Required:    true,
+				Description: "An IP address or range of IP addresses in CIDR notation.",
+			},
+			"name": {
+				Type:        schema.TypeString,
+				Optional:    true,
+				Description: "An optional name for the IP allow list entry.",
+			},
+			"is_active": {
+				Type:        schema.TypeBool,
+				Optional:    true,
+				Default:     true,
+				Description: "Whether the entry is currently active.",
+			},
+		},
+	}
+}
+
+func resourceGithubEnterpriseIpAllowListEntryCreate(d *schema.ResourceData, meta interface{}) error {
+	client := meta.(*Owner).v4client
+	ctx := context.WithValue(context.Background(), ctxId, d.Id())
+
+	// First, get the enterprise ID as we need it for the mutation
+	enterpriseSlug := d.Get("enterprise_slug").(string)
+	enterpriseID, err := getEnterpriseID(ctx, client, enterpriseSlug)
+	if err != nil {
+		return err
+	}
+
+	// Then create the IP allow list entry
+	var mutation struct {
+		CreateIpAllowListEntry struct {
+			IpAllowListEntry struct {
+				ID             githubv4.String
+				AllowListValue githubv4.String
+				Name           githubv4.String
+				IsActive       githubv4.Boolean
+				CreatedAt      githubv4.String
+				UpdatedAt      githubv4.String
+			}
+		} `graphql:"createIpAllowListEntry(input: $input)"`
+	}
+
+	name := d.Get("name").(string)
+	input := githubv4.CreateIpAllowListEntryInput{
+		OwnerID:        githubv4.ID(enterpriseID),
+		AllowListValue: githubv4.String(d.Get("ip").(string)),
+		IsActive:       githubv4.Boolean(d.Get("is_active").(bool)),
+	}
+
+	if name != "" {
+		input.Name = githubv4.NewString(githubv4.String(name))
+	}
+
+	err = client.Mutate(ctx, &mutation, input, nil)
+	if err != nil {
+		return err
+	}
+
+	d.SetId(string(mutation.CreateIpAllowListEntry.IpAllowListEntry.ID))
+
+	return resourceGithubEnterpriseIpAllowListEntryRead(d, meta)
+}
+
+func resourceGithubEnterpriseIpAllowListEntryRead(d *schema.ResourceData, meta interface{}) error {
+	client := meta.(*Owner).v4client
+	ctx := context.WithValue(context.Background(), ctxId, d.Id())
+
+	var query struct {
+		Node struct {
+			IpAllowListEntry struct {
+				ID             githubv4.String
+				AllowListValue githubv4.String
+				Name           githubv4.String
+				IsActive       githubv4.Boolean
+				CreatedAt      githubv4.String
+				UpdatedAt      githubv4.String
+				Owner          struct {
+					Enterprise struct {
+						Slug githubv4.String
+					} `graphql:"... on Enterprise"`
+				}
+			} `graphql:"... on IpAllowListEntry"`
+		} `graphql:"node(id: $id)"`
+	}
+
+	variables := map[string]interface{}{
+		"id": githubv4.ID(d.Id()),
+	}
+
+	err := client.Query(ctx, &query, variables)
+	if err != nil {
+		if strings.Contains(err.Error(), "Could not resolve to a node with the global id") {
+			log.Printf("[INFO] Removing IP allow list entry (%s) from state because it no longer exists in GitHub", d.Id())
+			d.SetId("")
+			return nil
+		}
+		return err
+	}
+
+	entry := query.Node.IpAllowListEntry
+
+	d.Set("ip", entry.AllowListValue)
+	d.Set("name", entry.Name)
+	d.Set("is_active", entry.IsActive)
+	d.Set("created_at", entry.CreatedAt)
+	d.Set("updated_at", entry.UpdatedAt)
+	d.Set("enterprise_slug", entry.Owner.Enterprise.Slug)
+
+	return nil
+}
+
+func resourceGithubEnterpriseIpAllowListEntryUpdate(d *schema.ResourceData, meta interface{}) error {
+	client := meta.(*Owner).v4client
+	ctx := context.WithValue(context.Background(), ctxId, d.Id())
+
+	var mutation struct {
+		UpdateIpAllowListEntry struct {
+			IpAllowListEntry struct {
+				ID             githubv4.String
+				AllowListValue githubv4.String
+				Name           githubv4.String
+				IsActive       githubv4.Boolean
+				UpdatedAt      githubv4.String
+			}
+		} `graphql:"updateIpAllowListEntry(input: $input)"`
+	}
+
+	name := d.Get("name").(string)
+	input := githubv4.UpdateIpAllowListEntryInput{
+		IPAllowListEntryID: githubv4.ID(d.Id()),
+		AllowListValue:     githubv4.String(d.Get("ip").(string)),
+		IsActive:           githubv4.Boolean(d.Get("is_active").(bool)),
+	}
+
+	if name != "" {
+		input.Name = githubv4.NewString(githubv4.String(name))
+	}
+
+	err := client.Mutate(ctx, &mutation, input, nil)
+	if err != nil {
+		return err
+	}
+
+	return resourceGithubEnterpriseIpAllowListEntryRead(d, meta)
+}
+
+func resourceGithubEnterpriseIpAllowListEntryDelete(d *schema.ResourceData, meta interface{}) error {
+	client := meta.(*Owner).v4client
+	ctx := context.WithValue(context.Background(), ctxId, d.Id())
+
+	var mutation struct {
+		DeleteIpAllowListEntry struct {
+			ClientMutationID githubv4.String
+		} `graphql:"deleteIpAllowListEntry(input: $input)"`
+	}
+
+	input := githubv4.DeleteIpAllowListEntryInput{
+		IPAllowListEntryID: githubv4.ID(d.Id()),
+	}
+
+	err := client.Mutate(ctx, &mutation, input, nil)
+	if err != nil {
+		return err
+	}
+
+	d.SetId("")
+	return nil
+}
+
+// Helper function to get Enterprise ID from slug
+func getEnterpriseID(ctx context.Context, client *githubv4.Client, enterpriseSlug string) (string, error) {
+	var query struct {
+		Enterprise struct {
+			ID githubv4.ID
+		} `graphql:"enterprise(slug: $slug)"`
+	}
+
+	variables := map[string]interface{}{
+		"slug": githubv4.String(enterpriseSlug),
+	}
+
+	err := client.Query(ctx, &query, variables)
+	if err != nil {
+		return "", err
+	}
+
+	return query.Enterprise.ID.(string), nil
+}
--- a/github/resource_github_enterprise_ip_allow_list_entry_test.go
+++ b/github/resource_github_enterprise_ip_allow_list_entry_test.go
@@ -1,0 +1,101 @@
+package github
+
+import (
+	"fmt"
+	"testing"
+
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/resource"
+)
+
+func TestAccGithubEnterpriseIpAllowListEntry_basic(t *testing.T) {
+	t.Skip("Acceptance test requires a real GitHub Enterprise environment")
+
+	resourceName := "github_enterprise_ip_allow_list_entry.test"
+	enterpriseSlug := "test-enterprise"
+	ip := "192.168.1.0/24"
+	name := "Test Entry"
+	isActive := true
+
+	resource.Test(t, resource.TestCase{
+		PreCheck: func() {
+			testAccPreCheck(t)
+			testAccPreCheckEnterprise(t)
+		},
+		Providers: testAccProviders,
+		Steps: []resource.TestStep{
+			{
+				Config: testAccGithubEnterpriseIpAllowListEntryConfig(enterpriseSlug, ip, name, isActive),
+				Check: resource.ComposeTestCheckFunc(
+					resource.TestCheckResourceAttr(resourceName, "enterprise_slug", enterpriseSlug),
+					resource.TestCheckResourceAttr(resourceName, "ip", ip),
+					resource.TestCheckResourceAttr(resourceName, "name", name),
+					resource.TestCheckResourceAttr(resourceName, "is_active", fmt.Sprintf("%t", isActive)),
+				),
+			},
+			{
+				ResourceName:      resourceName,
+				ImportState:       true,
+				ImportStateVerify: true,
+			},
+		},
+	})
+}
+
+func TestAccGithubEnterpriseIpAllowListEntry_update(t *testing.T) {
+	t.Skip("Acceptance test requires a real GitHub Enterprise environment")
+
+	resourceName := "github_enterprise_ip_allow_list_entry.test"
+	enterpriseSlug := "test-enterprise"
+	ip := "192.168.1.0/24"
+	name := "Test Entry"
+	isActive := true
+
+	updatedIP := "10.0.0.0/16"
+	updatedName := "Updated Entry"
+	updatedIsActive := false
+
+	resource.Test(t, resource.TestCase{
+		PreCheck: func() {
+			testAccPreCheck(t)
+			testAccPreCheckEnterprise(t)
+		},
+		Providers: testAccProviders,
+		Steps: []resource.TestStep{
+			{
+				Config: testAccGithubEnterpriseIpAllowListEntryConfig(enterpriseSlug, ip, name, isActive),
+				Check: resource.ComposeTestCheckFunc(
+					resource.TestCheckResourceAttr(resourceName, "enterprise_slug", enterpriseSlug),
+					resource.TestCheckResourceAttr(resourceName, "ip", ip),
+					resource.TestCheckResourceAttr(resourceName, "name", name),
+					resource.TestCheckResourceAttr(resourceName, "is_active", fmt.Sprintf("%t", isActive)),
+				),
+			},
+			{
+				Config: testAccGithubEnterpriseIpAllowListEntryConfig(enterpriseSlug, updatedIP, updatedName, updatedIsActive),
+				Check: resource.ComposeTestCheckFunc(
+					resource.TestCheckResourceAttr(resourceName, "enterprise_slug", enterpriseSlug),
+					resource.TestCheckResourceAttr(resourceName, "ip", updatedIP),
+					resource.TestCheckResourceAttr(resourceName, "name", updatedName),
+					resource.TestCheckResourceAttr(resourceName, "is_active", fmt.Sprintf("%t", updatedIsActive)),
+				),
+			},
+		},
+	})
+}
+
+func testAccGithubEnterpriseIpAllowListEntryConfig(enterpriseSlug, ip, name string, isActive bool) string {
+	return fmt.Sprintf(`
+resource "github_enterprise_ip_allow_list_entry" "test" {
+  enterprise_slug = "%s"
+  ip              = "%s"
+  name            = "%s"
+  is_active       = %t
+}
+`, enterpriseSlug, ip, name, isActive)
+}
+
+func testAccPreCheckEnterprise(t *testing.T) {
+	if v := testAccProvider.Meta().(*Owner).name; v == "" {
+		t.Fatal("The GITHUB_ENTERPRISE_SLUG environment variable must be set for enterprise tests")
+	}
+}
--- a/github/provider.go
+++ b/github/provider.go
@@ -195,6 +195,7 @@ func Provider() *schema.Provider {
 			"github_user_ssh_key":                                                   resourceGithubUserSshKey(),
 			"github_enterprise_organization":                                        resourceGithubEnterpriseOrganization(),
 			"github_enterprise_actions_runner_group":                                resourceGithubActionsEnterpriseRunnerGroup(),
+			"github_enterprise_ip_allow_list_entry":                                 resourceGithubEnterpriseIpAllowListEntry(),
 		},
 
 		DataSourcesMap: map[string]*schema.Resource{
--- a/website/docs/r/enterprise_ip_allow_list_entry.html.markdown
+++ b/website/docs/r/enterprise_ip_allow_list_entry.html.markdown
@@ -1,0 +1,38 @@
+---
+layout: "github"
+page_title: "GitHub: github_enterprise_ip_allow_list_entry"
+description: |-
+  Creates and manages IP allow list entries within a GitHub Enterprise
+---
+
+# github_enterprise_ip_allow_list_entry
+
+This resource allows you to create and manage IP allow list entries for a GitHub Enterprise account. IP allow list entries define IP addresses or ranges that are permitted to access private resources in the enterprise.
+
+## Example Usage
+
+```hcl
+resource "github_enterprise_ip_allow_list_entry" "test" {
+  enterprise_slug = "my-enterprise"
+  ip              = "192.168.1.0/20"
+  name            = "My IP Range Name"
+  is_active       = true
+}
+```
+
+## Argument Reference
+
+The following arguments are supported:
+
+* `enterprise_slug` - (Required) The slug of the enterprise.
+* `ip`              - (Required) An IP address or range of IP addresses in CIDR notation.
+* `name`            - (Optional) A descriptive name for the IP allow list entry.
+* `is_active`       - (Optional) Whether the entry is currently active. Default: true.
+
+## Import
+
+This resource can be imported using the ID of the IP allow list entry:
+
+```bash
+$ terraform import github_enterprise_ip_allow_list_entry.test IALE_kwHOC1234567890a
+```


nickfloyd (dismissed):

nickfloyd:
@ErikElkins Apologies for the delay on getting to this PR.  Would you be willing to fix the lint issue? Thank you!

github/resource_github_enterprise_ip_allow_list_entry.go
```diff
@@ -0,0 +1,218 @@
...
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+	"github.com/shurcooL/githubv4"
+)
+
+func resourceGithubEnterpriseIpAllowListEntry() *schema.Resource {
+	return &schema.Resource{
+		Create: resourceGithubEnterpriseIpAllowListEntryCreate,
+		Read:   resourceGithubEnterpriseIpAllowListEntryRead,
+		Update: resourceGithubEnterpriseIpAllowListEntryUpdate,
+		Delete: resourceGithubEnterpriseIpAllowListEntryDelete,
```
stevehipwell:
```suggestion
		CreateContext: resourceGithubEnterpriseIpAllowListEntryCreate,
		ReadContext:   resourceGithubEnterpriseIpAllowListEntryRead,
		UpdateContext: resourceGithubEnterpriseIpAllowListEntryUpdate,
		DeleteContext: resourceGithubEnterpriseIpAllowListEntryDelete,
```

Please could you use the context aware functions?

github/resource_github_enterprise_ip_allow_list_entry.go
```diff
@@ -0,0 +1,218 @@
...
+				Default:     true,
+				Description: "Whether the entry is currently active.",
+			},
+		},
+	}
+}
+
+func resourceGithubEnterpriseIpAllowListEntryCreate(d *schema.ResourceData, meta interface{}) error {
+	client := meta.(*Owner).v4client
+	ctx := context.WithValue(context.Background(), ctxId, d.Id())
```
stevehipwell:
```suggestion
```
This doesn't do anything.

github/resource_github_enterprise_ip_allow_list_entry.go
```diff
@@ -0,0 +1,218 @@
...
+	if err != nil {
+		return err
+	}
+
+	d.SetId("")
+	return nil
+}
+
+// Helper function to get Enterprise ID from slug
+func getEnterpriseID(ctx context.Context, client *githubv4.Client, enterpriseSlug string) (string, error) {
```
stevehipwell: AFAIK enterprise slug is the ID, the GraphQL ID shouldn't pollute TF.

ErikElkins:
Yep! Let me jump in here.

ErikElkins:
Oh, i see the other PR. Should we just close this? EDIT: NEVERMIND

ErikElkins:
> Is there a REST API for this functionality?

Doesn't look like they've added it since I opened the PR.

github/resource_github_enterprise_ip_allow_list_entry.go
```diff
@@ -0,0 +1,217 @@
...
+				Type:        schema.TypeString,
+				Optional:    true,
+				Description: "An optional name for the IP allow list entry.",
+			},
+			"is_active": {
+				Type:        schema.TypeBool,
+				Optional:    true,
+				Default:     true,
+				Description: "Whether the entry is currently active.",
+			},
```
gateixeira, line 47:
```suggestion
			},
			"created_at": {
				Type:        schema.TypeString,
				Computed:    true,
				Description: "Timestamp of when the entry was created.",
			},
			"updated_at": {
				Type:        schema.TypeString,
				Computed:    true,
				Description: "Timestamp of when the entry was last updated.",
			},
```

github/resource_github_enterprise_ip_allow_list_entry_test.go
```diff
@@ -0,0 +1,101 @@
...
+resource "github_enterprise_ip_allow_list_entry" "test" {
+  enterprise_slug = "%s"
+  ip              = "%s"
+  name            = "%s"
+  is_active       = %t
+}
+`, enterpriseSlug, ip, name, isActive)
+}
+
+func testAccPreCheckEnterprise(t *testing.T) {
```
gateixeira: I believe you should replace this by a call to `skipUnlessEnterprise(t)`

github/resource_github_enterprise_ip_allow_list_entry_test.go
```diff
@@ -0,0 +1,101 @@
...
+	resourceName := "github_enterprise_ip_allow_list_entry.test"
+	enterpriseSlug := "test-enterprise"
+	ip := "192.168.1.0/24"
+	name := "Test Entry"
+	isActive := true
+
+	resource.Test(t, resource.TestCase{
+		PreCheck: func() {
+			testAccPreCheck(t)
+			testAccPreCheckEnterprise(t)
```
gateixeira:
```suggestion
            skipUnlessEnterprise(t)
```

github/resource_github_enterprise_ip_allow_list_entry_test.go
```diff
@@ -0,0 +1,101 @@
...
+	isActive := true
+
+	updatedIP := "10.0.0.0/16"
+	updatedName := "Updated Entry"
+	updatedIsActive := false
+
+	resource.Test(t, resource.TestCase{
+		PreCheck: func() {
+			testAccPreCheck(t)
+			testAccPreCheckEnterprise(t)
```
gateixeira:
```suggestion
            skipUnlessEnterprise(t)
```

gateixeira (changes requested):
hi @ErikElkins 👋 

I was updating your PR to have the lint fixed and realized that the acceptance tests seem to also not pass. Can you please update the PR with the following? Feel free to validate on your end.

github/resource_github_enterprise_ip_allow_list_entry_test.go
ErikElkins: Done!

github/resource_github_enterprise_ip_allow_list_entry.go
ErikElkins: If i understand this correctly, you'd still want the resource to reference the slug (and not the GraphQL ID) like so:

```tf
resource "github_enterprise_ip_allow_list_entry" "example" {
  enterprise_slug = "my-enterprise-slug"
  ip              = "1.2.3.4"
}
```

If you pass `enterprise_slug` to the `CreateIpAllowListEntry` mutation it will fail with `Error: Could not resolve to a node with the global id of 'my-enterprise-slug'` (i tested with a real one). It looks to need the GraphQL ID, and not the slug. If we keep the lookup like this, we should be able to keep the GraphQL ID out of the terraform and have the lookup take place behind the scenes.

deiga: https://docs.github.com/en/enterprise-cloud@latest/graphql/reference/input-objects#createipallowlistentryinput

I think it makes sense that it's not the `enterprise_slug` as AFAIK that Mutation also works for Organizations, which could have the same slug as the enterprise

github/resource_github_enterprise_ip_allow_list_entry.go
```diff
@@ -0,0 +1,225 @@
...
+	return &schema.Resource{
+		CreateContext: resourceGithubEnterpriseIpAllowListEntryCreate,
+		ReadContext:   resourceGithubEnterpriseIpAllowListEntryRead,
+		UpdateContext: resourceGithubEnterpriseIpAllowListEntryUpdate,
+		DeleteContext: resourceGithubEnterpriseIpAllowListEntryDelete,
+		Importer: &schema.ResourceImporter{
+			StateContext: schema.ImportStatePassthroughContext,
+		},
+
+		Schema: map[string]*schema.Schema{
```
deiga, line 24: Please add a top-level `Description` field

github/resource_github_enterprise_ip_allow_list_entry.go
```diff
@@ -0,0 +1,225 @@
...
+	}
+
+	err = client.Mutate(ctx, &mutation, input, nil)
+	if err != nil {
+		return diag.FromErr(err)
+	}
+
+	d.SetId(string(mutation.CreateIpAllowListEntry.IpAllowListEntry.ID))
+
+	return resourceGithubEnterpriseIpAllowListEntryRead(ctx, d, meta)
```
deiga: Please remove the call to `Read` in `Create`. Instead set any computed values based on the response

github/resource_github_enterprise_ip_allow_list_entry.go
```diff
@@ -0,0 +1,225 @@
...
+	if name != "" {
+		input.Name = githubv4.NewString(githubv4.String(name))
+	}
+
+	err := client.Mutate(ctx, &mutation, input, nil)
+	if err != nil {
+		return diag.FromErr(err)
+	}
+
+	return resourceGithubEnterpriseIpAllowListEntryRead(ctx, d, meta)
```
deiga: Same for `Update` https://github.com/integrations/terraform-provider-github/pull/2649/changes#r2688260511

github/resource_github_enterprise_ip_allow_list_entry.go
```diff
@@ -0,0 +1,225 @@
...
+	}
+
+	variables := map[string]interface{}{
+		"id": githubv4.ID(d.Id()),
+	}
+
+	err := client.Query(ctx, &query, variables)
+	if err != nil {
+		if strings.Contains(err.Error(), "Could not resolve to a node with the global id") {
+			log.Printf("[INFO] Removing IP allow list entry (%s) from state because it no longer exists in GitHub", d.Id())
```
deiga: Please use `tflog.Info` instead of `log.Printf`

github/resource_github_enterprise_ip_allow_list_entry.go
```diff
@@ -0,0 +1,225 @@
...
+	if err != nil {
+		return diag.FromErr(err)
+	}
+
+	d.SetId("")
+	return nil
+}
+
+// Helper function to get Enterprise ID from slug
+func getEnterpriseID(ctx context.Context, client *githubv4.Client, enterpriseSlug string) (string, error) {
```
deiga: This function seems to exist already in resource_github_enterprise_organization.go, could you re-use that?

Maybe it could be moved to `util_v4.go`?

github/resource_github_enterprise_ip_allow_list_entry_test.go
```diff
@@ -0,0 +1,93 @@
...
+
+import (
+	"fmt"
+	"testing"
+
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/resource"
+)
+
+func TestAccGithubEnterpriseIpAllowListEntry_basic(t *testing.T) {
+	t.Skip("Acceptance test requires a real GitHub Enterprise environment")
```
deiga:
```suggestion
```

github/resource_github_enterprise_ip_allow_list_entry_test.go
```diff
@@ -0,0 +1,93 @@
...
+	enterpriseSlug := "test-enterprise"
+	ip := "192.168.1.0/24"
+	name := "Test Entry"
+	isActive := true
+
+	resource.Test(t, resource.TestCase{
+		PreCheck: func() {
+			skipUnlessEnterprise(t)
+		},
+		Providers: testAccProviders,
```
deiga:
```suggestion
		ProviderFactories: providerFactories,
```

github/resource_github_enterprise_ip_allow_list_entry_test.go
```diff
@@ -0,0 +1,93 @@
...
+		Steps: []resource.TestStep{
+			{
+				Config: testAccGithubEnterpriseIpAllowListEntryConfig(enterpriseSlug, ip, name, isActive),
+				Check: resource.ComposeTestCheckFunc(
+					resource.TestCheckResourceAttr(resourceName, "enterprise_slug", enterpriseSlug),
+					resource.TestCheckResourceAttr(resourceName, "ip", ip),
+					resource.TestCheckResourceAttr(resourceName, "name", name),
+					resource.TestCheckResourceAttr(resourceName, "is_active", fmt.Sprintf("%t", isActive)),
```
deiga:
```suggestion
					resource.TestCheckResourceAttr(resourceName, "is_active", strconv.FormatBool(isActive)),
```

github/resource_github_enterprise_ip_allow_list_entry_test.go
```diff
@@ -0,0 +1,93 @@
...
+				ResourceName:      resourceName,
+				ImportState:       true,
+				ImportStateVerify: true,
+			},
+		},
+	})
+}
+
+func TestAccGithubEnterpriseIpAllowListEntry_update(t *testing.T) {
+	t.Skip("Acceptance test requires a real GitHub Enterprise environment")
```
deiga:
```suggestion
```

github/resource_github_enterprise_ip_allow_list_entry_test.go
```diff
@@ -0,0 +1,93 @@
...
+
+	updatedIP := "10.0.0.0/16"
+	updatedName := "Updated Entry"
+	updatedIsActive := false
+
+	resource.Test(t, resource.TestCase{
+		PreCheck: func() {
+			skipUnlessEnterprise(t)
+		},
+		Providers: testAccProviders,
```
deiga:
```suggestion
		ProviderFactories: providerFactories,
```

github/resource_github_enterprise_ip_allow_list_entry.go
stevehipwell: The GraphQL ID may be required, but it shouldn't be the resource ID as that should be the slug. I'm not fully familiar with the GitHub GraphQL implementation but isn't the whole point of GraphQL that you can do everything with a single query; in which case why can't you use the slug to lookup the GraphQL ID in the same query?

deiga: Yeah, you can combine multiple "requests" in a single GQL query, but directly dependant things aren't possible AFAIK. 
But there one can then just use sequential queries. And one can't mix mutations and queries in a single request

stevehipwell: My original point stands, the "ID" here should be the slug. The GraphQL ID is an implementation detail that shouldn't leak into the provider. IMHO we'd be looking to swap the implementation out for a REST API at the first possible opportunity.

deiga: @stevehipwell I can't spot a place where the `enterpriseID` reaches state, what am I missing? Or is this thread resolved already?

stevehipwell: It's being put into `d.SetId()`.

deiga: Yes, the ID of the IP allowlist entry: `d.SetId(string(mutation.CreateIpAllowListEntry.IpAllowListEntry.ID))`
AFAIK your original comment was about the enterprise ID, but that is not being set into state AFAIK.


I can't think of any other reasonable thing to put as the ID of this resource which would make it unique enough.

Here are all the fields the entry has: https://docs.github.com/en/graphql/reference/objects#ipallowlistentry

stevehipwell: The enterprise slug should be the resource ID.

deiga: @stevehipwell an enterprise can have multiple IpAllowListEntries, if we'd use enterprise slug users could only ever configure one entry

stevehipwell: Are the names immutable? I also see there isn't a databaseid field.

deiga: Nope, name isn't immutable. IP is, but it's not a unique field

stevehipwell: I can't see a pending REST implementation. We may have to leak a GraphQL ID into the state after all. I'd suggest using `<graphql-id>:<name>` for the ID so if we get a REST implementation we can remove the GraphQL dependency.

deiga: I don't recommend using `name` as it's actually a description field and can contain long text. 

And since `name` is mutable, it would change the ID even if the actual resource didn't change.

Existing IP entry in UI: image
Add entry in UI: 
image
Edit entry in UI:
image

deiga:
@ErikElkins Are you able to continue finalizing this PR? :)

github/resource_github_enterprise_ip_allow_list_entry.go
stevehipwell: Are we sure IP isn't unique in the context of an enterprise?

deiga: Yes, we have had multiple duplicate IPs. We could treat it as unique even if the API allows duplicate IPS. One reason for allowing duplicate IPs would be ease of management. 

For example the IPs in api.github.com/meta have duplicates between products. If I'd be syncing these blindly to the allowlist with the product name (which we are doing) then it needs to enable duplicate IPS 😬

stevehipwell: Not that it's relevant to this PR, but wouldn't you run them through `distinct` by default?

ErikElkins:
> @ErikElkins Are you able to continue finalizing this PR? :)

Work picked up a bit, I might be able to return to it next week. If that's too long, feel free to fork or continue it otherwise.

deiga:
> > @ErikElkins Are you able to continue finalizing this PR? :)
> 
> Work picked up a bit, I might be able to return to it next week. If that's too long, feel free to fork or continue it otherwise.

No worries, we have a lot of other changes to finalize as well :D

ErikElkins:
> > > @ErikElkins Are you able to continue finalizing this PR? :)
> > 
> > 
> > Work picked up a bit, I might be able to return to it next week. If that's too long, feel free to fork or continue it otherwise.
> 
> No worries, we have a lot of other changes to finalize as well :D

Alright, I think that's everything from the feedback? Sorry if I missed anything, there was a lot.

github/resource_github_enterprise_ip_allow_list_entry.go
```diff
@@ -98,7 +99,7 @@ func resourceGithubEnterpriseIpAllowListEntryCreate(ctx context.Context, d *sche
 
 	d.SetId(string(mutation.CreateIpAllowListEntry.IpAllowListEntry.ID))
 
-	return resourceGithubEnterpriseIpAllowListEntryRead(ctx, d, meta)
+	return nil
```
deiga, line 112: You still need to set `updated_at` and `created_at` based on the `mutation.CreateIpAllowListEntry.IpAllowListEntry` fields

github/resource_github_enterprise_ip_allow_list_entry.go
```diff
@@ -179,7 +182,7 @@ func resourceGithubEnterpriseIpAllowListEntryUpdate(ctx context.Context, d *sche
 		return diag.FromErr(err)
 	}
 
-	return resourceGithubEnterpriseIpAllowListEntryRead(ctx, d, meta)
+	return nil
```
deiga: Here you need to set `updated_at` too

github/resource_github_enterprise_ip_allow_list_entry_test.go
```diff
@@ -2,14 +2,13 @@ package github
...
 	"testing"
 
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/resource"
 )
 
 func TestAccGithubEnterpriseIpAllowListEntry_basic(t *testing.T) {
-	t.Skip("Acceptance test requires a real GitHub Enterprise environment")
-
 	resourceName := "github_enterprise_ip_allow_list_entry.test"
 	enterpriseSlug := "test-enterprise"
```
deiga: You should be able to replace this with `testAccConf.enterpriseSlug`

github/resource_github_enterprise_ip_allow_list_entry_test.go
```diff
@@ -41,8 +40,6 @@ func TestAccGithubEnterpriseIpAllowListEntry_basic(t *testing.T) {
 }
 
 func TestAccGithubEnterpriseIpAllowListEntry_update(t *testing.T) {
-	t.Skip("Acceptance test requires a real GitHub Enterprise environment")
-
 	resourceName := "github_enterprise_ip_allow_list_entry.test"
 	enterpriseSlug := "test-enterprise"
```
deiga: You should be able to replace this with `testAccConf.enterpriseSlug`

ErikElkins: Done!

github/resource_github_enterprise_ip_allow_list_entry_test.go
ErikElkins: Done!

github/resource_github_enterprise_ip_allow_list_entry.go
ErikElkins: Done!

github/resource_github_enterprise_ip_allow_list_entry.go
ErikElkins: Done!

github/resource_github_enterprise_ip_allow_list_entry.go
```diff
@@ -0,0 +1,219 @@
...
+	}
+
+	variables := map[string]interface{}{
+		"id": githubv4.ID(d.Id()),
+	}
+
+	err := client.Query(ctx, &query, variables)
+	if err != nil {
+		if strings.Contains(err.Error(), "Could not resolve to a node with the global id") {
+			tflog.Info(ctx, "[INFO] Removing IP allow list entry (%s) from state because it no longer exists in GitHub", map[string]any{
```
deiga:
```suggestion
			tflog.Info(ctx, "Removing IP allow list entry from state because it no longer exists in GitHub", map[string]any{
```

github/resource_github_enterprise_ip_allow_list_entry.go
```diff
@@ -0,0 +1,219 @@
...
+	input := githubv4.DeleteIpAllowListEntryInput{
+		IPAllowListEntryID: githubv4.ID(d.Id()),
+	}
+
+	err := client.Mutate(ctx, &mutation, input, nil)
+	if err != nil {
+		return diag.FromErr(err)
+	}
+
+	d.SetId("")
```
deiga: This should be superfluous
```suggestion
```

deiga (changes requested):

deiga:
@ErikElkins We'll need you to rebase again 😬

ErikElkins:
> @ErikElkins We'll need you to rebase again 😬

Done!

github/resource_github_enterprise_ip_allow_list_entry.go
```diff
@@ -0,0 +1,218 @@
...
+		UpdateContext: resourceGithubEnterpriseIpAllowListEntryUpdate,
+		DeleteContext: resourceGithubEnterpriseIpAllowListEntryDelete,
+		Importer: &schema.ResourceImporter{
+			StateContext: schema.ImportStatePassthroughContext,
+		},
+
+		Schema: map[string]*schema.Schema{
+			"enterprise_slug": {
+				Type:        schema.TypeString,
+				Required:    true,
```
stevehipwell, line 27: This should have `ForceNew: true`.

github/resource_github_enterprise_ip_allow_list_entry.go
```diff
@@ -0,0 +1,218 @@
...
+				Description: "The slug of the enterprise to apply the IP allow list entry to.",
+			},
+			"ip": {
+				Type:        schema.TypeString,
+				Required:    true,
```
stevehipwell, line 33: This should probably have `ForceNew: true`.

github/resource_github_enterprise_ip_allow_list_entry.go
```diff
@@ -0,0 +1,218 @@
...
+	}
+
+	entry := query.Node.IpAllowListEntry
+
+	d.Set("ip", entry.AllowListValue)
+	d.Set("name", entry.Name)
+	d.Set("is_active", entry.IsActive)
+	d.Set("created_at", entry.CreatedAt)
+	d.Set("updated_at", entry.UpdatedAt)
+	d.Set("enterprise_slug", entry.Owner.Enterprise.Slug)
```
stevehipwell: This shouldn't be being updated.

github/resource_github_enterprise_ip_allow_list_entry.go
```diff
@@ -0,0 +1,218 @@
...
+			})
+			d.SetId("")
+			return nil
+		}
+		return diag.FromErr(err)
+	}
+
+	entry := query.Node.IpAllowListEntry
+
+	d.Set("ip", entry.AllowListValue)
```
stevehipwell: This probably shouldn't be being updated (see above).

github/resource_github_enterprise_ip_allow_list_entry.go
```diff
@@ -0,0 +1,218 @@
...
+	d.Set("name", entry.Name)
```
stevehipwell:
```suggestion
	if err := d.Set("name", entry.Name); err != nil {
		return diag.FromErr(err)
	}
```
Please use this pattern.

github/resource_github_enterprise_ip_allow_list_entry.go
```diff
@@ -0,0 +1,218 @@
...
+
+func resourceGithubEnterpriseIpAllowListEntry() *schema.Resource {
+	return &schema.Resource{
+		Description:   "Manage a GitHub Enterprise IP Allow List Entry.",
+		CreateContext: resourceGithubEnterpriseIpAllowListEntryCreate,
+		ReadContext:   resourceGithubEnterpriseIpAllowListEntryRead,
+		UpdateContext: resourceGithubEnterpriseIpAllowListEntryUpdate,
+		DeleteContext: resourceGithubEnterpriseIpAllowListEntryDelete,
+		Importer: &schema.ResourceImporter{
+			StateContext: schema.ImportStatePassthroughContext,
```
stevehipwell: Please use a custom import method.

github/resource_github_enterprise_ip_allow_list_entry_test.go
```diff
@@ -0,0 +1,88 @@
...
+
+import (
+	"fmt"
+	"strconv"
+	"testing"
+
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/resource"
+)
+
+func TestAccGithubEnterpriseIpAllowListEntry_basic(t *testing.T) {
```
stevehipwell:
```suggestion
func TestAccGithubEnterpriseIpAllowListEntry(t *testing.T) {
	t.Run("basic", func(t *testing.T) {
```

Please use the nested test case pattern.

github/resource_github_enterprise_ip_allow_list_entry_test.go
```diff
@@ -0,0 +1,88 @@
...
+					resource.TestCheckResourceAttr(resourceName, "ip", updatedIP),
+					resource.TestCheckResourceAttr(resourceName, "name", updatedName),
+					resource.TestCheckResourceAttr(resourceName, "is_active", fmt.Sprintf("%t", updatedIsActive)),
+				),
+			},
+		},
+	})
+}
+
+func testAccGithubEnterpriseIpAllowListEntryConfig(enterpriseSlug, ip, name string, isActive bool) string {
```
stevehipwell: Please keep the configuration in the test cases for readability.

github/resource_github_enterprise_ip_allow_list_entry_test.go
```diff
@@ -0,0 +1,89 @@
...
+			{
+				ResourceName:      resourceName,
+				ImportState:       true,
+				ImportStateVerify: true,
+			},
+		},
+	})
+}
+
+func TestAccGithubEnterpriseIpAllowListEntry_update(t *testing.T) {
```
stevehipwell: @ErikElkins this needs flattening under `TestAccGithubEnterpriseIpAllowListEntry` as `t.Run("update", func(t *testing.T) {`.

ErikElkins: Done!

github/resource_github_enterprise_ip_allow_list_entry.go
```diff
@@ -0,0 +1,257 @@
...
+
+	err := client.Mutate(ctx, &mutation, input, nil)
+	if err != nil {
+		return diag.FromErr(err)
+	}
+
+	return nil
+}
+
+func resourceGithubEnterpriseIpAllowListEntryImport(ctx context.Context, d *schema.ResourceData, meta any) ([]*schema.ResourceData, error) {
```
deiga, line 233: Import only needs to set the values that are needed for Read, which will be called by terraform directly after Import. 

I think the query is unnecessary in Import if you change the import ID to enterprise_slug:ipEntryID

github/resource_github_enterprise_ip_allow_list_entry.go
```diff
@@ -0,0 +1,257 @@
...
+		return diag.FromErr(err)
+	}
+
+	entry := query.Node.IpAllowListEntry
+	if err := d.Set("name", entry.Name); err != nil {
+		return diag.FromErr(err)
+	}
+	d.Set("is_active", entry.IsActive)
+	d.Set("created_at", entry.CreatedAt)
+	d.Set("updated_at", entry.UpdatedAt)
```
deiga: Add error handling to these

github/resource_github_enterprise_ip_allow_list_entry.go
```diff
@@ -0,0 +1,257 @@
...
+				Required:    true,
+				ForceNew:    true,
+				Description: "The slug of the enterprise to apply the IP allow list entry to.",
+			},
+			"ip": {
+				Type:        schema.TypeString,
+				Required:    true,
+				ForceNew:    true,
+				Description: "An IP address or range of IP addresses in CIDR notation.",
+			},
```
deiga, line 36: @stevehipwell While I agree that from a TF perspective it could make sense for this to be ForceNew, the API doesn't require it as the IP is editable on the entry. 
To make import simpler, I'd allow this field to be settable in Read

stevehipwell: I think it's simpler for this to be `ForceNew: true`, allowing it to be modified adds complexity.

github/resource_github_enterprise_ip_allow_list_entry.go
ErikElkins: Import happens much more quickly now, nice catch! Docs updated as well.

github/resource_github_enterprise_ip_allow_list_entry.go
```diff
@@ -0,0 +1,243 @@
...
+	err := client.Mutate(ctx, &mutation, input, nil)
+	if err != nil {
+		return diag.FromErr(err)
+	}
+
+	return nil
+}
+
+func resourceGithubEnterpriseIpAllowListEntryImport(ctx context.Context, d *schema.ResourceData, meta any) ([]*schema.ResourceData, error) {
+	// <enterprise_slug>:<ip_allow_list_entry_id>
```
stevehipwell: Based on the other functions it looks like you only need the `ip_allow_list_entry_id` to lookup the object? You need to do the lookup here as you need to set the `ip` field, so you should set all of them here.

github/resource_github_enterprise_ip_allow_list_entry_test.go
stevehipwell: I still see multiple top level test functions instead of a single `TestAccGithubEnterpriseIpAllowListEntry` function with multiple calls to `t.Run()`.

github/resource_github_enterprise_ip_allow_list_entry.go
ErikElkins: Alright now it does a full lookup in the import and sets all the fields. Let me know if that looks right.

stevehipwell:
@ErikElkins could you please rebase?

stevehipwell (approved):
LGTM

stevehipwell:
@deiga are you happy with this now? If so could you please approve (or clear your requested changes)?

github/resource_github_enterprise_ip_allow_list_entry.go
```diff
@@ -0,0 +1,282 @@
...
+	}
+	if err := d.Set("is_active", entry.IsActive); err != nil {
+		return nil, err
+	}
+	if err := d.Set("created_at", entry.CreatedAt); err != nil {
+		return nil, err
+	}
+	if err := d.Set("updated_at", entry.UpdatedAt); err != nil {
+		return nil, err
+	}
```
deiga, line 282: issue: Am I seeing things or did you re-add the API query to the `Import` func?

The ID should be enough as the `Read` func is the next thing to be called with futher Terraform operations and it does the exact same thing as `Import`. It doesn't seem useful to have an API interaction here

github/resource_github_enterprise_ip_allow_list_entry.go
```diff
@@ -0,0 +1,282 @@
...
+		} `graphql:"deleteIpAllowListEntry(input: $input)"`
+	}
+
+	input := githubv4.DeleteIpAllowListEntryInput{
+		IPAllowListEntryID: githubv4.ID(d.Id()),
+	}
+
+	err := client.Mutate(ctx, &mutation, input, nil)
+	if err != nil {
+		return diag.FromErr(err)
```
deiga, line 227: Please add error handling that a `404` doesn't unnecessarily create a Terraform error

github/resource_github_enterprise_ip_allow_list_entry.go
stevehipwell: I asked for this, the query is required otherwise it'll be recreated du to the IP being unset and the import is a complete waste of time.

ErikElkins: It was the recommended approach by @stevehipwell here: https://github.com/integrations/terraform-provider-github/pull/2649#discussion_r2840770420

deiga: @stevehipwell Oh, right. 
What if we add the IP to the import ID to circumvent the need for a API call? It's already QUITE difficult to run Import as finding the GQL ID of the allowlist entry requires one to query the API manually first 😬 

Actually, I wonder if there is a way to make `Import` more user friendly? 🤔

ErikElkins: I was hoping to do the import based on `ip`, but IP is not unique in the IP allow list. You could have the same IP ~~and description~~ multiple times in the list for one enterprise. :( 

EDIT: It's unique based on IP AND description

github/resource_github_enterprise_ip_allow_list_entry.go
ErikElkins: Added error handling and a comment. The GraphQL API (and github go client) means it'll look kinda janky. Let me know if you'd rather have me remove it.

github/resource_github_enterprise_ip_allow_list_entry.go
ErikElkins: Maybe to put a bow on this, i think we have two options for importing:
1. Use global ID like it's doing now `terraform import github_enterprise_ip_allow_list_entry.test IALE_kwHOC1234567890a`
2. Use both the IP and description to import (you can't import on IP alone as it isn't unique) `terraform import github_enterprise_ip_allow_list_entry.test 192.168.1.1:"Local Network Stack for HQ"`

Seeing how descriptions could make that especially gnarly, I think option 1 would be the way to go.

stevehipwell: Option one seems like the correct pattern.

deiga:
@ErikElkins Would you be available to rebase and test this again? :)

ErikElkins:
> @ErikElkins Would you be available to rebase and test this again? :)

Merged and tested, looks good!

stevehipwell (dismissed):
LGTM

stevehipwell:
@robert-crandall could you please review/approve this?

robert-crandall (dismissed):
This PR looks good. It looks like it's failing CI due to lint errors. Once those are fixed, this is good to go.

ErikElkins:
> This PR looks good. It looks like it's failing CI due to lint errors. Once those are fixed, this is good to go.

Done!

deiga:
@stevehipwell @robert-crandall please review again ☺️

LornaJoyce:
Hello, Could this get reviewed please?

deiga:
@ErikElkins could you please rebase, so that @stevehipwell can do final review?

ErikElkins:
> @ErikElkins could you please rebase, so that @stevehipwell can do final review?

Done!

ErikElkins:
Thanks everyone!

PR #2660, opened 2025-05-15
Add branch pattern example to repository_ruleset documentation

chemmi:
The pattern can be confusing since Github API and UI differ in this point. This is just a clarification for the documentation.

See also: https://github.com/orgs/community/discussions/119797
See also: https://docs.github.com/en/rest/repos/rules#create-a-repository-ruleset

Resolves #ISSUE_NUMBER

----

### Before the change?

* Intuitively, the user would use the simple branch pattern `main` oder `feat-*` to include or exclude branches in the ruleset. The provider will crash with a 402 when applying the change:

```
Error: POST https://api.github.com/repos/ORG/REPO/rulesets: 422 Validation Failed [{Resource: Field: Code: Message:Invalid target patterns: 'main'}]
```

### After the change?

* The documentation suggests to use the required pattern, e.g. `ref/heads/main`, in this case.

See also the example in https://docs.github.com/en/rest/repos/rules#create-a-repository-ruleset which uses this pattern.

### Pull request checklist

- [ ] Yes
- [x] No

----

### DIFF
--- a/website/docs/r/repository_ruleset.html.markdown
+++ b/website/docs/r/repository_ruleset.html.markdown
@@ -27,7 +27,7 @@ resource "github_repository_ruleset" "example" {
 
   conditions {
     ref_name {
-      include = ["~ALL"]
+      include = ["~ALL", "ref/heads/main"]
       exclude = []
     }
   }


deiga:
Hey @chemmi 👋 

I like your thinking here. The format is definitely confusing and we should help our users to not stumble on this.

I wonder if you'd be up for doing a few more modifications before this is good to merge?

- Could you change the example from. `refs/heads/main` to maybe `refs/heads/foo` to make it distinct from `~DEFAULT_BRANCH`?
- I think the documentation could be improved here by adding this information into the field description as well. Could you add note to both `description` fields [here](https://github.com/chemmi/terraform-provider-github/blob/main/github/resource_github_repository_ruleset.go#L103-L111) about the expected format?
- (optional) If you're feeling adventurous: Would you want to try to add validation to those fields? 😱

PR #2675, opened 2025-05-30
Adding a computed attribute for the repository owner to the github_repository

### LINKED ISSUE #2503: [FEAT]: Add `owner` attribute to `github_repository` data source and resource

AlanCitrix:
Resolves #2503

----

### Before the change?

* The repository owner has to either be parsed out of the `full_name`, or passed around as a separate variable.

### After the change?

* Repository owner is now a computed field in both the repository resource and data source.
* The data source can be queried by `name` + `owner` when using anonymous auth. `full_name` will also still work.

### Pull request checklist

- [ ] Yes
- [x] No

----

### DIFF
--- a/github/resource_github_repository.go
+++ b/github/resource_github_repository.go
@@ -412,6 +412,11 @@ func resourceGithubRepository() *schema.Resource {
 				Computed:    true,
 				Description: "A string of the form 'orgname/reponame'.",
 			},
+			"owner": {
+				Type:        schema.TypeString,
+				Computed:    true,
+				Description: "The owner of the repository.",
+			},
 			"html_url": {
 				Type:        schema.TypeString,
 				Computed:    true,
@@ -832,6 +837,9 @@ func resourceGithubRepositoryRead(ctx context.Context, d *schema.ResourceData, m
 	_ = d.Set("topics", flattenStringList(repo.Topics))
 	_ = d.Set("node_id", repo.GetNodeID())
 	_ = d.Set("repo_id", repo.GetID())
+	if repo.Owner != nil {
+		_ = d.Set("owner", repo.Owner.GetLogin())
+	}
 
 	// TODO: Validate this behavior as I can see these fields being returned even when archived
 	// GitHub API doesn't respond following parameters when repository is archived
--- a/website/docs/d/repository.html.markdown
+++ b/website/docs/d/repository.html.markdown
@@ -25,6 +25,8 @@ The following arguments are supported:
 
 * `full_name` - (Optional) Full name of the repository (in `org/name` format).
 
+* `owner` - (Optional) Owner of the repository. If not provided, the provider's default owner is used.
+
 ## Attributes Reference
 
 * `node_id` - the Node ID of the repository.
--- a/github/data_source_github_repository.go
+++ b/github/data_source_github_repository.go
@@ -22,13 +22,20 @@ func dataSourceGithubRepository() *schema.Resource {
 				Type:          schema.TypeString,
 				Optional:      true,
 				Computed:      true,
-				ConflictsWith: []string{"name"},
+				ConflictsWith: []string{"name", "owner"},
 			},
 			"name": {
 				Type:          schema.TypeString,
 				Optional:      true,
 				Computed:      true,
 				ConflictsWith: []string{"full_name"},
+			},
+			"owner": {
+				Type:          schema.TypeString,
+				Optional:      true,
+				Computed:      true,
+				ConflictsWith: []string{"full_name"},
+				Description:   "Owner of the repository. If not provided, the provider's default owner is used.",
 			},
 			"description": {
 				Type:     schema.TypeString,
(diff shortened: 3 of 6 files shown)


github/resource_github_repository.go
```diff
@@ -688,6 +693,7 @@ func resourceGithubRepositoryRead(d *schema.ResourceData, meta interface{}) erro
 	d.Set("topics", flattenStringList(repo.Topics))
 	d.Set("node_id", repo.GetNodeID())
 	d.Set("repo_id", repo.GetID())
+	d.Set("owner", repo.GetOwner().GetLogin())
```
nickfloyd: I belive this will panic if `nil` Recommend using go-github's safe accessor
```suggestion
if repo.Owner != nil {
    d.Set("owner", repo.Owner.GetLogin())
}
```

Alternatively we could defensively: 
```
if repo != nil && repo.Owner != nil {
    d.Set("owner", repo.Owner.GetLogin())
} else {
    d.Set("owner", "")
}
```

nickfloyd (changes requested):
Just a comment about making things a little more safe.

github/resource_github_repository.go
```diff
@@ -761,6 +766,7 @@ func resourceGithubRepositoryRead(d *schema.ResourceData, meta any) error {
 	_ = d.Set("topics", flattenStringList(repo.Topics))
 	_ = d.Set("node_id", repo.GetNodeID())
 	_ = d.Set("repo_id", repo.GetID())
+  _ = d.Set("owner", repo.GetOwner().GetLogin())
```
diofeher: Thanks for the contribution @AlanCitrix!

Minor fix: please run `go fmt .` to format your files.

website/docs/d/repository.html.markdown
```diff
@@ -25,6 +25,8 @@ The following arguments are supported:
 
 * `full_name` - (Optional) Full name of the repository (in `org/name` format).
 
+* `owner` - (Optional) Owner of the repository. If not provided, the owner specified in the provider configuration will be used.
```
diofeher:
```suggestion
* `owner` - (Optional)  Owner of the repository. If not provided, the provider's default owner is used.
```

This is the phrase used in other resources, since no owner is specified in the provider's configuration.

nickfloyd:
@AlanCitrix Please let us know if you still would like to see this change land.  Given it's been a bit we will move it out of the current target v6 and into a later one v7.  If you get the chance and are able to run `make lint` and fix the issues please let us know.

AlanCitrix:
> @AlanCitrix Please let us know if you still would like to see this change land. Given it's been a bit we will move it out of the current target v6 and into a later one v7. If you get the chance and are able to run `make lint` and fix the issues please let us know.

Sorry for the delay got it fixed now. That all sounds good thanks!

deiga:
Thanks for fixing these!

Since this has been targeted to the v7 release, would you be willing to try rebasing on the github-go-v68 branch?

Though it's not final and you'll possibly need to rebase once we get it merged to main

github/resource_github_repository.go
AlanCitrix: Addressed!

website/docs/d/repository.html.markdown
AlanCitrix: Addressed

AlanCitrix:
Ok had a little fight with git but I think I've got it correct now. @deiga is this what you had in mind to target the `github-go-v68` branch?

deiga:
@AlanCitrix Yes, exactly! Thank you for doing that :)

Now let's hope we can get v7 released ASAP

deiga:
@AlanCitrix 
We've now been able to upgrade the go-github version without needing a major version bump! Please rebase onto main now :)

AlanCitrix:
Done @deiga , please approve the github action

PR #2695, opened 2025-07-04
Feat/app token refresh

### LINKED ISSUE #977: app_auth credentials expire after an hour

sreejesh123:
Resolves #977

----

### Before the change?

* 

### After the change?

* 

### Pull request checklist

- [ ] Yes
- [x] No

----


kishaningithub:
@nickfloyd Would be great if this PR can be reviewed!

deiga (changes requested):
Hey @sreejesh123 👋 

Thank you for your contribution!

There are a few details I'd need your help with to understand your proposal :)

- You seem to be relying on ENV variables for the App identification, that won't work. You would need to use the parameters given to the provider
- You built a whole custom token refreshingSource, did you try using https://pkg.go.dev/golang.org/x/oauth2#Config.Client which looks like it would already include a lot of the logic you needed to create
- I don't think there should be a need to add a new library for doing the token refresh, but I might be wrong

sreejesh123:
Thanks a lot for the reply, appreciate taking time to review , please see answers inline 
> 
> * You seem to be relying on ENV variables for the App identification, that won't work. You would need to use the parameters given to the provider  

Been a while after I did this , I will take a look why I used os.env , not sure if it was for local
Setup, might be  I did overlook the actual runtime requirement . I will
Cross check and update the pr. 

> * You built a whole custom token refreshingSource, did you try using https://pkg.go.dev/golang.org/x/oauth2#Config.Client which looks like it would already include a lot of the logic you needed to create

I did explore oauth2.Config.Client, but GitHub App installation tokens do not follow standard OAuth2 refresh token mechanics.

GitHub Apps require this flow:
	1.	Create a short-lived JWT signed with the app’s private key
	2.	Exchange that JWT for an installation access token
	3.	Refresh by repeating steps 1–2 (no long-lived refresh token exists)

Since this flow isn’t compatible with oauth2.Config, I implemented a custom TokenSource that performs the GitHub-specific JWT → installation-token exchange. If this is my misunderstanding please let me know. Happy to adjust the structure if there’s a preferred pattern within the provider.

> * I don't think there should be a need to add a new library for doing the token refresh, but I might be wrong

The additional library is used only for securely generating the required JWT for GitHub App authentication.

If you prefer not to introduce this dependency, I can alternatively use the built-in Go crypto packages and re-implement the minimal signing logic.
However, the dedicated library reduces code size and avoids re-implementing JWT logic, so I’d appreciate guidance on which direction you prefer.

deiga:
@sreejesh123 Looking at the docs of go-github they have a recommendation to use https://github.com/jferrl/go-githubauth which would handle the refresh https://github.com/google/go-github/blob/master/README.md#as-a-github-app

Could you try implementing that instead of homebrewing the logic?

PR #2703, opened 2025-07-16
Max per page

### LINKED ISSUE #2557: [BUG]: github_organization_team_sync_groups causing unhandled 429 (Too Many Attempts) errors
### Expected Behavior

Running TF code using the `github_organization_team_sync_groups` resource to apply mapping works normally when running in multiple workspaces at the same time

### Actual Behavior

When running multiple TF runs sumulatiously (20+) causes the `github_organization_team_sync_groups` to fail with `429` status codes on the API:

```
Error: GET https://api.github.com/orgs/connectedbrewery/team-sync/groups?page=xxx&per_page=100: 429 []
with module.xxx.data.github_organization_team_sync_groups.okta_groups
on ../../modules/xxx/github.tf line 1, in data "github_organization_team_sync_groups" "okta_groups":
data "github_organization_team_sync_groups" "okta_groups" {}
```

We're been scaling out our HCP concurrent runners to 50, but when trigger a set of workspaces that use this data source, after the 25th or so workspace the rest fails with this error.

All other Providers handle this well, even the Github providers works decent but this is the only datasource that consistently fails with this error.

Since a 429 is not retried by default I've tries to work around it with the following provider configuration:

```
  max_retries      = 3
  retry_delay_ms   = 10000
  retryable_errors = [429, 500, 502, 503, 504]
```

But since solving a "Too many attempts" with more attempts is of course not a really good idea.... As expected this doesn't resolve the issue.

Is there something that can be implemented in the provider to solve/improve this? 

Since getting ALL groups for a simple mapping is already a bit overkill, maybe related feature request can help with this: https://github.com/integrations/terraform-provider-github/issues/1809

### Terraform Version

Terraform 1.10.5
HCP
integrations/github: 6.5.0

### Affected Resource(s)

- github_organization_team_sync_groups

```hcl

```

### Steps to Reproduce

_No response_

### Debug Output

```shell

```

### Panic Output

```shell

```

### Code of Conduct

dee-kryvenko:
Resolves #2557

----

### Before the change?

Added a new `max_per_page` configuration option to the provider block backed by `GITHUB_MAX_PER_PAGE` env variable as default fn.

### After the change?

Current behavior remains backward compatible as the default is still set to `100`.

### Pull request checklist

- [ ] Yes
- [x] No

----

### DIFF
--- a/github/provider.go
+++ b/github/provider.go
@@ -122,6 +122,13 @@ func Provider() *schema.Provider {
 					},
 				},
 			},
+			// https://developer.github.com/guides/traversing-with-pagination/#basics-of-pagination
+			"max_per_page": {
+				Type:        schema.TypeInt,
+				Optional:    true,
+				DefaultFunc: schema.EnvDefaultFunc("GITHUB_MAX_PER_PAGE", "100"),
+				Description: descriptions["max_per_page"],
+			},
 		},
 
 		ResourcesMap: map[string]*schema.Resource{
@@ -305,6 +312,8 @@ func init() {
 			"Defaults to [500, 502, 503, 504]",
 		"max_retries": "Number of times to retry a request after receiving an error status code" +
 			"Defaults to 3",
+		"max_per_page": "Number of items per page for pagination" +
+			"Defaults to 100",
 	}
 }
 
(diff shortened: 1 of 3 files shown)


dee-kryvenko:
I wanted to add it as a param to `Config` struct but the whole shenanigans with `Config` not being available in the runtime context and instead having the `Meta()` as `interface{}` and the casting - I am baffled what is going on in there, and didn't want to touch any of that. This should be safe enough since TF guarantees `ConfigureContextFunc` to be called thread safely.

PR #2710, opened 2025-07-24
Disable preemptive rate limiting

### LINKED ISSUE #2709: [FEAT]: Bypass preemptive rate limit check  by go-github
### Describe the need

## Proposal

### SDK Version

_No response_

### API Version

_No response_

### Relevant log output

```shell

```

### Code of Conduct

kishaningithub:
Resolves #2709

----

### Before the change?

* Preemptive rate limiting is by default enabled by go-github which makes the concurrent requests to fail

### After the change?

* Premptive rate limiting by go-github library is eliminated thereby giving the complete rate limiting control to this provider.

### Pull request checklist

- [ ] Yes
- [ ] No

----


stevehipwell:
@kishaningithub upgrading the SDK version is going to be complicated as there are going to be a significant number of related changes required.

kishaningithub:
I understand, would be great if we can simplify this and get it done

Let me know your ideas

stevehipwell:
@kishaningithub I've added a comment to your issue, we should probably have any discussions there.

kishaningithub:
Waiting on this PR to be merged https://github.com/integrations/terraform-provider-github/pull/2898

deiga:
@kishaningithub We've now upgraded to the latest go-github version.

If I'm understanding the change you are proposing, you are disabling the rate limit handling from go-github, without adding any handling to the provider. This doesn't sound feasible.

As suggested in the issue, could you implement https://github.com/gofri/go-github-ratelimit?

stevehipwell:
@deiga the provider is doing some rate limiting in the transport, but it's only ever been tested in coordination with the SDK rate limiting.

@kishaningithub if we were to make this change there will be a lot of testing required, are you in a position to add the tests and run them? My preference is to disable both the SDK rate limiter and the current provider implementation and to replace them with [gofri/go-github-ratelimit](https://github.com/gofri/go-github-ratelimit) integrated with the resource timeouts (that we've been adding as part of the switch to context based functions).

PR #2736, opened 2025-08-20
fix: improve error handling and define delete file options

### LINKED ISSUE #2735: [BUG]: Repository Delete File does not include author and comitter
### Expected Behavior

Options used for making a deleting a file differs from create and update, delete commits do not provide author and committer. Therefore, deletes will not be verified and not accepted for repositories enforcing verified commits.

see: https://github.com/integrations/terraform-provider-github/blob/main/github/resource_github_repository_file.go#L470

### Actual Behavior

Delete file should behave the same as create and update. If auther and committer is configured comits in github should have the verified batch, and if policy for only accepting verified commits is set should be accepted

### Terraform Version

Terraform v1.12.2
on linux_amd64

### Affected Resource(s)

- github_repository_file

```hcl
// Create a file:

resource "github_repository_file" "foo" {
  repository          = "tmp"
  branch              = "main"
  file                = ".gitignore"
  content             = "**/*.tfstate"
  commit_message      = "Managed by Terraform"
  commit_author       = "Terraform User"
  commit_email        = "terraform@example.com"
  overwrite_on_create = true
}
```

### Steps to Reproduce

// Delete created file by removing resource
// resource "github_repository_file" "foo" {
//   repository          = "tmp"
//   branch              = "main"
//   file                = ".gitignore"
//   content             = "**/*.tfstate"
//   commit_message      = "Managed by Terraform"
//   commit_author       = "Terraform User"
//   commit_email        = "terraform@example.com"
//   overwrite_on_create = true
// }}

### Debug Output

```shell

```

### Panic Output

```shell

```

### Code of Conduct

ivanmartos:
Resolves #2735

----

### Before the change?

* currently deleting `github_repository_file` fails silently
* when deleting`github_repository_file` then incomplete `RepositoryContentFileOptions` is passed to github sdk. This means that deletion commit is not signed and in case there is a ruleset requiring signed commits, this delete commit will fail silently
* currently when all tests are executed for `TestAccGithubRepositoryFile` they fail with duplicate repository name

### After the change?

* deletion of `github_repository_file` no longer fails silently, but if there is an error, it is retruned
* proper `RepositoryContentFileOptions` is passed to sdk call to `DeleteFile` (same as for create and update methods). This way commits are signed
* each `TestAccGithubRepositoryFile` test execution creates new repository specific for given test

### Pull request checklist

- [ ] Yes
- [ ] No

----

### DIFF
--- a/github/resource_github_repository_file.go
+++ b/github/resource_github_repository_file.go
@@ -460,16 +460,9 @@ func resourceGithubRepositoryFileDelete(d *schema.ResourceData, meta interface{}
 
 	var branch string
 
-	message := fmt.Sprintf("Delete %s", file)
-
-	if commitMessage, hasCommitMessage := d.GetOk("commit_message"); hasCommitMessage {
-		message = commitMessage.(string)
-	}
-
-	sha := d.Get("sha").(string)
-	opts := &github.RepositoryContentFileOptions{
-		Message: &message,
-		SHA:     &sha,
+	opts, err := resourceGithubRepositoryFileOptions(d)
+	if err != nil {
+		return err
 	}
 
 	if b, ok := d.GetOk("branch"); ok {
@@ -503,9 +496,9 @@ func resourceGithubRepositoryFileDelete(d *schema.ResourceData, meta interface{}
 		opts.Branch = &branch
 	}
 
-	_, _, err := client.Repositories.DeleteFile(ctx, owner, repo, file, opts)
+	_, _, err = client.Repositories.DeleteFile(ctx, owner, repo, file, opts)
 	if err != nil {
-		return nil
+		return err
 	}
 
 	return nil
--- a/github/resource_github_repository_file.go
+++ b/github/resource_github_repository_file.go
@@ -465,6 +465,11 @@ func resourceGithubRepositoryFileDelete(d *schema.ResourceData, meta interface{}
 		return err
 	}
 
+	if *opts.Message == fmt.Sprintf("Add %s", file) {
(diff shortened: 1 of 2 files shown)


nickfloyd:
This is technically a breaking change - prior to the change this would fail forward, after the change an error is thrown - functionally the operation does not work either way but user workflows could be disrupted even still.  

We'll schedule this one to go out in the up coming major release.  Thank you for working on this change - we'll work to get it out the door!

nickfloyd:
I'm rolling my previous comment back.  We implemented some fixes to more appropriately handle things like this for archived repos - the behavior defined in that change set and this one is, I feel, what is expected.

github/resource_github_repository_file_test.go
nickfloyd:
```suggestion
		`, acctest.RandStringFromCharSet(5, acctest.CharSetAlphaNum))
```

nickfloyd:
```suggestion
						`,acctest.RandStringFromCharSet(5, acctest.CharSetAlphaNum)),
```

github/resource_github_repository_file.go
nickfloyd:
```suggestion
	_, _, err = client.Repositories.DeleteFile(ctx, owner, repo, file, opts)
```

nickfloyd (changes requested):

github/resource_github_repository_file.go
nickfloyd:
```suggestion
	_, _, err = client.Repositories.DeleteFile(ctx, owner, repo, file, opts)
```

PR #2758, opened 2025-09-17
fix(resource/github_actions_environment_variable): handle existing va…

nebrass:
# fix(actions): Handle existing environment variables on create

### Description

This PR fixes an issue with the `github_actions_environment_variable` resource where it would fail with a `409 Conflict` error if the environment variable already existed in GitHub but not in the Terraform state. This typically occurs when a variable is created manually via the GitHub UI and then later managed by Terraform.

The `Create` function for the resource has been updated to handle this scenario gracefully. It now attempts to `Update` the variable if it already exists, effectively performing an "upsert" operation. This makes the resource more robust and prevents unexpected errors.

Fixes #2328

### How to test

An acceptance test has been added to cover this specific scenario. You can run the test with the following command:

```sh  
TF_ACC=1 go test -v ./github -run ^TestAccGithubActionsEnvironmentVariable_alreadyExists$
```

This test:
1.  Creates a repository and an environment.
2.  Creates an environment variable using the GitHub API to simulate an out-of-band creation.
3.  Applies a Terraform configuration that defines the same variable.
4.  Verifies that Terraform successfully "adopts" the variable without any errors.


github/resource_github_actions_environment_variable.go
```diff
@@ -73,33 +73,22 @@ func resourceGithubActionsEnvironmentVariableCreate(d *schema.ResourceData, meta
...
-
-	variable := &github.ActionsVariable{
-		Name:  name,
-		Value: d.Get("value").(string),
-	}
-
-	_, err := client.Actions.UpdateEnvVariable(ctx, owner, repoName, escapedEnvName, variable)
-	if err != nil {
-		return err
+		if ghErr, ok := err.(*github.ErrorResponse); ok {
```
mbarkiMohamed95, line 79: suggestion :
if ghErr, ok := err.(*github.ErrorResponse); ok && ghErr.Response.StatusCode == http.StatusConflict {
    _, err = client.Actions.UpdateEnvVariable(ctx, owner, repoName, escapedEnvName, variable)
}
return err

Issue #2755, opened 2025-09-16
[FEAT]: github_repository_ruleset - support bypass_mode "exempt"


audunsolemdal:
### Describe the need

https://github.blog/changelog/2025-09-10-github-ruleset-exemptions-and-repository-insights-updates/

this is now the recommended approach for bot users. The idea behind it is to reduce spam in the audit signals where bypassing is expected.

E.g.

```hcl

resource "github_repository_ruleset" "example" {
  name        = "example"
  repository  = github_repository.example.name

  bypass_actors {
    actor_id    = 13473
    actor_type  = "Integration"
    bypass_mode = "exempt" # new enum
  }
}

```

### Relevant log output

```shell

```

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

PR #2764, opened 2025-09-30
Add support for 'exempt' github_repository_ruleset bypass_mode

joshhunt:
Resolves [#ISSUE_NUMBER](https://github.com/integrations/terraform-provider-github/issues/2755)

I'm not familiar with terraform provider development, so I'm not sure if there's anything else I'm missing, but this looked right to me.

----

### Before the change?

* Terraform provider was unable to create ruleset bypasses with bypass_mode exempt, which was [recently added to Github](https://github.blog/changelog/2025-09-10-github-ruleset-exemptions-and-repository-insights-updates/)

### After the change?

* bypass_mode exempt is now supported!

### Pull request checklist

- [ ] Yes
- [x] No (I don't think so)

----

### DIFF
--- a/github/resource_github_organization_ruleset.go
+++ b/github/resource_github_organization_ruleset.go
@@ -64,7 +64,7 @@ func resourceGithubOrganizationRuleset() *schema.Resource {
 						"bypass_mode": {
 							Type:         schema.TypeString,
 							Required:     true,
-							ValidateFunc: validation.StringInSlice([]string{"always", "pull_request"}, false),
+							ValidateFunc: validation.StringInSlice([]string{"always", "pull_request", "exempt"}, false),
 							Description:  "When the specified actor can bypass the ruleset. pull_request means that an actor can only bypass rules on pull requests. Can be one of: `always`, `pull_request`.",
 						},
 					},
--- a/github/resource_github_repository_ruleset.go
+++ b/github/resource_github_repository_ruleset.go
@@ -69,7 +69,7 @@ func resourceGithubRepositoryRuleset() *schema.Resource {
 						"bypass_mode": {
 							Type:         schema.TypeString,
 							Required:     true,
-							ValidateFunc: validation.StringInSlice([]string{"always", "pull_request"}, false),
+							ValidateFunc: validation.StringInSlice([]string{"always", "pull_request", "exempt"}, false),
 							Description:  "When the specified actor can bypass the ruleset. pull_request means that an actor can only bypass rules on pull requests. Can be one of: `always`, `pull_request`.",
 						},
 					}},


krische:
Maybe don't replace the `*` with `-` in the markdown files? It makes it look like there are a lot more changes than there really are.

joshhunt:
Ahh, thanks for catching that @krische. My IDE auto-formatted the markdown upon save and I didn't catch it 😅

github/resource_github_organization_ruleset.go
```diff
@@ -64,7 +64,7 @@ func resourceGithubOrganizationRuleset() *schema.Resource {
 						"bypass_mode": {
 							Type:         schema.TypeString,
 							Required:     true,
-							ValidateFunc: validation.StringInSlice([]string{"always", "pull_request"}, false),
+							ValidateFunc: validation.StringInSlice([]string{"always", "pull_request", "exempt"}, false),
 							Description:  "When the specified actor can bypass the ruleset. pull_request means that an actor can only bypass rules on pull requests. Can be one of: `always`, `pull_request`.",
```
v1v: I think the description needs to be updated too:

```suggestion
 							Description:  "When the specified actor can bypass the ruleset. pull_request means that an actor can only bypass rules on pull requests. Can be one of: `always`, `pull_request`, `exempt`.",
```

github/resource_github_repository_ruleset.go
```diff
@@ -69,7 +69,7 @@ func resourceGithubRepositoryRuleset() *schema.Resource {
 						"bypass_mode": {
 							Type:         schema.TypeString,
 							Required:     true,
-							ValidateFunc: validation.StringInSlice([]string{"always", "pull_request"}, false),
+							ValidateFunc: validation.StringInSlice([]string{"always", "pull_request", "exempt"}, false),
 							Description:  "When the specified actor can bypass the ruleset. pull_request means that an actor can only bypass rules on pull requests. Can be one of: `always`, `pull_request`.",
```
v1v: I think the description needs to be updated too:

```suggestion
 							Description:  "When the specified actor can bypass the ruleset. pull_request means that an actor can only bypass rules on pull requests. Can be one of: `always`, `pull_request`, `exempt`.",
```

oikarinen:
@joshhunt are you working on requested changes? Would need this to go forward and happy to help.

joshhunt:
Thanks for the reminder @oikarinen. I fixed it!

oikarinen:
> We should follow up with some tests - there doesn't seem to be any coverage for any of the bypass_modes, this new one or otherwise 😬

Hope this helps https://github.com/integrations/terraform-provider-github/pull/2802

PR #2840, opened 2025-10-29
feat: Adds DiffSuppressFunc and DiffSuppressOnRefresh to resources that have etag properties to suppress etag-related diffs

### LINKED ISSUE #796: continual changes detected due to changing etag

nickfloyd:
Resolves #796 

----

### Before the change?

* Prior to this change the provider would detect etag diff changes 

### After the change?

* After a successful terraform apply, terraform plan or terraform apply detects no changes.

Details from #796 

Initially an approach was made but the implementation was incomplete due to needing to support update.  4 years later we're on a newer version of the SDK (terraform-plugin-sdk/v2 v2.34.0)  that supports `DiffSuppressOnRefresh`

We should now be able to go with the initial approach (even on computed fields) of:
```
DiffSuppressFunc: func(k, old, new string, d *schema.ResourceData) bool {
    return true
},
```

And add `DiffSuppressOnRefresh: true` as well to suppress the diff between config and state.

Lastly these fields have to be marked as optional=true - [SDK ref](https://github.com/integrations/terraform-provider-github/blob/16872b724254fdddc3441c713b087cb4d7005f83/vendor/github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema/schema.go#L841)
```
Computed-only fields: Computed: true, Optional: false → Can't have DiffSuppressFunc
Optional computed fields: Computed: true, Optional: true → Can have DiffSuppressFunc
```

I believe we can set this to always returns true - which should suppress all etag diffs; this should also work for refresh operations as well.

This approach should be safe because ETags are internal concerns, not user configuration, ETags are still stored and used for API optimization by the SDK / provider, and all user changes to actual resource attributes are still detected normally.

Let me put a quick PR together to illustrate this and see if everyone agrees with the approach. The better move would be to migrate to the Plugin Framework but for now we can get this cleaned up a bit as is.

### Pull request checklist

- [ ] Yes
- [x] No

----

### DIFF
--- a/github/migrate_github_actions_secret.go
+++ b/github/migrate_github_actions_secret.go
@@ -1,0 +1,36 @@
+package github
+
+import (
+	"fmt"
+	"log"
+
+	"github.com/hashicorp/terraform-plugin-sdk/v2/terraform"
+)
+
+func resourceGithubActionsSecretMigrateState(v int, is *terraform.InstanceState, meta interface{}) (*terraform.InstanceState, error) {
+	switch v {
+	case 0:
+		log.Printf("[INFO] Found GitHub Actions Secret State v0; migrating to v1")
+		return migrateGithubActionsSecretStateV0toV1(is)
+	default:
+		return is, fmt.Errorf("unexpected schema version: %d", v)
+	}
+}
+
+func migrateGithubActionsSecretStateV0toV1(is *terraform.InstanceState) (*terraform.InstanceState, error) {
+	if is.Empty() {
+		log.Printf("[DEBUG] Empty InstanceState; nothing to migrate.")
+		return is, nil
+	}
+
+	log.Printf("[DEBUG] GitHub Actions Secret Attributes before migration: %#v", is.Attributes)
+
+	// Add the destroy_on_drift field with default value true if it doesn't exist
+	if _, ok := is.Attributes["destroy_on_drift"]; !ok {
+		is.Attributes["destroy_on_drift"] = "true"
+	}
+
+	log.Printf("[DEBUG] GitHub Actions Secret Attributes after State Migration: %#v", is.Attributes)
+
+	return is, nil
+}
--- a/github/migrate_github_actions_secret_test.go
+++ b/github/migrate_github_actions_secret_test.go
@@ -1,0 +1,69 @@
+package github
+
+import (
+	"reflect"
+	"testing"
+
+	"github.com/hashicorp/terraform-plugin-sdk/v2/terraform"
+)
+
+func TestMigrateGithubActionsSecretStateV0toV1(t *testing.T) {
+	// Secret without destroy_on_drift should get default value
+	oldAttributes := map[string]string{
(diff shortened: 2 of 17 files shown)


Wirone:
Sorry for digging old PRs, but we struggle with this in our PR automation.

> Installed integrations/github v6.11.0 (signed by a HashiCorp partner, key ID 38027F80D7FD5FB2)

Even though we're using 6.11, we see _a lot_ `etag` changes for `github_team` resources in the output, which makes it cumbersome to find actual changes in the Atlantis plan's output. What can be the reason?

deiga:
@Wirone Could you share DEBUG logs of a plan to help us narrow it down?

Wirone:
@deiga it's in our company's proprietary project for IaaC, so it may be hard to share it as it contains a lot of information about the infra even in INFO mode... But in short: today I added 6 `github_actions_organization_variable`/ `github_actions_organization_secret` resources and the Atlantis plan's output was so long (because of etag stuff) that it reached Github's comment's length limit and had to be split in two, with the info I was interested in at the very end of it 🤷. We live with it, but it's pretty overwhelming.

deiga:
> @deiga it's in our company's proprietary project for IaaC, so it may be hard to share it as it contains a lot of information about the infra even in INFO mode... But in short: today I added 6 `github_actions_organization_variable`/ `github_actions_organization_secret` resources and the Atlantis plan's output was so long (because of etag stuff) that it reached Github's comment's length limit and had to be split in two, with the info I was interested in at the very end of it 🤷. We live with it, but it's pretty overwhelming.

@Wirone i get that it can be daunting to share something. 
I'm not asking for a full log, you're free to redact it. The thing is, you're the first in a long while reporting this and we don't have a clue on why it wouldn't work. 

Would you be able to use targeting to only one of these resources on DEBUG level? That could be enough to follow where the code is going. 
And feel free to obfuscate anything that could be sensitive ☺️

Wirone:
I will try to do it 👍.

Issue #2809, opened 2025-10-23
[MAINT]: Update to go-github v68


nickfloyd:
### Describe the need

⚠️ This is a significant breaking change around the projects API. 

The v68 [release](https://github.com/google/go-github/releases/tag/v68.0.0) has the following breaking changes:

feat!: Remove Project (classic) support (https://github.com/google/go-github/pull/3380)
BREAKING CHANGE: All support for "Project (classic)" is removed and Project responses are replaced by ProjectV2 (renamed from ProjectsV2).
fix!: Remove unused Stats field from Commit struct (https://github.com/google/go-github/pull/3395)
BREAKING CHANGE: The unused Stats field is removed from the Commit struct.

### Relevant log output

```shell

```

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

nickfloyd:
Post upgrade issues (expected):

```
# github.com/integrations/terraform-provider-github/v6/github
github/resource_github_organization_project.go:62:42: client.Organizations.CreateProject undefined (type *"github.com/google/go-github/v68/github".OrganizationsService has no field or method CreateProject)
github/resource_github_organization_project.go:64:11: undefined: github.ProjectOptions
github/resource_github_organization_project.go:95:31: client.Projects undefined (type *"github.com/google/go-github/v68/github".Client has no field or method Projects)
github/resource_github_organization_project.go:139:20: undefined: github.ProjectOptions
github/resource_github_organization_project.go:150:25: client.Projects undefined (type *"github.com/google/go-github/v68/github".Client has no field or method Projects)
github/resource_github_organization_project.go:170:18: client.Projects undefined (type *"github.com/google/go-github/v68/github".Client has no field or method Projects)
github/resource_github_project_card.go:75:20: undefined: github.ProjectCardOptions
github/resource_github_project_card.go:92:25: client.Projects undefined (type *"github.com/google/go-github/v68/github".Client has no field or method Projects)
github/resource_github_project_card.go:115:25: client.Projects undefined (type *"github.com/google/go-github/v68/github".Client has no field or method Projects)
github/resource_github_project_card.go:152:20: undefined: github.ProjectCardOptions
github/resource_github_project_card.go:152:20: too many errors
make: *** [build] Error 1
```

nickfloyd:
Possible path forward: It looks like in order to get the functionality that we need for projects v2 we will have to jump to go-github [v77](https://github.com/google/go-github/releases/tag/v77.0.0)

[...]

Issue #2602, opened 2025-03-21
[MAINT]: Upgrade google/go-github to v70


zepeng811:
### Describe the need

We are currently on v66 and the latest version for google/go-github is v70. We should upgrade the dependency to stay on the latest version.

Release Notes:
* v67: https://github.com/google/go-github/releases/tag/v67.0.0
* v68: https://github.com/google/go-github/releases/tag/v68.0.0
* v69: https://github.com/google/go-github/releases/tag/v69.0.0
* v70: https://github.com/google/go-github/releases/tag/v70.0.0

### SDK Version

google/go-github v70

### API Version

N/A

### Relevant log output

N/A

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

tobias-jenkner:
Any ETA on this update?

al3xanderschmidt:
Any updates on this? The newest version for go-github is already v74 and can't await to see a version upgrade in a new release of terraform-provider-github

eyalgal:
We're working on upgrading go-github to v77 (and beyond).  
There is a [PR](https://github.com/integrations/terraform-provider-github/pull/2891) tracking the changes.

This upgrade introduces a number of breaking changes.  
We expect to release it in the coming weeks.

eyalgal:
We’ve upgraded `google/go-github` beyond v70 and are now on v81 in the latest release. Closing as resolved.

tobias-jenkner:
Great! Thank you very much!

Issue #2889, opened 2025-11-13
[BUG]: Resource github_repository_ruleset inconsistence issue


neptune19821220:
### Expected Behavior

I saw the release 6.8.0 including the PR #2821 which said it fixed the issue #2818 I created.
But there is no related info about my issue in the PR change notes.

After I upgrade provider to registry.terraform.io/integrations/github v6.8.0 today, I found my issue still exists.
Could you double check it?

### Actual Behavior

Terraform would like to create a new resource with the same name, though it will fail due to the same name.

### Terraform Version

Terraform v1.13.5
on linux_arm64
+ provider registry.terraform.io/integrations/github v6.8.0

### Affected Resource(s)

- github_repository_ruleset

### Terraform Configuration Files

```hcl

```

### Steps to Reproduce

1. terraform apply -> create a ruleset resource
2. select the option "Automatically request Copilot code review"
3. terraform plan -> shows terraform will create a new ruleset
4. terraform apply -> failed to create because of ruleset name duplication
5. Remove the selection "Automatically request Copilot code review"
6. terraform plan -> show "no change"

### Debug Output

```shell

```

### Panic Output

```shell

```

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

neptune19821220:
This is the sample terraform resource block I used.
```js
resource "github_repository_ruleset" "ruleset" {
  name        = "TEST"
  repository  = "myrepo"
  target      = "branch"
  enforcement = "active"
  conditions {
    ref_name {
      include = ["~DEFAULT_BRANCH"]
      exclude = []
    }
  }
  rules {
  }
}

**Comment by stevehipwell** [comment]
@neptune19821220 AFAIK a ruleset can't be created without any rules so this is likely caused by that and we need to improve the validation. Could you check if you still get the issue if you always have at least one rule defined?

**Comment by neptune19821220** [comment]
@stevehipwell 
I still met the issue if there have some rules defined. The test template is as following:

```js
resource "github_repository_ruleset" "ruleset" {
  name        = "TEST"
  repository  = "myrepo"
  target      = "branch"
  enforcement = "active"
  conditions {
    ref_name {
      include = ["~DEFAULT_BRANCH"]
      exclude = []
    }
  }
  rules {
    required_linear_history = true
    branch_name_pattern {
        operator = "starts_with"
        pattern = "feature/"
        name = ""
        negate = false
    }
  }
}
```
[...]

Issue #2497, opened 2024-12-04
[FEAT]: Replace `team_id` with `team_slug`


stevehipwell:
### Describe the need

The REST APIs for interacting with team by ID are currently under a closing down notice ([ref](https://docs.github.com/en/rest/teams/teams?apiVersion=2022-11-28#get-a-team-legacy)) so we should switch to using `team_slug`. Although #2421 is currently still relevant and correct this would change this position with the eventual removal of the `team_id`.

This change should be able to be completed with minimal end user pain as we can add `team_slug` alongside `team_id`. I'm planning on opening a PR to achieve this once #2476 has been merged. I'm waiting for this so there can be automated acceptance tests in place.

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

stevehipwell:
This issue isn't stale! IMHO using a stale bot for a repository that hasn't had any meaningful interactions from the maintainers for almost a year is user hostile.

PR #2891, opened 2025-11-13
maint: Updates go-github dep to v77

nickfloyd:
Resolves: #2809, #2602
Relates to: #2898

This change set is large.  This was an intentional decision to drive forward our need to update go-github given the current provider was 10 or more versions behind the latest. The hope is that the release will encapsulate all of the breaking changes over the last 12 months.  This helps to make us just deal with them once if we choose to move forward on the major version bump.  The last two releases were packed with much needed changes - so users, if they choose, can migrate as time allows while using the updates from v6.7.x and v6.8.x.

Additionally, we'll most likely cherry-pick out feature sets from this branch that are not actual fixes and the like - i.e. projects APIs have stopped working so we can look at projects v2 as a new feature.

----

### Before the change?

* go-github v67

### After the change?

* go-github v77

### Checklist
- [x] Update go-github to >= v77
  - [x] Vendor / Mod
- ~[x] Convert/Replace projects to projects v2~ (#2898)
- [x] API renames edit > update
  - [x] resource_github_actions_organization_permissions.go
  - [x] resource_github_actions_repository_permissions.go
  - [x] resource_github_enterprise_actions_permissions.go
  - [x] resource_github_enterprise_actions_workflow_permissions.go
- [x] Fix CreateReference
- [x] Default Ruleset changes
  - [x] resource_github_organization_ruleset.go
  - [x] resource_github_repository_ruleset.go
  - [x] respository_rules_utils.go
- [x] Ensure test coverage and documentation (for new and deprecated)
- [x] Ensure schema migrations if needed
  - [x] RuleSet migrations

### Pull request checklist

- [x] Yes
- [ ] No

----

### DIFF
--- a/github/resource_github_repository_ruleset.go
+++ b/github/resource_github_repository_ruleset.go
@@ -713,13 +713,13 @@ func resourceGithubRepositoryRulesetUpdate(d *schema.ResourceData, meta any) err
 		if err != nil {
 			return err
 		}
-		ruleset, _, err = client.Repositories.UpdateRuleset(ctx, owner, repoName, rulesetID, *rulesetReq)
-	} else {
-		ruleset, _, err = client.Repositories.UpdateRuleset(ctx, owner, repoName, rulesetID, *rulesetReq)
 	}
+
+	ruleset, _, err = client.Repositories.UpdateRuleset(ctx, owner, repoName, rulesetID, *rulesetReq)
 	if err != nil {
 		return err
 	}
+
 	d.SetId(strconv.FormatInt(*ruleset.ID, 10))
 
 	return resourceGithubRepositoryRulesetRead(d, meta)


stevehipwell:
@nickfloyd I've noticed that you've swapped out team ids for slugs to deal with the deprecated methods but that will break the provider for anyone using it at any serious scale. The fix here is to deprecate the `team_id` field and introduce a parallel `team_slug` field; the cost of using a single field for both is significant and I previously optimised it for ID as that's what the provider was using in the REST API.

See #2497.

nickfloyd:
@stevehipwell Thanks for the heads up.  I'll definitely pull that out... I am pulling things out of this change set that are orthogonal to the upgrade - I'll pull those out and just throw the changes away instead.

stevehipwell:
@nickfloyd now that the provider is under maintenance I wouldn't be against adding the `team_slug` fields. I've got a few other team changes that are waiting on this PR.

stevehipwell:
@nickfloyd is this ready for a review?

nickfloyd:
> @nickfloyd is this ready for a review?

Yes it is ❤️ .  I'll need to sync the branch with #2902 once/if you approve my changes there.

stevehipwell:
@nickfloyd I'll review this shortly.

github/resource_github_organization_ruleset_test.go
```diff
@@ -597,22 +597,40 @@ func TestOrganizationPushRulesetSupport(t *testing.T) {
...
 	// Verify we have all expected push rule types
-	ruleTypes := make(map[string]bool)
-	for _, rule := range expandedRules {
-		ruleTypes[rule.Type] = true
+	ruleCount := 0
+	if expandedRules.FilePathRestriction != nil {
+		ruleCount++
+		if len(expandedRules.FilePathRestriction.RestrictedFilePaths) != 3 {
+			t.Errorf("Expected 3 restricted file paths, got %d", len(expandedRules.FilePathRestriction.RestrictedFilePaths))
+		}
```
deiga: question: Would it make sense to use something like `len(rulesMap["file_path_restriction"][0]["file_path_restriction"])` instead of hard-coding the number 3 here?

github/resource_github_organization_ruleset_test.go
```diff
@@ -597,22 +597,40 @@ func TestOrganizationPushRulesetSupport(t *testing.T) {
...
 	}
+	if expandedRules.MaxFilePathLength != nil {
+		ruleCount++
+		if expandedRules.MaxFilePathLength.MaxFilePathLength != 250 {
+			t.Errorf("Expected max file path length to be 250, got %d", expandedRules.MaxFilePathLength.MaxFilePathLength)
+		}
+	}
+	if expandedRules.FileExtensionRestriction != nil {
+		ruleCount++
+		if len(expandedRules.FileExtensionRestriction.RestrictedFileExtensions) != 4 {
```
deiga: question: Same for these conditionals. Would it make sense to refer the `rulesMap` values directly instead of hard-coding them?

website/docs/r/project_card.html.markdown
```diff
@@ -9,6 +9,8 @@ description: |-
 
 !> **Warning:** This resource no longer works as the [Projects (classic) REST API](https://docs.github.com/en/rest/projects/projects?apiVersion=2022-11-28) has been [removed](https://github.blog/changelog/2024-05-23-sunset-notice-projects-classic/) and as such has been deprecated. It will be removed in a future release.
 
+~> **Migration:** Please use the [`github_project_item`](./project_item.html) resource for GitHub Projects V2 instead.
```
deiga: issue: This file seems to be missing?

github/resource_github_organization_ruleset.go
```diff
@@ -588,57 +589,70 @@ func resourceGithubOrganizationRuleset() *schema.Resource {
...
 
+	org := meta.(*Owner).name
 	ctx := context.Background()
 
-	var ruleset *github.Ruleset
+	var ruleset *github.RepositoryRuleset
 	var err error
 
-	ruleset, _, err = client.Organizations.CreateOrganizationRuleset(ctx, owner, rulesetReq)
+	ruleset, _, err = client.Organizations.CreateRepositoryRuleset(ctx, org, *rulesetReq)
```
deiga, line 599: note: Boy, this choice of naming was poor from the SDK. I had to read the API and SDK contents to make sure it's still doing the right thing 🙃

github/resource_github_repository_ruleset.go
```diff
@@ -684,15 +685,20 @@ func resourceGithubRepositoryRulesetUpdate(d *schema.ResourceData, meta any) err
 
 	ctx := context.WithValue(context.Background(), ctxId, rulesetID)
 
-	var ruleset *github.Ruleset
+	var ruleset *github.RepositoryRuleset
 	// Use UpdateRulesetNoBypassActor here instead of UpdateRuleset *if* bypass_actors has changed.
 	// UpdateRuleset uses `omitempty` on BypassActors, causing empty arrays to be omitted from the JSON.
 	// UpdateRulesetNoBypassActor always includes the field so that bypass actors can actually be removed.
 	// See: https://github.com/google/go-github/blob/b6248e6f6aec019e75ba2c8e189bfe89f36b7d01/github/repos_rules.go#L196
```
deiga: question: This comment doesn't seem to be correct anymore, since we don't use `UpdateRulesetNoBypassActor` now. Is it otherwise still relevant?

github/data_source_github_organization_security_managers.go
```diff
@@ -51,6 +51,7 @@ func dataSourceGithubOrganizationSecurityManagersRead(d *schema.ResourceData, me
 
 	allTeams := make([]any, 0)
 
+	//nolint:staticcheck // SA1019: ListSecurityManagerTeams is deprecated but still needed for legacy compatibility
```
deiga, line 56: question: Do we have a plan when to migrate this to use `ListTeamsAssignedToOrgRole` instead? It seems Jan 1st `ListSecurityManagerTeams` might stop working?

website/docs/r/project_card.html.markdown
stevehipwell: I think this doc needs removing as the projects V2 resources are in a separate PR and there is no migration path.

github/resource_github_repository_collaborators.go
```diff
@@ -407,8 +407,8 @@ func matchUserCollaboratorsAndInvites(repoName string, want []any, hasUsers []us
 
 func matchTeamCollaborators(repoName string, want []any, has []teamCollaborator, meta any) error {
 	client := meta.(*Owner).v3client
-	orgID := meta.(*Owner).id
 	owner := meta.(*Owner).name
+	orgId := meta.(*Owner).id
```
stevehipwell:
```suggestion
	orgID := meta.(*Owner).id
```

Why did you change this? In Go the idiomatic capitalization is `<thing>ID`.

github/resource_github_repository_pull_request.go
```diff
@@ -289,13 +289,13 @@ func resourceGithubRepositoryPullRequestUpdate(d *schema.ResourceData, meta any)
 		return resourceGithubRepositoryPullRequestRead(d, meta)
 	}
 
-	errs := []string{fmt.Sprintf("could not update the Pull Request: %v", err)}
+	errors := []string{fmt.Sprintf("could not update the Pull Request: %v", err)}
```
stevehipwell:
```suggestion
	errs := []string{fmt.Sprintf("could not update the Pull Request: %v", err)}
```

Why did you change this? This would shadow the `errors` package in the stdlib and even if it's not currently being used in this file it's not idiomatic.

github/resource_github_team_repository.go
```diff
@@ -108,23 +109,20 @@ func resourceGithubTeamRepositoryRead(d *schema.ResourceData, meta any) error {
 	}
 
 	client := meta.(*Owner).v3client
-	orgId := meta.(*Owner).id
+	teamSlug := d.Get("slug").(string)
```
stevehipwell: I'm not sure where `slug` is defined? I suspect this may be left over from some previous changes?

github/resource_github_team_repository.go
```diff
@@ -108,23 +109,20 @@ func resourceGithubTeamRepositoryRead(d *schema.ResourceData, meta any) error {
...
 
 	teamIdString, repoName, err := parseTwoPartID(d.Id(), "team_id", "repository")
 	if err != nil {
 		return err
 	}
-	teamId, err := getTeamID(teamIdString, meta)
-	if err != nil {
-		return err
-	}
```
stevehipwell, line 120: I think this might need putting back?

github/resource_github_team_repository.go
```diff
@@ -108,23 +109,20 @@ func resourceGithubTeamRepositoryRead(d *schema.ResourceData, meta any) error {
...
+
 	orgName := meta.(*Owner).name
 	ctx := context.WithValue(context.Background(), ctxId, d.Id())
 	if !d.IsNewResource() {
 		ctx = context.WithValue(ctx, ctxEtag, d.Get("etag").(string))
 	}
 
-	repo, resp, repoErr := client.Teams.IsTeamRepoByID(ctx, orgId, teamId, orgName, repoName)
+	repo, resp, repoErr := client.Teams.IsTeamRepoBySlug(ctx, orgName, teamSlug, orgName, repoName)
```
stevehipwell: I think this might need undoing?

github/resource_github_organization_ruleset.go
```diff
@@ -23,7 +23,8 @@ func resourceGithubOrganizationRuleset() *schema.Resource {
 			State: resourceGithubOrganizationRulesetImport,
 		},
 
-		SchemaVersion: 1,
+		SchemaVersion: 2,
```
stevehipwell: I'm not sure why we need to do this? If we don't then I'd say that this goes against the principle of least surprise.

github/resource_github_organization_ruleset.go
```diff
@@ -588,57 +589,70 @@ func resourceGithubOrganizationRuleset() *schema.Resource {
...
 	}
 
-	var ruleset *github.Ruleset
+	var ruleset *github.RepositoryRuleset
 	var resp *github.Response
 
-	ruleset, resp, err = client.Organizations.GetOrganizationRuleset(ctx, owner, rulesetID)
+	ruleset, resp, err = client.Organizations.GetRepositoryRuleset(ctx, org, rulesetID)
 	if err != nil {
 		ghErr := &github.ErrorResponse{}
```
stevehipwell, line 627: This code pattern looks to be fundamentally flawed as we've got an error and are not handling it, I suspect we'd find it throughout the codebase. In this case we don't return the error and the ruleset is empty so we remove it from state which can result in issues like #2889.

github/resource_github_organization_ruleset.go
```diff
@@ -588,57 +589,70 @@ func resourceGithubOrganizationRuleset() *schema.Resource {
...
 	}
 
+	if ruleset == nil {
+		log.Printf("[INFO] Removing organization ruleset %s: %d from state because it no longer exists in GitHub (empty response)",
+			org, rulesetID)
+		d.SetId("")
+		return nil
+	}
+
+	if ruleset == nil {
```
stevehipwell, line 642: This duplication looks like a typo?

github/resource_github_repository_collaborators.go
```diff
@@ -444,6 +445,10 @@ func matchTeamCollaborators(repoName string, want []any, has []teamCollaborator,
 	for _, t := range want {
 		teamData := t.(map[string]any)
 		teamIDString := teamData["team_id"].(string)
+		teamSlug, err := getTeamSlug(teamIDString, meta)
```
stevehipwell, line 448: I think this might be left over from some removed changes?

github/resource_github_repository_collaborators.go
```diff
@@ -460,9 +465,10 @@ func matchTeamCollaborators(repoName string, want []any, has []teamCollaborator,
 		}
 		permission := teamData["permission"].(string)
 		// team needs to be added
-		log.Printf("[DEBUG] Adding team %s with permission %s for repo: %s.", teamIDString, permission, repoName)
+		log.Printf("[DEBUG] Adding team %s with permission %s for repo: %s.", teamSlug, permission, repoName)
```
stevehipwell, line 468: We shouldn't have the team slug due to the overhead of doing so, so this should be using the team id.

github/resource_github_repository_ruleset.go
```diff
@@ -23,7 +23,8 @@ func resourceGithubRepositoryRuleset() *schema.Resource {
 			State: resourceGithubRepositoryRulesetImport,
 		},
 
-		SchemaVersion: 1,
+		SchemaVersion: 2,
```
stevehipwell: As org ruleset comment.

github/resource_github_organization_ruleset_test.go
nickfloyd: This is really following the current idioms in the provider.  It appears that the pattern might've set up for readability and determinism, as well as simplicity of implementation. We can certainly look at changing these down the road to establish a pattern.

stevehipwell (changes requested):
I've added some review comments.

github/resource_github_organization_ruleset_test.go
nickfloyd: My [opinion](https://github.com/integrations/terraform-provider-github/pull/2891#discussion_r2543680477) above is why I would not favor changing these right now but if we deem that as important then I'd be up for considering it.

website/docs/r/project_card.html.markdown
nickfloyd: Yeah this was a left over artifact from my conversion and then later sundering it into it's own PR. I'll remove it. Great catch!

github/resource_github_organization_ruleset.go
nickfloyd: Agreed.  I got this one wrong about 3 times because of it. Other than making the changes in the SDK itself I think we want to be cautious with our APIs as a result - balancing pragmatism with DX.

github/resource_github_organization_ruleset_test.go
stevehipwell: @nickfloyd it's worth noting that the duplication of the expected value in the test and the error message caused quite a lot of overhead in the patch alignment PR. I'm also not sure I agree that this can be classed as an idiom and would be tempted to follow the boy scout rule and fix it as part of the changes.

github/resource_github_repository_ruleset.go
nickfloyd: Great catch / point.  This was a changeset merged in after the migration.  I'll clean it up ❤️

github/data_source_github_organization_security_managers.go
nickfloyd: Yes.  Well sort of, there is an [issue](https://github.com/integrations/terraform-provider-github/issues/2497).  @stevehipwell rightfully [called out](https://github.com/integrations/terraform-provider-github/pull/2891#issuecomment-3533950639) the issues with migrating to the team_slug from team_ids.   RE: #2497

github/resource_github_organization_ruleset_test.go
nickfloyd: Absolutely. I can totally get behind that.  My concern is that this PR is already huge - so I can scope the modifications to just this file and any others related to it - we should follow up with a cleaner approach across the entire provider.

github/resource_github_organization_ruleset.go
nickfloyd: Oof good catch.

github/resource_github_organization_ruleset.go
stevehipwell: Naming is hard so GitHub likes to reuse names with slight contextual modifiers. The new function name is clearer in the context of the GitHub API as the old name duplicated the service context (organization) into the function name and didn't distinguish between ruleset types (which in fairness wasn't a problem when it was originally named). FYI I work on the SDK.

github/resource_github_organization_ruleset_test.go
nickfloyd: Side note.  Since we moved away from `interface{}` the logic to prop dig is kinda gnarly - perhaps there is a better way to do this:

```
	filePathRule := rulesMap["file_path_restriction"].([]any)[0].(map[string]any)
	expectedFilePaths := len(filePathRule["restricted_file_paths"].([]any))
```

stevehipwell: > Side note. Since we moved away from interface{} the logic to prop dig is kinda gnarly - perhaps there is a better way to do this:

Do you mean the use of `any` over `interface{}`?

github/resource_github_organization_ruleset.go
nickfloyd: Fair point.  I think I am/was being overly sensitive to the idea of breaking changes and really not thinking about log health and the DX around it. 

My thinking was that there are significant underlying implementation changes and often times the Terraform best practice is to bump the schema version even if the state structure remains the same. I guess in isolation this change would warrant a signal like this but perhaps it's just extra noise here.  I'll remove it.

github/resource_github_repository_collaborators.go
nickfloyd: Looks like I missed one during the rollback.. thx for the heads up.

github/resource_github_repository_collaborators.go
nickfloyd: Yeah, great catch... it was missed when I was making changes after your comment.

github/resource_github_organization_ruleset_test.go
nickfloyd: Yeah... I feel like there might be a better (cleaner) way to handle things like that - orthogonal to this change but I'd like to include it in our patterns handbook (when we write one 😉)

github/resource_github_repository_collaborators.go
nickfloyd: Yeah this is the curse of the polyglot.  Being the maintainer of all of the SDKs over the past 4 years I've had to jump between languages constantly - at least +5 by my count - changing hour by hour.  I sometimes loose context and do stupid things like this out of habit.  Thanks for calling me on it.   I'll try to be more diligent.

github/resource_github_team_repository.go
nickfloyd: Cleaned up!

github/resource_github_repository_ruleset.go
nickfloyd: Cleaned up!

github/resource_github_team.go
```diff
@@ -267,40 +269,41 @@ func resourceGithubTeamUpdate(d *schema.ResourceData, meta any) error {
 	}
 
 	client := meta.(*Owner).v3client
-	orgId := meta.(*Owner).id
```
deiga, line 270: nitpick: removing this looks inconsistent with the existing convention

github/resource_github_repository_collaborators.go
deiga: @nickfloyd I think this change might need cleanup?

github/resource_github_organization_ruleset.go
```diff
@@ -656,18 +662,19 @@ func resourceGithubOrganizationRulesetRead(d *schema.ResourceData, meta any) err
...
+	rulesetReq := resourceGithubRulesetObject(d, meta.(*Owner).name)
 
+	org := meta.(*Owner).name
 	rulesetID, err := strconv.ParseInt(d.Id(), 10, 64)
 	if err != nil {
 		return unconvertibleIdErr(d.Id(), err)
 	}
 
-	ctx := context.WithValue(context.Background(), ctxId, d.Id())
+	ctx := context.WithValue(context.Background(), ctxId, rulesetID)
```
stevehipwell, line 673: Why are we putting the ID in the context?

github/resource_github_repository_collaborators.go
stevehipwell: Yeah, still looks like it needs removing.

github/resource_github_branch_validation_test.go
```diff
@@ -20,6 +21,7 @@ func TestGithubBranchIsRecreatedWhenRepositoryChanges(t *testing.T) {
 	repositorySchema := resource.Schema["repository"]
 	if repositorySchema == nil {
 		t.Fatal("repository field should exist in schema")
+		return
```
diofeher, line 24: Why did you add these `return`s? `t.Fatal` is supposed to stop the execution of the test, so adding it is redundant.

github/resource_github_branch_validation_test.go
```diff
@@ -32,6 +34,7 @@ func TestGithubBranchIsRecreatedWhenSourceBranchChanges(t *testing.T) {
 	sourceBranchSchema := resource.Schema["source_branch"]
 	if sourceBranchSchema == nil {
 		t.Fatal("source_branch field should exist in schema")
+		return
```
diofeher, line 37:
```suggestion
```

github/resource_github_branch_validation_test.go
```diff
@@ -44,6 +47,7 @@ func TestGithubBranchIsRecreatedWhenSourceSHAChanges(t *testing.T) {
 	sourceSHASchema := resource.Schema["source_sha"]
 	if sourceSHASchema == nil {
 		t.Fatal("source_sha field should exist in schema")
+		return
```
diofeher, line 50:
```suggestion
```

github/resource_github_branch_validation_test.go
```diff
@@ -8,6 +8,7 @@ func TestGithubBranchIsUpdatedWhenBranchChanges(t *testing.T) {
 	branchSchema := resource.Schema["branch"]
 	if branchSchema == nil {
 		t.Fatal("branch field should exist in schema")
+		return
```
diofeher, line 11:
```suggestion
```

github/resource_github_organization_ruleset_test.go
```diff
@@ -597,22 +597,50 @@ func TestOrganizationPushRulesetSupport(t *testing.T) {
 	// Test expand functionality (organization rulesets use org=true)
 	expandedRules := expandRules(input, true)
 
-	if len(expandedRules) != 4 {
-		t.Fatalf("Expected 4 expanded rules for organization push ruleset, got %d", len(expandedRules))
+	if expandedRules == nil {
+		t.Fatal("Expected expanded rules to not be nil")
+		return
```
diofeher, line 602:
```suggestion
```

github/resource_github_repository_ruleset.go
```diff
@@ -705,15 +705,17 @@ func resourceGithubRepositoryRulesetUpdate(d *schema.ResourceData, meta any) err
...
-		ruleset, _, err = client.Repositories.UpdateRulesetNoBypassActor(ctx, owner, repoName, rulesetID, rulesetReq)
+		// Clear bypass actors first, then update with new ruleset
+		_, err = client.Repositories.UpdateRulesetClearBypassActor(ctx, owner, repoName, rulesetID)
+		if err != nil {
+			return err
+		}
+		ruleset, _, err = client.Repositories.UpdateRuleset(ctx, owner, repoName, rulesetID, *rulesetReq)
 	} else {
-		ruleset, _, err = client.Repositories.UpdateRuleset(ctx, owner, repoName, rulesetID, rulesetReq)
+		ruleset, _, err = client.Repositories.UpdateRuleset(ctx, owner, repoName, rulesetID, *rulesetReq)
```
diofeher: Can we consolidate that in a single line?

github/resource_github_team.go
```diff
@@ -154,8 +154,9 @@ func resourceGithubTeamCreate(d *schema.ResourceData, meta any) error {
 		on the parent team, the operation might still fail to set the parent team.
 	*/
 	if newTeam.ParentTeamID != nil && githubTeam.Parent == nil {
+		//nolint:staticcheck // SA1019: EditTeamByID is deprecated but still needed for legacy compatibility
 		_, _, err := client.Teams.EditTeamByID(ctx,
-			*githubTeam.Organization.ID,
+			meta.(*Owner).id,
```
diofeher, line 159: Just double-checking if this is right. Would it be possible to change teams in another organization?

github/respository_rules_utils.go
```diff
@@ -1,16 +1,42 @@
...
+func toInt(v any) int {
+	switch val := v.(type) {
+	case int:
+		return val
+	case float64:
+		return int(val)
+	case int64:
+		return int(val)
+	default:
+		return 0
```
diofeher, line 21: I'm not a huge fan of returning 0 when passing other types to this function, as it can turn later debugging errors into a nightmare.

github/respository_rules_utils.go
```diff
@@ -1,16 +1,42 @@
...
+	switch val := v.(type) {
+	case int:
+		return int64(val)
+	case int64:
+		return val
+	case float64:
+		return int64(val)
+	default:
+		return 0
+	}
```
diofeher, line 36: I'm not a huge fan of returning 0 when passing other types to this function, as it can turn later debugging errors into a nightmare. I usually prefer an explicit error or a panic

github/resource_github_branch_validation_test.go
nickfloyd: This is because of staticcheck linter - specifically the SA5011 rule about nil pointer dereferences.

Technically the code is unreachable and I didn't want to put another ignore in there.

github/resource_github_branch_validation_test.go
nickfloyd: See above

github/resource_github_branch_validation_test.go
nickfloyd: See above

github/resource_github_organization_ruleset_test.go
nickfloyd: See above

github/resource_github_repository_ruleset.go
nickfloyd: Nice and tidy!  Great catch!

github/resource_github_branch_validation_test.go
diofeher: That's strange. Is this raised by `make lint`?

I removed those lines here and ran the linter; it didn't find any issues.

github/resource_github_repository_ruleset.go
stevehipwell, line 721:
```suggestion
	if d.HasChange("bypass_actors") && len(rulesetReq.BypassActors) == 0 {
		// Clear bypass actors first, then update with new ruleset
		_, err := client.Repositories.UpdateRulesetClearBypassActor(ctx, owner, repoName, rulesetID)
		if err != nil {
			return err
		}
	}

	ruleset, _, err = client.Repositories.UpdateRuleset(ctx, owner, repoName, rulesetID, *rulesetReq)
	if err != nil {
		return err
	}
```
This pattern doesn't make much sense, it should look something like this.

github/resource_github_branch_validation_test.go
diofeher: From what I'm reading here: https://staticcheck.dev/docs/checks#SA5011, it seems `staticcheck` can catch this kind of error without introducing `return`. I'm just curious in which lint you're using to see this error to see if I could reproduce here.

github/resource_github_repository_ruleset.go
nickfloyd: > This pattern doesn't make much sense, it should look something like this.

That should've been handled in an earlier set of clean up.

github/resource_github_branch_validation_test.go
nickfloyd: Interesting... I am definitely running `make lint` - I'll need to see if I can track down what the variation might be.

stevehipwell: I was picking them up when I updated the linter, I think it's a golangci-lint issue and not directly a static check issue. Also it looks like these ones didn't require the return in my PR which seems slightly odd. There is a new linter version literally just released, so it might be worth giving it a try with that?

stevehipwell: I was picking them up when I updated the linter, I think it's a golangci-lint issue and not directly a static check issue. Also it looks like these ones didn't require the return in my PR which seems slightly odd. There is a new linter version literally just released, so it might be worth giving it a try with that?

stevehipwell: I was picking them up when I updated the linter, I think it's a golangci-lint issue and not directly a static check issue. Also it looks like these ones didn't require the return in my PR which seems slightly odd. There is a new linter version literally just released, so it might be worth giving it a try with that?

diofeher: @stevehipwell Thanks for the suggestion! 

Just tried with `2.7.0` (the latest release) and was still able to run the lint without errors after removing the returns.

github/resource_github_repository_ruleset.go
```diff
@@ -705,19 +705,21 @@ func resourceGithubRepositoryRulesetUpdate(d *schema.ResourceData, meta any) err
...
 	}
 
-	var ruleset *github.Ruleset
-	// Use UpdateRulesetNoBypassActor here instead of UpdateRuleset *if* bypass_actors has changed.
-	// UpdateRuleset uses `omitempty` on BypassActors, causing empty arrays to be omitted from the JSON.
-	// UpdateRulesetNoBypassActor always includes the field so that bypass actors can actually be removed.
-	// See: https://github.com/google/go-github/blob/b6248e6f6aec019e75ba2c8e189bfe89f36b7d01/github/repos_rules.go#L196
+	var ruleset *github.RepositoryRuleset
+
 	if d.HasChange("bypass_actors") {
```
stevehipwell, line 710:
```suggestion
	if d.HasChange("bypass_actors") && len(rulesetReq.BypassActors) == 0 {
```
The clear bypass actors should only need running if there are none left. This pattern likely needs copying to the other ruleset implementations.

stevehipwell (changes requested):

PR #2895, opened 2025-11-14
feat: Add github_organization_network_configuration resource

austenstone:
## Overview

This PR adds a new `github_organization_network_configuration` resource to manage GitHub organization network configurations for hosted compute services.

It also adds the missing linkage for GitHub-hosted private networking on `github_actions_runner_group` via `network_configuration_id`, so hosted runners can consume private networking through their assigned runner group.

## What this PR adds

- New resource: `github_organization_network_configuration`
- Full CRUD support for organization network configurations
- Import support for organization network configurations
- Support for `github_actions_runner_group.network_configuration_id`
- Acceptance coverage for:
  - create, update, and import flows for `github_organization_network_configuration`
  - create, update, import, and clear flows for runner group network configuration association
- Documentation updates for:
  - `github_organization_network_configuration`
  - `github_actions_hosted_runner`
- Example updates for hosted runners and public IP configuration

### `github_organization_network_configuration`

- `name` (Required) - Network configuration name (1-100 chars, alphanumeric plus `._-`)
- `compute_service` (Optional) - Hosted compute service: `none` or `actions` (default: `none`)
- `network_settings_ids` (Required) - Exactly one network settings ID
- `created_on` (Computed) - Creation timestamp
- `id` (Computed) - Resource ID

### `github_actions_runner_group`

- `network_configuration_id` (Optional) - The ID of the hosted compute network configuration associated with the runner group for GitHub-hosted private networking

## Notes

- This resource is organization-scoped and validates organization mode via `checkOrganization()`.
- GitHub private networking for GitHub-hosted runners is configured on the runner group, not directly on `github_actions_hosted_runner`.
- For Azure private networking, `network_settings_ids` must use the GitHub-facing network settings identifier (`GitHubId`), not the Azure ARM resource ID.
- Network configurations are available for GitHub Enterprise Cloud organizations.

## Validation

- `go test ./github -run TestAccGithubOrganizationNetworkConfiguration -count=1`
- `go test ./github -run TestAccGithubActionsRunnerGroup -count=1`
- `go test ./github -run '^$' -count=1`

### DIFF
--- a/github/resource_github_actions_runner_group.go
+++ b/github/resource_github_actions_runner_group.go
@@ -125,7 +125,7 @@ func getOrganizationRunnerGroup(client *github.Client, ctx context.Context, org 
 	runnerGroup, resp, err := client.Actions.GetOrganizationRunnerGroup(ctx, org, groupID)
 	if err != nil {
 		var ghErr *github.ErrorResponse
-		if errors.As(err, &ghErr) {
+		if errors.As(err, &ghErr) && ghErr.Response != nil && ghErr.Response.StatusCode == http.StatusNotModified {
 			// ignore error StatusNotModified
 			return runnerGroup, resp, nil
 		}
@@ -161,6 +161,47 @@ func setGithubActionsRunnerGroupNetworkingState(d *schema.ResourceData, runnerGr
 			return err
 		}
 	}
+	return nil
+}
+
+func setGithubActionsRunnerGroupState(d *schema.ResourceData, runnerGroup *github.RunnerGroup, etag string, selectedRepositoryIDs []int64) error {
+	if err := d.Set("etag", etag); err != nil {
+		return err
+	}
+	if err := d.Set("allows_public_repositories", runnerGroup.GetAllowsPublicRepositories()); err != nil {
+		return err
+	}
+	if err := d.Set("default", runnerGroup.GetDefault()); err != nil {
+		return err
+	}
+	if err := d.Set("id", strconv.FormatInt(runnerGroup.GetID(), 10)); err != nil {
+		return err
+	}
+	if err := d.Set("inherited", runnerGroup.GetInherited()); err != nil {
+		return err
+	}
+	if err := d.Set("name", runnerGroup.GetName()); err != nil {
+		return err
+	}
+	if err := d.Set("runners_url", runnerGroup.GetRunnersURL()); err != nil {
+		return err
+	}
+	if err := d.Set("selected_repositories_url", runnerGroup.GetSelectedRepositoriesURL()); err != nil {
+		return err
+	}
+	if err := d.Set("visibility", runnerGroup.GetVisibility()); err != nil {
+		return err
+	}
+	if err := d.Set("selected_repository_ids", selectedRepositoryIDs); err != nil {
+		return err
+	}
+	if err := d.Set("restricted_to_workflows", runnerGroup.GetRestrictedToWorkflows()); err != nil {
+		return err
+	}
+	if err := d.Set("selected_workflows", runnerGroup.SelectedWorkflows); err != nil {
+		return err
+	}
+
 	return nil
 }
 
@@ -216,52 +257,28 @@ func resourceGithubActionsRunnerGroupCreate(d *schema.ResourceData, meta any) er
 		return err
 	}
 	d.SetId(strconv.FormatInt(runnerGroup.GetID(), 10))
-	if err = d.Set("etag", resp.Header.Get("ETag")); err != nil {
-		return err
-	}
-	if err = d.Set("allows_public_repositories", runnerGroup.GetAllowsPublicRepositories()); err != nil {
-		return err
-	}
-	if err = d.Set("default", runnerGroup.GetDefault()); err != nil {
-		return err
-	}
-
-	if err = d.Set("id", strconv.FormatInt(runnerGroup.GetID(), 10)); err != nil {
-		return err
-	}
-	if err = d.Set("inherited", runnerGroup.GetInherited()); err != nil {
-		return err
-	}
-	if err = d.Set("name", runnerGroup.GetName()); err != nil {
-		return err
-	}
-	if err = d.Set("runners_url", runnerGroup.GetRunnersURL()); err != nil {
-		return err
-	}
-	if err = d.Set("selected_repositories_url", runnerGroup.GetSelectedRepositoriesURL()); err != nil {
-		return err
-	}
-	if err = d.Set("visibility", runnerGroup.GetVisibility()); err != nil {
-		return err
-	}
-	if err = d.Set("selected_repository_ids", selectedRepositoryIDs); err != nil { // Note: runnerGroup has no method to get selected repository IDs
-		return err
-	}
-	if err = d.Set("restricted_to_workflows", runnerGroup.GetRestrictedToWorkflows()); err != nil {
-		return err
-	}
-	if err = d.Set("selected_workflows", runnerGroup.SelectedWorkflows); err != nil {
+	if err = setGithubActionsRunnerGroupState(d, runnerGroup, resp.Header.Get("ETag"), selectedRepositoryIDs); err != nil {
 		return err
 	}
 
 	if networkConfigurationID, ok := d.GetOk("network_configuration_id"); ok {
 		networkConfigurationIDValue := networkConfigurationID.(string)
+		// The create endpoint does not accept network_configuration_id, so private networking
+		// must be attached with a follow-up PATCH after the runner group has been created.
 		if _, err = updateOrganizationRunnerGroupNetworking(client, ctx, orgName, runnerGroup.GetID(), &networkConfigurationIDValue); err != nil {
+			return err
+		}
+
+		if err = setGithubActionsRunnerGroupNetworkingState(d, &organizationRunnerGroup{NetworkConfigurationID: &networkConfigurationIDValue}); err != nil {
+			return err
+		}
+	} else {
+		if err = setGithubActionsRunnerGroupNetworkingState(d, nil); err != nil {
 			return err
 		}
 	}
 
-	return resourceGithubActionsRunnerGroupRead(d, meta)
+	return nil
 }
 
 func resourceGithubActionsRunnerGroupRead(d *schema.ResourceData, meta any) error {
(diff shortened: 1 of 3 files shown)


deiga:
@austenstone We've now upgraded `go-github` to latest. Could you rebase onto `main` now? :)

github/resource_github_organization_network_configuration.go
```diff
@@ -0,0 +1,193 @@
...
+)
+
+var organizationNetworkConfigurationNamePattern = regexp.MustCompile(`^[a-zA-Z0-9._-]+$`)
+
+func resourceGithubOrganizationNetworkConfiguration() *schema.Resource {
+	return &schema.Resource{
+		Create: resourceGithubOrganizationNetworkConfigurationCreate,
+		Read:   resourceGithubOrganizationNetworkConfigurationRead,
+		Update: resourceGithubOrganizationNetworkConfigurationUpdate,
+		Delete: resourceGithubOrganizationNetworkConfigurationDelete,
```
deiga: Please use Context-aware functions

github/resource_github_organization_network_configuration.go
```diff
@@ -0,0 +1,193 @@
...
+		Importer: &schema.ResourceImporter{
+			StateContext: schema.ImportStatePassthroughContext,
+		},
+
+		Schema: map[string]*schema.Schema{
```
deiga, line 31: Please add a top-level description

github/resource_github_organization_network_configuration.go
```diff
@@ -0,0 +1,193 @@
...
+	ctx := context.Background()
+
+	configuration, _, err := client.Organizations.CreateNetworkConfiguration(ctx, orgName, buildOrganizationNetworkConfigurationRequest(d))
+	if err != nil {
+		return err
+	}
+
+	d.SetId(configuration.GetID())
+
+	return resourceGithubOrganizationNetworkConfigurationRead(d, meta)
```
deiga: Please don't call any CRUD function directly. All fields need to be set directly in Create

github/resource_github_organization_network_configuration.go
```diff
@@ -0,0 +1,193 @@
...
+
+	client := meta.(*Owner).v3client
+	orgName := meta.(*Owner).name
+	networkConfigurationID := d.Id()
+	ctx := context.WithValue(context.Background(), ctxId, networkConfigurationID)
+
+	configuration, resp, err := client.Organizations.GetNetworkConfiguration(ctx, orgName, networkConfigurationID)
+	if err != nil {
+		if ghErr, ok := err.(*github.ErrorResponse); ok && ghErr.Response.StatusCode == http.StatusNotFound {
+			log.Printf("[WARN] Removing organization network configuration %s from state because it no longer exists in GitHub", networkConfigurationID)
```
deiga: Please use tflog

github/resource_github_organization_network_configuration.go
```diff
@@ -0,0 +1,193 @@
...
+		}
+
+		return err
+	}
+
+	if resp != nil && resp.StatusCode == http.StatusNotModified {
+		return nil
+	}
+
+	return setOrganizationNetworkConfigurationState(d, configuration)
```
deiga: Please inline this

github/resource_github_organization_network_configuration_test.go
```diff
@@ -0,0 +1,116 @@
...
+
+		config := testAccOrganizationNetworkConfigurationConfig(configurationName, "actions", networkSettingsID)
+
+		resource.Test(t, resource.TestCase{
+			PreCheck:          func() { skipUnlessHasPaidOrgs(t) },
+			ProviderFactories: providerFactories,
+			Steps: []resource.TestStep{
+				{
+					Config: config,
+					Check: resource.ComposeTestCheckFunc(
```
deiga: Please replace Check with ConfigStsteChecks

deiga (changes requested):

austenstone:
Copilot note 🤖

I ran this against a real `octodemo` setup using a locally built provider from this PR branch and was able to validate the `github_organization_network_configuration` resource end-to-end:

- create ✅
- update ✅
- import ✅
- delete ✅

One important gotcha from the live test:

For **org-scoped** `github_organization_network_configuration`, the backing Azure `GitHub.Network/networkSettings` resource needs to be created with the **organization** `databaseId`, not the enterprise `databaseId`.

If the Azure `networkSettings` resource is created with the enterprise-backed `businessId`, the org create call fails with:

`422 The business ID is invalid or does not match the configuration's.`

Once I recreated the Azure `networkSettings` resource using the **organization** `databaseId`, the resource worked correctly and the network configuration showed up in the org settings UI.

So in practice the distinction seems to be:

- enterprise-level network configuration flow -> enterprise-backed Azure `networkSettings`
- organization-level network configuration flow -> organization-backed Azure `networkSettings`

Leaving this here since it seems like exactly the kind of thing that can make acceptance/manual testing look broken when the provider behavior is actually fine.

austenstone:
Thanks — I addressed the review feedback and pushed follow-up changes.

Summary of updates:

- switched the resource to context-aware CRUD functions
- added the top-level resource description
- updated `Create` to set state directly instead of calling another CRUD function
- replaced logging with `tflog`
- aligned acceptance tests to use `ConfigStateChecks`
- added docs/error-handling guidance for the Azure network settings scope gotcha that came up in live validation

I also re-tested the resource end-to-end against a real org/Azure-backed setup after the refactor.

Would appreciate another pass when you have a chance.

austenstone:
Healthy from my side ✅

I re-tested the full flow and everything is behaving as expected: create/update/destroy all work cleanly, CI is green, and the acceptance coverage passes. I also validated the real-world integration path with org network configurations, runner group association, and hosted runner private networking.

github/resource_github_actions_runner_group.go
```diff
@@ -92,10 +96,74 @@ func resourceGithubActionsRunnerGroup() *schema.Resource {
...
+
+	return &runnerGroup, resp, nil
+}
+
+func getOrganizationRunnerGroup(client *github.Client, ctx context.Context, org string, groupID int64) (*github.RunnerGroup, *github.Response, error) {
+	runnerGroup, resp, err := client.Actions.GetOrganizationRunnerGroup(ctx, org, groupID)
+	if err != nil {
+		var ghErr *github.ErrorResponse
+		if errors.As(err, &ghErr) {
+			// ignore error StatusNotModified
```
deiga, line 129: question: this doesn't seem to check the error type, how are you sure it's only ignoring that specific error?

github/resource_github_actions_runner_group.go
```diff
@@ -186,19 +254,14 @@ func resourceGithubActionsRunnerGroupCreate(d *schema.ResourceData, meta any) er
...
-			return runnerGroup, resp, nil
+	if networkConfigurationID, ok := d.GetOk("network_configuration_id"); ok {
+		networkConfigurationIDValue := networkConfigurationID.(string)
+		if _, err = updateOrganizationRunnerGroupNetworking(client, ctx, orgName, runnerGroup.GetID(), &networkConfigurationIDValue); err != nil {
+			return err
 		}
 	}
-	return runnerGroup, resp, err
+
+	return resourceGithubActionsRunnerGroupRead(d, meta)
```
deiga: Issue: nothing should call any of the CRUD functions. If computed values need to be set it needs to happen directly in here

github/resource_github_actions_runner_group.go
```diff
@@ -186,19 +254,14 @@ func resourceGithubActionsRunnerGroupCreate(d *schema.ResourceData, meta any) er
...
-func getOrganizationRunnerGroup(client *github.Client, ctx context.Context, org string, groupID int64) (*github.RunnerGroup, *github.Response, error) {
-	runnerGroup, resp, err := client.Actions.GetOrganizationRunnerGroup(ctx, org, groupID)
-	if err != nil {
-		var ghErr *github.ErrorResponse
-		if errors.As(err, &ghErr) {
-			// ignore error StatusNotModified
-			return runnerGroup, resp, nil
+	if networkConfigurationID, ok := d.GetOk("network_configuration_id"); ok {
+		networkConfigurationIDValue := networkConfigurationID.(string)
+		if _, err = updateOrganizationRunnerGroupNetworking(client, ctx, orgName, runnerGroup.GetID(), &networkConfigurationIDValue); err != nil {
```
deiga, line 268: This is the Create function, if it needs to call multiple API endpoints please add comments explaining the logic

github/resource_github_actions_runner_group_test.go
```diff
@@ -97,6 +120,115 @@ func TestAccGithubActionsRunnerGroup(t *testing.T) {
...
+			}
+		`, networkConfigurationName, networkSettingsID, runnerGroupName)
+
+		resource.Test(t, resource.TestCase{
+			PreCheck:          func() { skipUnlessHasPaidOrgs(t) },
+			ProviderFactories: providerFactories,
+			Steps: []resource.TestStep{
+				{
+					Config: configWithoutNetworkConfiguration,
+					Check: resource.ComposeTestCheckFunc(
```
deiga: Please replace with ConfigStsteChecks

github/resource_github_actions_runner_group.go
austenstone: Addressed — `Create` no longer calls `Read`. I moved the shared attribute population into `setGithubActionsRunnerGroupState(...)` and now set the computed fields directly in `Create` instead of chaining another CRUD function.

github/resource_github_actions_runner_group.go
austenstone: Good catch — I tightened this so it only suppresses the error when the response is actually `http.StatusNotModified`. Other `*github.ErrorResponse` cases now propagate normally instead of being implicitly ignored.

github/resource_github_actions_runner_group_test.go
austenstone: Updated the new private-networking acceptance steps to use `ConfigStateChecks`. The cross-resource assertion now uses `statecheck.CompareValuePairs(...)`, and I re-ran `go test ./github -run TestAccGithubActionsRunnerGroup -count=1` after the change.

github/resource_github_actions_runner_group.go
austenstone: Added a clarifying comment here. The create endpoint does not accept `network_configuration_id`, so the runner group has to be created first and then patched in a follow-up call to attach the private networking association.

austenstone:
I opened a separate pull request for the Enterprise level. We can either stack these PRS or just merge the Enterprise one.

austenstone:
closing in favor of #3274

Issue #2077, opened 2024-01-04
[BUG]: 5.43 upgrade failing due to signoff issues


mattklein123:
### Expected Behavior

Just got the 5.43 upgrade and I'm seeing failures like: 

```
Error: PATCH https://api.github.com/repos/***: 422 Commit signoff is enforced by the organization and cannot be disabled []
```

I assume this is somehow due to https://github.com/integrations/terraform-provider-github/pull/2007/files. I've tried manually adding `web_commit_signoff_required` but this still seems to fail:

```
 # module.***.github_repository.repo will be updated in-place
  ~ resource "github_repository" "repo" {
      + web_commit_signoff_required = true
        # (35 unchanged attributes hidden)
    }
```

Is this related to something else or is there something else I should be doing here? Thank you.

### Actual Behavior

Doesn't fail / backwards compatible. 

### Terraform Version

1.2.6, 5.43

### Affected Resource(s)

- github_repository (I think)

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

nnellanspdl:
I know it's not a fix, but ignoring that field might be a workaround for now?

```terraform
resource "github_repository" "repo" {
  lifecycle {
    ignore_changes = [
      web_commit_signoff_required
    ]
  }
}
```

nyanhp:
> I know it's not a fix, but ignoring that field might be a workaround for now?
> 
> ```terraform
> resource "github_repository" "repo" {
>   lifecycle {
>     ignore_changes = [
>       web_commit_signoff_required
>     ]
>   }
> }
> ```

Unfortunately, this does not seem to help either. The error also occurs on a completely new deployment, as soon as the repository resource is applied.

brunooon:
> > I know it's not a fix, but ignoring that field might be a workaround for now?
> > ```terraform
> > resource "github_repository" "repo" {
> >   lifecycle {
> >     ignore_changes = [
> >       web_commit_signoff_required
> >     ]
> >   }
> > }
> > ```
> 
> Unfortunately, this does not seem to help either. The error also occurs on a completely new deployment, as soon as the repository resource is applied.

I was still facing the same issue, so I had to remove the commit signoff from the org and I'm setting to true only on the repos now, its working so far.

mrclrchtr:
Same here... 😕

mvarchdev:
Same here, using pulumi/github

mvarchdev:
@iwahbe FYI

[...]

PR #2763, opened 2025-09-30
fix: commit signoff is enforced by the organization

steveteuber:
Recreate the pull request #2222 to resolve issue #2077.

----

### Before the change?

* There is a bug in the GitHub API 2022-11-28 version that causes a 422 error when updating a repository within an organization that has "Require sign off on web-based commits" enabled.

### After the change?

* On a repository update, we check if the organization has "Require sign off on web-based commits" enabled.
If it's enabled, we do not send the `web_commit_signoff_required` field in the update request to avoid the 422 error.

### Pull request checklist
- [x] Tests for the changes have been added (for bug fixes / features)
- [x] Docs have been reviewed and added / updated if needed (for bug fixes / features)

### Does this introduce a breaking change?

Please see our docs on [breaking changes](https://github.com/octokit/.github/blob/master/community/breaking_changes.md) to help!

- [x] No

----

PR #2896, opened 2025-11-14
fix: check if organization and requires sign off commits

steveteuber:
This is a follow-up pull request to fix the issue described in https://github.com/integrations/terraform-provider-github/issues/2077.

----

### Before the change?

There is a bug in the GitHub API 2022-11-28 version that causes a 422 error when updating a repository within an organization that has "Require sign off on web-based commits" enabled.

### After the change?

On a repository update, we check if the organization has "Require sign off on web-based commits" enabled.
If it's enabled, we do not send the web_commit_signoff_required field in the update request to avoid the 422 error.

### Pull request checklist

- [ ] Yes
- [x] No

----


ArlonAntonius:
Hi @steveteuber,

I was reading into this issue and was wondering the following:
At this moment only changes to the update method have been provided, does that mean that this issue does not seem to happen on the creation of a GitHub repository?

steveteuber:
@ArlonAntonius Yes, correct. After #2763 was merged, this issue now occurs only when updating a repository.

github/resource_github_repository.go
```diff
@@ -867,18 +867,11 @@ func resourceGithubRepositoryUpdate(d *schema.ResourceData, meta any) error {
...
-		organization, _, err := client.Organizations.Get(ctx, owner)
-		if err == nil {
-			if organization != nil && organization.GetWebCommitSignoffRequired() {
-				repoReq.WebCommitSignoffRequired = nil
-			}
-		}
+	// When the organization has "Require contributors to sign off on web-based commits" enabled,
+	// the API doesn't allow you to send `web_commit_signoff_required` or it returns a 422 error.
+	// As a workaround, check if the organization requires it, and if so, remove it from the request.
+	if meta.(*Owner).IsOrganization && meta.(*Owner).IsWebCommitSignoffRequired {
```
nickfloyd, line 951: I think removing the `d.HasChange("web_commit_signoff_required")` conditional might be problematic.  This change might cause issues when archiving repositories in organizations that don't have signoff required but the repository itself has it set, potentially causing a 422.  Check me on that though.

github/resource_github_repository.go
```diff
@@ -982,8 +975,9 @@ func resourceGithubRepositoryDelete(d *schema.ResourceData, meta any) error {
 				return err
 			}
 			repoReq := resourceGithubRepositoryObject(d)
-			// Always remove `web_commit_signoff_required` when archiving, to avoid 422 error
-			repoReq.WebCommitSignoffRequired = nil
+			if meta.(*Owner).IsOrganization && meta.(*Owner).IsWebCommitSignoffRequired {
```
nickfloyd, line 1037: Same as above.

github/resource_github_repository.go
steveteuber: From my point of view the 422 error only occurs when the organization has "Require sign off on web-based commits" enabled. If the organization does not have this setting enabled, then sending the `web_commit_signoff_required` field should not cause any issues, even if the repository itself has it set.

steveteuber:
@nickfloyd Did you read my comment in the discussion above? It seems like you might have missed it.

IaroslavSapak:
currently I have to play around lifecycle{ ignore_changes = all} or reimporting updated repo. So this PR would greatly imrove workflow. So it would be great go with proposed patch instead of waiting v7

deiga:
@steveteuber 
We've been discussing internally, that modifying the `Owner` struct isn't something we'd like to do. And we want to move the provider to be more explicit instead of implicit, the proposed solution doesn't look like one we want to adopt.

And for the changes to `github/resource_github_repository.go` I'm having trouble understading how it changes any behaviour. There seem to be a few more cases where we send the field, but that doesn't sound like it would fix the issue you're reporting.

Could you add a testcase which fails with the current code and works with your code?

Edit: Once you have a testcase, could you change the logic to use `d.GetOkExists` so that the field is unset, if it's unset in the TF config and otherwise the value of the TF config?

steveteuber:
@deiga Thanks for the feedback. Regarding the changes to `github/resource_github_repository.go`, the main change is to check if the owner is an organization and has "Require sign off on web-based commits" enabled before sending the `web_commit_signoff_required` field in the update request.

To clarify the behavior, here are some scenarios:

1. Owner is an individual user account:
   The `web_commit_signoff_required` field can be sent without any issues.

2. Owner is an organization:
   The `web_commit_signoff_required` field can be sent without any issues.

3. Owner is an organization and has "Require sign off on web-based commits" enabled:
   The `web_commit_signoff_required` field cannot be sent, otherwise it will cause a 422 error. To avoid this, check if the owner is an organization and has "Require sign off on web-based commits" enabled. When this is the case, do not send the `web_commit_signoff_required` field in the update request.

steveteuber:
@deiga I've created a new PR #3165 for an alternative approach that only sets the `web_commit_signoff_required` field if it's explicitly configured in the repository configuration.

I'll close this PR here because the alternative approach doesn't modifying the Owner struct, and it's more in line with making the provider more explicit.

PR #2902, opened 2025-11-14
Integrate v6.8.3 into main

nickfloyd:
This integrates the changes from the release patch v6.8.2 into main. 

Note: this also encapsulates the patch from v6.7.5 as well that we're missed after the release was made.

----

### DIFF
--- a/github/resource_github_organization_ruleset_test.go
+++ b/github/resource_github_organization_ruleset_test.go
@@ -639,7 +639,7 @@ func TestOrganizationPushRulesetSupport(t *testing.T) {
 		t.Fatalf("Expected 1 max_file_size rule, got %d", len(maxFileSizeRules))
 	}
 	if maxFileSizeRules[0]["max_file_size"] != int64(100) {
-		t.Errorf("Expected max_file_size to be 104857600, got %v", maxFileSizeRules[0]["max_file_size"])
+		t.Errorf("Expected max_file_size to be 100, got %v", maxFileSizeRules[0]["max_file_size"])
 	}
 
 	// Verify max_file_path_length
--- a/github/respository_rules_utils_test.go
+++ b/github/respository_rules_utils_test.go
@@ -302,8 +302,8 @@ func TestMaxFilePathLengthWithOtherRules(t *testing.T) {
 	}
 
 	maxFileSizeRules := flattenedRulesMap["max_file_size"].([]map[string]any)
-	if len(maxFileSizeRules) != 1 || maxFileSizeRules[0]["max_file_size"] != int64(1048576) {
-		t.Error("Expected max_file_size rule with value 1048576")
+	if len(maxFileSizeRules) != 1 || maxFileSizeRules[0]["max_file_size"] != int64(1) {
+		t.Error("Expected max_file_size rule with value 1")
 	}
 }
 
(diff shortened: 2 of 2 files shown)


github/resource_github_actions_organization_secret.go
```diff
@@ -72,6 +71,7 @@ func resourceGithubActionsOrganizationSecret() *schema.Resource {
 				},
 				Set:         schema.HashInt,
 				Optional:    true,
+				ForceNew:    true,
```
stevehipwell, line 74: I don't think this is the right approach, both because it's [not necessary](https://docs.github.com/en/enterprise-cloud@latest/rest/actions/secrets?apiVersion=2022-11-28#set-selected-repositories-for-an-organization-secret) and because it takes the provider further away from the current idiomatic pattern for secrets.

github/resource_github_organization_ruleset_test.go
```diff
@@ -576,7 +577,7 @@ func TestOrganizationPushRulesetSupport(t *testing.T) {
 		},
 		"max_file_size": []any{
 			map[string]any{
-				"max_file_size": float64(10485760), // 10MB
+				"max_file_size": 10485760, // 10MB
```
stevehipwell: This doesn't look right, the [GitHub API](https://docs.github.com/en/rest/repos/rules?apiVersion=2022-11-28#create-a-repository-ruleset) specifies that the max size is in megabytes and supports a value between 1 & 100.

github/respository_rules_utils_test.go
```diff
@@ -349,7 +350,7 @@ func TestCompletePushRulesetSupport(t *testing.T) {
 		},
 		"max_file_size": []any{
 			map[string]any{
-				"max_file_size": float64(5242880), // 5MB
+				"max_file_size": 5242880, // 5MB
```
stevehipwell: Same as above.

github/resource_github_actions_organization_secret.go
nickfloyd: Now that this is working in the v6.8.3 patch.  I am hesitant to change anything else about this unless it is causing pain or problems.  I'd like to take a bit more time to look over our entire implementation and focus on your idea [here](https://github.com/integrations/terraform-provider-github/issues/2913).  Let me know what you think.

github/respository_rules_utils_test.go
stevehipwell:
```suggestion
				"max_file_size": 5, // 5MB
```

github/respository_rules_utils_test.go
```diff
@@ -255,7 +256,7 @@ func TestMaxFilePathLengthWithOtherRules(t *testing.T) {
 		},
 		"max_file_size": []any{
 			map[string]any{
-				"max_file_size": float64(1048576), // 1MB
+				"max_file_size": 1048576, // 1MB
```
stevehipwell:
```suggestion
				"max_file_size": 1, // 1MB
```

github/resource_github_organization_ruleset_test.go
```diff
@@ -637,8 +638,8 @@ func TestOrganizationPushRulesetSupport(t *testing.T) {
 	if len(maxFileSizeRules) != 1 {
 		t.Fatalf("Expected 1 max_file_size rule, got %d", len(maxFileSizeRules))
 	}
-	if maxFileSizeRules[0]["max_file_size"] != int64(10485760) {
-		t.Errorf("Expected max_file_size to be 10485760, got %v", maxFileSizeRules[0]["max_file_size"])
+	if maxFileSizeRules[0]["max_file_size"] != int64(100) {
+		t.Errorf("Expected max_file_size to be 104857600, got %v", maxFileSizeRules[0]["max_file_size"])
```
stevehipwell:
```suggestion
		t.Errorf("Expected max_file_size to be 100, got %v", maxFileSizeRules[0]["max_file_size"])
```

github/respository_rules_utils_test.go
stevehipwell:
```suggestion
if maxFileSizeRules[0]["max_file_size"] != int64(5) {
t.Errorf("Expected max_file_size to be 5, got %v", maxFileSizeRules[0]["max_file_size"])
}
```

stevehipwell:
```suggestion
if len(maxFileSizeRules) != 1 || maxFileSizeRules[0]["max_file_size"] != int64(1) {
t.Error("Expected max_file_size rule with value 1")
}
```

PR #2904, opened 2025-11-15
chore(actions): Add release stack and extra validation test on Release (tag event)

### LINKED ISSUE #2856: [MAINT]: Add a CI workflow that runs terraform configuration validation on a generic config to ensure no regressions happen
### Describe the need

In light of both #2843 and #2855 and the v6.7.x releases we need a way to continually validate that fundamental things such as secret destroy on drift workflows and other fundamental mechanisms of how the provider initializes don't regress. 

ex.
```
 # Build validation
      - name: Build Provider
        run: |
          go build -o terraform-provider-github

      # Terraform validation with actual binary
      - name: Terraform Init and Validate
        run: |
          # Create test configuration
          cat > test.tf << 'EOF'
          terraform {
            required_providers {
              github = {
              }
            }
          }
          
          provider "github" {
            token = "fake_token_for_validation"
          }
          
          # Test both resources with different configurations
          resource "github_actions_secret" "test" {
            repository       = "test_repo"
            secret_name      = "test_secret"
            plaintext_value  = "test_value"
            destroy_on_drift = true
          }
          
          resource "github_actions_organization_secret" "test" {
            secret_name     = "org_secret"
            encrypted_value = "dGVzdA=="
            visibility      = "private"
            destroy_on_drift = false
          }
          EOF
          
          # Setup dev overrides
          mkdir -p ~/.terraformrc
          cat > ~/.terraformrc << 'EOF'
          provider_installation {
            dev_overrides {
              "integrations/github" = "$(pwd)"
            }
            direct {}
          }
          EOF
          
          # Test with terraform (this should NOT fail with SDK validation errors)
          terraform init || echo "Init failed - expected for dev overrides"
          terraform validate
          
          # Clean up
          rm -f test.tf ~/.terraformrc terraform-provider-github
```

### SDK Version

_No response_

### API Version

_No response_

### Relevant log output

```shell

```

### Code of Conduct

ViacheslavKudinov:
Resolves #2856

----

### Before the change?

* ❌ validation doesn't run on exactly release/tag commit at this moment

### After the change?

* ✅ add extra job on "Release" workflow to run validation (own Terraform stack) before release happens
* ⚙️ add separate Terraform stack (for now only provider) which makes sense to run on "Release" 
* ✔️ result of the run of new version of workflow can be seen [there](https://github.com/vk-or/terraform-provider-github/actions/runs/19391362778/job/55485271596) 

### Pull request checklist

- [ ] Yes
- [x] No

----

⚠️ PAY ATTENTION: If validation on release workflow fails it means "tag" most likely has to be deleted or moved to the fixed commit as provider is "broken" ! ⚠️

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -14,7 +14,63 @@ permissions:
   contents: write
 
 jobs:
+  pre-release-tests:
+    name: Run tests before release
+    runs-on: ubuntu-latest
+    env:
+      test_stacks_directory: test_tf_stacks # root directory for test stacks
+      pre_release_tests: provider_only # directory name for pre-release tests
+    permissions:
+      contents: read
+      
+    steps:
+      - name: Checkout
+        uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # v5.0.0
+
+      - name: Setup Go
+        uses: actions/setup-go@44694675825211faa026b3c33043df3e48a5fa00 # v6.0.0
+        with:
+          go-version-file: go.mod
+          cache: true
+
+      - name: Build provider
+        run: go build -o terraform-provider-github
+
+      - name: Setup dev overrides
+        run: |
+          ROOT_DIR=$(pwd)
+          cat > ~/.terraformrc << EOF
+          provider_installation {
+            dev_overrides {
+              "integrations/github" = "${ROOT_DIR}"
+            }
+            direct {}
+          }
+          EOF
+
+      - name: Verify dev overrides setup
+        run: cat ~/.terraformrc
+
+      - name: Setup Terraform
+        uses: hashicorp/setup-terraform@b9cd54a3c349d3f38e8881555d616ced269862dd # v3.1.2
+        with:
+          terraform_version: 1.x
+
+      - name: Check Terraform version
+        run: terraform version
+
+      - name: Terraform init
+        continue-on-error: true # continue even if init fails
+        run: terraform -chdir=./${{ env.test_stacks_directory }}/${{ env.pre_release_tests }} init
+
+      - name: Terraform validate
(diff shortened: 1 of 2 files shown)


ViacheslavKudinov:
@nick Please, feel free to suggest any improvements or other way how we can catch issues on actual release artifact.

ViacheslavKudinov:
Hi @nickfloyd 
Our CI without this extra validation will not catch broken provider if a new tag is not based on main branch,  like the last release.
It may  help (at least my hope) to detect provider doesn't pass validation before new release will be done.

nickfloyd:
> Hi @nickfloyd Our CI without this extra validation will not catch broken provider if a new tag is not based on main branch, like the last release. It may help (at least my hope) to detect provider doesn't pass validation before new release will be done.

Yeah (thank you for making this change), catching it before attempting releasing is the ideal - but with patches it has made things messy.

ViacheslavKudinov:
You are welcome and I really hope it may help to prevent to happen releases when this type of validation didn't pass.

We are as organization rely on this provider and I am interested to make my best to make it more stable. 

I appreciate your work and happy to see that development is happening again.
There are some gaps and we are behind in such things like the latest go github client, Terraform framework and other aspects.

ViacheslavKudinov:
I see its actually even more tricky.
Our workflows are run from these patch branches and if new versions of workflows aren't merged it will run in context of this branch.

There is a need to get new release workflow in patch branch or make it to use reusable workflow from @main branch. 
I'm afraid it will complicate CI if we start to introduce reusable workflows, but it may help for some use cases when patch branch is behind main. But still at least it should use reusable workflows to fetch new code of workflow from main.

ViacheslavKudinov:
@nickfloyd Maybe we need to add something to run tests on release event to do some regression when release has happened and it wasn't yet synced from hashicorp registry site.

PR #2931, opened 2025-11-21
Add values_editable_by support to github_organization_custom_properties

### LINKED ISSUE #2930: [FEAT]: Add 'Allow repository actors to set this property' support for github_organization_custom_properties
### Describe the need

For org-level custom properties (`github_organization_custom_properties`), one can configure `Allow repository actors to set this property` to be Enabled or Disabled on the UI. The default is disabled, which means only org-owners can change these properties on an individual repository. Regular repository admins can't.

Currently this can not be configured with the terraform provider. This is a request to make it configurable. 

The library go-github [supports this field](https://github.com/google/go-github/blob/903265bfcfe4c499bec1973f7075b8a60503c19b/github/orgs_properties.go#L36) as `ValuesEditableBy` which was added in https://github.com/google/go-github/pull/3164 and included in the [v63 release](https://github.com/google/go-github/pull/3205).

### SDK Version

_No response_

### API Version

_No response_

### Relevant log output

```shell

```

### Code of Conduct

markszabo:
Resolves #2930

----

### Before the change?

The `github_organization_custom_properties` resource does not support the `values_editable_by` field. On the UI this is called `Allow repository actors to set this property` and can be set to either `Enabled` or `Disabled`. [In the API](https://docs.github.com/en/rest/orgs/custom-properties?apiVersion=2022-11-28#create-or-update-custom-properties-for-an-organization) (and [go-github](https://github.com/google/go-github/pull/3164)) it is called `values_editable_by` and can be set to `org_actors` (default), `org_and_repo_actors`.

### After the change?

The `github_organization_custom_properties` resource now supports the `values_editable_by` field with values `org_actors` (default) or `org_and_repo_actors`.

> * `values_editable_by` - (Optional) Who can edit the values of the custom property. Can be one of `org_actors` or `org_and_repo_actors`. When set to `org_actors` (the default), only organization owners can edit the property values on repositories. When set to `org_and_repo_actors`, both organization owners and repository administrators with the custom properties permission can edit the values.

### Pull request checklist

- [ ] Yes
- [x] No

This adds a new, optional field who's default value matches the current behavior.


sheplu:
I was looking to add the same PR to fix this missing property on the provider!  
+1

github/resource_github_organization_custom_properties.go
```diff
@@ -79,14 +85,22 @@ func resourceGithubCustomPropertiesCreate(d *schema.ResourceData, meta any) erro
...
 		Description:   &description,
 		AllowedValues: allowedValuesString,
-	})
+	}
+
+	// Set ValuesEditableBy if provided
+	if valuesEditableBy, ok := d.GetOk("values_editable_by"); ok {
+		valuesEditableByStr := valuesEditableBy.(string)
+		customProperty.ValuesEditableBy = &valuesEditableByStr
+	}
```
deiga: question: What caused you to choose to follow a different pattern than what is used for the other optional properties?

markszabo: I wanted to ensure that the field missing and the field being set to an empty string are handled differently, but looking at the existing fields now, it doesn't look like this is necessary. Let me rewrite this to match the rest of the codebase.

markszabo:
@deiga thank you for the review, I made the requested changes. Let me know what you think :)

deiga (approved):
Amazing, Thank you!

nickfloyd:
> @nickfloyd are you running local tests on PRs before merging them?

Always.  It's a bit of a slog, but it's currently the best approach for verification that we have.

stevehipwell:
> Always. It's a bit of a slog, but it's currently the best approach for verification that I we have.

Hopefully #2946 will be able to make this less of a challenge.

PR #2487, opened 2024-11-29
feat: Added organization role support

stevehipwell:
Resolves #2314
Related #2224

----

### Before the change?

* No support for organization org roles

### After the change?

* Added support for organization or roles
* Data sources
  * `github_organization_role`
  * `github_organization_role_teams`
  * `github_organization_users`
  * `github_organization_roles`
* Resources
  * `github_organization_role`
  * `github_organization_role_team`
  * `github_organization_role_user`

### Pull request checklist
- [ ] Tests for the changes have been added (for bug fixes / features)
- [ ] Docs have been reviewed and added / updated if needed (for bug fixes / features)

### Does this introduce a breaking change?

Please see our docs on [breaking changes](https://github.com/octokit/.github/blob/master/community/breaking_changes.md) to help!

- [ ] Yes
- [x] No

----

PR #2933, opened 2025-11-21
chore: Deprecate duplicate org team role resources

stevehipwell:
Resolves #ISSUE_NUMBER

----

### Before the change?

There are multiple resources for managing organization role teams and the deprecated security managers API is still in use.

### After the change?

I've deprecated the duplicate resources and the data source using the deprecated API.

### Pull request checklist

- [ ] Yes
- [x] No

----


github/data_source_github_organization_security_managers.go
```diff
@@ -9,6 +9,8 @@ import (
 
 func dataSourceGithubOrganizationSecurityManagers() *schema.Resource {
 	return &schema.Resource{
+		DeprecationMessage: "This data source is deprecated in favour of using the github_organization_role_team data source.",
```
deiga: There is only a plural data source 
```suggestion
		DeprecationMessage: "This data source is deprecated in favour of using the github_organization_role_teams data source.",
```

github/resource_github_organization_role_team_assignment.go
```diff
@@ -11,6 +11,8 @@ import (
 
 func resourceGithubOrganizationRoleTeamAssignment() *schema.Resource {
 	return &schema.Resource{
+		DeprecationMessage: "This resource is deprecated in favor of the github_organization_role_team resource.",
```
deiga, line 14: question: Do we have a naming convention for these kinds of resources? I personally think that `github_organization_role_team_assignment` is clearer than `github_organization_role_team`, especially since GitHub uses the "Role assignment" terminology

And a second note: The docs for https://registry.terraform.io/providers/integrations/github/latest/docs/resources/organization_role_team_assignment are more verbose and user-friendly than https://registry.terraform.io/providers/integrations/github/latest/docs/resources/organization_role_team

So, if we want to deprecate `github_organization_role_team_assignment` can we copy over the docs from it? :)

deiga (changes requested):

github/data_source_github_organization_security_managers.go
stevehipwell: Good spot.

stevehipwell:
@nickfloyd do we have a v6 release branch (e.g. `release-v6`) yet? I've been opening v6 PRs onto `main` with the expectation that they'd be cherry picked into the release branch, which I think is right for changes like this.

github/resource_github_organization_role_team_assignment.go
ViacheslavKudinov: Naming for the resources and data is a good thing to have.

There is for example another thing which may need to be addressed as well due to naming.

The data and resource for organization custom properties are named in plural `properties`, but neither data not resource works with more than a single property.

The resource [organization_custom_properties](https://registry.terraform.io/providers/integrations/github/latest/docs/resources/organization_custom_properties) and  data [organization_custom_properties](https://registry.terraform.io/providers/integrations/github/latest/docs/data-sources/organization_custom_properties)

It means it most likely needs to be done depreciation of these "plurals" and introducing of "singulars" with `property` in the names. 

It could be that "singular" named resources should be introduced in `v6` and  "plurals" removed in `v7`.

I saw @mroyme done naming of repsource with [singular property](https://registry.terraform.io/providers/nexthink-oss/github/latest/docs/resources/organization_custom_property) in their fork, which is more correct if it adds only a single custom property.

PS Added the issue https://github.com/integrations/terraform-provider-github/issues/2936

stevehipwell: @deiga the resource `github_organization_role_team` was implemented with all of the other related resources and data sources using consistent naming and implementation patterns.

I also don't agree that the assignment suffix is necessarily an improvement to the name, it adds additional characters to parse and fit on a single line. In this context the point is consistency with all of the other related resources. If you want to discuss changing the naming conventions I think that's a separate issue.

@ViacheslavKudinov I agree with your points on pluralization and duplication of context in names. I'm not sure which way you're arguing here but in this context my interpretation of your logic would be that the assignment suffix isn't required as the context is already complete.

ViacheslavKudinov: i have no strong opinion about this concrete resource, tbh.
 
PS At the same time as it was introduced Enterprise Roles  and  (Custom) Enterprise Role can be assigned to Enterprise team(s), it would be great if we use same "naming" for those as well to keep consistency, when we introduce new data/resources.

stevehipwell: > PS At the same time as it was introduced Enterprise Roles and (Custom) Enterprise Role can be assigned to Enterprise team(s), it would be great if we use same "naming" for those as well to keep consistency, when we introduce new data/resources.

FYI the PR (#2487) was opened almost a year ago, long before enterprise roles were a thing.

ViacheslavKudinov: @stevehipwell thanks for sharing. 

Now it looks like (from my perspective) is better to deprecate `organization_role_team_assignment`, exactly what PR addresses (it was introduced a little bit more than a month ago in [v6.7.0](https://github.com/integrations/terraform-provider-github/releases/tag/v6.7.0). 

As it (`organization_role_team_assignment`) was added "recently" may need less changes for provider consumers instead of `organization_role_team`, which is present for almost one year.

stevehipwell: @ViacheslavKudinov they were all introduced in the same release, the PR freeze left everything backed up which meant that overlapping PRs got merged. The `organization_role_team` resource is part of a set of 11 resources or data sources while `organization_role_team_assignment` was in a PR on it's own.

ViacheslavKudinov: @stevehipwell thank you for sharing the context.

I totally fine with this PR as I got better understanding. Have no any concern.

nickfloyd:
> @nickfloyd do we have a v6 release branch (e.g. `release-v6`) yet? I've been opening v6 PRs onto `main` with the expectation that they'd be cherry picked into the release branch, which I think is right for changes like this.

Hey @stevehipwell help me understand the thinking behind the need to cherry pick our changes into a v6.x branch?  Currently main is clean and shippable for any future v6.x changes - no breaking changes have been introduced in main and we should be able to treat the last v6.x release as just a normal release with one caveat - we've finally create a v6 release branch before merging anything v7 into main.

I could definitely be missing something in the line of thinking though - I want to make sure have a complete understanding.

stevehipwell:
@nickfloyd I was assuming that your v77 branch was heading to `main` imminently and that other work for v7 would want to be integrated sooner than later. Obviously until v7 changes are actually merged we don't need to do anything. Once v7 changes are merged I was thinking that the best course of action would be for PRs to target `main`, unless they were v6 only, and then to cherry pick the commits back into the v6 release branch.

nickfloyd:
@stevehipwell Got it, thank you for the clarity.  I was trying to hold off on that as long as we possibly could to avoid the additional complexity of maintaining two "main" branches.  I know that it will cause extra work on me on the other side but I'd prefer to simply roll as much as we can from this [milestone](https://github.com/integrations/terraform-provider-github/milestone/61) targeting a merge of v77 next week.

PR #2944, opened 2025-11-25
Bugfix/#2943 fix data residency regex

### LINKED ISSUE #2943: [BUG]: Data residency does not use correct api url
### Expected Behavior

Use baseURL for api requests

### Actual Behavior

Uses api.github.com since regex is wrong

### Terraform Version

opentofu 1.10.6
integrations/github v6.8.3

### Affected Resource(s)

github_organization_role 
but guessing all

```hcl

```

### Steps to Reproduce

set baseURL in provider to a data residency and run opentofu init && opentofu plan

### Debug Output

```shell
GET https://<redacted>.ghe.com/api/v3/orgs/banken/organization-roles/
```

### Panic Output

```shell
No panic just error
```

### Code of Conduct

tehho:
Add actual support for data residency in regex

Resolves #2943

----

### Before the change?

* Does not check baseURL correctly

### After the change?

* Should check baseURL for data residency match

### Pull request checklist

- [ ] Yes
- [x] No

----

### DIFF
--- a/github/config.go
+++ b/github/config.go
@@ -39,7 +39,7 @@ type Owner struct {
 
 // GHECDataResidencyMatch is a regex to match a GitHub Enterprise Cloud data residency URL:
 // https://[hostname].ghe.com instances expect paths that behave similar to GitHub.com, not GitHub Enterprise Server.
-var GHECDataResidencyMatch = regexp.MustCompile(`^https:\/\/[a-zA-Z0-9.\-]*\.ghe\.com\/?$`)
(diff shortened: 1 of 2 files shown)


stevehipwell (commented):
I don't think this is the right way to fix this issue, it looks like the real cause of the issue is that the provider logic is brittle and uses strings where a `url.URL` should be being used.

stevehipwell:
@tehho I've opened #2951 as a replacement for this PR.

PR #2951, opened 2025-11-26
chore: Refactor base url logic

### LINKED ISSUE #2943: [BUG]: Data residency does not use correct api url

stevehipwell:
Resolves #2943
Closes #2944

----

### Before the change?

* The base URL is often handled as a string with manual concatenations
* Base URL validation isn't well defined
* Base URL format is rigid and doesn't actually work for GHEC

### After the change?

* The base URL is handled as a strongly typed `url.URL`
* The base URL is validated with a single function
* GHEC should work correctly

### Pull request checklist

- [ ] Yes
- [x] No

----

### DIFF
--- a/github/config.go
+++ b/github/config.go
@@ -37,9 +37,12 @@ type Owner struct {
 	IsOrganization bool
 }
 
-// GHECDataResidencyMatch is a regex to match a GitHub Enterprise Cloud data residency URL:
-// https://[hostname].ghe.com instances expect paths that behave similar to GitHub.com, not GitHub Enterprise Server.
-var GHECDataResidencyMatch = regexp.MustCompile(`^https:\/\/[a-zA-Z0-9.\-]*\.ghe\.com$`)
+// DotComHost is the hostname for GitHub.com API.
+const DotComHost = "api.github.com"
+
+// GHECDataResidencyHostMatch is a regex to match a GitHub Enterprise Cloud data residency host:
+// https://[hostname].ghe.com/ instances expect paths that behave similar to GitHub.com, not GitHub Enterprise Server.
+var GHECDataResidencyHostMatch = regexp.MustCompile(`^[a-zA-Z0-9.\-]+\.ghe\.com\/?$`)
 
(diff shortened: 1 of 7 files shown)


github/config.go
```diff
@@ -82,7 +85,8 @@ func (c *Config) NewGraphQLClient(client *http.Client) (*githubv4.Client, error)
 		return nil, err
 	}
 
-	if uv4.String() != "https://api.github.com/" && !GHECDataResidencyMatch.MatchString(uv4.String()) {
+	hostname := uv4.Hostname()
+	if hostname != DotComHost && !GHECDataResidencyHostMatch.MatchString(hostname) {
```
deiga, line 89: suggestion: this condition keeps repeating in multiple places. Would there be any value in refactoring it to a function?

stevehipwell: I think it's used in 2 places like that and 1 other place negated, so it could be turned into a function. The question is would a function called something like `isDotComOrGHEC()` (please give me a better name) provide as much clarity as seeing the code above? I was thinking that from a Go perspective this would be premature drying, but after writing my replay I'm leaning more towards the function approach. Especially as I could then add some tests.

nickfloyd: I'd go with the rule of 3 (3 or more implementations to trigger an abstraction) on this one given the logic is not significantly complex and abstraction feels like an over correction.

If you are going to abstract it I'd suggest getting rid of the negation (nots in sequence always tend to make things harder to reason IMO - though I might be alone on that 😉 ). Something like:

```
func isGitHubHosted(hostname string) bool {
    return hostname == DotComHost || GHECDataResidencyHostMatch.MatchString(hostname)
}
```

then 

```
if !isGitHubHosted(hostname) {
   // Add the prefix or segment
}
```

nickfloyd (approved):
Just a comment on the discussion - other than that this is g2g

Issue #2952, opened 2025-11-26
[BUG]: Removing bypass actors from Rulesets does not have any effect


yamatt:
### Expected Behavior

When removing `bypass_actors` from a `github_organization_ruleset` this is recognised in the plan and removed from the rulsets.

### Actual Behavior

When removing `bypass_actors` from a `github_organization_ruleset` this is recognised in the plan but not removed from the rulsets.

### Terraform Version

Terraform v1.14.0
on windows_amd64
+ provider registry.terraform.io/hashicorp/azurerm v4.49.0
+ provider registry.terraform.io/hashicorp/time v0.13.1
+ provider registry.terraform.io/integrations/github v6.8.3

### Affected Resource(s)

- github_organization_ruleset

### Terraform Configuration Files

```hcl
resource "github_organization_ruleset" "review_required_protected_branches" {
  name        = "Review required on protected branches"
  enforcement = "active"
  target      = "branch"

  conditions {
    ref_name {
      exclude = []
      include = [
        "~DEFAULT_BRANCH",
        "refs/heads/main",
      ]
    }
    repository_name {
      exclude   = []
      include   = local.repos
      protected = false
    }
  }

  rules {
    creation                = false
    deletion                = true
    non_fast_forward        = true
    required_linear_history = false
    required_signatures     = false
    update                  = false

    pull_request {
      dismiss_stale_reviews_on_push     = true
      require_code_owner_review         = true
      require_last_push_approval        = true
      required_approving_review_count   = 1
      required_review_thread_resolution = true
    }
  }

  lifecycle {
    prevent_destroy = true
  }
}
```

### Steps to Reproduce

Running Terraform plan in my pipeline I can see this output

```
# github_organization_rulesetreview_required_protected_branches will be updated in-place
  ~ resource "github_organization_ruleset" "review_required_protected_branches" {
        id          = "8406***"
        name        = "Git Flow: Review required on protected branches"
        # (5 unchanged attributes hidden)

      - bypass_actors {
          - actor_id    = 13588*** -> null
          - actor_type  = "Team" -> null
          - bypass_mode = "always" -> null
        }
      - bypass_actors {
          - actor_id    = 13588*** -> null
          - actor_type  = "Team" -> null
          - bypass_mode = "always" -> null
        }

        # (2 unchanged blocks hidden)
    }
```

And in the apply:

```
github_organization_ruleset.review_required_protected_branches: Modifying... [id=8406***]
...
[...]

Issue #2467, opened 2024-11-19
[BUG]: Ruleset, required_check should be optional


marcomansi-ahold:
### Expected Behavior

I want to create a ruleset with a required status check with only "Require branches to be up to date before merging" enables, like this (created through the UI):

![image](https://github.com/user-attachments/assets/9f6c0cbb-a944-4aa2-a93f-40b44b3e362c)
quired and the context too 

### Actual Behavior

Error: 

│ Error: PUT https://api.github.com/repos/RoyalAholdDelhaize/gso-tech-test-marco_mansi-1/rulesets/2669259: 422 Validation Failed [{Resource: Field: Code: Message:Invalid rule 'required_status_checks': Invalid parameter required_status_checks: Invalid array contents. Errors at index 0: Expected context to be present}]
│ 
│   with module.gso_github_repo["gso-tech-test-marco_mansi-1"].github_repository_ruleset.ruleset,
│   on gso_github_repo/main.tf line 84, in resource "github_repository_ruleset" "ruleset":
│   84: resource "github_repository_ruleset" "ruleset" {

### Terraform Version

Terraform 1.9.8
integrations/github 6.3.1

### Affected Resource(s)

- github_repository_ruleset

### Steps to Reproduce

I am using this code: 

```
resource "github_repository_ruleset" "ruleset" {
  name        = "PRs & conventional commits"
  repository  = github_repository.repo.name
  target      = "branch"
  enforcement = "active"

  conditions {
    ref_name {
      include = ["~DEFAULT_BRANCH"]
      exclude = []
    }
  }

  rules {
    creation                = true
    update                  = true
    deletion                = true
    required_linear_history = true
    non_fast_forward        = true

    pull_request {
      require_last_push_approval        = true
      dismiss_stale_reviews_on_push     = true
      required_review_thread_resolution = true
      required_approving_review_count   = 1
    }

    required_status_checks {
      required_check {
        context = ""
      }
      strict_required_status_checks_policy = true
    }
  }
}
```

The required_check block is required and context too. But it cannot be empty

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

kris-evans:
I'm currently experience this as well. This is actually blocking our ability to configure repositories with the Terraform provider as it overwrites the rules on deploy.

cam-bond:
Same here kris-evans. We want to block PR merges on any failing status checks, but we don't care what those status checks actually are, as they are so repo-specific across our org.

[...]

PR #2705, opened 2025-07-17
fix(organization_ruleset): handle other error responses

skeggse:
Resolves #2450

----

### Before the change?

* `SIGSEVG`

### After the change?

* Valid error response

### Pull request checklist

- [ ] Docs have been reviewed and added / updated if needed (for bug fixes / features)

### Does this introduce a breaking change?

- [ ] Yes
- [x] No

----

PR #2958, opened 2025-12-02
[MAINT] Fix `github_organization_ruleset` and `github_repository_ruleset` with `push` target

### LINKED ISSUE #2929: [BUG]: github_organization_ruleset doesn't work with push rulesets
### Expected Behavior

`github_organization_ruleset` should support push rulesets but the github provider (v6.8.3) fails.

using the exact example from the documentation:
https://registry.terraform.io/providers/integrations/github/latest/docs/resources/organization_ruleset

```hcl
# Example with push ruleset  
resource "github_organization_ruleset" "example_push" {
  name        = "example_push"
  target      = "push"
  enforcement = "active"

  conditions {
    ref_name {
      include = ["~ALL"]
      exclude = []
    }
    repository_name {
      include = ["~ALL"] 
      exclude = []
    }
  }

  rules {
    file_path_restriction {
      restricted_file_paths = [".github/workflows/*", "*.env"]
    }

    max_file_size {
      max_file_size = 104857600  # 100 MB in bytes
    }

    max_file_path_length {
      max_file_path_length = 255
    }

    file_extension_restriction {
      restricted_file_extensions = ["*.exe", "*.dll", "*.so"]
    }
  }
}
```
should create a new push ruleset.

### Actual Behavior

instead it causes an error or the provider crashes:

```
│ Error: POST https://api.github.com/orgs/xxx/rulesets: 422 Validation Failed [{Resource: Field: Code: Message:Target ref_name is not supported for push rulesets}]
```

removing the `ref_name` object from the condition causes the provider to crash:

```
Stack trace from the terraform-provider-github_v6.8.3 plugin:

goroutine 52 [running]:
github.com/integrations/terraform-provider-github/v6/github.expandConditions({0xc000559d70, 0xf35f06?, 0x0?}, 0x1)
	github.com/integrations/terraform-provider-github/v6/github/respository_rules_utils.go:105 +0xae8
github.com/integrations/terraform-provider-github/v6/github.resourceGithubRulesetObject(0xc0001fde00, {0xc0003f82c0, 0x10})
	github.com/integrations/terraform-provider-github/v6/github/respository_rules_utils.go:32 +0x27c
github.com/integrations/terraform-provider-github/v6/github.resourceGithubOrganizationRulesetCreate(0xc0001fde00, {0xd49780, 0xc0006da0c0})
	github.com/integrations/terraform-provider-github/v6/github/resource_github_organization_ruleset.go:591 +0x5d
github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema.(*Resource).create(0x1096870?, {0x1096870?, 0xc000504ba0?}, 0xd?, {0xd49780?, 0xc0006da0c0?})
	github.com/hashicorp/terraform-plugin-sdk/v2@v2.34.0/helper/schema/resource.go:794 +0x15f
github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema.(*Resource).Apply(0xc00037b500, {0x1096870, 0xc000504ba0}, 0xc00050a9c0, 0xc0001fdc80, {0xd49780, 0xc0006da0c0})
	github.com/hashicorp/terraform-plugin-sdk/v2@v2.34.0/helper/schema/resource.go:937 +0xa89
github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema.(*GRPCProviderServer).ApplyResourceChange(0xc00034c660, {0x1096870?, 0xc000504ae0?}, 0xc0002356d0)
	github.com/hashicorp/terraform-plugin-sdk/v2@v2.34.0/helper/schema/grpc_provider.go:1153 +0xd5c
github.com/hashicorp/terraform-plugin-go/tfprotov5/tf5server.(*server).ApplyResourceChange(0xc000249900, {0x1096870?, 0xc000504120?}, 0xc0001f6ee0)
	github.com/hashicorp/terraform-plugin-go@v0.23.0/tfprotov5/tf5server/server.go:865 +0x3d0
github.com/hashicorp/terraform-plugin-go/tfprotov5/internal/tfplugin5._Provider_ApplyResourceChange_Handler({0xef6da0, 0xc000249900}, {0x1096870, 0xc000504120}, 0xc0001fd180, 0x0)
	github.com/hashicorp/terraform-plugin-go@v0.23.0/tfprotov5/internal/tfplugin5/tfplugin5_grpc.pb.go:518 +0x1a6
google.golang.org/grpc.(*Server).processUnaryRPC(0xc000173000, {0x1096870, 0xc000504090}, {0x109b280, 0xc00025a300}, 0xc0004fd440, 0xc000293110, 0x173c578, 0x0)
	google.golang.org/grpc@v1.63.2/server.go:1369 +0xdf8
google.golang.org/grpc.(*Server).handleStream(0xc000173000, {0x109b280, 0xc00025a300}, 0xc0004fd440)
	google.golang.org/grpc@v1.63.2/server.go:1780 +0xe8b
google.golang.org/grpc.(*Server).serveStreams.func2.1()
	google.golang.org/grpc@v1.63.2/server.go:1019 +0x8b
created by google.golang.org/grpc.(*Server).serveStreams.func2 in goroutine 15
	google.golang.org/grpc@v1.63.2/server.go:1030 +0x125

Error: The terraform-provider-github_v6.8.3 plugin crashed!

This is always indicative of a bug within the plugin. It would be immensely
helpful if you could report the crash with the plugin's maintainers so that it
can be fixed. The output above should help diagnose the issue.

```

### Terraform Version

OpenTofu v1.10.6
GitHub Provider v6.8.3

### Affected Resource(s)

- github_organization_ruleset

```hcl
# Example with push ruleset  
resource "github_organization_ruleset" "example_push" {
  name        = "example_push"
  target      = "push"
  enforcement = "active"

  conditions {
    ref_name {
      include = ["~ALL"]
      exclude = []
    }
    repository_name {
      include = ["~ALL"] 
      exclude = []
    }
  }

  rules {
    file_path_restriction {
      restricted_file_paths = [".github/workflows/*", "*.env"]
    }

    max_file_size {
      max_file_size = 104857600  # 100 MB in bytes
    }

    max_file_path_length {
      max_file_path_length = 255
    }

    file_extension_restriction {
      restricted_file_extensions = ["*.exe", "*.dll", "*.so"]
    }
  }
}
```

### Steps to Reproduce

_No response_

### Debug Output

```shell

```

### Panic Output

```shell

```

### Code of Conduct

deiga:
Resolves #2929, #2467

----

### Before the change?

* The provider would crash from a 422 error response when following the example in the docs
* Removing `ref_name` would cause the provider to Panic as `ref_name` is a required field
* One could add `rules` which weren't valid for `push` rulesets

### After the change?

* The provider should correctly apply `push` rulesets to an organization
* `ref_name` should no longer be needed to be set for `push` target
* `conditions` & `target` validation logic should ensure correct fields are populated

### Pull request checklist

- [ ] Yes
- [x] No

----


github/resource_github_organization_ruleset.go
```diff
@@ -713,3 +714,50 @@ func resourceGithubOrganizationRulesetImport(d *schema.ResourceData, meta any) (
 
 	return []*schema.ResourceData{d}, nil
 }
+
+func validateConditionsFieldForBranchAndTagTargets(d *schema.ResourceDiff, meta interface{}) error {
+	conditions := d.Get("conditions").([]any)[0].(map[string]any)
+	if conditions["ref_name"] == nil || conditions["repository_name"] == nil || conditions["repository_id"] == nil {
+		return fmt.Errorf("ref_name and repository_name or ref_name and repository_id must be set for branch and tag targets")
+	}
```
nickfloyd: This should be simplified to make what is required a bit more clear.

```suggestion
    if conditions["ref_name"] == nil {
        return fmt.Errorf("ref_name must be set for branch and tag targets")
    }
    if conditions["repository_name"] == nil && conditions["repository_id"] == nil {
        return fmt.Errorf("repository_name or repository_id must be set for branch and tag targets")
    }
```

github/resource_github_organization_ruleset.go
```diff
@@ -86,12 +88,12 @@ func resourceGithubOrganizationRuleset() *schema.Resource {
...
 				Optional:    true,
 				MaxItems:    1,
-				Description: "Parameters for an organization ruleset condition. `ref_name` is required alongside one of `repository_name` or `repository_id`.",
+				Description: "Parameters for an organization ruleset condition. For `branch` and `tag` targets, `ref_name` is required alongside one of `repository_name` or `repository_id`. The `push` target does not require `ref_name` conditions. For `repository` target, the conditions object should only contain the `repository_name` and the `repository_id`.",
 				Elem: &schema.Resource{
 					Schema: map[string]*schema.Schema{
 						"ref_name": {
 							Type:     schema.TypeList,
-							Required: true,
+							Optional: true,
```
nickfloyd: This is a non breaking change ➕

github/resource_github_organization_ruleset.go
```diff
@@ -713,3 +714,50 @@ func resourceGithubOrganizationRulesetImport(d *schema.ResourceData, meta any) (
...
+}
+
+func validateConditionsFieldBasedOnTarget(ctx context.Context, d *schema.ResourceDiff, meta interface{}) error {
+	target := d.Get("target").(string)
+	conditionsRaw := d.Get("conditions").([]any)
+
+	// Handle empty conditions - branch and tag targets require conditions with ref_name
+	if len(conditionsRaw) == 0 {
+		return fmt.Errorf("conditions block is required for %s target", target)
+	}
```
nickfloyd: This block is problematic and should be considered a breaking change.  Existing configs with targets using `push` and `repository` might not have conditions defined.

Currently conditions is marked as `Optional:    true,` which I belive returning this error is a contradiction of that.  If we have to make this move I think the log would be better as a warn with a deprecation warning that this will be different in 7.0 or something - depending on where you see this change going - if this is strictly for v7.0 then perhaps your breaking change is ok if properly documented.

github/resource_github_organization_ruleset.go
```diff
@@ -35,8 +37,8 @@ func resourceGithubOrganizationRuleset() *schema.Resource {
 			"target": {
 				Type:         schema.TypeString,
 				Required:     true,
-				ValidateFunc: validation.StringInSlice([]string{"branch", "tag", "push"}, false),
-				Description:  "Possible values are `branch`, `tag` and `push`. Note: The `push` target is in beta and is subject to change.",
+				ValidateFunc: validation.StringInSlice([]string{"branch", "tag", "push", "repository"}, false),
+				Description:  "The target of the ruleset. Possible values are `branch`, `tag`, `push` and `repository`.",
```
stevehipwell: The SDK doesn't support `repository` target rules, mainly because I've only just discovered that despite being in a separate UI area they're merged into the existing API. This is made even less discoverable by the top example of targets not having `repository` as an option, the schemas not being updated, and nobody at GitHub being able to give me a [definitive answer](https://github.com/orgs/community/discussions/148564).

github/resource_github_organization_ruleset.go
```diff
@@ -86,12 +88,12 @@ func resourceGithubOrganizationRuleset() *schema.Resource {
 				Type:        schema.TypeList,
 				Optional:    true,
 				MaxItems:    1,
-				Description: "Parameters for an organization ruleset condition. `ref_name` is required alongside one of `repository_name` or `repository_id`.",
+				Description: "Parameters for an organization ruleset condition. For `branch` and `tag` targets, `ref_name` is required alongside one of `repository_name` or `repository_id`. The `push` target does not require `ref_name` conditions. For `repository` target, the conditions object should only contain the `repository_name` and the `repository_id`.",
```
stevehipwell: This doesn't look quite right, all org rulesets support complex repository targeting (SDK dependent). Push and repository (see above) targets don't support a ref target.

stevehipwell (commented):
I think this change likely wants to wait for the SDK upgrade as a lot of this area is modified in future versions.

FYI the error behaviours previously seen should have been mitigated by #2705 so if there is an error the provider should handle it gracefully.

github/resource_github_organization_ruleset.go
deiga: After some review of the API, this isn't even wanted. Changed it's function entirely

github/resource_github_organization_ruleset.go
deiga: v79 of the SDK does support those: https://github.com/google/go-github/blob/master/github/rules.go#L220

I honestly wanted to already put them there, so that it's ready when we get to upgrade. But I can remove it for now

github/resource_github_organization_ruleset.go
deiga: I'm not sure what you mean, I extrapolated this based on the REST API description. AFAIK there isn't anything more we can currently use (since `custom_property` targeting is not an option yet)
Though it's true that `ref_name` is optional even for `tag` and `branch` targets

github/resource_github_organization_ruleset.go
stevehipwell: I literally just added support for them in https://github.com/google/go-github/pull/3850 and they've just been released in [v80.0.0](https://github.com/google/go-github/releases/tag/v80.0.0).

github/resource_github_organization_ruleset.go
stevehipwell: The wording isn't right even if the field names are, the REST API makes a complete mess of this too. My point was that it'd be worth making the comment clearer as to what the fields actually do.

deiga: You are right that we can try to describe it better than the API docs. I'll try my hand at that.

github/resource_github_organization_ruleset.go
```diff
@@ -659,46 +662,46 @@ func resourceGithubOrganizationRulesetRead(d *schema.ResourceData, meta any) err
...
-		return unconvertibleIdErr(d.Id(), err)
+		return diag.FromErr(unconvertibleIdErr(d.Id(), err))
 	}
 
-	ctx := context.WithValue(context.Background(), ctxId, rulesetID)
+	ctx = context.WithValue(ctx, ctxId, rulesetID)
 
 	var ruleset *github.RepositoryRuleset
 
 	ruleset, _, err = client.Organizations.UpdateRepositoryRuleset(ctx, org, rulesetID, *rulesetReq)
```
stevehipwell: Checking if the bypass actors have changed and the length is `0` here would allow you to use the clear function to resolve #2952. See https://github.com/integrations/terraform-provider-github/pull/2891#discussion_r2591993478.

github/config.go
```diff
@@ -47,7 +47,7 @@ var GHECDataResidencyHostMatch = regexp.MustCompile(`^[a-zA-Z0-9.\-]+\.ghe\.com\
 func RateLimitedHTTPClient(client *http.Client, writeDelay, readDelay, retryDelay time.Duration, parallelRequests bool, retryableErrors map[int]bool, maxRetries int) *http.Client {
 	client.Transport = NewEtagTransport(client.Transport)
 	client.Transport = NewRateLimitTransport(client.Transport, WithWriteDelay(writeDelay), WithReadDelay(readDelay), WithParallelRequests(parallelRequests))
-	client.Transport = logging.NewSubsystemLoggingHTTPTransport("GitHub", client.Transport)
+	client.Transport = logging.NewLoggingHTTPTransport(client.Transport)
```
stevehipwell, line 62: Could you explain why we need this change?

github/resource_github_organization_ruleset.go
stevehipwell: This needs removing.

github/resource_github_organization_ruleset.go
```diff
@@ -45,7 +50,7 @@ func resourceGithubOrganizationRuleset() *schema.Resource {
 				Description:  "Possible values for Enforcement are `disabled`, `active`, `evaluate`. Note: `evaluate` is currently only supported for owners of type `organization`.",
 			},
 			"bypass_actors": {
-				Type:             schema.TypeList,
+				Type:             schema.TypeList, // TODO: These are returned from GH API sorted by actor_id, we might want to investigate if we want to include sorting
```
stevehipwell, line 56: I think we should sort the returned values based on the inputs which would stop churn.

stevehipwell (changes requested):
Would it be possible to either de-scope this PR or open a new PR with the smallest number of changes possible to fix the outstanding bugs?

deiga:
@stevehipwell Yes, I agree. I've done that already in this PR: https://github.com/integrations/terraform-provider-github/pull/2976

But I can't switch the base of this PR to point to that :)

github/config.go
deiga: We don't inititialize the `GitHub` subsystem anywhere and from my understanding subsystems are usually per-resource things.
With an unitialized Subsystem the logs get spammed with warnings about it.

But this implementation is "actually" in this PR https://github.com/integrations/terraform-provider-github/pull/2976 and any modifications to it should happen there

github/resource_github_organization_ruleset.go
deiga: That sounds sensible, need to investigate where the best place for the sorting is

github/resource_github_organization_ruleset.go
deiga: @stevehipwell The problem here is, that go-github doesn't have `UpdateRulesetClearBypassActor` for Organization rulesets :/

github/resource_github_organization_ruleset.go
deiga: @stevehipwell after some investigation, it seems that `schema.TypeSet` would be the correct way to implement unsorted. What do you think?

github/util_ruleset_validation.go
```diff
@@ -0,0 +1,157 @@
+package github
```
stevehipwell, line 1: I'm not sure why we need this new file instead of using the _util\_rules.go_ file?

github/repository_rules_validation_utils_test.go
```diff
@@ -0,0 +1,225 @@
+package github
+
+import (
+	"testing"
+)
+
+func TestValidateConditionsFieldForPushTarget(t *testing.T) {
```
stevehipwell:
```suggestion
func Test_validateConditionsFieldForPushTarget(t *testing.T) {
```

This would be the idiomatic name for an unexported method test.

github/resource_github_organization_ruleset.go
```diff
@@ -26,6 +28,11 @@ func resourceGithubOrganizationRuleset() *schema.Resource {
 
 		SchemaVersion: 1,
 
+		CustomizeDiff: customdiff.All(
+			validateConditionsFieldBasedOnTarget,
+			validateOrganizationRulesetRules,
+		),
```
stevehipwell: I don't think we should use separate function here as we're writing our own logic, that pattern is only really useful when we're composing. Could you take a look at how I've structured the schema and done this in #3069?

github/resource_github_organization_ruleset.go
```diff
@@ -831,3 +844,37 @@ func resourceGithubOrganizationRulesetImport(ctx context.Context, d *schema.Reso
 
 	return []*schema.ResourceData{d}, nil
 }
+
+func validateConditionsFieldBasedOnTarget(ctx context.Context, d *schema.ResourceDiff, meta any) error {
```
stevehipwell: If this is scoped to the organization ruleset then the name should reflect that, otherwise it should probably be in the _util\_rules.go_ file (idiomatically I'd expect this to be called _util\_ruleset.go_).

github/resource_github_organization_ruleset_test.go
```diff
@@ -493,6 +489,333 @@ resource "github_organization_ruleset" "test" {
...
+					# Push rulesets only support push-specific rules:
+					# file_path_restriction, max_file_path_length, file_extension_restriction, max_file_size
+					max_file_size {
+						max_file_size = 100
+					}
+				}
+			}
+		`, resourceName, randomID)
+
+		check := resource.ComposeTestCheckFunc(
```
stevehipwell: Can we please inline these instead of creating a separate var, this makes the code so much harder to read.

github/resource_github_organization_ruleset_test.go
```diff
@@ -493,6 +489,333 @@ resource "github_organization_ruleset" "test" {
...
+			resource.TestCheckResourceAttr(
+				fmt.Sprintf("github_organization_ruleset.%s", resourceName),
+				"name",
+				fmt.Sprintf("test-push-%s", randomID),
+			),
```
stevehipwell:
```suggestion
			resource.TestCheckResourceAttr(fmt.Sprintf("github_organization_ruleset.%s", resourceName), "name", fmt.Sprintf("test-push-%s", randomID)),
```
Please can these be made a single line, it's the idiomatic Go pattern (no introducing line breaks) and makes the code much easier to read.

stevehipwell (changes requested):
I've added some review comments, mainly about the code structure.

github/util_ruleset_validation.go
deiga: I felt that `util_rules.go` is already getting too long to reason about nicely. But if this is an anti-pattern I can move these there

github/repository_rules_validation_utils_test.go
deiga: Oh! Thanks 🙏

github/util_ruleset_validation.go
stevehipwell: If you need a new file it'd make sense to keep it located near the other util files and out of the middle of the main provider code. So for example _util\_rule\_validation.go_?

github/resource_github_organization_ruleset.go
deiga: Honestly, I prefer the composition pattern as it's easier to read and easier to extend when new needs arise.

Does either pattern have downsides on the code-level?

github/resource_github_organization_ruleset.go
deiga: Ugh, right. I keep forgetting that all functions within a package are visible 😬

github/resource_github_organization_ruleset.go
```diff
@@ -34,16 +37,17 @@ func resourceGithubOrganizationRuleset() *schema.Resource {
 				Description:  "The name of the ruleset.",
 			},
 			"target": {
-				Type:         schema.TypeString,
-				Required:     true,
-				ValidateFunc: validation.StringInSlice([]string{"branch", "tag", "push"}, false),
-				Description:  "Possible values are `branch`, `tag` and `push`. Note: The `push` target is in beta and is subject to change.",
+				Type:     schema.TypeString,
+				Required: true,
+				// The API accepts an `repository` target, but any rule created with that doesn't show up in the UI, nor does it have any rules.
```
stevehipwell: It shows up in the UI, just in a different UI.

github/resource_github_organization_ruleset.go
```diff
@@ -826,7 +835,7 @@ func resourceGithubOrganizationRulesetImport(ctx context.Context, d *schema.Reso
 			"owner":      owner,
 			"ruleset_id": rulesetID,
 		})
-		return []*schema.ResourceData{d}, fmt.Errorf("`ruleset_id` must be present")
+		return []*schema.ResourceData{d}, errors.New("`ruleset_id` must be present")
```
stevehipwell: Please use `fmt.Errorf()` for consistency.

github/resource_github_organization_ruleset.go
```diff
@@ -853,3 +862,17 @@ func resourceGithubOrganizationRulesetImport(ctx context.Context, d *schema.Reso
 
 	return []*schema.ResourceData{d}, nil
 }
+
+func resourceGithubOrganizationRulesetValidate(ctx context.Context, d *schema.ResourceDiff, _ any) error {
```
stevehipwell:
```suggestion
func resourceGithubOrganizationRulesetDiff(ctx context.Context, d *schema.ResourceDiff, _ any) error {
```

github/resource_github_repository.go
```diff
@@ -1019,7 +1019,7 @@ func resourceGithubRepositoryUpdate(ctx context.Context, d *schema.ResourceData,
 		}
 	}
 
-	if d.HasChanges("visibility", "private") {
+	if d.HasChanges("visibility", "private") && !d.Get("archived").(bool) {
```
stevehipwell: This should be in a separate PR, and I'm also not sure if it's valid to disable changing the visibility level when archiving (the execution order issue is a separate concern).

github/resource_github_repository_ruleset.go
```diff
@@ -792,3 +791,17 @@ func resourceGithubRepositoryRulesetImport(ctx context.Context, d *schema.Resour
 
 	return []*schema.ResourceData{d}, nil
 }
+
+func resourceGithubRepositoryRulesetValidate(ctx context.Context, d *schema.ResourceDiff, meta any) error {
```
stevehipwell:
```suggestion
func resourceGithubRepositoryRulesetDiff(ctx context.Context, d *schema.ResourceDiff, meta any) error {
```

github/resource_github_repository_test.go
```diff
@@ -1315,19 +1315,19 @@ func Test_expandPages(t *testing.T) {
 }
 
 func TestGithubRepositoryTopicPassesValidation(t *testing.T) {
-	resource := resourceGithubRepository()
-	schema := resource.Schema["topics"].Elem.(*schema.Schema)
-	diags := schema.ValidateDiagFunc("ef69e1a3-66be-40ca-bb62-4f36186aa292", cty.Path{cty.GetAttrStep{Name: "topic"}})
+	repositoryResourceSchema := resourceGithubRepository()
```
stevehipwell: Shouldn't these changes be in a separate PR?

github/resource_github_team_repository.go
```diff
@@ -138,7 +138,7 @@ func resourceGithubTeamRepositoryRead(d *schema.ResourceData, meta any) error {
 				return nil
 			}
 		}
-		return err
+		return repoErr
```
stevehipwell, line 141: Shouldn't this fix be in a separate PR?

stevehipwell (changes requested):
@deiga I've added some comments but the main concern here is that we keep only the relevant changes in the PR. If you could tidy the PR up, rebase, and then I'll give this a full PR with the aim of getting it merged ASAP.

github/resource_github_organization_ruleset.go
deiga: What's the benefit of consistency for this? AFAIK `fmt.Errorf` on a static string has an unnecessary performance overhead over `errors.New` and their outcome is exactly the same, right?
Sure, the performance overhead is miniscule, granted.

deiga:
> @deiga I've added some comments but the main concern here is that we keep only the relevant changes in the PR. If you could tidy the PR up, rebase, and then I'll give this a full PR with the aim of getting it merged ASAP.

Shuffled the unrelated changes out!

github/resource_github_organization_ruleset.go
stevehipwell: Go is about readability, so the code should be consistent as a priority. FYI the previous performance overhead, which is minimal in the context of this TF provider, will [no longer be present in Go 1.26](https://antonz.org/go-1-26/#fmt-errorf).

github/util_rules.go
```diff
@@ -1,14 +1,24 @@
...
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
 )
 
+type Target string
+
+const (
+	TargetPush   Target = "push"
+	TargetBranch Target = "branch"
+	TargetTag    Target = "tag"
+)
```
stevehipwell: Use `github.RulesetTarget`.

github/util_rules.go
```diff
@@ -221,19 +231,26 @@ func expandConditions(input []any, org bool) *github.RepositoryRulesetConditions
 }
 
 func flattenConditions(conditions *github.RepositoryRulesetConditions, org bool) []any {
-	if conditions == nil || conditions.RefName == nil {
+	return flattenConditionsWithContext(context.TODO(), conditions, org)
+}
+
+func flattenConditionsWithContext(ctx context.Context, conditions *github.RepositoryRulesetConditions, org bool) []any {
```
stevehipwell: Could you rename this function to `flattenConditions` and replace the old function with this one?

github/util_rules.go
```diff
@@ -576,7 +593,7 @@ func flattenRules(rules *github.RepositoryRulesetRules, org bool) []any {
 			"required_review_thread_resolution": rules.PullRequest.RequiredReviewThreadResolution,
 			"allowed_merge_methods":             rules.PullRequest.AllowedMergeMethods,
 		})
-		log.Printf("[DEBUG] Flattened Pull Request rules slice request slice: %#v", pullRequestSlice)
+		log.Printf("[DEBUG] Flattened Pull Request rules slice: %#v", pullRequestSlice)
```
stevehipwell: If we always have a valid context I think we could use `tflog`?

github/util_ruleset_validation.go
```diff
@@ -0,0 +1,182 @@
...
+//
+// To verify/maintain this list:
+//  1. Check the GitHub API documentation for organization rulesets:
+//     https://docs.github.com/en/rest/orgs/rules?apiVersion=2022-11-28#create-an-organization-repository-ruleset
+//  2. The API docs don't clearly separate push vs branch/tag rules. To verify,
+//     attempt to create a push ruleset via API or UI with each rule type.
+//     Push rulesets will reject branch/tag rules with "Invalid rule '<name>'" error.
+//  3. Generally, push rules deal with file content (paths, sizes, extensions),
+//     while branch/tag rules deal with ref lifecycle and merge requirements.
+var branchTagOnlyRules = []string{
```
stevehipwell: Please use `github.RepositoryRuleType`.

github/util_ruleset_validation.go
```diff
@@ -0,0 +1,182 @@
...
+// to repositories. They are not supported for branch or tag rulesets.
+//
+// To verify/maintain this list:
+//  1. Check the GitHub API documentation for organization rulesets:
+//     https://docs.github.com/en/rest/orgs/rules?apiVersion=2022-11-28#create-an-organization-repository-ruleset
+//  2. The API docs don't clearly separate push vs branch/tag rules. To verify,
+//     attempt to create a branch ruleset via API or UI with each rule type.
+//     Branch rulesets will reject push-only rules with an error.
+//  3. Push rules control file content: paths, sizes, extensions, path lengths.
+var pushOnlyRules = []string{
```
stevehipwell: As above.

stevehipwell (changes requested):
This is looking good. I've added a few more comments, if you could address these then I think we're ready to review the new validation logic.

deiga:
@stevehipwell Thanks for the comments! I addressed them all now :)

github/resource_github_organization_ruleset.go
```diff
@@ -34,16 +37,17 @@ func resourceGithubOrganizationRuleset() *schema.Resource {
...
 			},
 			"target": {
-				Type:         schema.TypeString,
-				Required:     true,
-				ValidateFunc: validation.StringInSlice([]string{"branch", "tag", "push"}, false),
-				Description:  "Possible values are `branch`, `tag` and `push`. Note: The `push` target is in beta and is subject to change.",
+				Type:     schema.TypeString,
+				Required: true,
+				// The API accepts an `repository` target, but we don't support it yet.
+				ValidateFunc: validation.StringInSlice([]string{string(github.RulesetTargetBranch), string(github.RulesetTargetTag), string(github.RulesetTargetPush)}, false),
```
stevehipwell:
```suggestion
				ValidateDiagFunc: validation.ToDiagFunc(validation.StringInSlice([]string{string(github.RulesetTargetBranch), string(github.RulesetTargetTag), string(github.RulesetTargetPush)}, false)),
```

Could you update all of these in the modified files? You'll also want to replace `ValidateDiagFunc: toDiagFunc()` with `ValidateDiagFunc: validation.ToDiagFunc()` as the SDK function correctly configures the path in the diagnostics.

github/resource_github_organization_ruleset.go
```diff
@@ -488,9 +495,10 @@ func resourceGithubOrganizationRuleset() *schema.Resource {
 													Description: "The repository in which the workflow is defined.",
 												},
 												"path": {
-													Type:        schema.TypeString,
-													Required:    true,
-													Description: "The path to the workflow YAML definition file.",
+													Type:             schema.TypeString,
+													Required:         true,
+													ValidateDiagFunc: toDiagFunc(validation.StringMatch(regexp.MustCompile(`^\.github\/workflows\/.*$`), "Path must be in the .github/workflows directory"), "path"),
```
stevehipwell:
```suggestion
													ValidateDiagFunc: validation.ToDiagFunc(validation.StringMatch(regexp.MustCompile(`^\.github\/workflows\/.*$`), "Path must be in the .github/workflows directory")),
```
As an example of a change from the comment above.

github/resource_github_organization_ruleset.go
```diff
@@ -645,7 +654,7 @@ func resourceGithubOrganizationRulesetCreate(ctx context.Context, d *schema.Reso
 	_ = d.Set("ruleset_id", ruleset.ID)
 	_ = d.Set("node_id", ruleset.GetNodeID())
 	_ = d.Set("etag", resp.Header.Get("ETag"))
-	_ = d.Set("rules", flattenRules(ruleset.Rules, true))
+	_ = d.Set("rules", flattenRules(ctx, ruleset.Rules, true))
```
stevehipwell:
```suggestion
	if err := d.Set("rules", flattenRules(ctx, ruleset.Rules, true)); err != nil {
		return diag.FromErr(err)
	}
```
Ideally all of these would be updated.

stevehipwell (changes requested):
@deiga this is looking good, just a couple of review comments to look at.

github/resource_github_organization_ruleset.go
```diff
@@ -651,10 +663,18 @@ func resourceGithubOrganizationRulesetCreate(ctx context.Context, d *schema.Reso
 	}
 
 	d.SetId(strconv.FormatInt(*ruleset.ID, 10))
```
stevehipwell:
```go
d.SetId(strconv.FormatInt(ruleset.GetID(), 10))
```

You caught the one repo rules one.

github/resource_github_organization_ruleset.go
```diff
@@ -651,10 +663,18 @@ func resourceGithubOrganizationRulesetCreate(ctx context.Context, d *schema.Reso
...
+		return diag.FromErr(err)
+	}
+	if err := d.Set("etag", resp.Header.Get("ETag")); err != nil {
+		return diag.FromErr(err)
+	}
+	if err := d.Set("rules", flattenRules(ctx, ruleset.Rules, true)); err != nil {
+		return diag.FromErr(err)
+	}
 
 	tflog.Info(ctx, fmt.Sprintf("Created organization ruleset: %s/%s (ID: %d)", owner, name, *ruleset.ID), map[string]any{
```
stevehipwell: ruleset.GetID()

github/resource_github_organization_ruleset.go
```diff
@@ -769,9 +807,15 @@ func resourceGithubOrganizationRulesetUpdate(ctx context.Context, d *schema.Reso
 	}
 
 	d.SetId(strconv.FormatInt(*ruleset.ID, 10))
```
stevehipwell: `ruleset.GetID()`

github/resource_github_repository_ruleset.go
```diff
@@ -740,9 +778,15 @@ func resourceGithubRepositoryRulesetUpdate(ctx context.Context, d *schema.Resour
 		return diag.FromErr(err)
 	}
 	d.SetId(strconv.FormatInt(*ruleset.ID, 10))
```
stevehipwell: `ruleset.GetID()`

github/resource_github_repository_ruleset.go
```diff
@@ -653,11 +665,19 @@ func resourceGithubRepositoryRulesetCreate(ctx context.Context, d *schema.Resour
 		return diag.FromErr(err)
 	}
 
-	d.SetId(strconv.FormatInt(*ruleset.ID, 10))
-	_ = d.Set("ruleset_id", ruleset.ID)
-	_ = d.Set("node_id", ruleset.GetNodeID())
-	_ = d.Set("etag", resp.Header.Get("ETag"))
-	_ = d.Set("rules", flattenRules(ctx, ruleset.Rules, false))
+	d.SetId(strconv.FormatInt(ruleset.GetID(), 10))
+	if err := d.Set("ruleset_id", ruleset.GetID()); err != nil {
```
stevehipwell:
```suggest
	if err := d.Set("ruleset_id", ruleset.ID); err != nil {
```
This would be consistent with the rest of the code.

github/resource_github_repository_ruleset.go
```diff
@@ -699,15 +719,33 @@ func resourceGithubRepositoryRulesetRead(ctx context.Context, d *schema.Resource
...
-	_ = d.Set("ruleset_id", ruleset.ID)
-	_ = d.Set("name", ruleset.Name)
-	_ = d.Set("target", ruleset.GetTarget())
-	_ = d.Set("enforcement", ruleset.Enforcement)
-	_ = d.Set("bypass_actors", flattenBypassActors(ruleset.BypassActors))
-	_ = d.Set("conditions", flattenConditions(ctx, ruleset.GetConditions(), false))
-	_ = d.Set("rules", flattenRules(ctx, ruleset.Rules, false))
-	_ = d.Set("node_id", ruleset.GetNodeID())
-	_ = d.Set("etag", resp.Header.Get("ETag"))
+	if err := d.Set("ruleset_id", ruleset.GetID()); err != nil {
```
stevehipwell:
```suggest
	if err := d.Set("ruleset_id", ruleset.ID); err != nil {
```
This would be consistent with the rest of the code.

stevehipwell (changes requested):
This PR is looking great @deiga; I think we're down to the last minor consistency issue.

stevehipwell (approved):
LGTM

stevehipwell:
@deiga could you please rebase this again so we can get it merged?

nickfloyd (approved):

PR #2962, opened 2025-12-03
fix: missing error returns

### LINKED ISSUE #2957: [BUG]: Fix missing return when using `errors.As`
### Expected Behavior

When an error is returned from the GitHub API it should result in a provider error unless the behaviour is explicitly expected and coded for.

### Actual Behavior

There are code paths in the provider where an error is checked for nil and then only handled if it's a specific type as there is no catch-all `return err`.

### Terraform Version

n/a

### Affected Resource(s)

n/a

```hcl

```

### Steps to Reproduce

_No response_

### Debug Output

```shell

```

### Panic Output

```shell

```

### Code of Conduct

nickfloyd:
Resolves #2957
Relates to: #2705

----

### Before the change?

* When an error is returned from the GitHub API it should result in a provider error unless the behavior is explicitly expected and coded for.

### After the change?

Added catch all for:
- resource_github_user_ssh_key.go

### Pull request checklist

- [ ] Yes
- [x] No

----

### DIFF
--- a/github/resource_github_actions_environment_variable.go
+++ b/github/resource_github_actions_environment_variable.go
@@ -78,14 +78,11 @@ func resourceGithubActionsEnvironmentVariableCreateOrUpdate(d *schema.ResourceDa
 	_, err := client.Actions.CreateEnvVariable(ctx, owner, repoName, escapedEnvName, variable)
 	if err != nil {
 		ghErr := &github.ErrorResponse{}
-		if errors.As(err, &ghErr) {
-			if ghErr.Response.StatusCode == http.StatusConflict {
-				// Variable already exists, try to update instead
-				_, err = client.Actions.UpdateEnvVariable(ctx, owner, repoName, escapedEnvName, variable)
-				if err != nil {
-					return err
-				}
-			} else {
+		if errors.As(err, &ghErr) && ghErr.Response.StatusCode == http.StatusConflict {
+			// Variable already exists, try to update instead
+			// If it fails here, we want to return the error otherwise continue
+			_, err = client.Actions.UpdateEnvVariable(ctx, owner, repoName, escapedEnvName, variable)
+			if err != nil {
 				return err
 			}
 		} else {


stevehipwell:
I'm not sure all of these need the extra return as it looks like they're falling through to an error return outside of the type check. Basically the pattern looks subtly different to the ones we saw for the rulesets. I'm on my phone so will look again in the morning on a big screen to double check.

github/resource_github_actions_environment_secret.go
```diff
@@ -139,6 +139,7 @@ func resourceGithubActionsEnvironmentSecretRead(d *schema.ResourceData, meta any
 				d.SetId("")
 				return nil
 			}
+			return err
```
diofeher: This one looks redundant, since we already have `return err` below.

github/resource_github_user_ssh_key.go
```diff
@@ -96,7 +96,9 @@ func resourceGithubUserSshKeyRead(d *schema.ResourceData, meta any) error {
 				d.SetId("")
 				return nil
 			}
+			return err
 		}
+		return err
```
diofeher, line 100: This is the only one that doesn't return a redundant error on this PR, I think. But I would still remove the `return err` above.

nickfloyd:
@stevehipwell @diofeher let me do another sweep to make sure I didn't over index on my search of missing instances. Thx for the quick 👀

github/resource_github_actions_environment_variable.go
```diff
@@ -78,14 +78,11 @@ func resourceGithubActionsEnvironmentVariableCreateOrUpdate(d *schema.ResourceDa
...
-				}
-			} else {
+		if errors.As(err, &ghErr) && ghErr.Response.StatusCode == http.StatusConflict {
+			// Variable already exists, try to update instead
+			// If it fails here, we want to return the error otherwise continue
+			_, err = client.Actions.UpdateEnvVariable(ctx, owner, repoName, escapedEnvName, variable)
+			if err != nil {
 				return err
 			}
 		} else {
```
deiga, line 88: question: it looks like the else might be redundant here or are there legitimate use-case where it's fine to swallow `err`?

github/resource_github_actions_environment_variable.go
stevehipwell, line 90:
```suggestion
}

return err
```

I think the else wants removing, this would be more idiomatic.

stevehipwell: Looks like the UI isn't working correctly.

github/resource_github_actions_environment_variable.go
nickfloyd: I think there is.  When create fails but update succeeds.

github/resource_github_actions_environment_variable.go
nickfloyd: I might be looking at this wrong but, since this is a multi op function - it tries to create first and then update I don't think we should return the error outside of the block given there are assignments after and if an update occurs then we want to continue with a clean exit.

stevehipwell: Yeah, I agree. I think the whole block need looking at again at some point as i'm not sure why we're allowing an auto-import of an existing resource?

PR #2968, opened 2025-12-04
fix: Fix org roles implementation

### LINKED ISSUE #2805: [BUG]: Inconsistent result when moving from `github_organization_custom_role` to `github_organization_repository_role`
### Expected Behavior

The `github_organization_repository_role` resource should be created successfully on the first terraform apply and tracked correctly in state, allowing subsequent applies to be idempotent.

### Actual Behavior

On the first apply, Terraform fails with: Provider produced inconsistent result after apply: Root object was present, but now absent. Despite the error, the role is actually created in GitHub.

A second apply fails with: 422 Name has already been taken
because Terraform attempts to recreate the role it failed to record in state.

### Terraform Version

Terraform v1.11.1
on darwin_arm64
+ provider registry.terraform.io/integrations/github v6.7.0

### Affected Resource(s)

- github_organization_repository_role

```hcl
repository_roles = {
    "codeowner" = {
      description = "Maintain plus bypass branch protection and resolve scanning alerts"
      base_role   = "maintain"
      permissions = [
        "bypass_branch_protection",
        "delete_alerts_code_scanning",
        "resolve_secret_scanning_alerts",
        "view_secret_scanning_alerts",
      ]
    }
    "developer" = {
      description = "Write plus view secret scanning alerts"
      base_role   = "write"
      permissions = ["view_secret_scanning_alerts"]
    }
  }

resource "github_organization_repository_role" "this" {
  for_each    = var.repository_roles
  name        = each.key
  description = each.value.description
  base_role   = each.value.base_role
  permissions = each.value.permissions
}
```

### Steps to Reproduce

```
$ terraform apply
```

### Debug Output

```shell

```

### Panic Output

```shell
╷
│ Error: Provider produced inconsistent result after apply
│ 
│ When applying changes to module.org.github_organization_repository_role.this["developer"], provider "provider[\"registry.terraform.io/integrations/github\"]" produced an unexpected new value: Root object was present, but now absent.
│ 
│ This is a bug in the provider, which should be reported in the provider's own issue tracker.
╵


╷
│ Error: error creating GitHub organization repository role (prisamedia-training-sandbox/developer): POST https://api.github.com/orgs/prisamedia-training-sandbox/custom-repository-roles: 422 Name has already been taken []
│ 
│   with module.org.github_organization_repository_role.this["developer"],
│   on ../../modules/github-organization/main.tf line 98, in resource "github_organization_repository_role" "this":
│   98: resource "github_organization_repository_role" "this" {
│ 
╵
```

### Code of Conduct

### LINKED ISSUE #2960: [BUG]: github_organization_role base_role effectively can't be set to `none`
### Expected Behavior

`github_organization_role` contains both org-scoped and repo-scoped permissions. A base role is only required if setting repo-scoped permissions; and is not desirable in orgs where the base repository role for the org is set to "None".

However, the module doesn't seem to support either omitting this argument, or setting it to `none`.

Additionally, `ignore_changes` on `base_role` doesn't help because _other_ changes to the resource will still attempt to set this on a `PATCH` and trigger the below error messages.

### Actual Behavior

I can't omit the `base_role` argument or explicitly null it:

```
│ Error: error updating GitHub custom organization role ([REDACTED]): PATCH https://api.github.com/orgs/[REDACTED]: 422 Invalid request.
│ 
│ Invalid property /base_role: `` is not a possible value. Must be one of the following: none, read, triage, write, maintain, admin. []
│ 
│   with github_organization_role.base,
│   on roles.tf line 4, in resource "github_organization_role" "base":
│    4: resource "github_organization_role" "base" {
│ 
╵
```

But I also can't set it to `none`:

```
╷
│ Error: Error
│ 
│   with github_organization_role.base,
│   on roles.tf line 14, in resource "github_organization_role" "base":
│   14:   base_role = "none"
│ 
│ none is an invalid value for argument [{{} base_role}]
```

Or set it to an empty string:

```
╷
│ Error: Error
│ 
│   with github_organization_role.base,
│   on roles.tf line 14, in resource "github_organization_role" "base":
│   14:   base_role = ""
│ 
│  is an invalid value for argument [{{} base_role}]
╵
```

### Terraform Version

```
Terraform v1.13.5
on linux_amd64
+ provider registry.terraform.io/integrations/github v6.8.3

Your version of Terraform is out of date! The latest version
is 1.14.0. You can update by downloading from https://developer.hashicorp.com/terraform/install
```

### Affected Resource(s)

- github_organization_role

```hcl
resource "github_organization_role" "base" {
  name        = "Organization base permissions"

  base_role = "none"
  permissions = [
    "edit_org_custom_properties_values",
    "read_custom_properties_for_organizations",
  ]
  lifecycle {
      ignore_changes = [base_role]
  }
}
```

### Steps to Reproduce

_No response_

### Debug Output

```shell

```

### Panic Output

```shell

```

### Code of Conduct

stevehipwell:
Resolves #2805
Resolves #2960

This PR improves on the code in #2487.

----

### Before the change?

* Org roles re-ran the read logic
* Org roles didn't use a context correctly
* `github_organization_role` didn't correctly support org only permissions

### After the change?

* Org roles don't re-read
* Org roles use context
* `github_organization_role` can support org only permissions or no permissions

### Pull request checklist

- [ ] Yes
- [x] No

----

### DIFF
--- a/github/resource_github_organization_role.go
+++ b/github/resource_github_organization_role.go
@@ -9,6 +9,7 @@ import (
 	"strconv"
 
 	"github.com/google/go-github/v67/github"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
 )
 
@@ -16,10 +17,10 @@ func resourceGithubOrganizationRole() *schema.Resource {
 	return &schema.Resource{
 		Description: "Manage a custom organization role.",
 
-		Create: resourceGithubOrganizationRoleCreate,
-		Read:   resourceGithubOrganizationRoleRead,
-		Update: resourceGithubOrganizationRoleUpdate,
-		Delete: resourceGithubOrganizationRoleDelete,
+		CreateContext: resourceGithubOrganizationRoleCreate,
+		ReadContext:   resourceGithubOrganizationRoleRead,
+		UpdateContext: resourceGithubOrganizationRoleUpdate,
+		DeleteContext: resourceGithubOrganizationRoleDelete,
 		Importer: &schema.ResourceImporter{
 			StateContext: schema.ImportStatePassthroughContext,
 		},
@@ -41,32 +42,30 @@ func resourceGithubOrganizationRole() *schema.Resource {
 				Optional:    true,
 			},
 			"base_role": {
-				Description:      "The base role for the organization role.",
+				Description:      "The system role from which this role inherits permissions.",
 				Type:             schema.TypeString,
 				Optional:         true,
-				Computed:         true,
-				ValidateDiagFunc: validateValueFunc([]string{"read", "triage", "write", "maintain", "admin"}),
+				Default:          "none",
+				ValidateDiagFunc: validateValueFunc([]string{"none", "read", "triage", "write", "maintain", "admin"}),
 			},
 			"permissions": {
 				Description: "The permissions for the organization role.",
 				Type:        schema.TypeSet,
 				Required:    true,
 				Elem:        &schema.Schema{Type: schema.TypeString},
-				MinItems:    1,
 			},
 		},
 	}
 }
 
-func resourceGithubOrganizationRoleCreate(d *schema.ResourceData, meta any) error {
+func resourceGithubOrganizationRoleCreate(ctx context.Context, d *schema.ResourceData, meta any) diag.Diagnostics {
 	err := checkOrganization(meta)
 	if err != nil {
-		return err
+		return diag.FromErr(err)
 	}
 
 	client := meta.(*Owner).v3client
 	orgName := meta.(*Owner).name
-	ctx := context.Background()
 
 	permissions := d.Get("permissions").(*schema.Set).List()
 	permissionsStr := make([]string, len(permissions))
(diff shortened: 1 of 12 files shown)


github/resource_github_organization_role.go
```diff
@@ -41,11 +42,10 @@ func resourceGithubOrganizationRole() *schema.Resource {
 				Optional:    true,
 			},
 			"base_role": {
-				Description:      "The base role for the organization role.",
+				Description:      "The system role from which this role inherits permissions.",
 				Type:             schema.TypeString,
-				Optional:         true,
-				Computed:         true,
-				ValidateDiagFunc: validateValueFunc([]string{"read", "triage", "write", "maintain", "admin"}),
+				Required:         true,
```
deiga: question: GH API doesn't mark this as required, if you found that it IS required, could you add a comment here?

github/resource_github_organization_role.go
```diff
@@ -41,11 +42,10 @@ func resourceGithubOrganizationRole() *schema.Resource {
...
+				ValidateDiagFunc: validateValueFunc([]string{"none", "read", "triage", "write", "maintain", "admin"}),
```
deiga, line 49: issue: There doesn't seem to be an option for `none` in https://docs.github.com/en/enterprise-cloud@latest/rest/orgs/organization-roles?apiVersion=2022-11-28#create-a-custom-organization-role

But is this an oversight on GH API docs?

deiga (commented):
Would you be able to add a test case which reproduces the error from the reported issue?

github/resource_github_organization_role.go
stevehipwell: The issue on this covers the error. Note that the GitHub REST API docs aren't great, especially in cases like this where the spec has changed.

This change is intentional, if `none` had been an option when I created the apartment for this resource this would have been the design I'd have chosen. Basically optional+ computed is best avoided if possible (which it often isn't). But if it works in some cases without this being set then this would be a major version change so I can change the implementation. I'm hoping the automation testing PRs are merged soon so this can be tested.

github/resource_github_organization_role.go
stevehipwell: Yeah, read the description text above.

stevehipwell:
> Would you be able to add a test case which reproduces the error from the reported issue?

@deiga there is already a test for the org role error, that worked prior to the API changing. The other error is a distributed system error that can't easily be tested, as it'd be proving a negative; but we do have tests. We've seen the same problem in other areas in the API and need to get rid of the re-read pattern. Interestingly the resource failing here is having to use a list for get pattern as the get API was a later addition and needs v68. Also interesting is that at least one report is from GHES.

github/resource_github_organization_role.go
stevehipwell: OK, I've changed the implementation to have a default as this will be non-breaking. Based on the previous logic and the reported error it looks like the implementation broke when the API swapped out `""` for `"none"`.

I've added extra test coverage and I've tested this manually using the API/UI, but as this is an enterprise feature I can't directly run the unit tests.

PR #2970, opened 2025-12-05
chore: PR for moving secrets

stevehipwell:
**DO NOT MERGE!**

@nickfloyd if you add a PAT token to the `API_TOKEN` secret in the `hack` environment you should be able to re-run the hack workflow to move the secret. This can all happen on this PR, no need to merge anything to `main`.


nickfloyd:
✅ Soft approve for running workflow.  The PAT has been updated.

PR #2941, opened 2025-11-24
chore: Refactor integration tests

stevehipwell:
This PR replaces #2476.

----

### Before the change?

* The acceptance tests are complicated and don't all pass

### After the change?

- Acceptance tests have been refactored
  - Fewer tests
  - Simpler test structure
  - Test env setup has been made more granular and documented
- Acceptance tests have been tested
  - Tests introduced after [v6.4.0](https://github.com/integrations/terraform-provider-github/releases/tag/v6.4.0) haven't been tested
  - Enterprise tests and external user tests have been skipped

### Outstanding

- [ ] Re-run all tests locally

### Pull request checklist
- [x] Schema migrations have been created if needed ([example](https://github.com/integrations/terraform-provider-github/pull/2820/files))
- [x] Tests for the changes have been added (for bug fixes / features)
- [x] Docs have been reviewed and added / updated if needed (for bug fixes / features)

### Does this introduce a breaking change?

Please see our docs on [breaking changes](https://github.com/octokit/.github/blob/master/community/breaking_changes.md) to help!

- [ ] Yes
- [x] No

----

PR #2986, opened 2025-12-08
chore: Refactor integration tests

### LINKED ISSUE #2983: [BUG]: The `github_organization_role_teams` & `github_organization_role_users` data sources error
### Expected Behavior

The `github_organization_role_teams` & `github_organization_role_users` data sources should query the organization without erroring

### Actual Behavior

The `github_organization_role_teams` & `github_organization_role_users` data sources contain a typo and error.

### Terraform Version

n/a

### Affected Resource(s)

- `github_organization_role_teams` (data source)
- `github_organization_role_users` (data source)

```hcl

```

### Steps to Reproduce

_No response_

### Debug Output

```shell

```

### Panic Output

```shell

```

### Code of Conduct

### LINKED ISSUE #2984: [BUG]: The `github_repository_deployment_branch_policies` data source errors
### Expected Behavior

The `github_repository_deployment_branch_policies` data source should correctly lookup he policies.

### Actual Behavior

The `github_repository_deployment_branch_policies` data source errors due to a typo.

### Terraform Version

n/a

### Affected Resource(s)

- `github_repository_deployment_branch_policies` (data source)

```hcl

```

### Steps to Reproduce

_No response_

### Debug Output

```shell

```

### Panic Output

```shell

```

### Code of Conduct

### LINKED ISSUE #2985: [BUG]: The `github_organization_custom_properties` resource can't be created with `values_editable_by` unset
### Expected Behavior

I should be able to create a `github_organization_custom_properties` resource and get the API generated value for `values_editable_by` back.

### Actual Behavior

I can't create a `github_organization_custom_properties` resource without explicitly setting the `values_editable_by` field.

### Terraform Version

n/a

### Affected Resource(s)

- `github_organization_custom_properties`

```hcl

```

### Steps to Reproduce

_No response_

### Debug Output

```shell

```

### Panic Output

```shell

```

### Code of Conduct

### LINKED ISSUE #3006: [BUG]: Acceptance Tests Failing
### Expected Behavior

Acceptance tests should pass

### Actual Behavior

Fails on the test `creates and manages files on default branch if branch is omitted` for both the individual and organization account.

It seems to be failing the test cleanup step because Github throws an error when it attempts to delete the default branch. 

```
=== RUN   TestAccGithubRepositoryFile/creates_and_manages_files_on_default_branch_if_branch_is_omitted/with_an_individual_account
    testing_new.go:85: Error running post-test destroy, there may be dangling resources: exit status 1
        
        Error: error deleting GitHub branch reference test-os-org/tf-acc-test-i03pq (refs/heads/test): DELETE https://api.github.com/repos/####/tf-acc-test-i03pq/git/refs/heads/test: 422 Cannot delete the default branch []
```

### Terraform Version

Terraform v1.14.2
on darwin_arm64

### Affected Resource(s)

- github_repository

```hcl
None
```

### Steps to Reproduce

Set env variables 
```go test -v ./... -run ^TestAccGithubRepositoryFile```

### Debug Output

```shell

```

### Panic Output

```shell

```

### Code of Conduct

stevehipwell:
Resolves #2983
Resolves #2984 
Resolves #2985
Resolves #3006
Closes #2941

----

### Before the change?

- The acceptance tests are complicated and some are broken
- Some of the code is incorrect

### After the change?

- Acceptance tests have been refactored
  - Fewer tests
  - Simpler test structure
  - Test env setup has been made more granular and documented
- Acceptance tests have been run
  - Some tests that require paid features haven't been run yet

### Pull request checklist

- [ ] Yes
- [x] No

----


github/data_source_github_branch_test.go
```diff
@@ -31,76 +31,49 @@ func TestAccGithubBranchDataSource(t *testing.T) {
...
+	// 		PreCheck:          func() { skipUnauthenticated(t) },
+	// 		ProviderFactories: providerFactories,
+	// 		Steps: []resource.TestStep{
+	// 			{
+	// 				Config: config,
+	// 				Check:  check,
+	// 			},
+	// 		},
+	// 	})
+	// })
```
deiga, line 78: question: Is there value in keeping this test? The API for refs returns 404 if a ref doesn't exist. I don't understand why this behaviour would be expected from the provider

github/data_source_github_ref_test.go
```diff
@@ -2,104 +2,73 @@ package github
...
+	// 		PreCheck:          func() { skipUnauthenticated(t) },
+	// 		ProviderFactories: providerFactories,
+	// 		Steps: []resource.TestStep{
+	// 			{
+	// 				Config: config,
+	// 				Check:  check,
+	// 			},
+	// 		},
+	// 	})
+	// })
```
deiga, line 73: Same as in https://github.com/integrations/terraform-provider-github/pull/2986/files#r2601299558

github/data_source_github_repository_file_test.go
```diff
@@ -133,376 +141,105 @@ func TestAccGithubRepositoryFileDataSource(t *testing.T) {
...
+	// 					resource.TestCheckNoResourceAttr("data.github_repository_file.test", "sha"),
+	// 					resource.TestCheckNoResourceAttr("data.github_repository_file.test", "commit_sha"),
+	// 					resource.TestCheckNoResourceAttr("data.github_repository_file.test", "commit_message"),
+	// 					resource.TestCheckNoResourceAttr("data.github_repository_file.test", "commit_author"),
+	// 					resource.TestCheckNoResourceAttr("data.github_repository_file.test", "commit_email"),
+	// 				),
+	// 			},
+	// 		},
+	// 	})
+	// })
```
deiga, line 205: question: Similar to https://github.com/integrations/terraform-provider-github/pull/2986/files#r2601299558, is there value in keeping this as it's not something the API supports?

github/data_source_github_rest_api_test.go
```diff
@@ -127,64 +98,34 @@ func TestAccGithubRestApiDataSource(t *testing.T) {
...
+	// 		PreCheck:          func() { skipUnauthenticated(t) },
+	// 		ProviderFactories: providerFactories,
+	// 		Steps: []resource.TestStep{
+	// 			{
+	// 				Config:      config,
+	// 				ExpectError: regexp.MustCompile("Error: GET https://api.github.com/xx.*: 414"),
+	// 			},
+	// 		},
+	// 	})
+	// })
```
deiga, line 130: question: Why is this commented out?

github/data_source_github_team_test.go
```diff
@@ -192,33 +138,72 @@ func TestAccGithubTeamDataSource(t *testing.T) {
...
+			repository = github_repository.test.name
+			permission = "admin"
+		}
+
+		data "github_team" "test" {
+			slug = github_team.test.slug
 
-		t.Run("with an organization account", func(t *testing.T) {
-			testCase(t, organization)
+			depends_on = ["github_repository.test", "github_team.test", "github_team_repository.test"]
```
deiga: question: Do all of these dependents need to be set? I would assume that `github_team_repository` would already depend on `github_team` and `github_repository` implicitly.

If it doesn't should that be fixed?

Also: I don't think resources can be quoted there
```suggestion
			depends_on = [github_team_repository.test]
```

github/data_source_github_users_test.go
```diff
@@ -9,44 +9,34 @@ import (
 
 // TODO: this is failing.
 func TestAccGithubUsersDataSource(t *testing.T) {
+	if len(testAccConf.testExternalUser) == 0 {
+		t.Skip("No external user provided")
+	}
```
deiga, line 14: question: Would we benefit from this moving to `acc_test.go` as a `skipUnlessExternal` function?

deiga (commented):
Half-way through, looking good so far!

github/data_source_github_branch_test.go
stevehipwell: The question about the behaviour is why I've left them commented out. I think we'll need a few issues opened once this is merged to make iterative improvements.

github/data_source_github_rest_api_test.go
stevehipwell: The current REST API implementation swallows `404` errors, and I can't off the top of my head think of a way to repeatedly get a different error response with an anonymous call. This is another behaviour that needs discussing.

github/data_source_github_team_test.go
stevehipwell: The depends on needs to be set as it's a data source under test so the resources need to be applied first. Given the behaviour of depends on it's good practice to be explicit here as if the pattern is re-used in a multi step test not having all the resources in there could cause an issue or require additional work to validate if it's the reason for a failure.

> Also: I don't think resources can be quoted there

At first glance I'd have agreed with you (if I didn't know the tests were passing), but HCL is very permissive and TF has a lot of legacy baggage.

I'll update these ones, but this type of stylistic cleanup is something that can happen once the PR has landed.

github/data_source_github_users_test.go
stevehipwell: It would do, once we've got the tests actually running in CI and the concepts are correct. I'm pretty sure the org app pattern in the tests at the moment is wrong and will need replacing, but all of this would be easier being incremental changes on top of a largely working test suite.

scripts/fmtcheck.sh
```diff
@@ -0,0 +1,13 @@
+#!/usr/bin/env bash
+
+# Check formatting
+echo "==> Checking that code complies with fmt requirements..."
+fmt_diff="$(golangci-lint fmt --diff ./...)"
```
deiga: question: I presume that `golangci-lint fmt` is equally fast as `golangci-lint fmt --diff`, with that in mind, do we get any benefit from doing the diff before build? What if we just did the formatting before build automatically?

That could remove the need for this file as well

github/resource_github_organization_custom_properties.go
```diff
@@ -85,16 +85,19 @@ func resourceGithubCustomPropertiesCreate(d *schema.ResourceData, meta any) erro
...
-		ValuesEditableBy: &valuesEditableBy,
+		PropertyName:  &propertyName,
+		ValueType:     valueType,
+		Required:      &required,
+		DefaultValue:  &defaultValue,
+		Description:   &description,
+		AllowedValues: allowedValuesString,
+	}
+
+	if val, ok := d.GetOk("values_editable_by"); ok {
```
deiga, line 98: question: Since this is a computed value, aren't there cases where `ok` would be false, but `val` has a value? Should we ignore the value in those cases?

github/resource_github_organization_project_test.go
deiga, line 1: question: Is there a reason to keep this file around? Are we planning to re-use it for projectsV2?

github/resource_github_organization_ruleset_test.go
```diff
@@ -91,132 +88,115 @@ func TestGithubOrganizationRulesets(t *testing.T) {
...
+			Steps: []resource.TestStep{
+				{
+					Config: config,
+					Check: resource.ComposeTestCheckFunc(
+						resource.TestCheckResourceAttr("github_organization_ruleset.test", "name", "test"),
+						resource.TestCheckResourceAttr("github_organization_ruleset.test", "enforcement", "active"),
+						resource.TestCheckResourceAttr("github_organization_ruleset.test", "rules.0.required_workflows.0.do_not_enforce_on_create", "true"),
+						resource.TestCheckResourceAttr("github_organization_ruleset.test", "rules.0.required_workflows.0.required_workflow.0.path", "path/to/workflow.yaml"),
+						resource.TestCheckResourceAttr("github_organization_ruleset.test", "rules.0.required_workflows.0.required_workflow.0.repository_id", "1234"),
+						resource.TestCheckResourceAttr("github_repository_ruleset.test", "rules.0.required_code_scanning.0.required_code_scanning_tool.0.alerts_threshold", "errors"),
```
deiga:
```suggestion
						resource.TestCheckResourceAttr("github_organization_ruleset.test", "rules.0.required_code_scanning.0.required_code_scanning_tool.0.alerts_threshold", "errors"),
```

github/resource_github_organization_ruleset_test.go
```diff
@@ -91,132 +88,115 @@ func TestGithubOrganizationRulesets(t *testing.T) {
...
+						resource.TestCheckResourceAttr("github_repository_ruleset.test", "rules.0.required_code_scanning.0.required_code_scanning_tool.0.security_alerts_threshold", "high_or_higher"),
```
deiga:
```suggestion
						resource.TestCheckResourceAttr("github_organization_ruleset.test", "rules.0.required_code_scanning.0.required_code_scanning_tool.0.security_alerts_threshold", "high_or_higher"),
```

github/resource_github_organization_ruleset_test.go
```diff
@@ -91,132 +88,115 @@ func TestGithubOrganizationRulesets(t *testing.T) {
...
+						resource.TestCheckResourceAttr("github_repository_ruleset.test", "rules.0.required_code_scanning.0.required_code_scanning_tool.0.tool", "CodeQL"),
```
deiga:
```suggestion
						resource.TestCheckResourceAttr("github_organization_ruleset.test", "rules.0.required_code_scanning.0.required_code_scanning_tool.0.tool", "CodeQL"),
```

github/resource_github_project_card_test.go
deiga, line 1: Same question as for project tests

github/resource_github_project_column_test.go
deiga, line 1: Same question as for project tests

github/resource_github_repository_collaborators_test.go
```diff
@@ -52,9 +42,61 @@ func TestAccGithubRepositoryCollaborators(t *testing.T) {
...
+
+							collaborators := state.RootModule().Resources["github_repository_collaborators.test_repo_collaborators"].Primary
+							for name, val := range collaborators.Attributes {
+								if strings.HasPrefix(name, "user.") && strings.HasSuffix(name, ".username") && val != testAccConf.testExternalUser {
+									return fmt.Errorf("expected user.*.username to be set to %s, was %s", testAccConf.testExternalUser, val)
+								}
+								if strings.HasPrefix(name, "user.") && strings.HasSuffix(name, ".permission") && val != "push" {
+									return fmt.Errorf("expected user.*.permission to be set to push, was %s", val)
+								}
+							}
```
deiga, line 69: question: Could this be replaced with a `resource.TestCheckResourceAttr` function call?
Or is there a purpose to this which I'm not seeing?

Maybe something like:
```go
resource.TestCheckResourceAttr("github_repository_collaborators.test_repo_collaborators", "user.0.username", testAccConf.testExternalUser)
resource.TestCheckResourceAttr("github_repository_collaborators.test_repo_collaborators", "user.0.permission", "push")

```

github/resource_github_repository_collaborators_test.go
```diff
@@ -52,9 +42,61 @@ func TestAccGithubRepositoryCollaborators(t *testing.T) {
...
+
+							invites, _, err := conn.Repositories.ListInvitations(context.TODO(), owner, repoName, nil)
```
deiga: Do we have access to `t` in here?
```suggestion
							invites, _, err := conn.Repositories.ListInvitations(t.Context(), owner, repoName, nil)
```

scripts/fmtcheck.sh
stevehipwell: This is intended to error if the formatting is incorrect. But actually `golangci-lint run ./...` will handle this and is currently missing from the CI, so i can remove and add this check.

github/resource_github_organization_custom_properties.go
stevehipwell: This is in the resource create, we only want to send the value if it's been explicitly set. The computed part happenes when we set it after it's been created, at which point we can use that value. Without this change the implementation requires the value to be set as otherwise we're sending `""` instead of `nil`.

github/resource_github_organization_project_test.go
stevehipwell: The other files exist, so this should be removed with them.

github/resource_github_repository_collaborators_test.go
stevehipwell: It's been a while since I wrote this, but I'm pretty sure that it solves the unknown ordering issue.

github/resource_github_repository_collaborators_test.go
stevehipwell: We didn't have it when I created the PR, but we do now so I'll have a look at where we can use it.

github/resource_github_repository_collaborators_test.go
deiga: I get that, but there is only 1 collaborator in this config, right? If it's meant to be a helper for the sorting case, maybe it should be extracted for re-use?

github/resource_organization_block_test.go
```diff
@@ -64,10 +72,14 @@ func testAccCheckOrganizationBlockExists(n string) resource.TestCheckFunc {
...
-		orgName := testAccProvider.Meta().(*Owner).name
+		meta, err := getTestMeta()
+		if err != nil {
+			return err
+		}
+		conn := meta.v3client
+		orgName := meta.name
 
-		blocked, _, err := conn.Organizations.IsBlocked(context.TODO(), orgName, username)
+		blocked, _, err := conn.Organizations.IsBlocked(context.Background(), orgName, username)
```
deiga, line 82: question: would we get access to `t.Context()` if we changed the function definition to `func testAccCheckOrganizationBlockExists(ctx context.Context, n string)`?

github/resource_github_repository_collaborators_test.go
stevehipwell: Yeah it could be, but given the sheer amount of resources I've put into this PR I'd just like to get it merged and then it can be iterated on. I'd hate to lose something I spent time on that turns out later to be important. Once the code is merged anything removed can be retrieved, I think I've already lost work due to needing to get the tests passing.

github/resource_organization_block_test.go
stevehipwell: I cascaded `t.Context()` everywhere I could (other than for projects where I just got rid of `context.TODO()`), but some usage comes from typed functions being called directly. We could wrap up these function calls to pass it in, but IMHO this is a problem for another day.

github/resource_github_repository_collaborators_test.go
deiga: Sure! Most of my suggestions are usually "address now, if capacity, otherwise do a follow-up". 

I should learn to write that, since usually people can't rind my mind! 😬

github/resource_organization_block_test.go
deiga: Definitely, let's get this shippable!

github/resource_github_repository_collaborators_test.go
stevehipwell: That's fine. If this was a smaller piece of work or even if it was all fresh in my mind I'd be making more changes and being less cautious about removing code.

CONTRIBUTING.md
```diff
@@ -137,87 +145,81 @@ The following provider development overrides are set in the CLI configuration:
...
-			"name": "Attach to Process",
-			"type": "go",
-			"request": "attach",
-			"mode": "local",
-			"processId": 0
-		}
-	]
+  "go.testEnvVars": {
+    "TF_ACC": "1",
+    "GITHUB_TOKEN": "<TOKEN>",
```
deiga, line 202:
```suggestion
    "GH_TOKEN": "<TOKEN>",
```

github/resource_github_repository_collaborator_test.go
```diff
@@ -154,10 +122,10 @@ func TestAccGithubRepositoryCollaboratorArchivedRepo(t *testing.T) {
...
 
 	t.Run("can delete collaborators from archived repositories without error", func(t *testing.T) {
-		// Note: This test requires GITHUB_TEST_COLLABORATOR to be set to a valid GitHub username
-		testCollaborator := os.Getenv("GITHUB_TEST_COLLABORATOR")
+		// Note: This test requires GH_TEST_COLLABORATOR to be set to a valid GitHub username
+		testCollaborator := os.Getenv("GH_TEST_COLLABORATOR")
 		if testCollaborator == "" {
-			t.Skip("GITHUB_TEST_COLLABORATOR not set, skipping archived repository collaborator test")
+			t.Skip("GH_TEST_COLLABORATOR not set, skipping archived repository collaborator test")
 		}
```
deiga: question: Should this use `testAccConf.testExternalUser` as well?

github/acc_test.go
```diff
@@ -0,0 +1,291 @@
...
+		testExternalUser:                  os.Getenv("GH_TEST_EXTERNAL_USER"),
+		testExternalUserToken:             os.Getenv("GH_TEST_EXTERNAL_USER_TOKEN"),
+		testExternalUser2:                 os.Getenv("GH_TEST_EXTERNAL_USER2"),
+		testAdvancedSecurity:              os.Getenv("GH_TEST_ADVANCED_SECURITY") == "true",
+	}
+
+	if config.authMode != anonymous {
+		config.owner = os.Getenv("GITHUB_OWNER")
+		config.username = os.Getenv("GITHUB_USERNAME")
+		config.token = os.Getenv("GITHUB_TOKEN")
```
deiga, line 127: question: The [CONTRIBUTING.md](https://github.com/integrations/terraform-provider-github/pull/2986/changes#diff-eca12c0a30e25b4b46522ebf89465a03ba72a03f540796c979137931d8f92055) talks about `GH_OWNER`, `GH_USERNAME` and `GH_TOKEN` which one is it?

GNUmakefile
```diff
@@ -27,8 +29,8 @@ test:
 	CGO_ENABLED=0 go test ./...
 	# commenting this out for release tooling, please run testacc instead
 
-testacc: fmtcheck
-	TF_ACC=1 CGO_ENABLED=0 go test $(TEST) -v $(TESTARGS) -timeout 120m
+testacc:
+	TF_ACC=1 CGO_ENABLED=0 go test -run "^TestAcc*" $(TEST) -v $(TESTARGS) -timeout 120m -count=1
```
deiga, line 33: question: With this format there is no way to run only a specific subset of acctests, is that intentional?

deiga (commented):
There are a couple tests commented out in this PR. They seem to cover functionality which isn't covered otherwise, are those just WIP refactoring or is there some other purpose?

stevehipwell:
@nickfloyd @deiga I think this PR is ready to be merged. All of the tests I'm able to run locally have been run and passed.

GNUmakefile
stevehipwell: Yes, the other tests are run in the CI workflow.

github/acc_test.go
stevehipwell: Good spot.

github/resource_github_repository_collaborator_test.go
stevehipwell: Good spot.

CONTRIBUTING.md
stevehipwell: `GITHUB_TOKEN` is correct.

github/test-fixtures/README.md
```diff
@@ -1,4 +1,4 @@
-# Hi fellow bots and humans :wave:
+# Hi fellow bots and humans :wave
```
nickfloyd:
```suggestion
# Hi fellow bots and humans :wave:
```

github/resource_github_repository_collaborator_test.go
```diff
@@ -79,7 +60,7 @@ func TestAccGithubRepositoryCollaborator(t *testing.T) {
 				username   = "<username>"
```
deiga:
```suggestion
				username   = "%s"
```

deiga (approved):

github/resource_github_repository_collaborator_test.go
stevehipwell: I applied the change but it broke the automation. I'm now trying to figure if it was correct before or if it needed the index. I'm on my phone so tricky to follow.

stevehipwell: I applied the change but it broke the automation. I'm now trying to figure if it was correct before or if it needed the index. I'm on my phone so tricky to follow.

stevehipwell: Fixed.

nickfloyd (approved):

PR #2547, opened 2025-01-24
Add support for GitHub Enterprise Cloud Data residency

EricDales:
Resolves #2546

----

### Before the change?

* setting BaseURL to a value other than api.github.com implies the instance is considered as GHES instance

### After the change?

* provider is able to identify GHEC data residency URL and use the correct API paths

### Pull request checklist
- [ ] Tests for the changes have been added (for bug fixes / features)
- [ ] Docs have been reviewed and added / updated if needed (for bug fixes / features)

### Does this introduce a breaking change?

Please see our docs on [breaking changes](https://github.com/octokit/.github/blob/master/community/breaking_changes.md) to help!

- [ ] Yes
- [X] No

----

PR #2997, opened 2025-12-12
fix: append /api/v3 and /api/graphql paths for GHEC Data Residency hosts

JiayangZhou:
## Description

This PR fixes a regression introduced in v6.6.0 where GHEC Data Residency hosts (`*.ghe.com`) were incorrectly treated like `github.com`, resulting in the API being unreachable.

## Problem

In PR #2547, support for GHEC Data Residency was added with the assumption that `*.ghe.com` hosts behave like `github.com` (i.e., they don't need `/api/v3` or `/api/graphql/` paths). However, this assumption is incorrect - these hosts actually behave like GitHub Enterprise Server and **do require** the API paths.

### Symptoms
- With v6.5.0: Users configured `base_url = "https://customer.ghe.com/api/v3"` and it worked
- With v6.6.0: Validation requires removing `/api/v3`, but then the provider fails to reach the API
- Error: `this resource can only be used in the context of an organization, "ORG_NAME" is a user`

## Solution

Remove the `GHECDataResidencyHostMatch` exclusion from the path construction logic in:
- `config.go`: `NewGraphQLClient()` and `NewRESTClient()`
- `apps.go`: `getInstallationAccessToken()`

This ensures all non-`api.github.com` hosts (including `*.ghe.com`) will have the correct API paths appended.

## Testing

- [x] All existing tests pass
- [x] Manually tested with a real GHEC Data Residency environment (`hidden-staging.ghe.com`)
- [x] `terraform plan` successfully refreshes all GitHub resources

## Related Issues

Fixes the issue reported in the diff comparison: https://github.com/integrations/terraform-provider-github/compare/v6.5.0...v6.6.0#diff-062b7a475c9ddd7765fd96c620a3f7854307a4b68c0ff57db24b9d912b5bcec9R88


github/apps.go
```diff
@@ -34,7 +34,8 @@ func GenerateOAuthTokenFromApp(baseURL *url.URL, appID, appInstallationID, pemDa
 
 func getInstallationAccessToken(baseURL *url.URL, jwt, installationID string) (string, error) {
 	hostname := baseURL.Hostname()
-	if hostname != DotComHost && !GHECDataResidencyHostMatch.MatchString(hostname) {
+	if hostname != DotComHost {
+		// GHE Server and GHEC Data Residency hosts need /api/v3 path
 		baseURL.Path = path.Join(baseURL.Path, "api/v3/")
```
JiayangZhou: should we start to append `api/v3/` for GHEC Data Residency hosts? @deiga please have a look at this Pr when you get time thanks!

JiayangZhou: sorry just pulled upstream, i think it already contained the fix

PR #3011, opened 2025-12-18
[BUG] `github_organization_role` Ensure `role_id` is set after Create

### LINKED ISSUE #3004: [BUG]: github_organization_role setting base_role = "none" causes permanent diff
### Expected Behavior

Given a configuration like
```
resource "github_organization_role" "audit_viewer" {
  name        = "Audit Viewer"
  description = "Access the Organization Audit Log"
  base_role   = "none"
  permissions = [
    "read_audit_logs"
  ]
}
```
it should get applied once and on subsequent plans do not cause any diff.

### Actual Behavior

```
  # github_organization_role.audit_viewer will be updated in-place
  ~ resource "github_organization_role" "audit_viewer" ***
      + base_role   = "none"
        id          = "12345"
        name        = "Audit Viewer"
    ***
```

That change is applied on every subsequent run.

### Terraform Version

```
$ terraform -v
Terraform v1.14.2
on linux_amd64
+ provider registry.terraform.io/hashicorp/google v7.13.0
+ provider registry.terraform.io/hashicorp/googleworkspace v0.7.0
+ provider registry.terraform.io/integrations/github v6.9.0
```

### Affected Resource(s)

- github_organization_role

```hcl

```

### Steps to Reproduce

Run `terraform apply` twice or more.

### Debug Output

```shell

```

### Panic Output

```shell

```

### Code of Conduct

### LINKED ISSUE #3009: [BUG]: `github_organization_role_team` doesn't wait for role_id
### Expected Behavior

When creating a `github_organization_role`-resource and trying to use the `role_id` afterwards, the resource fails with the following error:
```
│   43:   role_id   = github_organization_role.example.role_id
│ 
│ The argument "role_id" is required, but no definition was found.
```
When re-runing apply with the created `github_organization_role` (from first run) it finds the `role_id`.

I would expect the resource to provide the `role_id` also when creating.

### Actual Behavior

See above

### Terraform Version

Terraform v1.11.4
on darwin_arm64

### Affected Resource(s)

- github_organization_role_team

```hcl
resource "github_organization_role" "example" {
  name      = "Example Reader"
  base_role = "read"
}

resource "github_organization_role_team" "example" {
  role_id   = github_organization_role.example.role_id
  team_slug = "example"
}

```

### Steps to Reproduce

_No response_

### Debug Output

```shell

```

### Panic Output

```shell

```

### Code of Conduct

deiga:
Resolves #3009
Resolves #3004

----

### Before the change?

* `role_id` wouldn't not be set during `apply`, only on next `plan`, thus rendering the resource unusable for references

### After the change?

* `role_id` is set in Create allowing it to be used in references

### Pull request checklist

- [ ] Yes
- [x] No

----

### DIFF
--- a/github/resource_github_organization_role.go
+++ b/github/resource_github_organization_role.go
@@ -84,6 +84,9 @@ func resourceGithubOrganizationRoleCreate(ctx context.Context, d *schema.Resourc
 	}
 
 	d.SetId(fmt.Sprint(role.GetID()))
+	if err = d.Set("role_id", role.GetID()); err != nil {
+		return diag.FromErr(err)
+	}
 	return nil
 }
 
--- a/github/resource_github_organization_role.go
+++ b/github/resource_github_organization_role.go
@@ -73,12 +73,18 @@ func resourceGithubOrganizationRoleCreate(ctx context.Context, d *schema.Resourc
 		permissionsStr[i] = v.(string)
 	}
 
-	role, _, err := client.Organizations.CreateCustomOrgRole(ctx, orgName, &github.CreateOrUpdateOrgRoleOptions{
+	createOrUpdateOrgRoleOptions := &github.CreateOrUpdateOrgRoleOptions{
 		Name:        github.String(d.Get("name").(string)),
 		Description: github.String(d.Get("description").(string)),
-		BaseRole:    github.String(d.Get("base_role").(string)),
 		Permissions: permissionsStr,
-	})
+	}
+
+	baseRole := d.Get("base_role").(string)
+	if baseRole != "none" {
+		createOrUpdateOrgRoleOptions.BaseRole = github.String(baseRole)
+	}
+
+	role, _, err := client.Organizations.CreateCustomOrgRole(ctx, orgName, createOrUpdateOrgRoleOptions)
 	if err != nil {
 		return diag.FromErr(fmt.Errorf("error creating organization role (%s/%s): %w", orgName, d.Get("name").(string), err))
 	}
@@ -126,8 +132,14 @@ func resourceGithubOrganizationRoleRead(ctx context.Context, d *schema.ResourceD
 	if err = d.Set("description", role.Description); err != nil {
 		return diag.FromErr(err)
 	}
-	if err = d.Set("base_role", role.BaseRole); err != nil {
-		return diag.FromErr(err)
+	if role.BaseRole != nil {
+		if err = d.Set("base_role", role.BaseRole); err != nil {
+			return diag.FromErr(err)
+		}
+	} else {
+		if err = d.Set("base_role", "none"); err != nil {
+			return diag.FromErr(err)
+		}
 	}
 	if err = d.Set("permissions", role.Permissions); err != nil {
 		return diag.FromErr(err)
--- a/github/resource_github_organization_role_test.go
+++ b/github/resource_github_organization_role_test.go
@@ -12,6 +12,7 @@ func TestAccGithubOrganizationRole(t *testing.T) {
 	t.Run("can create an empty organization role", func(t *testing.T) {
 		randomID := acctest.RandStringFromCharSet(5, acctest.CharSetAlphaNum)
 		name := fmt.Sprintf("tf-acc-org-role-%s", randomID)
+		baseRole := "none"
 		config := fmt.Sprintf(`
 			resource "github_organization_role" "test" {
 				name        = "%s"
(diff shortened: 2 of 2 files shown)


github/resource_github_organization_role_test.go
```diff
@@ -12,6 +12,7 @@ func TestAccGithubOrganizationRole(t *testing.T) {
 	t.Run("can create an empty organization role", func(t *testing.T) {
 		randomID := acctest.RandStringFromCharSet(5, acctest.CharSetAlphaNum)
 		name := fmt.Sprintf("tf-acc-org-role-%s", randomID)
+		baseRole := "none"
```
stevehipwell: This test is testing the optional base role defaulting to none, so this change breaks the test.

github/resource_github_organization_role_test.go
```diff
@@ -25,12 +26,12 @@ func TestAccGithubOrganizationRole(t *testing.T) {
...
 					Config: config,
-					Check: resource.ComposeTestCheckFunc(
+					Check: resource.ComposeAggregateTestCheckFunc(
 						resource.TestCheckResourceAttrSet("github_organization_role.test", "id"),
 						resource.TestCheckResourceAttrSet("github_organization_role.test", "role_id"),
 						resource.TestCheckResourceAttr("github_organization_role.test", "name", name),
-						resource.TestCheckResourceAttr("github_organization_role.test", "base_role", "none"),
-						resource.TestCheckResourceAttrSet("github_organization_role.test", "permissions.#"),
+						resource.TestCheckResourceAttr("github_organization_role.test", "base_role", baseRole),
+						resource.TestCheckNoResourceAttr("github_organization_role.test", "permissions"),
```
stevehipwell, line 33: Permissions should be an empty list.

github/resource_github_organization_role.go
```diff
@@ -73,17 +73,26 @@ func resourceGithubOrganizationRoleCreate(ctx context.Context, d *schema.Resourc
...
 		Description: github.String(d.Get("description").(string)),
-		BaseRole:    github.String(d.Get("base_role").(string)),
 		Permissions: permissionsStr,
-	})
+	}
+
+	baseRole := d.Get("base_role").(string)
+	if baseRole != "none" {
+		createOrUpdateOrgRoleOptions.BaseRole = github.String(baseRole)
+	}
```
stevehipwell, line 85: This doesn't match the API docs.

stevehipwell (changes requested):
The part of this PR fixing the role ID looks good, but I'm not sure about the other changes.

According to the [API docs](https://docs.github.com/en/enterprise-cloud@latest/rest/orgs/organization-roles?apiVersion=2022-11-28#create-a-custom-organization-role) the base role should be set to `"none"`. This is the fix applied in #2968 which this PR appears to be undoing?

github/resource_github_organization_role_test.go
deiga: I ran the tests on GHEC, the check kept failing until I changed it to this

deiga:
> The part of this PR fixing the role ID looks good, but I'm not sure about the other changes.
> 
> According to the [API docs](https://docs.github.com/en/enterprise-cloud@latest/rest/orgs/organization-roles?apiVersion=2022-11-28#create-a-custom-organization-role) the base role should be set to `"none"`. This is the fix applied in #2968 which this PR appears to be undoing?

Yeah, I see those docs, but running the tests on GHEC gives me a 422 Validation error if I'm passing in `"none"`.

When running a manual curl with `base_role: "none"` I get this response:
```json
{
	"message": "Invalid request.\n\nInvalid property /base_role: `none` is not a possible value. Must be one of the following: read, triage, write, maintain, admin.",
	"documentation_url": "https://docs.github.com/rest/orgs/organization-roles#create-a-custom-organization-role",
	"status": "422"
}
```

github/resource_github_organization_role.go
deiga: It doesn't, but it matches API behaviour on GHEC. https://github.com/integrations/terraform-provider-github/pull/3011#issuecomment-3675455499

github/resource_github_organization_role_test.go
deiga: This change doesn't modify the value of `base_role` in the resource, just the expected output. I did this, when I chose to do some modifications to the defaults of `base_role`.

But then I decided against doing changes to the resource API and this change was left in as harmless. I'll revert it, since it has no real function any more

github/resource_github_organization_role_test.go
stevehipwell: I think you're right.

stevehipwell (approved):
LGTM

nickfloyd (approved):

PR #3020, opened 2025-12-25
[MAINT]: Update GNUmakefile to enhance test command functionality

deiga:
- Renamed TESTARGS to COVERAGEARGS for clarity.
- Added variable references and examples for test-specific variables.
- Improved test and acceptance test commands to include branch information and coverage options.

Inspired by: https://github.com/hashicorp/terraform-provider-aws/blob/main/GNUmakefile

----

### Before the change?

* No way to declare a granular test pattern for `make testacc`
* Coverage ran for all test commands

### After the change?

* Able to declare a granular test pattern for `make testacc T=TestAccFoo`
* Coverage is only run if `COV=true` when calling a make command

### Pull request checklist
- [ ] ~Tests for the changes have been added (for bug fixes / features)~

- [ ] Yes
- [x] No

----

### DIFF
--- a/GNUmakefile
+++ b/GNUmakefile
@@ -60,7 +60,7 @@ test:
 testacc:
 	@branch=$$(git rev-parse --abbrev-ref HEAD); \
 	printf "==> Running acceptance tests on branch: \033[1m%s\033[0m...\n" "🌿 $$branch 🌿"
-	TF_ACC=1 CGO_ENABLED=0 go test $(TEST) -v -run '^TestAcc' $(RUNARGS) $(TESTARGS) -timeout 120m -count=1
+	TF_ACC=1 go test $(TEST) -v -run '^TestAcc' $(RUNARGS) $(TESTARGS) -timeout 120m -count=1
 
 test-compile:
 	@if [ "$(TEST)" = "./..." ]; then \


GNUmakefile
```diff
@@ -26,11 +46,20 @@ lintcheck:
 	golangci-lint run ./...
 
 test:
-	CGO_ENABLED=0 go test ./...
-	# commenting this out for release tooling, please run testacc instead
+	@branch=$$(git rev-parse --abbrev-ref HEAD); \
+	printf "==> Running acceptance tests on branch: \033[1m%s\033[0m...\n" "🌿 $$branch 🌿"
+	go test $(TEST) \
```
stevehipwell, line 52: It looks like `CGO_ENABLED=0` has been missed from here?

deiga: True. Question: Is there a reason we don't just put `export CGO_ENABLED=0` at the top of the makefile?

stevehipwell: Likely for readability/portability; the primary intent is much clearer with it defined inline.

GNUmakefile
```diff
@@ -2,7 +2,28 @@ TEST?=$$(go list ./... |grep -v 'vendor')
 WEBSITE_REPO=github.com/hashicorp/terraform-website
 PKG_NAME=github
 
-export TESTARGS=-race -coverprofile=coverage.txt -covermode=atomic
+COVERAGEARGS?=-race -coverprofile=coverage.txt -covermode=atomic
+export CGO_ENABLED=0
```
stevehipwell, line 6: I don't think this is the best pattern for this kind of repo (public with more consumers than contributors). Also this is still inlined for some commands.

GNUmakefile
deiga: I get that having a global export in the makefile obfuscates the usage of `CGO_ENABLED` slightly, but I don't see how it creates a problem with readability and portability. 
It's not like we need/should encourage people to use commands outside of the makefile.

And it makes maintenance of the makefile a lot simpler, since it's unlikely to be forgotten from a new command that way

GNUmakefile
deiga: I'll update the inlined one, let's continue the pattern discussion in the other thread

stevehipwell:
@deiga I'm not sure how I managed to close this (the GH UI has been causing me all kinds of trouble), would you like me to reopen it?

PR #3024, opened 2025-12-26
[BUG] Ensure repository creation works if vulnerability-alerts isn't modifiable on repo level

### LINKED ISSUE #2988: [BUG]: creation of github_repository fails from v6.7.0 onwards
### Expected Behavior

Applying the Terraform plan for  a new `github_repository` with `vulnerability_alerts = true` should create the repository without producing an error.

This is running in an GitHub Enterprise Cloud account.

It works fine with provider version `6.6.0` but fails with `6.7.0` to `6.9.0` (error below).

Possibly related to https://github.com/integrations/terraform-provider-github/pull/2501.

### Actual Behavior

Terraform errors, creates the repository but the state is broken with some attributes missing. Just applying Terraform again does not fix the issues.

```
 Error: DELETE https://api.github.com/repos/lessonnine/jts-test.test/vulnerability-alerts: 422 An enforced security configuration prevented modifying dependabot alerts enablement. Contact your organization owner for details. []
│ 
│   with github_repository.lessonnine-jts-test-test,
│   on generated.tf line 50028, in resource "github_repository" "lessonnine-jts-test-test":
│ 50028: resource "github_repository" "lessonnine-jts-test-test" {
```

### Terraform Version

Terraform version `1.11.4` on `linux_amd64`.

### Affected Resource(s)

- `github_repository`

```hcl
resource "github_repository" "lessonnine-jts-test-test" {
  name        = "jts-test.test"
  description = "Just for testing GitHub api (Infrastructure)"

  topics = [
    "infrastructure",
  ]

  has_issues   = false
  has_projects = true
  has_wiki     = false

  allow_rebase_merge = false
  allow_squash_merge = false

  visibility = "private"

  vulnerability_alerts = true # test adding this attribute

  auto_init              = true
  delete_branch_on_merge = true
  allow_update_branch    = true
  allow_auto_merge       = true

  lifecycle {
    prevent_destroy = true

    ignore_changes = [
      # vulnerability alerts are managed via org-wide settings
      vulnerability_alerts,
    ]
  }
}


with provider version:


    github = {
      version = "6.7.0"
    }
```

### Steps to Reproduce

```
$ terraform apply
```

### Debug Output

```shell
Error: DELETE https://api.github.com/repos/lessonnine/jts-test.test/vulnerability-alerts: 422 An enforced security configuration prevented modifying dependabot alerts enablement. Contact your organization owner for details. []
│ 
│   with github_repository.lessonnine-jts-test-test,
│   on generated.tf line 50028, in resource "github_repository" "lessonnine-jts-test-test":
│ 50028: resource "github_repository" "lessonnine-jts-test-test" {
```

### Panic Output

```shell
none
```

### Code of Conduct

deiga:
Resolves #2988 

----

### Before the change?

* `resource_github_repository.go` was using Context-unaware functions
* All errors in `updateVulnerabilityAlerts` would cause terraform to fail

### After the change?

* `resource_github_repository.go` is using Context-aware functions
* Ignore specific 422 response error in  `updateVulnerabilityAlerts` to ensure repo creation works if vulnerability alerts are not modifiable

### Pull request checklist

- [ ] Yes
- [x] No

----

### DIFF
--- a/github/resource_github_repository.go
+++ b/github/resource_github_repository.go
@@ -10,6 +10,7 @@ import (
 	"strings"
 
 	"github.com/google/go-github/v67/github"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/customdiff"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/validation"
@@ -17,12 +18,12 @@ import (
 
 func resourceGithubRepository() *schema.Resource {
 	return &schema.Resource{
-		Create: resourceGithubRepositoryCreate,
-		Read:   resourceGithubRepositoryRead,
-		Update: resourceGithubRepositoryUpdate,
-		Delete: resourceGithubRepositoryDelete,
+		CreateContext: resourceGithubRepositoryCreate,
+		ReadContext:   resourceGithubRepositoryRead,
+		UpdateContext: resourceGithubRepositoryUpdate,
+		DeleteContext: resourceGithubRepositoryDelete,
 		Importer: &schema.ResourceImporter{
-			State: func(d *schema.ResourceData, meta any) ([]*schema.ResourceData, error) {
+			StateContext: func(ctx context.Context, d *schema.ResourceData, meta any) ([]*schema.ResourceData, error) {
 				if err := d.Set("auto_init", false); err != nil {
 					return nil, err
 				}
@@ -620,18 +621,17 @@ func resourceGithubRepositoryObject(d *schema.ResourceData) *github.Repository {
 	return repository
 }
 
-func resourceGithubRepositoryCreate(d *schema.ResourceData, meta any) error {
+func resourceGithubRepositoryCreate(ctx context.Context, d *schema.ResourceData, meta any) diag.Diagnostics {
 	client := meta.(*Owner).v3client
 
 	if branchName, hasDefaultBranch := d.GetOk("default_branch"); hasDefaultBranch && (branchName != "main") {
-		return fmt.Errorf("cannot set the default branch on a new repository to something other than 'main'")
+		return diag.Errorf("cannot set the default branch on a new repository to something other than 'main'")
 	}
 
 	repoReq := resourceGithubRepositoryObject(d)
 	owner := meta.(*Owner).name
 
 	repoName := repoReq.GetName()
-	ctx := context.Background()
 
 	// determine if repository should be private. assume public to start
 	isPrivate := false
@@ -657,7 +657,7 @@ func resourceGithubRepositoryCreate(d *schema.ResourceData, meta any) error {
 		for _, templateConfigBlock := range templateConfigBlocks {
 			templateConfigMap, ok := templateConfigBlock.(map[string]any)
 			if !ok {
-				return errors.New("failed to unpack template configuration block")
+				return diag.FromErr(errors.New("failed to unpack template configuration block"))
 			}
 
 			templateRepo := templateConfigMap["repository"].(string)
@@ -678,7 +678,7 @@ func resourceGithubRepositoryCreate(d *schema.ResourceData, meta any) error {
 				&templateRepoReq,
 			)
 			if err != nil {
-				return err
+				return diag.FromErr(err)
 			}
 
 			d.SetId(*repo.Name)
@@ -691,7 +691,7 @@ func resourceGithubRepositoryCreate(d *schema.ResourceData, meta any) error {
 		log.Printf("[INFO] Creating fork of %s/%s in %s", sourceOwner, sourceRepo, owner)
 
 		if sourceOwner == "" || sourceRepo == "" {
-			return fmt.Errorf("source_owner and source_repo must be provided when forking a repository")
+			return diag.Errorf("source_owner and source_repo must be provided when forking a repository")
 		}
 
 		// Create the fork using the GitHub client library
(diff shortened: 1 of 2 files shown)


github/resource_github_repository.go
```diff
@@ -1243,11 +1243,17 @@ func resourceGithubParseFullName(resourceDataLike interface {
 }
 
 func updateVulnerabilityAlerts(d *schema.ResourceData, client *github.Client, ctx context.Context, owner, repoName string) error {
-	updateVulnerabilityAlerts := client.Repositories.DisableVulnerabilityAlerts
-	if vulnerabilityAlerts, ok := d.GetOk("vulnerability_alerts"); ok && vulnerabilityAlerts.(bool) {
-		updateVulnerabilityAlerts = client.Repositories.EnableVulnerabilityAlerts
+	updateVulnerabilityAlertsSDK := client.Repositories.DisableVulnerabilityAlerts
+	vulnerabilityAlerts, ok := d.GetOk("vulnerability_alerts")
+	if ok && vulnerabilityAlerts.(bool) {
```
stevehipwell, line 1251: Would it not be even cleaner to use a pattern more like the following?

```go
	updateVulnerabilityAlertsSDK := client.Repositories.DisableVulnerabilityAlerts
  vulnerabilityAlerts, ok := d.GetOk("vulnerability_alerts")
  
  if !ok {
    return nil
  }

	if vulnerabilityAlerts.(bool) {
		updateVulnerabilityAlertsSDK = client.Repositories.EnableVulnerabilityAlerts
	}
```

deiga: Good question! Unfortunately, not. New repos will have `vulnerability_alerts` enabled by default, so they need to be disabled unless it's actually set by the user to be enabled.

Of course we could change it to behave differently, but that'd be a breaking change

deiga: Maybe it could make sense to extract `vulnerability_alerts` into it's own resource in v7.x and make that breaking change there

stevehipwell: > Unfortunately, not. New repos will have vulnerability_alerts enabled by default, so they need to be disabled unless it's actually set by the user to be enabled.

I don't think this is the case anymore as AFAIK this can be configured at the organization level and is the cause of this defect. Given this GitHub change the code shouldn't make any changes to the default unless explicitly specified by the user.

> Of course we could change it to behave differently, but that'd be a breaking change

IMHO this is a required change and while it could be considered breaking leaving the current behaviour isn't an option.

> Maybe it could make sense to extract vulnerability_alerts into it's own resource in v7.x and make that breaking change there

I'd say this should be part of `v6.10.0` if possible with the inline repo variable deprecated.

github/resource_github_repository.go
```diff
@@ -1243,11 +1243,20 @@ func resourceGithubParseFullName(resourceDataLike interface {
...
+	// Only if the vulnerability alerts are specifically set to true, enable them.
+	// Otherwise, disable them as GitHub defaults to enabled and we have not wanted to introduce a breaking change for this yet.
+	if ok && vulnerabilityAlerts.(bool) {
+		updateVulnerabilityAlertsSDK = client.Repositories.EnableVulnerabilityAlerts
 	}
 
-	_, err := updateVulnerabilityAlerts(ctx, owner, repoName)
+	resp, err := updateVulnerabilityAlertsSDK(ctx, owner, repoName)
+	if err != nil {
+		if resp.StatusCode == http.StatusUnprocessableEntity && strings.Contains(err.Error(), "An enforced security configuration prevented modifying") && !ok {
```
stevehipwell, line 1259: This would be a brittle fix, the documented REST API is hard enough to work with so using an error message to control logic isn't something I'd be comfortable doing.

I think the current behaviour is defective given the subsequent GitHub changes. Based on my understanding of `d.GetOK()` a change like the one I suggested above would only impact new public repos where there is no organization controlled default.

stevehipwell (changes requested):

github/resource_github_repository.go
deiga: I did try this with my own non-paid Org and without having anything configured both `private` and `public` repos had `vulnerability_alerts` by default.

I agree that we should deprecate this ASAP, as having this enabled by default makes a lot of sense

github/resource_github_repository.go
deiga: I did try using `d.GetOk()` and skipping Delete/Enable with a non-paid Org and private repo. That left the state in a weird state, as vulnerability_alerts was enabled by default and TF had trouble consolidating configuration vs remote state.

I agree that using the error message is problematic, but I can't see any other way to do this with our current limitations.
There are many other cases where a 422 status code shouldn't be ignored, so we can't base it on that either

github/resource_github_repository.go
stevehipwell: The comment in code says that this should be off by default for private repos but the [docs](https://docs.github.com/en/code-security/dependabot/dependabot-alerts/about-dependabot-alerts#access-to-dependabot-alerts) suggest that this has been modified to always be on by default.

github/resource_github_repository.go
stevehipwell: Actually there isn't a good solution for this, even with my pattern the API will be hit every time there is an unrelated repo change. I guess we could remove computed and leave it null if not set, but that is a bigger change. Based on that I think we'll have to go with the error checking logic, but I think we should do the following as part of this PR.

- Add warning comment in the code that it's a temporary workaround
- Add issue to track adding a new resource for vulnerability alerts, adding a new `ignore_vulnerability_alerts` field and deprecating `vulnerability_alerts` & `ignore_vulnerability_alerts_during_read` for `v6.10.0`
- Add issue to track removal of `ignore_vulnerability_alerts`, `vulnerability_alerts` & `ignore_vulnerability_alerts_during_read` for `v7.0.0`

deiga: @stevehipwell Done!

I've added #3038 and #3039 could you assign them to correct milestones?

stevehipwell (approved):
LGTM

nickfloyd (approved):

Issue #2996, opened 2025-12-10
[MAINT] Migrate all resources and data sources to Context-aware CRUD functions


deiga:
## Summary

This issue tracks the migration of all resources and data sources from legacy CRUD functions (`Create`, `Read`, `Update`, `Delete`) to Context-aware CRUD functions (`CreateContext`, `ReadContext`, `UpdateContext`, `DeleteContext`).

## Why This Migration Is Needed

Per the [Terraform Plugin SDK v2 Upgrade Guide](https://developer.hashicorp.com/terraform/plugin/sdkv2/guides/v2-upgrade-guide#more-support-for-context-context):

1. **Deprecation**: The legacy CRUD function types (`CreateFunc`, `ReadFunc`, `UpdateFunc`, `DeleteFunc`) are deprecated in favor of their context-aware counterparts
2. **Reliable Timeouts & Cancellation**: Context-aware functions plumb `context.Context` throughout the SDK, enabling better request lifecycle management
3. **Improved Error Handling**: Migration to `diag.Diagnostics` allows returning multiple errors and warnings while associating them with specific fields

### Before (Legacy)
```go
func resourceGithubExample() *schema.Resource {
    return &schema.Resource{
        Create: resourceGithubExampleCreate,
        Read:   resourceGithubExampleRead,
        Update: resourceGithubExampleUpdate,
        Delete: resourceGithubExampleDelete,
        // ...
    }
}

func resourceGithubExampleCreate(d *schema.ResourceData, meta interface{}) error {
    // ...
    return nil
}
```

### After (Context-Aware)
```go
func resourceGithubExample() *schema.Resource {
    return &schema.Resource{
        CreateContext: resourceGithubExampleCreate,
        ReadContext:   resourceGithubExampleRead,
        UpdateContext: resourceGithubExampleUpdate,
        DeleteContext: resourceGithubExampleDelete,
        // ...
    }
}

func resourceGithubExampleCreate(ctx context.Context, d *schema.ResourceData, meta interface{}) diag.Diagnostics {
    // Use ctx in API calls for proper cancellation support
    // ...
    return nil
}
```

## Current Status

- **Total files requiring migration**: 148
- **Already migrated**: 1 (`data_source_github_repository_file.go`)
- **Partially migrated** (StateContext in Importer only): 43

#### Actions
- [ ] `resource_github_actions_environment_secret.go`
- [ ] `resource_github_actions_environment_variable.go`
[...]

PR #3036, opened 2026-01-03
chore: migrate releases resource and data source to context-aware fun…

tempora-mutantur:
Contributes to #2996

Local tests passed for anonymous, individual, and organization modes.

----

### After the change?
CRUD functions were switched to context-aware versions
https://developer.hashicorp.com/terraform/plugin/sdkv2/guides/v2-upgrade-guide#more-support-for-context-context
* 

### Pull request checklist

- [ ] Yes
- [x] No

----

### DIFF
--- a/github/data_source_github_release.go
+++ b/github/data_source_github_release.go
@@ -2,10 +2,10 @@ package github
 
 import (
 	"context"
-	"fmt"
 	"strconv"
 	"strings"
 
+	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/validation"
 
 	"github.com/google/go-github/v81/github"
@@ -14,7 +14,7 @@ import (
 
 func dataSourceGithubRelease() *schema.Resource {
 	return &schema.Resource{
-		Read:        dataSourceGithubReleaseRead,
+		ReadContext: dataSourceGithubReleaseRead,
 		Description: "Use this data source to retrieve information about a GitHub release in a specific repository.",
 		Schema: map[string]*schema.Schema{
 			"repository": {
@@ -156,12 +156,11 @@ func dataSourceGithubRelease() *schema.Resource {
 	}
 }
 
-func dataSourceGithubReleaseRead(d *schema.ResourceData, meta any) error {
+func dataSourceGithubReleaseRead(ctx context.Context, d *schema.ResourceData, meta any) diag.Diagnostics {
 	repository := d.Get("repository").(string)
 	owner := d.Get("owner").(string)
 
 	client := meta.(*Owner).v3client
-	ctx := context.Background()
 
 	var err error
 	var release *github.RepositoryRelease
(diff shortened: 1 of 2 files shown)


github/resource_github_release.go
```diff
@@ -137,12 +138,7 @@ func resourceGithubRelease() *schema.Resource {
 	}
 }
 
-func resourceGithubReleaseCreateUpdate(d *schema.ResourceData, meta any) error {
-	ctx := context.Background()
-	if !d.IsNewResource() {
-		ctx = context.WithValue(ctx, ctxId, d.Id())
```
tempora-mutantur, line 143: I would be grateful if this change could be verified, as I found that there are only writes and no reads of `ctxId`. I couldn't find the reason why it was added this way from the git history.

deiga: While this isn't actively used at this time, I wouldn't remove the `context.WithValue` call. It could prove to be useful in debugging

tempora-mutantur: Restored the calls as requested. Thank you.

deiga (approved):
Amazing, thank you! 🎉

deiga:
@tempora-mutantur Could you please resolve the conflicts? 🙏

tempora-mutantur:
> @tempora-mutantur Could you please resolve the conflicts? 🙏

Done. Thanks

tempora-mutantur:
@deiga 

After rebasing, I re-ran tests including the `GH_TEST_AUTH_MODE=anonymous` configuration. I found that the provider now requires either `token` or `app_auth` to be specified (e.g., via `GITHUB_TOKEN` environment variable). Without it, I get:

```
        Error: Invalid combination of arguments
        
          with provider["registry.terraform.io/hashicorp/github"],
          on terraform_plugin_test.tf line 2, in provider "github":
           2:           provider "github" {
        
        "token": one of `app_auth,token` must be specified
```

The issue appears to be caused by using `ExactlyOneOf` validation in the provider schema:

https://github.com/integrations/terraform-provider-github/blob/main/github/provider.go#L25 

`ExactlyOneOf` requires that exactly one of the specified fields must be set, which prevents anonymous mode from working when neither `token` nor `app_auth` are set.

I tested the "can be configured to run anonymously" test case by removing the line

https://github.com/integrations/terraform-provider-github/blob/main/github/provider_test.go#L57

to test this

https://github.com/integrations/terraform-provider-github/blob/main/website/docs/index.html.markdown?plain=1#L32

It failed with: `"app_auth": one of 'app_auth,token' must be specified`

```
        Error: Invalid combination of arguments
        
          with provider["registry.terraform.io/hashicorp/github"],
          on terraform_plugin_test.tf line 2, in provider "github":
           2:           provider "github" {
        
        "token": one of `app_auth,token` must be specified
```

Maybe we could use `ConflictsWith` instead of `ExactlyOneOf` if we want to allow providers with no auth config.

deiga:
@tempora-mutantur I think I actually made the decision conciously that for anonymous access one would need to set `token = ""` in the provider.

But I didn't document that and I didn't run this past the other maintainers properly either

deiga:
There is now a PR to address that #3071

github/data_source_github_release.go
```diff
@@ -172,85 +171,85 @@ func dataSourceGithubReleaseRead(d *schema.ResourceData, meta any) error {
...
-		return err
+		return diag.FromErr(err)
 	}
 
 	d.SetId(strconv.FormatInt(release.GetID(), 10))
 	err = d.Set("release_tag", release.GetTagName())
 	if err != nil {
-		return err
+		return diag.FromErr(err)
 	}
```
stevehipwell:
```go
  if err := d.Set("release_tag", release.GetTagName()); err != nil {
		return diag.FromErr(err)
	}
```

Please could you use this pattern here.

github/resource_github_release.go
```diff
@@ -197,25 +197,25 @@ func resourceGithubReleaseCreateUpdate(d *schema.ResourceData, meta any) error {
...
 
 	transformResponseToResourceData(d, release, repoName)
 	return nil
 }
 
-func resourceGithubReleaseRead(d *schema.ResourceData, meta any) error {
+func resourceGithubReleaseRead(ctx context.Context, d *schema.ResourceData, meta any) diag.Diagnostics {
 	repository := d.Get("repository").(string)
-	ctx := context.WithValue(context.Background(), ctxId, d.Id())
+	ctx = context.WithValue(ctx, ctxId, d.Id())
```
stevehipwell:
```suggestion
```
This value doesn't do anything.

github/resource_github_release.go
```diff
@@ -228,31 +228,31 @@ func resourceGithubReleaseRead(d *schema.ResourceData, meta any) error {
...
+		return diag.FromErr(err)
 	}
 	transformResponseToResourceData(d, release, repository)
 	return nil
 }
 
-func resourceGithubReleaseDelete(d *schema.ResourceData, meta any) error {
-	ctx := context.WithValue(context.Background(), ctxId, d.Id())
+func resourceGithubReleaseDelete(ctx context.Context, d *schema.ResourceData, meta any) diag.Diagnostics {
+	ctx = context.WithValue(ctx, ctxId, d.Id())
```
stevehipwell:
```suggestion
```

As above.

github/resource_github_release.go
```diff
@@ -228,31 +228,31 @@ func resourceGithubReleaseRead(d *schema.ResourceData, meta any) error {
...
-		return fmt.Errorf("`release_id` must be present")
+		return diag.Errorf("`release_id` must be present")
 	}
 
 	_, err = client.Repositories.DeleteRelease(ctx, owner, repository, releaseID)
 	if err != nil {
-		return fmt.Errorf("error deleting GitHub release reference %s/%s (%s): %w",
-			fmt.Sprint(releaseID), repository, owner, err)
+		return diag.FromErr(fmt.Errorf("error deleting GitHub release reference %s/%s (%s): %w",
+			fmt.Sprint(releaseID), repository, owner, err))
```
stevehipwell:
```suggestion
		return diag.Errorf("error deleting GitHub release reference %s/%s (%s): %v", fmt.Sprint(releaseID), repository, owner, err)
```

stevehipwell (changes requested):
Thanks for the PR @tempora-mutantur, I've added some comments.

github/resource_github_release.go
tempora-mutantur: No longer setting `ctxId`

https://github.com/integrations/terraform-provider-github/pull/3036#discussion_r2659100448

@deiga FYI

stevehipwell:
@nickfloyd could you please review this.

stevehipwell:
@tempora-mutantur could you please rebase this PR?

nickfloyd (approved):

PR #3051, opened 2026-01-07
[MAINT]: Update GNUmakefile to enhance test command functionality

deiga:
Supersedes #3020

----

### Before the change?

* No way to declare a granular test pattern for `make testacc`
* Coverage ran for all test commands

### After the change?

* Able to declare a granular test pattern for `make testacc T=TestAccFoo`
  * Inspired by: https://github.com/hashicorp/terraform-provider-aws/blob/main/GNUmakefile
* Coverage is only run if `COV=true` when calling a make command

### Pull request checklist
- [ ] ~Tests for the changes have been added (for bug fixes / features)~

- [ ] Yes
- [x] No

----

### DIFF
--- a/GNUmakefile
+++ b/GNUmakefile
@@ -62,14 +62,6 @@ testacc:
 	printf "==> Running acceptance tests on branch: \033[1m%s\033[0m...\n" "🌿 $$branch 🌿"
 	TF_ACC=1 CGO_ENABLED=0 go test $(TEST) -v -run '^TestAcc' $(RUNARGS) $(TESTARGS) -timeout 120m -count=1
 
-test-compile:
-	@if [ "$(TEST)" = "./..." ]; then \
-		echo "ERROR: Set TEST to a specific package. For example,"; \
-		echo "  make test-compile TEST=./$(PKG_NAME)"; \
-		exit 1; \
-	fi
-	CGO_ENABLED=0 go test -c $(TEST) $(TESTARGS)
-
 sweep:
 	@echo "WARNING: This will destroy infrastructure. Use only in development accounts."
 	go test $(TEST) -v -sweep=$(SWEEP) $(SWEEPARGS)
@@ -92,4 +84,4 @@ ifeq (,$(wildcard $(GOPATH)/src/$(WEBSITE_REPO)))
 endif
 	@$(MAKE) -C $(GOPATH)/src/$(WEBSITE_REPO) website-provider-test PROVIDER_PATH=$(shell pwd) PROVIDER_NAME=$(PKG_NAME)
 
-.PHONY: build test testacc fmt lint lintcheck tools test-compile website website-lint website-test sweep
+.PHONY: build test testacc fmt lint lintcheck tools website website-lint website-test sweep


stevehipwell (changes requested):
I like the test changes but I still think we need the inline `CGO_ENABLED=0` to make it as easy as possible to understand.

GNUmakefile
```diff
@@ -26,19 +47,28 @@ lintcheck:
 	golangci-lint run ./...
 
 test:
-	CGO_ENABLED=0 go test ./...
-	# commenting this out for release tooling, please run testacc instead
+	@branch=$$(git rev-parse --abbrev-ref HEAD); \
+	printf "==> Running acceptance tests on branch: \033[1m%s\033[0m...\n" "🌿 $$branch 🌿"
+	go test $(TEST) \
```
stevehipwell:
```suggestion
	go test ./... \
```
Vendor has been ignored since about Go 1.9.

GNUmakefile
```diff
@@ -26,19 +47,28 @@ lintcheck:
...
+		-timeout=30s \
+		-parallel=4 \
+		-v \
+		-run '^Test[^A]|^TestA[^c]|^TestAc[^c]' \
```
stevehipwell: What is the purpose of this?

GNUmakefile
```diff
@@ -26,19 +47,28 @@ lintcheck:
...
+		$(RUNARGS) $(TESTARGS) \
+		-count 1;
 
 testacc:
-	TF_ACC=1 CGO_ENABLED=0 go test -run "^TestAcc*" $(TEST) -v $(TESTARGS) -timeout 120m -count=1
+	@branch=$$(git rev-parse --abbrev-ref HEAD); \
+	printf "==> Running acceptance tests on branch: \033[1m%s\033[0m...\n" "🌿 $$branch 🌿"
+	TF_ACC=1 go test $(TEST) -v -run '^TestAcc' $(RUNARGS) $(TESTARGS) -timeout 120m -count=1
```
stevehipwell:
```suggestion
	TF_ACC=1 go test ./... -v -run '^TestAcc' $(RUNARGS) $(TESTARGS) -timeout 120m -count=1
```
As above.

GNUmakefile
```diff
@@ -26,19 +47,28 @@ lintcheck:
...
 
 test-compile:
 	@if [ "$(TEST)" = "./..." ]; then \
 		echo "ERROR: Set TEST to a specific package. For example,"; \
 		echo "  make test-compile TEST=./$(PKG_NAME)"; \
 		exit 1; \
 	fi
-	CGO_ENABLED=0 go test -c $(TEST) $(TESTARGS)
+	go test -c $(TEST) $(TESTARGS)
```
stevehipwell: I'm not sure why we even need this task?

GNUmakefile
deiga: Honestly, I just copied it from https://github.com/hashicorp/terraform-provider-aws/blob/main/GNUmakefile#L773

It selects all tests that don't start with `TestAcc`, I think it could be simplified by using `-skip '^TestAcc'` instead 🤔

GNUmakefile
deiga: I don't either, but I didn't want to remove it yet 😬

GNUmakefile
```diff
@@ -1,8 +1,28 @@
-TEST?=$$(go list ./... |grep -v 'vendor')
-WEBSITE_REPO=github.com/hashicorp/terraform-website
 PKG_NAME=github
+TEST?=./$(PKG_NAME)/...
```
stevehipwell, line 3: I don't see the value here or even how it'd work without an additional undefined variable?

stevehipwell (changes requested):

GNUmakefile
deiga: If we use `./...` `go test` will see 2 packages and run tests for both. In which case terminal output is printed only when all tests have finished. When using `./github/...` it will print out each test as it is finished, which is SUPER useful when running `make testacc`

I don't understand the question about "how it'd work without an additional undefined variable?"

GNUmakefile
stevehipwell: If it's not referenced anywhere else then I'd suggest removing it.

GNUmakefile
stevehipwell: Why not hardcode `./github/...` then as we don't currently have any other packages? Readability is important and I think this fits into the YAGNI definition.

> I don't understand the question about "how it'd work without an additional undefined variable?"

Ignore that question, I was convinced I couldn't see `PKG_NAME` defined. I'm guessing it was because when I first saw these changes it was still on line 3.

deiga: Yeah, I'm okay with that. I was honestly just keeping this structure because I didn't want to just remove everything I don't understand entirely 😆

stevehipwell: I'm not sure the structure has been kept relevant given the workaround for tests in _vendor_ having been outdated for 8+ years. Also checkout the [current](https://github.com/hashicorp/terraform-provider-scaffolding/blob/main/GNUmakefile) (at time of archiving 3 years ago) version of this file; if you look at the history you can see how this was provided by the template and then effectively orphaned as it was removed from the template.

PR #3054, opened 2026-01-07
[BUG] Enable importing `github_emu_group_mapping` for Group with multiple teams

### LINKED ISSUE #3030: [BUG]: Importing GitHub EMU group mapping fails for more than 1 team
### Expected Behavior

Importing a EMU group mapping should require the group ID and team slug and should not fail if the group is connected to multiple teams.

Documentation reference: 
https://registry.terraform.io/providers/integrations/github/latest/docs/resources/emu_group_mapping#group_id-1

### Actual Behavior

Importing a EMU group mapping fails if the group is connected to multiple teams.

See error returned: https://github.com/integrations/terraform-provider-github/blob/63076f76129a825fc4068ff2cebd91202ef0b742/github/resource_github_emu_group_mapping.go#L34

### Terraform Version

Terraform v1.14.3
on linux_amd64
+provider registry.terraform.io/integrations/github v6.8.3

### Affected Resource(s)

- github_emu_group_mapping

```hcl

```

### Steps to Reproduce

_No response_

### Debug Output

```shell

```

### Panic Output

```shell

```

### Code of Conduct

deiga:
Superseded #3043 

Resolves #3030

----

### Before the change?

* No tests for `github_emu_group_mapping`
* Import set wrong ID
* Not able to import Groups with multiple teams mapped
* Using legacy CRUD provider functions

### After the change?

* Adds tests for `github_emu_group_mapping`
* Fixes import logic
* Enabled importing Group with multiple team mappings
* Refactor to use Context-aware functions

### Pull request checklist

- [ ] Yes
- [x] No

----


github/resource_github_emu_group_mapping.go
```diff
@@ -185,3 +250,117 @@ func getInt64FromInterface(val any) (int64, error) {
...
+
+	tflog.Debug(ctx, "Parsed two-part import ID", map[string]any{
+		"import_id": importID,
+		"group_id":  groupID,
+		"team_slug": teamSlug,
+	})
+
+	tflog.Trace(ctx, "Setting state attribute: group_id", map[string]any{
+		"group_id": groupID,
+	})
```
stevehipwell: Why do we need both of these?

github/resource_github_emu_group_mapping.go
```diff
@@ -185,3 +250,117 @@ func getInt64FromInterface(val any) (int64, error) {
...
+	}
+
+	tflog.Debug(ctx, "Parsed integer import ID", map[string]any{
+		"import_id": importID,
+		"group_id":  groupID,
+	})
+
+	tflog.Trace(ctx, "Setting state attribute: group_id", map[string]any{
+		"group_id": groupID,
+	})
```
stevehipwell: Why do we need both of these?

github/resource_github_emu_group_mapping.go
```diff
@@ -185,3 +250,117 @@ func getInt64FromInterface(val any) (int64, error) {
 	}
 	return id64, nil
 }
+
+func importWithTwoPartID(ctx context.Context, d *schema.ResourceData, _ any) ([]*schema.ResourceData, error) {
```
stevehipwell: I'd suggest moving this to _util_import.go_.

github/resource_github_emu_group_mapping.go
```diff
@@ -185,3 +250,117 @@ func getInt64FromInterface(val any) (int64, error) {
...
+
+	resourceID := fmt.Sprintf("teams/%s/external-groups", teamSlug)
+	tflog.Trace(ctx, "Setting resource ID", map[string]any{
+		"resource_id": resourceID,
+	})
+	d.SetId(resourceID)
+	return []*schema.ResourceData{d}, nil
+}
+
+func importWithIntegerID(ctx context.Context, d *schema.ResourceData, meta any) ([]*schema.ResourceData, error) {
```
stevehipwell: I'd suggest moving this to _util_import.go_.

stevehipwell (changes requested):
As we as the inline comments, we need to be careful when using `tflog` that it can only be used with contexts injected by TF (e.g. `CreateContext`).

github/resource_github_emu_group_mapping.go
deiga: I liked the idea of having a DEBUG statement "Parsing successful" and separately TRACE statements for each operation that isn't covered and can lead to an interruption.

But that might be too verbose and unnecessary. I'm trying to think about how we can get better information from future DEBUG or TRACE logs from bug reports 😬

github/resource_github_emu_group_mapping.go
deiga: Same: https://github.com/integrations/terraform-provider-github/pull/3054#discussion_r2673996204

github/resource_github_emu_group_mapping.go
stevehipwell: I don't think there is value in two log entries without anything between them.

github/resource_github_emu_group_mapping.go
stevehipwell: Sorry, I was pretty sure that the two functions I suggested moving were generic helpers?

github/resource_github_emu_group_mapping.go
deiga: Even if they are for different log levels?

stevehipwell: > Even if they are for different log levels?

Yes, as long as the payload isn't significantly different for the lower level log. Logs need to have a purpose and there isn't a purpose for logging to two different messages at the same point; it doesn't make it any easier for me to debug the system and arguably it makes it harder.

In this case the higher level log (trace) actually contains a smaller payload than the lower level one.

github/resource_github_emu_group_mapping.go
```diff
@@ -60,110 +46,189 @@ func resourceGithubEMUGroupMapping() *schema.Resource {
 	}
 }
 
-func resourceGithubEMUGroupMappingCreate(d *schema.ResourceData, meta any) error {
-	return resourceGithubEMUGroupMappingUpdate(d, meta)
+func resourceGithubEMUGroupMappingCreate(ctx context.Context, d *schema.ResourceData, meta any) diag.Diagnostics {
```
stevehipwell: It would be consistent to merge the create and update functions.

github/resource_github_emu_group_mapping.go
```diff
@@ -60,110 +46,189 @@ func resourceGithubEMUGroupMapping() *schema.Resource {
...
+	newResourceID := fmt.Sprintf("teams/%s/external-groups", teamSlugStr)
+	tflog.Trace(ctx, "Setting resource ID", map[string]any{
+		"resource_id": newResourceID,
+	})
+	d.SetId(newResourceID)
+
+	tflog.Trace(ctx, "Transitioning to read operation", map[string]any{
+		"resource_id": newResourceID,
+	})
+	return resourceGithubEMUGroupMappingRead(ctx, d, meta)
```
stevehipwell: We shouldn't call read, we can just set the etag directly here.

stevehipwell (changes requested):

github/resource_github_emu_group_mapping.go
```diff
@@ -4,41 +4,27 @@ import (
...
+		CreateContext: resourceGithubEMUGroupMappingCreateOrUpdate,
+		ReadContext:   resourceGithubEMUGroupMappingRead,
+		UpdateContext: resourceGithubEMUGroupMappingCreateOrUpdate,
+		DeleteContext: resourceGithubEMUGroupMappingDelete,
 		Importer: &schema.ResourceImporter{
-			State: func(d *schema.ResourceData, meta any) ([]*schema.ResourceData, error) {
-				id, err := strconv.Atoi(d.Id())
-				if err != nil {
-					return nil, err
+			StateContext: func(ctx context.Context, d *schema.ResourceData, meta any) ([]*schema.ResourceData, error) {
```
stevehipwell: Could you make this a declared function instead of inlined?

github/resource_github_emu_group_mapping.go
```diff
@@ -60,110 +46,193 @@ func resourceGithubEMUGroupMapping() *schema.Resource {
...
 	}
 	id64, err := getInt64FromInterface(id)
 	if err != nil {
-		return err
+		return diag.FromErr(err)
 	}
 
-	ctx := context.WithValue(context.Background(), ctxId, d.Id())
+	ctx = context.WithValue(ctx, ctxId, d.Id())
+
```
stevehipwell:
```suggestion
```
As far as I can tell this does nothing so I've been removing them.

github/util_import.go
```diff
@@ -0,0 +1,115 @@
...
+import (
+	"context"
+	"fmt"
+	"strconv"
+
+	"github.com/hashicorp/terraform-plugin-log/tflog"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+)
+
+func importEMUGroupMappingWithTwoPartID(ctx context.Context, d *schema.ResourceData, _ any) ([]*schema.ResourceData, error) {
```
stevehipwell: These need to be moved back into the resource file, sorry if one of my comments was confusing.

stevehipwell (changes requested):

github/resource_github_emu_group_mapping.go
```diff
@@ -60,110 +40,188 @@ func resourceGithubEMUGroupMapping() *schema.Resource {
...
+	tflog.Trace(ctx, "Setting state attribute: etag", map[string]any{
+		"etag": etag,
+	})
+	if err = d.Set("etag", etag); err != nil {
+		return diag.FromErr(err)
 	}
-	if err = d.Set("group_id", int(*group.GroupID)); err != nil {
-		return err
+
+	groupIDInt := int(*group.GroupID)
```
stevehipwell:
```suggestion
	groupIDInt := int(group.GetGroupID())
```
The SDK has helper methods for pointers that should be used. I know this was the code you're modifying but could you check the rest of this file to see if it was used incorrectly anywhere else?

stevehipwell (changes requested):

stevehipwell (changes requested):
I've just realized what you're doing here, which is what I've been working on with the other team resources. I think you need to make the following changes (none of which I'd consider breaking).

- Resource ID format to `<team-id>:<team-slug>:<group-id>`
  - State migration to update the ID
- Import to accept `<team-slug>:<group-id>`
  - Lookup team ID and update resource ID to correct format
- Make group ID force new
  - Otherwise changing the group would leave orphaned links
- Add custom diff function
  - Force new if team slug points at new ID
- Separate create and update functions

The top 3 steps would be good enough to fix the bug and stop records being orphaned (unless the team slug was changed to point at a completely different team).

deiga:
> I've just realized what you're doing here, which is what I've been working on with the other team resources. I think you need to make the following changes (none of which I'd consider breaking).
> 
> * Resource ID format to `<team-id>:<team-slug>:<group-id>`
>   
>   * State migration to update the ID
> * Import to accept `<team-slug>:<group-id>`
>   
>   * Lookup team ID and update resource ID to correct format
> * Make group ID force new
>   
>   * Otherwise changing the group would leave orphaned links
> * Add custom diff function
>   
>   * Force new if team slug points at new ID
> * Separate create and update functions
> 
> The top 3 steps would be good enough to fix the bug and stop records being orphaned (unless the team slug was changed to point at a completely different team).

@stevehipwell These definitely sound like worthwhile changes, but I wonder if we need to do them before merging this? 
I'd rather get this PR merged now, to enable these imports and then create a follow-up PR to address those changes.

stevehipwell:
@deiga you need the first 2 steps above to fix the bug as otherwise you'll have duplicate IDs when teams have multiple groups attached. The 3rd step costs nothing and fixes another set of current defects. I think maybe we move this PR and the issue to the v6.11.0 milestone to give us a bit more time?

deiga:
> @deiga you need the first 2 steps above to fix the bug as otherwise you'll have duplicate IDs when teams have multiple groups attached. The 3rd step costs nothing and fixes another set of current defects. I think maybe we move this PR and the issue to the v6.11.0 milestone to give us a bit more time?

@stevehipwell The ID uses team slug as the ID, so duplicate IDs shouldn't be an issue as each team can have exactly 1 mapping.

Like I said, I'll happily address those in a follow-up as I don't see any of them necessary for getting this merged

stevehipwell (changes requested):
Ok, at a bare minimum can we please only support two part import IDs. That way import doesn't have to lookup the team and we don't get any drift. It can save the original ID pattern for now and we can fix up the rest of my points later

stevehipwell (changes requested):
Sorry @deiga I don't see how this resource works at all? The read just checks that the group has a team linked, not that the relationship is the one being managed by TF. It looks like the SDK is missing the [required endpoint](https://docs.github.com/en/enterprise-cloud@latest/rest/teams/external-groups?apiVersion=2022-11-28#list-a-connection-between-an-external-group-and-a-team).

Let's push this PR into the next release. I'll open a PR to get the endpoint supported in the SDK.

stevehipwell (approved):
Thanks @deiga. This PR fixes the import bug and the other defects are pre existing so I think we should get it merged and fix the bugs in a separate PR.

nickfloyd (approved):

PR #3055, opened 2026-01-07
[MAINT] Properly enable test sweeper

deiga:
Supersedes #3035

----

### Before the change?

* Repository names in tests didn't follow a standard pattern
* Team names in tests didn't follow a standard pattern
* Repository sweeper didn't actually work

### After the change?

* Repository and Team names in tests use `testResourcePrefix` to ensure that sweepers can be used
* Added make target for sweep

### Pull request checklist

- [ ] Yes
- [x] No

----

### DIFF
--- a/GNUmakefile
+++ b/GNUmakefile
@@ -1,3 +1,4 @@
+TEST?=$$(go list ./... |grep -v 'vendor')
 SWEEP?=repositories,teams
 WEBSITE_REPO=github.com/hashicorp/terraform-website
 PKG_NAME=github


deiga:
@stevehipwell Unfortunately, I don't :/

- One option could be to create a custom linter for `golangci-lint`, but I don't know how involved that would be.
- Another that comes to mind, would be to create a Builder-pattern for repo and team resources, but that falls short quickly

One very simplistic check could be to run something like `grep 'name\s+=\s+%s' -v *_test.go` 😬

stevehipwell:
@deiga I was really thinking about keeping the PR valid before it's merged. Once it's in the code we'll just need to be good at reviewing. Custom linters are powerful, but the UX for contributors isn't great.

deiga:
Ah, right. I was already thinking structural :D

stevehipwell:
@deiga if you get this ready to merge, we could look to get it prioritised so the burden lands on the new PRs (I know this pain from the tests PR)?

deiga:
> @deiga if you get this ready to merge, we could look to get it prioritised so the burden lands on the new PRs (I know this pain from the tests PR)?

I'll try get it tonight

GNUmakefile
```diff
@@ -1,4 +1,4 @@
-TEST?=$$(go list ./... |grep -v 'vendor')
+SWEEP?=repositories,teams
```
stevehipwell, line 2: This looks like it might break the acceptance tests?

deiga: Woops, that's an oversight

PR #3061, opened 2026-01-07
[MAINT] Improve provider tests

deiga:
Supersedes #3033

----

### Before the change?

* It was possible to inadvertently set `token` and `app_auth` at the same time
* Some provider test's weren't actually testing anything

### After the change?

* Ensures that `token` (or `GITHUB_TOKEN`) and `app_auth` can't be set at the same time
* Updated some provider tests to actually initialize the provider and verify changes

### Pull request checklist

- [ ] Yes
- [x] No

----

### DIFF
--- a/github/provider.go
+++ b/github/provider.go
@@ -18,10 +18,11 @@ func Provider() *schema.Provider {
 	p := &schema.Provider{
 		Schema: map[string]*schema.Schema{
 			"token": {
-				Type:        schema.TypeString,
-				Optional:    true,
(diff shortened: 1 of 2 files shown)


stevehipwell (commented):
This looks good. It might be worth adding support for app auth via env variable? This is something we require to better be able to run integration tests.

deiga:
@stevehipwell Yeah, that could be good, but I don't think it's best to do in this PR. There is an open Feature request for that as well

stevehipwell (approved):
LGTM

PR #3067, opened 2026-01-07
fix: Correct private forking implementation

stevehipwell:
Fixes issue introduced with #3056

----

### Before the change?

* Public repo creation is broken

### After the change?

* Public repo creation fixed

### Pull request checklist

- [ ] Yes
- [x] No

----

### DIFF
--- a/github/resource_github_repository.go
+++ b/github/resource_github_repository.go
@@ -249,7 +249,7 @@ func resourceGithubRepository() *schema.Resource {
 			"allow_forking": {
 				Type:        schema.TypeBool,
 				Optional:    true,
-				Default:     false,
+				Computed:    true,
 				Description: "Set to 'true' to allow private forking on the repository; this is only relevant if the repository is owned by an organization and is private or internal.",
 			},
 			"squash_merge_commit_title": {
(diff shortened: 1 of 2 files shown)


deiga (approved):

github/resource_github_repository.go
```diff
@@ -1016,6 +1012,7 @@ func resourceGithubRepositoryUpdate(ctx context.Context, d *schema.ResourceData,
 	}
 
 	if d.HasChange("visibility") {
+		repoReq.AllowForking = allowForking
```
deiga, line 1017: question: should this first check if visibility is private?

stevehipwell: The check has already happened above.

nickfloyd (approved):

github/resource_github_repository.go
```diff
@@ -249,7 +249,7 @@ func resourceGithubRepository() *schema.Resource {
 			"allow_forking": {
 				Type:        schema.TypeBool,
 				Optional:    true,
-				Default:     false,
```
lebenitza, line 252: Why was this default value changed? It's not backwards compatible.

stevehipwell: @lebenitza how can this change not be backwards compatible when the `allow_forking` wasn't released until after this follow up PR was merged?

DevopsAdrian: They may be referring to https://github.com/integrations/terraform-provider-github/issues/3088, which has several other respondents.

In my use case, v6.10.1 is broken with a HTTP 422 "This organization does not allow private repository forking" error on a null allow_forking value for unrelated repository config updates. The operation being performed was a simple removal of the has_downloads parameter from our repository configs which should be a no-op from the perspective of the backing APIs.

Rolling back to 6.9.1 was required for successful apply operations.

stevehipwell: This PR is fixing an issue that showed up in testing with #3067, but the feature is net new. It looks like the feature introduced a regression where repo updates will fail when forking is configured at the org level (which there is no API for); this will be fixed by #3094.

lebenitza: @stevehipwell indeed, I thought it was already released, sorry about that. I'll give the code a closer look but the problem with 6.10.1 is that even if I set `allow_forking = false/null`, when creating a new repo, the provider cannot be used in an Organization that disables forking entirely and I get a 422.
Now, I am not familiar with terraform-plugin-sdk at all but considering the above I thought this change forced its value to `true` from the previous default `false`. It seems like this is just another instance of an edge case in GitHub API, which is full of, and you're working on a fix #3094 . Thanks!

PR #3068, opened 2026-01-08
Team slug

### LINKED ISSUE #3044: [FEAT]: Add explicit support for `team_slug`
### Describe the need

We should be using the team slug pattern rather than the legacy team ID pattern for referencing teams. The current pattern attempts to support both slug and ID but it has a number of problems including a high API overhead and an inability to be "correct".

This is required for #2497

### SDK Version

_No response_

### API Version

_No response_

### Relevant log output

```shell

```

### Code of Conduct

stevehipwell:
Resolves #3044

----

### Before the change?

* 

### After the change?

* 

### Pull request checklist

- [ ] Yes
- [ ] No

----

### DIFF
--- a/github/resource_github_repository_environment.go
+++ b/github/resource_github_repository_environment.go
@@ -14,10 +14,10 @@ import (
 
 func resourceGithubRepositoryEnvironment() *schema.Resource {
 	return &schema.Resource{
-		Create: resourceGithubRepositoryEnvironmentCreate,
-		Read:   resourceGithubRepositoryEnvironmentRead,
-		Update: resourceGithubRepositoryEnvironmentUpdate,
-		Delete: resourceGithubRepositoryEnvironmentDelete,
+		CreateContext: resourceGithubRepositoryEnvironmentCreate,
+		ReadContext:   resourceGithubRepositoryEnvironmentRead,
+		UpdateContext: resourceGithubRepositoryEnvironmentUpdate,
+		DeleteContext: resourceGithubRepositoryEnvironmentDelete,
 		Importer: &schema.ResourceImporter{
 			StateContext: schema.ImportStatePassthroughContext,
 		},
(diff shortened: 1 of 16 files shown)


github/resource_github_repository_environment.go
```diff
@@ -55,21 +55,39 @@ func resourceGithubRepositoryEnvironment() *schema.Resource {
...
 				Optional:    true,
-				MaxItems:    6,
+				MaxItems:    1,
 				Description: "The environment reviewers configuration.",
 				Elem: &schema.Resource{
 					Schema: map[string]*schema.Schema{
+						"team_slugs": {
+							Type:        schema.TypeSet,
+							Optional:    true,
+							MaxItems:    6,
```
deiga, line 65: Do you think we could add validation that none of these fields together are more than 6? AFAIK the API will fail with 422 or possibly silently fail 😬

github/resource_github_team.go
```diff
@@ -112,14 +113,15 @@ func resourceGithubTeam() *schema.Resource {
 	}
 }
 
-func resourceGithubTeamCreate(ctx context.Context, d *schema.ResourceData, meta any) diag.Diagnostics {
-	err := checkOrganization(meta)
+func resourceGithubTeamCreate(ctx context.Context, d *schema.ResourceData, m any) diag.Diagnostics {
```
deiga, line 116: question: what's the benefit of this change?

deiga (commented):
Just did a pass to see what you are up to. Ignore any  comments that aren't applicable

github/resource_github_repository_environment.go
stevehipwell: Good point, yes we should be able to add a diff function to handle the check for the aggregate count.

github/resource_github_team.go
stevehipwell: We get a typed `meta` var that we can pass around and fewer casts in the code. Note that in the framework the provider data is typed so this is tending towards consistency.

acouvreur:
@stevehipwell @deiga anything I can do to help you guys finishing/merging this pull request ?

PR #2499, opened 2024-12-05
Fixed incorrect secrets creation on externally updated secret

jorgecarleitao:
Resolves #2288

----

### Before the change?

Given a plan with a resource with any secret
When the plan is applied
And the secret is externally modified
And a new plan is planned
Then the new plan results in the creation of a new secret

### After the change?

Given a plan with a resource with any secret
When the plan is applied
And the secret is externally modified
And a new plan is created
Then the plan results in the update of the existing secret

In particular, this behavior enables the use of the lifecycle meta argument to ignore changes to externally updated secrets.

### Pull request checklist
- [ ] Tests for the changes have been added (for bug fixes / features)
- [x] Docs have been reviewed and added / updated if needed (for bug fixes / features)

NOTE: I could not find any test that I could re-use to introduce an external change to an existing resource. Need support.

### Does this introduce a breaking change?

Please see our docs on [breaking changes](https://github.com/octokit/.github/blob/master/community/breaking_changes.md) to help!

- [ ] Yes
- [x] No

----

PR #2462, opened 2024-11-13
[Feature] add environment secret import

whgiles:
Resolves #2461 

----

### Before the change?
After I run an import command

```
Error: Configuration for import target does not exist

The configuration for the given import <module>.github_actions_environment_secret.<name> <repository>/<environment>/<secret_name> does not exist. All target instances must have an associated configuration to be imported.
```

* 

### After the change?
I should be able to run:
 ```
Terraform import '<module>.github_actions_environment_secret.<name> <repository>/<environment>/<secret_name>'
```
 or use and import block
```tf
import {
    to = <module>.github_actions_environment_secret.<name> <repository>/<environment>/<secret_name>
    id = "<repository>/<environment>/<secret_name>"
}
```

* 

### Pull request checklist
- [ ] Tests for the changes have been added (for bug fixes / features)
- [ ] Docs have been reviewed and added / updated if needed (for bug fixes / features)

### Does this introduce a breaking change?
No

Please see our docs on [breaking changes](https://github.com/octokit/.github/blob/master/community/breaking_changes.md) to help!

- [ ] Yes
- [ ] No

----

PR #3069, opened 2026-01-09
fix: Correct secret drift implementation

### LINKED ISSUE #2047: [BUG]: RANDOMLY github_actions_variable can't be created - root object was present, but now absent

### LINKED ISSUE #2518: [BUG]: 'plaintext_value' forces delete and create instead of inplace update

### LINKED ISSUE #2782: [BUG]: Changes in secret values for `github_actions_organization_secret` should occur in-place
### Expected Behavior

When changing the `plaintext_value` or `encrypted_value` of an existing `github_actions_organization_secret` resource, the secret value should be updated in-place as per the GitHub API spec here: https://docs.github.com/en/rest/actions/secrets?apiVersion=2022-11-28#create-or-update-an-organization-secret.

### Actual Behavior

Terraform detects the `github_actions_organization_secret` as needing to be replaced and will delete and recreate the secret resource.

### Terraform Version

Irrelevant to the bug in question

### Affected Resource(s)

- github_actions_organization_secret

```hcl

```

### Steps to Reproduce

Create an `github_actions_organization_secret` resource with an initial value on the first apply. Then change the value of the secret stored and generate a plan. Terraform will show in its plan that it intends to replace the secret rather than update it in-place.

### Debug Output

```shell

```

### Panic Output

```shell

```

### Code of Conduct

### LINKED ISSUE #2913: [MAINT]: Replace the various secret patterns implemented in the provider with a more concrete and predictable solution
### Describe the need

From #2903 

Problem:
We keep getting impacted by `Error: Invalid provider configuration` resulting from the various ways the provider checks and creates secrets.

This particular issue has plagued our last few deployments and it originated from a very old PR being introduced and released which changed how we handle secrets.  

We have about 4 different implementations, that exist in 6 different resources, of how we handle secrets scoped to specific areas.  These various implementations have made it difficult to determining what is idiomatic for this provider given it's been done differently in the resources here: 

```
resource_github_actions_environment_secret.go
resource_github_actions_secret.go
resource_github_actions_organization_secret.go
resource_github_dependabot_organization_secret.go
resource_github_codespaces_organization_secret.go
resource_github_codespaces_user_secret.go
```

The destroy_on_drift pattern is problematic at best - its unpredictable, does not fit our idioms, and the detection is unreliable.

Some of our implementations have `Update` functions + `ForceNew` (which are causing InternalValidate errors), some perform recreation ForceNew , some have internalized logic into `destroy_on_drift` , not to mention that there are different field evaluations and configurations across resources.

@stevehipwell tossed out the idea around having a "secret version field" - that approach seems to fit the provider well and it would let us rip out the convoluted conditional logic around setting and unsetting fields as well as be much more deterministic. This field would control when the secret value is updated and would give user's far more control.

### SDK Version

v6.8.3

### API Version

_No response_

### Relevant log output

```shell

```

### Code of Conduct

### LINKED ISSUE #3049: [BUG]: Provider produced inconsistent result after apply (v6.9.1)
### Expected Behavior

When creating github_actions_environment_variable and github_environment_secret I expect terraform to track the state and create the resources successfully.

note: 
since I could not find a better place in the form to put this: I assume this bug is highly related to #2843

### Actual Behavior

The resources are created but after applying the plan terraform outputs:
 > provider "module.old_main.provider[\"registry.terraform.io/integrations/github\"]" produced
│ an unexpected new value: Root object was present, but now absent.

### Terraform Version

Terraform v1.14.3
on linux_amd64
...
+ provider registry.terraform.io/integrations/github v6.9.1

### Affected Resource(s)

- github_actions_environment_secret
- github_actions_environment_variable

```hcl
terraform {
  required_providers {
    github = {
      version = "6.9.1"
    }
  }
}

provider "github" {}




resource "github_repository_environment" "environment" {
  repository = "my-repo"
  environment = "default"
}

resource "github_actions_environment_secret" "deployment_key" {
  repository      = "my-repo"
  environment     = github_repository_environment.environment.environment
  secret_name     = "DEPLOYMENT_PRIVATE_KEY"
  plaintext_value = "my-super-secret-key"
}

```

### Steps to Reproduce

run `terraform apply` to create the resources.

### Debug Output

```shell
....
  # module.old_main.github_actions_environment_secret.deployment_key will be created
  + resource "github_actions_environment_secret" "deployment_key" {
      + created_at      = (known after apply)
      + environment     = "default"
      + id              = (known after apply)
      + plaintext_value = (sensitive value)
      + repository      = "my-repo"
      + secret_name     = "DEPLOYMENT_PRIVATE_KEY"
      + updated_at      = (known after apply)
    }

...
 
```

### Panic Output

```shell
...
╷
│ Error: Provider produced inconsistent result after apply
│
│ When applying changes to module.old_main.github_actions_environment_secret.deployment_key, provider "module.old_main.provider[\"registry.terraform.io/integrations/github\"]" produced an
│ unexpected new value: Root object was present, but now absent.
│
│ This is a bug in the provider, which should be reported in the provider's own issue tracker.
╵
```

### Code of Conduct

stevehipwell:
Resolves #3049
Resolves #2782
Resolves #2913
Resolves #2047
Resolves #2518

Closes #2462
Closes #2499

----

### Before the change?

- Secret and variable implementations were inconsistent and in some cases incorrect
- Secret drift detection wasn't using an idiomatic TF pattern

### After the change?

- Actions and Dependabot secrets/variables have a consistent implementation
- Secret drift detection uses an idiomatic TF pattern
- The `destroy_on_drift` pattern is ignored & deprecated
- Force new logic is correctly implemented and works with repo renaming
- There are working acceptance tests for secrets/variables

### Pull request checklist

- [ ] Yes
- [x] No

----


github/resource_github_actions_environment_secret.go
```diff
@@ -66,161 +100,172 @@ func resourceGithubActionsEnvironmentSecret() *schema.Resource {
...
-	// be.
-	//
-	// If the resource is changed externally in the meantime then reading back
-	// the last update timestamp will return a result different than the
-	// timestamp we've persisted in the state. In this case, we can no longer
-	// trust that the value matches what is in the state file.
-	//
-	// To solve this, we must unset the values and allow Terraform to decide whether or
-	// not this resource should be modified or left as-is (ignore_changes).
-	if updatedAt, ok := d.GetOk("updated_at"); ok && updatedAt != secret.UpdatedAt.String() {
```
nickfloyd, line 188: I belive this will be a breaking change due to `ignore_changes` - there will be forced recreations now.  Not sure if we want a migration path for the users who rely on that or just mark it as a breaking change.

github/resource_github_actions_environment_secret.go
```diff
@@ -9,15 +9,41 @@ import (
...
 func resourceGithubActionsEnvironmentSecret() *schema.Resource {
 	return &schema.Resource{
-		Create: resourceGithubActionsEnvironmentSecretCreateOrUpdate,
-		Read:   resourceGithubActionsEnvironmentSecretRead,
-		Delete: resourceGithubActionsEnvironmentSecretDelete,
+		CreateContext: resourceGithubActionsEnvironmentSecretCreate,
+		ReadContext:   resourceGithubActionsEnvironmentSecretRead,
+		DeleteContext: resourceGithubActionsEnvironmentSecretDelete,
+
+		CustomizeDiff: func(ctx context.Context, diff *schema.ResourceDiff, m any) error {
```
nickfloyd: I really like this approach.  Any concerns around external changes to the resource forcing the destruction and recreation of the secret? 

For instance...
Let's say a user config has: 
```
resource "github_actions_environment_secret" "api_key" {
  ...
  plaintext_value = "secret-value-123"
  
  lifecycle {
    ignore_changes = [plaintext_value]
  }
}
```

1. User runs terraform apply
2. Other user updates the secret via UI
3. User runs terraform apply

The old way the apply would detect no changes.  This new approach `updated_at` and `remote_updated_at` would be different and therefore force a recreation.

github/acc_test.go
```diff
@@ -74,15 +74,21 @@ type testAccConfig struct {
 
 var testAccConf *testAccConfig
 
+var testAccProviders map[string]*schema.Provider = map[string]*schema.Provider{
+	"github": Provider(),
+}
```
deiga: AFAIK we should be getting rid of using `testAccProviders` the recommended approach is to only use ProviderFactories.

Could you mark this as deprecated?

github/resource_github_actions_environment_secret.go
deiga: suggestion: I think we should be able to forgo adding the `remote_updated_at` field as we could use `customdiff.ForceNewIfChange` (https://pkg.go.dev/github.com/hashicorp/terraform-plugin-sdk/v2/helper/customdiff#ForceNewIfChange) here

Something along the lines of:
```
CustomizeDiff: 
	customdiff.ForceNewIfChange(
		"updated_at", 
		func(ctx context.Context, oldValue, newValue, meta any) bool {
			remoteUpdatedAt := newValue
			if len(remoteUpdatedAt) == 0 {
				return false
			}

			updatedAt := oldValue
			if updatedAt != remoteUpdatedAt {

				if len(updatedAt) != 0 {
					return true
				}
			}

			return false
		}
)
```

deiga: AFAIK setting `ignore_changes = [plaintext_value, updated_at, remote_updated_at]` should resolve that

github/acc_test.go
stevehipwell: I've made the change to do this already (and stashed them), but I'm holding off adding them in to the PR until the actual changes have been reviewed.

github/resource_github_actions_environment_secret.go
stevehipwell: IMHO this is a documentation issue, what @deiga added should be the correct solution. If it's not then we'd need to add in the explicit `destroy_on_drift` field that I'd planned on deprecating.

stevehipwell: Unfortunately you can't do this (I actually coded it up to make sure my mental model was correct). If we're doing drift detection we can't rely on a field updated in the read path as it will never have a change in the pan path. In this case `update_at` is only updated in read if it's not been able to be calculated as part of the create path, this contrasts with `remote_updated_at` which is always updated.

github/resource_github_actions_environment_secret.go
stevehipwell: Sorry, I'm not sure I follow how the new code would be breaking? The comment here is incorrect as this will always force new (note this resource doesn't have the `destroy_on_drift` input to suppress this behaviour).

stevehipwell:
I've updated the implementation to not require a recreate now that the diff is being handled correctly. I've also updated the pattern to correctly handle repo identity and support repo renames.

github/resource_github_actions_environment_secret.go
stevehipwell: Just looking into this further and it looks like this resource was relying on a lifecycle block to ignore changes. I've updated the docs to show this correctly being implemented on the new `remote_updated_at` field. I wouldn't consider a change to how meta arguments work a breaking change.

github/resource_github_actions_environment_secret.go
stevehipwell: The correct setting is `ignore_changes = [remote_updated_at]` and there is now a test to validate this.

deiga:
There is an open PR for adding import of `resourceGithubActionsEnvironmentSecret`
https://github.com/integrations/terraform-provider-github/pull/2462/changes#diff-bd7e54bc1e5ce2c06ae7d38565efee3c760620c960497f9c1d6db2f230415cbc

should that be taken into account or closed?

deiga:
Further prior art on this topic: https://github.com/integrations/terraform-provider-github/pull/2499

deiga:
This might resolve https://github.com/integrations/terraform-provider-github/issues/2047

stevehipwell:
Thanks for the context @deiga. PR #2462 can be closed by this PR; FYI it has the same special character escaping issues seen across all of the environment resources (& data sources). PR #2499 can also be closed by this PR as it looks like it's just a refactoring of the existing flawed pattern. This PR should fix #2047, that's one of the failure mode it's intended to resolve.

stevehipwell:
@deiga FYI I've updated this PR to use your ID as computed pattern, as well as shared diff functions.

github/resource_github_actions_environment_secret.go
```diff
@@ -67,165 +81,301 @@ func resourceGithubActionsEnvironmentSecret() *schema.Resource {
...
 		return diag.FromErr(err)
 	}
 
-	id, err := buildID(repoName, escapeIDPart(envName), secretName)
-	if err != nil {
+	if id, err := buildID(repoName, escapeIDPart(envName), secretName); err != nil {
 		return diag.FromErr(err)
 	} else {
 		d.SetId(id)
 	}
```
deiga: Not your change, but I wonder if the else is superfluous here

github/resource_github_actions_environment_secret.go
```diff
@@ -67,165 +81,301 @@ func resourceGithubActionsEnvironmentSecret() *schema.Resource {
...
-				log.Printf("[INFO] Removing environment secret %s from state because it no longer exists in GitHub",
-					d.Id())
-				d.SetId("")
-				return nil
-			}
+	if id, err := buildID(repoName, escapeIDPart(envName), secretName); err != nil {
+		return diag.FromErr(err)
+	} else {
+		d.SetId(id)
+	}
```
deiga: question: does this else increase clarity?

github/resource_github_actions_environment_secret.go
```diff
@@ -67,165 +81,301 @@ func resourceGithubActionsEnvironmentSecret() *schema.Resource {
...
-	// not this resource should be modified or left as-is (ignore_changes).
-	if updatedAt, ok := d.GetOk("updated_at"); ok && updatedAt != secret.UpdatedAt.String() {
-		log.Printf("[INFO] The environment secret %s has been externally updated in GitHub", d.Id())
-		_ = d.Set("encrypted_value", "")
-		_ = d.Set("plaintext_value", "")
-	} else if !ok {
-		if err = d.Set("updated_at", secret.UpdatedAt.String()); err != nil {
+	// GitHub API does not return on update so we have to lookup the secret to get timestamps
+	if secret, _, err := client.Actions.GetEnvSecret(ctx, repoID, escapedEnvName, secretName); err == nil {
+		if err := d.Set("created_at", secret.CreatedAt.String()); err != nil {
```
deiga, line 290: question: Is there a case where setting created_at is necessary in `Update`? I'm thinking that it's only and import and create concern

github/resource_github_actions_environment_secret.go
```diff
@@ -67,165 +81,301 @@ func resourceGithubActionsEnvironmentSecret() *schema.Resource {
...
+			return diag.FromErr(err)
+		}
+		if err := d.Set("updated_at", secret.UpdatedAt.String()); err != nil {
+			return diag.FromErr(err)
+		}
+		if err := d.Set("remote_updated_at", secret.UpdatedAt.String()); err != nil {
+			return diag.FromErr(err)
+		}
+	} else {
```
deiga, line 299: question: so, if the API returns an error when fetching secret, we set timestamps to nil. Why not I'd to nil as well? Isn't the secret then not existing?

github/resource_github_actions_environment_secret.go
```diff
@@ -67,165 +81,301 @@ func resourceGithubActionsEnvironmentSecret() *schema.Resource {
...
+func resourceGithubActionsEnvironmentSecretDelete(ctx context.Context, d *schema.ResourceData, m any) diag.Diagnostics {
+	meta := m.(*Owner)
+	client := meta.v3client
 
-	repoName, envNamePart, secretName, err := parseID3(d.Id())
+	repoID := d.Get("repository_id").(int)
+	envName := d.Get("environment").(string)
+	secretName := d.Get("secret_name").(string)
+
+	log.Printf("[INFO] Deleting environment secret: %s", d.Id())
```
deiga: question: would you be willing to make the tflog refactoring in the same PR? They should be fairly safe changes

github/resource_github_actions_environment_secret.go
```diff
@@ -67,165 +81,301 @@ func resourceGithubActionsEnvironmentSecret() *schema.Resource {
...
+}
+
+func resourceGithubActionsEnvironmentSecretImport(ctx context.Context, d *schema.ResourceData, m any) ([]*schema.ResourceData, error) {
+	meta := m.(*Owner)
+	client := meta.v3client
+	owner := meta.name
+
+	repoName, envNamePart, secretName, err := parseID3(d.Id())
+	if err != nil {
+		return nil, err
```
deiga, line 335: question: Should the error message include the import ID format, to make the UX nicer?

github/resource_github_actions_environment_secret_migration_test.go
```diff
@@ -0,0 +1,53 @@
+package github
+
+// TODO: Enable this test one there is a way to mock the metadata.
```
deiga: question: which metadata are you referring to here?

github/resource_github_actions_environment_variable.go
```diff
@@ -57,114 +63,209 @@ func resourceGithubActionsEnvironmentVariable() *schema.Resource {
...
 
-	// Try to create the variable first
-	_, err := client.Actions.CreateEnvVariable(ctx, owner, repoName, url.PathEscape(envName), variable)
+	_, err = client.Actions.CreateEnvVariable(ctx, owner, repoName, escapedEnvName, &github.ActionsVariable{
+		Name:  varName,
+		Value: d.Get("value").(string),
+	})
 	if err != nil {
-		var ghErr *github.ErrorResponse
-		if errors.As(err, &ghErr) && ghErr.Response.StatusCode == http.StatusConflict {
```
deiga, line 80: Love this! Implicit import can feel nice for the user but is a TF antipattern

github/resource_github_actions_environment_secret.go
stevehipwell: It's required as `id` is scoped to the `if`.

```go
	if id, err := buildID(repoName, escapeIDPart(envName), secretName); err != nil {
		return diag.FromErr(err)
	} else {
		d.SetId(id)
	}
```

It is my code, but on reflection I'll switch to pattern below.

```go
	id, err := buildID(repoName, escapeIDPart(envName), secretName); err != nil {
	if err != nil {
		return diag.FromErr(err)
	}
	d.SetId(id)
```

github/resource_github_actions_environment_secret.go
stevehipwell: As above.

github/resource_github_actions_environment_secret.go
stevehipwell: The API is eventually consistent, so unless we want to repeatedly hit the API waiting for the secret to have been committed (we don't) this is a requirement. This design provides a consistent API cost for operations (2 calls) and mirrors the eventually consistent API. In testing gihub.com I've not seen this be required, but I have seen indicators in production behaviour during high load events. We do have bug reports for GHES where this is likely happening (and the same for other APIs).

github/resource_github_actions_environment_secret.go
stevehipwell: Because we literally just called create/update to make the secret exist, this is another eventually consistent behaviour we need to support. If the create/update endpoint returned a body we wouldn't need to call this API.

github/resource_github_actions_environment_secret.go
stevehipwell: I'd rather not given the scope of the PR (all secrets and vars).

github/resource_github_actions_environment_secret.go
stevehipwell: Yes it could do.

github/resource_github_actions_environment_secret.go
deiga: That sounds understandable! Could you add a short version of that in the comments?

nickfloyd (dismissed):

github/resource_github_actions_environment_secret.go
```diff
@@ -67,165 +81,302 @@ func resourceGithubActionsEnvironmentSecret() *schema.Resource {
...
+	meta := m.(*Owner)
+	client := meta.v3client
+	owner := meta.name
 
 	repoName := d.Get("repository").(string)
 	envName := d.Get("environment").(string)
 	secretName := d.Get("secret_name").(string)
-	plaintextValue := d.Get("plaintext_value").(string)
-	var encryptedValue string
+	keyID := d.Get("key_id").(string)
```
deiga, line 114: question: Since `key_id` is Optional and Computed, would we need `GetOk` here?

github/resource_github_actions_environment_secret.go
deiga: Do you want to do that here or create a follow up?

github/resource_github_actions_environment_secret_migration_test.go
deiga: suggestion: Could you make the comment more verbose to know when the TODO can be resolved?

github/resource_github_actions_organization_secret_test.go
```diff
@@ -1,346 +1,607 @@
...
 	"github.com/hashicorp/terraform-plugin-sdk/v2/terraform"
 )
 
 func TestAccGithubActionsOrganizationSecret(t *testing.T) {
-	t.Run("creates and updates secrets without error", func(t *testing.T) {
-		secretValue := base64.StdEncoding.EncodeToString([]byte("super_secret_value"))
-		updatedSecretValue := base64.StdEncoding.EncodeToString([]byte("updated_super_secret_value"))
+	t.Run("create_update_plaintext", func(t *testing.T) {
+		randomID := acctest.RandStringFromCharSet(5, acctest.CharSetAlpha)
+		secretName := fmt.Sprintf("test_%s", randomID)
```
deiga, line 19: question: Should this be using `testResourcePrefix` to enable using sweepers in the future?

github/resource_github_actions_organization_variable_test.go
```diff
@@ -1,142 +1,229 @@
...
+	"github.com/hashicorp/terraform-plugin-sdk/v2/terraform"
 )
 
 func TestAccGithubActionsOrganizationVariable(t *testing.T) {
-	t.Run("creates and updates a private organization variable without error", func(t *testing.T) {
-		value := "my_variable_value"
-		updatedValue := "my_updated_variable_value"
+	t.Run("create_update_visibility_all", func(t *testing.T) {
+		randomID := acctest.RandStringFromCharSet(5, acctest.CharSetAlpha)
+		varName := fmt.Sprintf("test_%s", randomID)
```
deiga, line 18: question: Should we use `testResourcePrefix` here to enable future sweepers?

github/resource_github_actions_variable.go
```diff
@@ -7,17 +7,19 @@ import (
...
-		Delete: resourceGithubActionsVariableDelete,
-		Importer: &schema.ResourceImporter{
-			StateContext: schema.ImportStatePassthroughContext,
+		SchemaVersion: 1,
+		StateUpgraders: []schema.StateUpgrader{
+			{
+				Type:    resourceGithubActionsVariableV0().CoreConfigSchema().ImpliedType(),
+				Upgrade: resourceGithubActionsVariableStateUpgradeV0,
+				Version: 0,
+			},
```
deiga, line 22: question: Does this need a state migration? There is only an addition of a Computed field

github/resource_github_actions_variable_migration_test.go
```diff
@@ -0,0 +1,51 @@
+package github
+
+// TODO: Enable this test one there is a way to mock the metadata.
```
deiga: suggestion: Please update this TODO to be more descriptive, so that future travellers can know when/how to resolve it

github/resource_github_dependabot_organization_secret_test.go
```diff
@@ -1,103 +1,545 @@
...
+	"github.com/hashicorp/terraform-plugin-sdk/v2/terraform"
 )
 
 func TestAccGithubDependabotOrganizationSecret(t *testing.T) {
-	t.Run("creates and updates secrets without error", func(t *testing.T) {
-		secretValue := base64.StdEncoding.EncodeToString([]byte("super_secret_value"))
-		updatedSecretValue := base64.StdEncoding.EncodeToString([]byte("updated_super_secret_value"))
+	t.Run("create_update_plaintext", func(t *testing.T) {
+		randomID := acctest.RandStringFromCharSet(5, acctest.CharSetAlpha)
+		secretName := fmt.Sprintf("test_%s", randomID)
```
deiga, line 19: suggestion: Let's use `testResourcePrefix` here to enable future sweepers

github/resource_github_actions_organization_secret.go
```diff
@@ -210,77 +247,173 @@ func resourceGithubActionsOrganizationSecretRead(ctx context.Context, d *schema.
...
+	return nil
+}
+
+func resourceGithubActionsOrganizationSecretUpdate(ctx context.Context, d *schema.ResourceData, m any) diag.Diagnostics {
+	meta := m.(*Owner)
+	client := meta.v3client
+	owner := meta.name
+
+	secretName := d.Get("secret_name").(string)
+	keyID := d.Get("key_id").(string)
```
deiga, line 263: question: `key_id` isn't set in `Read` and there is no state migration to set `key_id`. Does this need an err check?

github/resource_github_actions_environment_secret.go
stevehipwell: It returns a zero value, so all good. I'm pretty sure I made the explicit decision to use `Get` over `GetOk`.

github/resource_github_actions_environment_secret_migration_test.go
stevehipwell: I'd rather not given that the PR is approved, there are a number of TODOs to pick up as the scope of required changes is immense. Basically your mock response pattern can be used here to skip making the meta directly mockable.

github/resource_github_actions_organization_secret_test.go
stevehipwell: The `testResourcePrefix` value is not compatible as is, so this is fine for now. The sweeper could use `test_` or we could refactor.

github/resource_github_actions_variable_migration_test.go
stevehipwell: As above.

github/resource_github_actions_variable.go
stevehipwell: It does, but I think I may have missed the other ones.

github/resource_github_actions_organization_secret.go
stevehipwell: As above.

github/resource_github_dependabot_organization_secret_test.go
stevehipwell: As above.

github/resource_github_actions_organization_variable_test.go
stevehipwell: As above.

github/resource_github_actions_environment_secret.go
stevehipwell: Follow up as it'd be a new provider pattern and best practice.

github/resource_github_actions_variable_migration_test.go
stevehipwell: Done.

github/resource_github_actions_environment_secret_migration_test.go
stevehipwell: Done.

github/resource_github_actions_organization_secret_migration.go
```diff
@@ -8,8 +8,10 @@ import (
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/validation"
 )
 
-func resourceGithubActionsOrganizationSecretResourceV0() *schema.Resource {
+func resourceGithubActionsOrganizationSecretV0() *schema.Resource {
 	return &schema.Resource{
+		SchemaVersion: 0,
```
deiga, line 13: I think this might an accidental addition
```suggestion
```

github/resource_github_actions_secret_migration.go
```diff
@@ -53,8 +56,64 @@ func resourceGithubActionsSecretResourceV0() *schema.Resource {
 	}
 }
 
-func resourceGithubActionsSecretInstanceStateUpgradeV0(ctx context.Context, rawState map[string]any, meta any) (map[string]any, error) {
+func resourceGithubActionsSecretV1() *schema.Resource {
+	return &schema.Resource{
+		SchemaVersion: 1,
```
deiga, line 61: question: We don't use this field and the function name has this info. Is it useful to add this here?

deiga (approved):

github/resource_github_actions_organization_secret_migration.go
stevehipwell: No, intentional.

github/resource_github_actions_secret_migration.go
stevehipwell: It's in the spec to set it.

stevehipwell: If not set it will be implicitly set to 0.

Piccirello:
Tested this in a project and confirmed it's working as intended - updating `selected_repositories` now results in updating existing org secrets rather then replacing them. Eagerly awaiting the release of v6.11.0.

Piccirello:
Scratch that - secrets showed that they would be updated in place, but the actual secret value was cleared (edit: unsure if it was cleared or set to something else, but it was definitely changed).

Example output from the apply of a secret whose value was wiped:
```
22:07:24.464 STDOUT terraform:   # github_actions_organization_secret.this["VERCEL_TOKEN"] will be updated in-place
22:07:24.464 STDOUT terraform:   ~ resource "github_actions_organization_secret" "this" {
22:07:24.464 STDOUT terraform:         id                      = "VERCEL_TOKEN"
22:07:24.464 STDOUT terraform:       ~ selected_repository_ids = [
22:07:24.464 STDOUT terraform:           + 1148597881,
22:07:24.464 STDOUT terraform:             # (1 unchanged element hidden)
22:07:24.464 STDOUT terraform:         ]
22:07:24.464 STDOUT terraform:         # (8 unchanged attributes hidden)
22:07:24.464 STDOUT terraform:     }
```

stevehipwell:
@Piccirello the secret will always be set to the value that you provide in the config unless you've used a lifecycle ignore changes? But even if you're using ignore changes, changing the selected repos will update the secret to the value in config. How come you're not using the specific resources for managing selected repos outside of the actual secret?

Piccirello:
The secret is defined with `ignore_changes = [plaintext_value, encrypted_value, destroy_on_drift]`.

> How come you're not using the specific resources for managing selected repos outside of the actual secret?

We're using the `selected_repository_ids` field of the `github_actions_organization_secret` resource. Should we be using `github_actions_organization_secret_repositories`? Our current goal is to create the secrets and manage their selected repositories via TF, but otherwise set the raw value via github.com.

stevehipwell:
@Piccirello your current pattern won't work as changing the selected repos will cause the secret value to be updated even if it's in lifecycle ignore changes. You should be using `github_actions_organization_secret_repositories`, but as far as I can tell this is currently causing drift to `github_actions_organization_secret` in both the released code and `main`. I've just opened #3155 to fix this, so it should be in the next release alongside these changes.

evanrappe:
qq - what's the platform-side impact of applying the changes to `destroy_on_drift` when going from v5 to v6 now? 

https://registry.terraform.io/providers/integrations/github/latest/docs/resources/actions_organization_secret#destroy_on_drift-1

Applying the below causes the secret to be updated in some way on the GH side, trying to understand what, or if it's just a one-time push of the same secret values?

```hcl
  # github_actions_organization_secret.EVAN_TEST will be updated in-place
  ~ resource "github_actions_organization_secret" "EVAN_TEST" {
      - destroy_on_drift  = true -> null
        id                = "EVAN_TEST"
        # (7 unchanged attributes hidden)
    }

stevehipwell:
@evanrappe it's a no-op as `destroy_on_drift` no longer serves any purpose. It should clean itself up?

evanrappe:
@stevehipwell yeah its a clean apply the second time, but I'm just curious what it's actually doing on the GH api when you apply the above one-time drift, as it updates the "last updated" time for the secret on the page that lists the secrets.

stevehipwell:
@evanrappe for clarity it's not a no-op, as it is putting the same secret back (it doesn't know what the secret it), but it should have no effect on your configuration. I removed the default value for `destroy_on_drift` so it gets cleaned up. If you want an actual no-op you could chose to set `destroy_on_drift`, but then you'd need to remove it before the next major version release, or add a lifecycle ignore for it.

PR #3073, opened 2026-01-11
[FEAT] Add ruleset rule for pull request required reviewers

### LINKED ISSUE #3081: [FEAT]: PR policy - require reviewer from specific team
### Describe the need

It was suggested to raise an issue here for this back in December. It never got done by the original commenter so I'm doing it.

Since this policy is not available at the enterprise level (or so it seems) we would need to manage it per-org instead, via Terraform.

### SDK Version

_No response_

### API Version

_No response_

### Relevant log output

```shell

```

### Code of Conduct

deiga:
Resolves #3081

----

### Before the change?

- It was not possible to configure PR Required Reviewers through Rulesets

### After the change?

- It is now possible to configure PR Required Reviewers through Rulesets

### Pull request checklist

- [ ] Yes
- [x] No

----

### DIFF
--- a/github/resource_github_organization_ruleset.go
+++ b/github/resource_github_organization_ruleset.go
@@ -249,7 +249,48 @@ func resourceGithubOrganizationRuleset() *schema.Resource {
 										Default:     false,
 										Description: "All conversations on code must be resolved before a pull request can be merged. Defaults to `false`.",
 									},
-									"required_reviewers": requiredReviewersSchema(),
+									"required_reviewers": {
+										Type:        schema.TypeList,
+										Optional:    true,
+										Description: "Require specific reviewers to approve pull requests targeting matching branches. Note: This feature is in beta and subject to change.",
+										Elem: &schema.Resource{
+											Schema: map[string]*schema.Schema{
+												"reviewer": {
+													Type:        schema.TypeList,
+													Required:    true,
+													MaxItems:    1,
+													Description: "The reviewer that must review matching files.",
+													Elem: &schema.Resource{
+														Schema: map[string]*schema.Schema{
+															"id": {
+																Type:        schema.TypeInt,
+																Required:    true,
+																Description: "The ID of the reviewer that must review.",
+															},
+															"type": {
+																Type:             schema.TypeString,
+																Required:         true,
+																ValidateDiagFunc: toDiagFunc(validation.StringInSlice([]string{"Team"}, false), "type"),
+																Description:      "The type of reviewer. Currently only `Team` is supported.",
+															},
+														},
+													},
+												},
+												"file_patterns": {
+													Type:        schema.TypeList,
+													Required:    true,
+													MinItems:    1,
+													Description: "File patterns (fnmatch syntax) that this reviewer must approve.",
+													Elem:        &schema.Schema{Type: schema.TypeString},
+												},
+												"minimum_approvals": {
+													Type:        schema.TypeInt,
+													Required:    true,
+													Description: "Minimum number of approvals required from this reviewer. Set to 0 to make approval optional.",
+												},
+											},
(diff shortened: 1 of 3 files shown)


github/resource_github_organization_ruleset.go
```diff
@@ -249,6 +249,7 @@ func resourceGithubOrganizationRuleset() *schema.Resource {
 										Default:     false,
 										Description: "All conversations on code must be resolved before a pull request can be merged. Defaults to `false`.",
 									},
+									"required_reviewers": requiredReviewersSchema(),
```
stevehipwell: Please could we inline this for consistency, making the schemas composable is something that should be discussed before being implemented.

github/resource_github_organization_ruleset_test.go
```diff
@@ -934,3 +934,204 @@ func TestOrganizationPushRulesetSupport(t *testing.T) {
 		t.Errorf("Expected 4 restricted file extensions, got %d", len(restrictedExts))
 	}
 }
+
+func TestAccGithubOrganizationRuleset_requiredReviewers(t *testing.T) {
```
stevehipwell: Could this be nested in the main test function, there should only be 1 top level function per resource.

stevehipwell (changes requested):

github/resource_github_organization_ruleset_test.go
deiga: I don't agree with that rule. The recommendation from Hashicorp is to use `TestAcc<ResourceName>_<test case>` (https://developer.hashicorp.com/terraform/plugin/sdkv2/testing/acceptance-tests/testcase)

It makes it easier to run tests which have the same prefix. In our current naming scheme the test pattern `TestAccGithubRepository/creates` would run ANY main test function which has `TestAccGithubRepository` and try to find any subtest with `creates` in the name instead of just the ones in `TestAccGithubRepository$`

Here is another example of how the pattern should be used https://github.com/bflad/tfproviderlint/tree/main/passes/AT003

github/resource_github_organization_ruleset.go
deiga: Of course I can.
We now (soon) have 3 Ruleset resources and 3 Ruleset data sources and we already have the issue that we are getting out of sync with the SDK/API as until now updates have been made to a single resource in need.

What are the major drawbacks, especially for entirely new fields, of starting with schema composition that require a formal discussion?

stevehipwell: > What are the major drawbacks, especially for entirely new fields, of starting with schema composition that require a formal discussion?

I'd suggest the discussion needs to be framed the other way, what problem is big enough that we'd sacrifice readability to solve it? For me there isn't one in a well governed codebase, so this kind of discussion can't really take place until we've got a better structured codebase. The idiom in Go is to prefer copying over attempting to refactor the code to be as DRY as possible.

> We now (soon) have 3 Ruleset resources and 3 Ruleset data sources and we already have the issue that we are getting out of sync with the SDK/API as until now updates have been made to a single resource in need.

The problem isn't the code structure, it's the governance structure. If resources are related then code changes should only be accepted as an atomic unit rather than piecemeal.

github/resource_github_organization_ruleset_test.go
stevehipwell: The test structure in this repo is nested, which allows us to easily run all tests for a specific resource or data source. The discoverability story for tests where there are multiple functions is significantly worse and the naming convention proposed by Hashicorp falls over as soon as you use it for more than a hello world; tests don't generally fall into a single category, which is why Go uses a regex for test discovery.

stevehipwell:

deiga:
@stevehipwell Done

PR #3080, opened 2026-01-15
feat: add enterprise (inherited) runner group org setting resource

### LINKED ISSUE #2768: [FEAT]: Enterprise Runner Group Repository Selection
### Describe the need

We use the [github_enterprise_actions_runner_group](https://registry.terraform.io/providers/integrations/github/latest/docs/resources/enterprise_actions_runner_group) resource to create Enterprise Runner Groups & we want to configure those for use with selected repositories in specific Orgs. We use the `selected_organization_ids` field of that resource to select the required Orgs however we would like a way to automate the selection of repositories for Enterprise Runner Groups inside each Org.

The only relevant resource we found was [github_actions_runner_group](https://registry.terraform.io/providers/integrations/github/latest/docs/resources/actions_runner_group) which does have the ability to select repository IDs, however that one creates a new Org-level runner group of the same name.

Perhaps there should be a new resource that allows repository selection for Enterprise Runner Groups? e.g.

```
# prerequisite - an Org
  name = "github"
}

# prerequisite - a Repo
  name = "my-repository"
}

# new resource type
resource "github_enterprise_actions_runner_group_repositories" "example" {
  organization = data.github_organization.example.id
  selected_repositories = [github_repository.example.repo_id]
}
```

That said, if I've missed some other way to add repository selections on an Enterprise Runner Group please do correct me.

### SDK Version

latest

### API Version

latest

### Relevant log output

```shell

```

### Code of Conduct

coopernetes:
Add new resource `github_enterprise_actions_runner_group_org_settings`.

This resource allows managing organization-level settings for enterprise
Actions runner groups that are inherited by an organization. It configures
various org-specific settings such as allowed repositories, visibility and
workflow restrictions. The existing org-level resource assumes that an
organization runner group is to be created & fully managed by Terraform so
it does not work for an enterprise-scope group that is shared with one or
more organizations within a GitHub Enterprise environment.

Resolves #2768

----

### Before the change?

* N/A, this is a new resource and new behaviour in the provider.

### After the change?

* A new resource called `github_enterprise_actions_runner_group_org_settings` can be created
  to manage an enterprise-scoped, inherited runner group for a GitHub organization. Note that a pre-requisite resource `github_enterprise_actions_runner_group` is required and is depended on existing before org settings can be created & managed.

### Pull request checklist

- [ ] Yes
- [x] No

----

### DIFF
--- a/github/provider.go
+++ b/github/provider.go
@@ -247,6 +247,7 @@ func NewProvider(version, commit string) func() *schema.Provider {
 				"github_enterprise_ip_allow_list_entry":                                 resourceGithubEnterpriseIpAllowListEntry(),
 				"github_enterprise_actions_workflow_permissions":                        resourceGithubEnterpriseActionsWorkflowPermissions(),
 				"github_actions_organization_workflow_permissions":                      resourceGithubActionsOrganizationWorkflowPermissions(),
+				"github_organization_inherited_runner_group_settings":                   resourceGithubOrganizationInheritedRunnerGroupSettings(),
 				"github_enterprise_security_analysis_settings":                          resourceGithubEnterpriseSecurityAnalysisSettings(),
 				"github_workflow_repository_permissions":                                resourceGithubWorkflowRepositoryPermissions(),
 			},
--- a/github/resource_github_organization_inherited_runner_group_settings.go
+++ b/github/resource_github_organization_inherited_runner_group_settings.go
@@ -1,0 +1,370 @@
+package github
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"net/http"
+	"strconv"
+
+	"github.com/google/go-github/v88/github"
+	"github.com/hashicorp/terraform-plugin-log/tflog"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/validation"
+)
+
+func resourceGithubOrganizationInheritedRunnerGroupSettings() *schema.Resource {
+	return &schema.Resource{
+		Description: "Manages organization-level settings for an enterprise Actions runner group inherited by the organization.",
+
+		CreateContext: resourceGithubOrganizationInheritedRunnerGroupSettingsCreate,
+		ReadContext:   resourceGithubOrganizationInheritedRunnerGroupSettingsRead,
+		UpdateContext: resourceGithubOrganizationInheritedRunnerGroupSettingsUpdate,
+		DeleteContext: resourceGithubOrganizationInheritedRunnerGroupSettingsDelete,
+		Importer: &schema.ResourceImporter{
+			StateContext: resourceGithubOrganizationInheritedRunnerGroupSettingsImport,
+		},
+
+		Schema: map[string]*schema.Schema{
+			"organization": {
+				Type:        schema.TypeString,
+				Required:    true,
+				ForceNew:    true,
+				Description: "The GitHub organization name.",
+			},
+			"enterprise_runner_group_name": {
+				Type:        schema.TypeString,
+				Required:    true,
+				Description: "The name of the enterprise runner group inherited by the organization.",
+			},
+			"runner_group_id": {
+				Type:        schema.TypeInt,
+				Computed:    true,
+				Description: "The ID of the inherited enterprise runner group in the organization.",
+			},
+			"inherited": {
+				Type:        schema.TypeBool,
+				Computed:    true,
+				Description: "Whether this runner group is inherited from the enterprise.",
+			},
+			"visibility": {
+				Type:             schema.TypeString,
+				Optional:         true,
+				Default:          "selected",
+				Description:      "The visibility of the runner group. Can be 'all', 'selected', or 'private'.",
+				ValidateDiagFunc: validation.ToDiagFunc(validation.StringInSlice([]string{"all", "selected", "private"}, false)),
+			},
+			"selected_repository_ids": {
+				Type: schema.TypeSet,
+				Elem: &schema.Schema{
+					Type: schema.TypeInt,
+				},
+				Set:         schema.HashInt,
+				Optional:    true,
+				Description: "List of repository IDs that can access the runner group. Only applicable when visibility is set to 'selected'.",
+			},
+			"allows_public_repositories": {
+				Type:        schema.TypeBool,
+				Optional:    true,
+				Default:     false,
+				Description: "Whether public repositories can be added to the runner group.",
+			},
+			"restricted_to_workflows": {
+				Type:        schema.TypeBool,
+				Optional:    true,
+				Default:     false,
+				Description: "If 'true', the runner group will be restricted to running only the workflows specified in the 'selected_workflows' array. Defaults to 'false'.",
+			},
+			"selected_workflows": {
+				Type:        schema.TypeList,
+				Elem:        &schema.Schema{Type: schema.TypeString},
+				Optional:    true,
+				Description: "List of workflows the runner group should be allowed to run. This setting will be ignored unless restricted_to_workflows is set to 'true'.",
+			},
+		},
+	}
+}
+
+func findInheritedEnterpriseRunnerGroupByName(client *github.Client, ctx context.Context, org, name string) (*github.RunnerGroup, error) {
+	for group, err := range client.Actions.ListOrganizationRunnerGroupsIter(ctx, org, nil) {
+		if err != nil {
+			return nil, err
+		}
+		if group.GetInherited() && group.GetName() == name {
+			return group, nil
+		}
+	}
+
+	return nil, fmt.Errorf("inherited enterprise runner group '%s' not found in organization '%s'. Ensure the enterprise runner group is shared with this organization", name, org)
+}
+
+func resourceGithubOrganizationInheritedRunnerGroupSettingsCreate(ctx context.Context, d *schema.ResourceData, meta any) diag.Diagnostics {
+	client := meta.(*Owner).v3client
+
+	org := d.Get("organization").(string)
+	enterpriseRunnerGroupName := d.Get("enterprise_runner_group_name").(string)
+	visibility := d.Get("visibility").(string)
+	allowsPublicRepositories := d.Get("allows_public_repositories").(bool)
+	restrictedToWorkflows := d.Get("restricted_to_workflows").(bool)
+	selectedRepositories, hasSelectedRepositories := d.GetOk("selected_repository_ids")
+
+	selectedWorkflows := []string{}
+	if workflows, ok := d.GetOk("selected_workflows"); ok {
+		for _, workflow := range workflows.([]any) {
+			selectedWorkflows = append(selectedWorkflows, workflow.(string))
+		}
+	}
+
+	// Find the inherited enterprise runner group by name
+	runnerGroup, err := findInheritedEnterpriseRunnerGroupByName(client, ctx, org, enterpriseRunnerGroupName)
+	if err != nil {
+		return diag.FromErr(err)
+	}
+
+	runnerGroupID := runnerGroup.GetID()
+	id, err := buildID(org, strconv.FormatInt(runnerGroupID, 10))
+	if err != nil {
+		return diag.FromErr(err)
+	}
+	d.SetId(id)
+
+	if err := d.Set("runner_group_id", int(runnerGroupID)); err != nil {
+		return diag.FromErr(err)
+	}
+	if err := d.Set("inherited", runnerGroup.GetInherited()); err != nil {
+		return diag.FromErr(err)
+	}
+
+	// Update runner group settings
+	updateReq := github.UpdateRunnerGroupRequest{
+		Visibility:               new(visibility),
+		AllowsPublicRepositories: new(allowsPublicRepositories),
+		RestrictedToWorkflows:    new(restrictedToWorkflows),
+		SelectedWorkflows:        selectedWorkflows,
+	}
+
+	_, _, err = client.Actions.UpdateOrganizationRunnerGroup(ctx, org, runnerGroupID, updateReq)
+	if err != nil {
+		return diag.FromErr(err)
+	}
+
+	// Set repository access if visibility is "selected"
+	if visibility == "selected" && hasSelectedRepositories {
+		selectedRepositoryIDs := []int64{}
+		for _, id := range selectedRepositories.(*schema.Set).List() {
+			selectedRepositoryIDs = append(selectedRepositoryIDs, int64(id.(int)))
+		}
+
+		repoAccessReq := github.SetRepoAccessRunnerGroupRequest{
+			SelectedRepositoryIDs: selectedRepositoryIDs,
+		}
+
+		_, err = client.Actions.SetRepositoryAccessRunnerGroup(ctx, org, runnerGroupID, repoAccessReq)
+		if err != nil {
+			return diag.FromErr(err)
+		}
+	}
+
+	return nil
+}
+
+func resourceGithubOrganizationInheritedRunnerGroupSettingsRead(ctx context.Context, d *schema.ResourceData, meta any) diag.Diagnostics {
+	client := meta.(*Owner).v3client
+
+	org := d.Get("organization").(string)
+	runnerGroupID := int64(d.Get("runner_group_id").(int))
+
+	// Get the runner group details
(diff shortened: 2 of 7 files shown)


deiga:
Thanks for opening a PR!

~Considering [this](https://github.com/integrations/terraform-provider-github/issues/2768#issuecomment-3756825392), I wonder if it would make sense to just add some options to the existing resource 🤔~

~What do you think?~

I was proven wrong, separate resources are better

github/resource_github_enterprise_actions_runner_group_org_settings.go
```diff
@@ -0,0 +1,402 @@
...
+)
+
+func resourceGithubEnterpriseActionsRunnerGroupOrgSettings() *schema.Resource {
+	return &schema.Resource{
+		Create: resourceGithubEnterpriseActionsRunnerGroupOrgSettingsCreate,
+		Read:   resourceGithubEnterpriseActionsRunnerGroupOrgSettingsRead,
+		Update: resourceGithubEnterpriseActionsRunnerGroupOrgSettingsUpdate,
+		Delete: resourceGithubEnterpriseActionsRunnerGroupOrgSettingsDelete,
+		Importer: &schema.ResourceImporter{
+			State: resourceGithubEnterpriseActionsRunnerGroupOrgSettingsImport,
```
deiga: issue: Please use Context-aware provider functions

github/resource_github_organization_inherited_runner_group_settings.go
```diff
@@ -0,0 +1,402 @@
...
+func resourceGithubEnterpriseActionsRunnerGroupOrgSettings() *schema.Resource {
+	return &schema.Resource{
+		Create: resourceGithubEnterpriseActionsRunnerGroupOrgSettingsCreate,
+		Read:   resourceGithubEnterpriseActionsRunnerGroupOrgSettingsRead,
+		Update: resourceGithubEnterpriseActionsRunnerGroupOrgSettingsUpdate,
+		Delete: resourceGithubEnterpriseActionsRunnerGroupOrgSettingsDelete,
+		Importer: &schema.ResourceImporter{
+			State: resourceGithubEnterpriseActionsRunnerGroupOrgSettingsImport,
+		},
+
```
deiga, line 28: Please add a top-level `Description` field

github/provider.go
```diff
@@ -213,6 +213,7 @@ func Provider() *schema.Provider {
 			"github_enterprise_actions_runner_group":                                resourceGithubActionsEnterpriseRunnerGroup(),
 			"github_enterprise_actions_workflow_permissions":                        resourceGithubEnterpriseActionsWorkflowPermissions(),
 			"github_actions_organization_workflow_permissions":                      resourceGithubActionsOrganizationWorkflowPermissions(),
+			"github_enterprise_actions_runner_group_org_settings":                   resourceGithubEnterpriseActionsRunnerGroupOrgSettings(),
```
deiga: I'm not a fan of using abbreviations in the resource names, what about something like `resource_github_organization_inherited_runner_group_settings`?

github/resource_github_enterprise_actions_runner_group_org_settings.go
```diff
@@ -0,0 +1,402 @@
...
+			"organization": {
+				Type:        schema.TypeString,
+				Required:    true,
+				ForceNew:    true,
+				Description: "The GitHub organization name.",
+			},
+			"enterprise_runner_group_name": {
+				Type:        schema.TypeString,
+				Required:    true,
+				ForceNew:    true,
```
deiga: I'm not sure if `ForceNew` here is correct, since the name of a runner group can change without changing the ID. Maybe this should instead have a `CustomizeDiff` function which checks if name was changed AND `runner_group_id` changed => `ForceNew`

github/resource_github_enterprise_actions_runner_group_org_settings.go
```diff
@@ -0,0 +1,402 @@
...
+
+	for {
+		groups, resp, err := client.Actions.ListOrganizationRunnerGroups(ctx, org, opts)
+		if err != nil {
+			return nil, err
+		}
+
+		for _, group := range groups.RunnerGroups {
+			// Only match runner groups that are inherited from the enterprise
+			if group.GetName() == name && group.GetInherited() {
```
deiga: suggestion: A minor optimization would be to swat the conditions here. Checking `GetInherited()` first should be "cheaper" than doing a string comparison

github/resource_github_enterprise_actions_runner_group_org_settings.go
```diff
@@ -0,0 +1,402 @@
...
+		for _, workflow := range workflows.([]any) {
+			selectedWorkflows = append(selectedWorkflows, workflow.(string))
+		}
+	}
+
+	// Find the inherited enterprise runner group by name
+	runnerGroup, err := findInheritedEnterpriseRunnerGroupByName(client, ctx, org, enterpriseRunnerGroupName)
+	if err != nil {
+		return err
+	}
```
deiga: suggestion: moving this earlier in the function reduces unnecessary cycles if an error does happen

github/resource_github_enterprise_actions_runner_group_org_settings.go
```diff
@@ -0,0 +1,402 @@
...
+
+	// Verify it's actually inherited
+	if !runnerGroup.GetInherited() {
```
deiga: question: This is already done inside `findInheritedEnterpriseRunnerGroupByName`, is this duplication necessary?

github/resource_github_enterprise_actions_runner_group_org_settings.go
```diff
@@ -0,0 +1,402 @@
...
+		return fmt.Errorf("runner group '%s' exists but is not inherited from the enterprise. This resource only manages inherited enterprise runner groups", enterpriseRunnerGroupName)
+	}
+
+	runnerGroupID := runnerGroup.GetID()
+	d.SetId(fmt.Sprintf("%s:%d", org, runnerGroupID))
```
deiga: We have helpers for structuring the IDs :)
```suggestion
	d.SetId(buildID(org, runnerGroupID))
```

github/resource_github_enterprise_actions_runner_group_org_settings.go
```diff
@@ -0,0 +1,402 @@
...
+	if err := d.Set("runner_group_id", int(runnerGroupID)); err != nil {
+		return err
+	}
+	if err := d.Set("inherited", runnerGroup.GetInherited()); err != nil {
+		return err
+	}
+
+	// Update runner group settings
+	updateReq := github.UpdateRunnerGroupRequest{
+		Visibility:               github.String(visibility),
```
deiga:
```suggestion
		Visibility:               github.Ptr(visibility),
```

github/resource_github_enterprise_actions_runner_group_org_settings.go
```diff
@@ -0,0 +1,402 @@
...
+		AllowsPublicRepositories: &allowsPublicRepositories,
+		RestrictedToWorkflows:    &restrictedToWorkflows,
```
deiga:
```suggestion
		AllowsPublicRepositories: github.Ptr(allowsPublicRepositories),
		RestrictedToWorkflows:    github.Ptr(restrictedToWorkflows),
```

github/resource_github_enterprise_actions_runner_group_org_settings.go
```diff
@@ -0,0 +1,402 @@
...
+			return fmt.Errorf("failed to set repository access: %w", err)
+		}
+	}
+
+	return resourceGithubEnterpriseActionsRunnerGroupOrgSettingsRead(d, meta)
+}
+
+func resourceGithubEnterpriseActionsRunnerGroupOrgSettingsRead(d *schema.ResourceData, meta interface{}) error {
+	client := meta.(*Owner).v3client
+	ctx := context.WithValue(context.Background(), ctxId, d.Id())
```
deiga: This is not used
```suggestion
```

github/resource_github_enterprise_actions_runner_group_org_settings.go
```diff
@@ -0,0 +1,402 @@
...
+				d.SetId("")
+				return nil
+			}
+		}
+		return err
+	}
+
+	if runnerGroup == nil {
+		return nil
+	}
```
deiga: question: In what scenario would we end up in this code path?

github/resource_github_enterprise_actions_runner_group_org_settings.go
```diff
@@ -0,0 +1,402 @@
...
+
+	// Verify it's still inherited from the enterprise
+	if !runnerGroup.GetInherited() {
+		log.Printf("[WARN] Runner group %s is no longer inherited from the enterprise", d.Id())
+	}
```
deiga: question: Is this a valid scenario?

github/resource_github_enterprise_actions_runner_group_org_settings.go
```diff
@@ -0,0 +1,402 @@
...
+			SelectedRepositoryIDs: selectedRepositoryIDs,
+		}
+
+		_, err = client.Actions.SetRepositoryAccessRunnerGroup(ctx, org, runnerGroupID, repoAccessReq)
+		if err != nil {
+			return fmt.Errorf("failed to set repository access: %w", err)
+		}
+	}
+
+	return resourceGithubEnterpriseActionsRunnerGroupOrgSettingsRead(d, meta)
```
deiga: If necessary set any computed fields directly in `Create` or `Update`
```suggestion
	return nil
```

github/resource_github_enterprise_actions_runner_group_org_settings.go
```diff
@@ -0,0 +1,402 @@
...
+			SelectedRepositoryIDs: selectedRepositoryIDs,
+		}
+
+		_, err := client.Actions.SetRepositoryAccessRunnerGroup(ctx, org, runnerGroupID, repoAccessReq)
+		if err != nil {
+			return fmt.Errorf("failed to set repository access: %w", err)
+		}
+	}
+
+	return resourceGithubEnterpriseActionsRunnerGroupOrgSettingsRead(d, meta)
```
deiga: If necessary set any computed fields directly in `Create` or `Update`
```suggestion
	return nil
```

github/resource_github_enterprise_actions_runner_group_org_settings.go
```diff
@@ -0,0 +1,402 @@
...
+	if err := d.Set("selected_workflows", runnerGroup.SelectedWorkflows); err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func resourceGithubEnterpriseActionsRunnerGroupOrgSettingsUpdate(d *schema.ResourceData, meta interface{}) error {
+	client := meta.(*Owner).v3client
+	ctx := context.WithValue(context.Background(), ctxId, d.Id())
```
deiga: Not used
```suggestion
```

github/resource_github_enterprise_actions_runner_group_org_settings.go
```diff
@@ -0,0 +1,402 @@
...
+
+	org := d.Get("organization").(string)
+	runnerGroupID := int64(d.Get("runner_group_id").(int))
+
+	log.Printf("[INFO] Removing repository access for runner group: %s", d.Id())
+
+	// Reset to "all" visibility and clear repository access
+	updateReq := github.UpdateRunnerGroupRequest{
+		Visibility: github.String("all"),
```
deiga:
```suggestion
		Visibility: github.Ptr("all"),
```

github/resource_github_enterprise_actions_runner_group_org_settings.go
```diff
@@ -0,0 +1,402 @@
...
+			}
+		}
+		return fmt.Errorf("failed to reset runner group visibility: %w", err)
+	}
+
+	return nil
+}
+
+func resourceGithubEnterpriseActionsRunnerGroupOrgSettingsImport(d *schema.ResourceData, meta interface{}) ([]*schema.ResourceData, error) {
+	parts := strings.Split(d.Id(), ":")
```
deiga:
```suggestion
	org, identifier, err := parseID2(d.Id())
```

github/resource_github_enterprise_actions_runner_group_org_settings.go
```diff
@@ -0,0 +1,402 @@
...
+	if len(parts) != 2 {
+		return nil, fmt.Errorf("invalid import ID format, expected 'organization:enterprise_runner_group_name' or 'organization:runner_group_id'")
```
deiga: The user needs to understand that they can't use the runner group ID from the enterprise, but the Org. They are different 😬 
```suggestion
		return nil, fmt.Errorf("invalid import ID format, expected 'organization:enterprise_runner_group_name' or 'organization:organization_runner_group_id'")
```

github/resource_github_enterprise_actions_runner_group_org_settings_test.go
```diff
@@ -0,0 +1,355 @@
...
+		t.Skip("Skipping because `ENTERPRISE_ACCOUNT` is not set or set to false")
+	}
+
+	if testEnterprise == "" {
+		t.Skip("Skipping because `ENTERPRISE_SLUG` is not set")
+	}
+
+	if testOrganization == "" {
+		t.Skip("Skipping because `GITHUB_ORGANIZATION` is not set")
+	}
```
deiga: Please use the new `skipUnless*` functions, you don't need to worry about checking these fields
```suggestion
```

github/resource_github_enterprise_actions_runner_group_org_settings_test.go
```diff
@@ -0,0 +1,355 @@
...
+				slug = "%s"
+			}
+
+			data "github_organization" "org" {
+				name = "%s"
+			}
+
+			# Create a test repository
+			resource "github_repository" "test" {
+				name        = "tf-acc-test-repo-%s"
```
deiga: Please use `testResourcePrefix` to generate the repo name

github/resource_github_enterprise_actions_runner_group_org_settings_test.go
```diff
@@ -0,0 +1,355 @@
...
+						Config: config,
+						Check:  check,
+					},
+				},
+			})
+		}
+
+		t.Run("with an enterprise account", func(t *testing.T) {
+			testCase(t, enterprise)
+		})
```
deiga: This structure is no longer supported
```suggestion
```

deiga (changes requested):

github/provider.go
coopernetes: Makes sense to me, will update the name. `inherited_runner_group` is also more descriptive, appreciate the feedback!

coopernetes:
Thanks for the review @deiga - comments addressed, this should be ready for another :eyes:

github/resource_github_organization_inherited_runner_group_settings.go
```diff
@@ -0,0 +1,383 @@
...
+	org := d.Get("organization").(string)
+	runnerGroupID := int64(d.Get("runner_group_id").(int))
+
+	// Get the runner group details
+	runnerGroup, _, err := client.Actions.GetOrganizationRunnerGroup(ctx, org, runnerGroupID)
+	if err != nil {
+		if ghErr, ok := err.(*github.ErrorResponse); ok {
+			if ghErr.Response.StatusCode == http.StatusNotFound {
+				log.Printf("[INFO] Removing actions organization runner group %s from state because it no longer exists in GitHub",
```
deiga: Please refactor to use `tflog`

github/resource_github_organization_inherited_runner_group_settings.go
```diff
@@ -0,0 +1,383 @@
...
+
+	// Get repository access list only if visibility is "selected"
+	if runnerGroup.GetVisibility() == "selected" {
+		selectedRepositoryIDs := []int64{}
+		opts := &github.ListOptions{
+			PerPage: maxPerPage,
+		}
+
+		for {
+			repos, resp, err := client.Actions.ListRepositoryAccessRunnerGroup(ctx, org, runnerGroupID, opts)
```
deiga: Does this have the new Iter accessor? Use that if possible

deiga (changes requested):
Partial review

coopernetes:
Hey @deiga sorry for the delay on this. This PR should be following better conventions now as per your prior review comments. Please have a look when you get a chance and if you don't mind re-running CI, that would be appreciated!

I also noticed that I missed documentation check list so I'll be making one more set of changes to convert the static markdown into the expected template with example TF code. 

Thanks again!

PR #3084, opened 2026-01-16
[BUG] Repository creation as `archived` would fail

deiga:
----

### Before the change?

- Creating a new `github_repository` resource with `archived = true` and `visibility = "private"` would fail with 403
- Docs stated that unarchiving a repository via API wasn't supported

### After the change?

- Creating a new `github_repository` resource with `archived = true` will "stop" execution in `Create` to only set `archived` on the repo.
- Docs updated to show that unarchiving is supported

### Pull request checklist

- [ ] Yes
- [x] No

----

### DIFF
--- a/github/resource_github_repository.go
+++ b/github/resource_github_repository.go
@@ -793,7 +793,7 @@ func resourceGithubRepositoryCreate(ctx context.Context, d *schema.ResourceData,
 		}
 	}
 
-	err := updateVulnerabilityAlerts(d, client, ctx, owner, repoName)
+	err := updateVulnerabilityAlerts(ctx, d, client, owner, repoName)
 	if err != nil {
 		return diag.FromErr(err)
 	}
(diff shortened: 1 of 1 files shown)


deiga:
> @deiga could we not just add a diff check function and disable the creation of archived repositories full stop? We could re-visit the functionality once the repository resource has been refactored.
> 
> On a separate note, the repository resource really needs to be refactored to avoid the calling of `create` -> `update` -> `read` or `update` -> `read`.

Yes please to the latter. 

Fair point to the former. I'll need to revisit this PR as my solution probably didn't consider "what should we allow"

stevehipwell:
@deiga do you already have an issue tracking the refactoring work?

deiga:
> @deiga do you already have an issue tracking the refactoring work?

@stevehipwell Not for the whole repo changes, just the vulnerability alerts

stevehipwell:
@deiga FYI https://github.com/integrations/terraform-provider-github/issues/3133

PR #3087, opened 2026-01-16
[BUG]: have an option to set org level forking policy

lee2sh:
Resolves [#ISSUE_NUMBER](https://github.com/integrations/terraform-provider-github/issues/3088)

----

### Before the change?

When managing a repository in an organization that has disabled forking, setting allow_forking on the github_repository resource would cause API errors (HTTP 422) because the GitHub API rejects the allow_forking parameter when the organization doesn't permit forking.

- 

### After the change?

Added a new org_allow_forking attribute (defaults to true) that controls whether the allow_forking setting is sent to the Github API. When `org_allow_forking` = `false`, the provider skips sending allow_forking to the API, preventing errors for organizations that have disabled forking

- 

### Pull request checklist

- [X ] Tests for the changes have been added (for bug fixes / features)

- [ ] Yes
- [ ] No

----

### DIFF
--- a/github/config.go
+++ b/github/config.go
@@ -134,6 +134,14 @@ func (c *Config) ConfigureOwner(owner *Owner) (*Owner, error) {
 			if remoteOrg != nil {
 				owner.id = remoteOrg.GetID()
 				owner.IsOrganization = true
+				// MembersCannotForkPrivateRepos is intentionally inverted from the API's
+				// MembersCanForkPrivateRepos field. This ensures the default zero value (false)
+				// means "no restriction" which is correct for:
+				// - Personal accounts (non-org): forking is always allowed, no org policy applies
+				// - Organizations that allow forking: MembersCanForkPrivateRepos=true → inverted to false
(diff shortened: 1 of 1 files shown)


stevehipwell (changes requested):
I don't think the PR is necessary, nor do I think the UX works.

FYI #3094 fixes this issue as it's just an implementation bug.

lee2sh:
> I don't think the PR is necessary, nor do I think the UX works.
> 
> FYI #3094 fixes this issue as it's just an implementation bug.

The PR has an edge case. If a user explicitly sets `allow_forking` (true or false) on an organization that has disabled private repo forking, they'll likely still get a 422 error. The GetOk() fix only handles the case where `allow_forking` is unset. I believe with this approach, it would proactively respect org settings by omitting the field from the request entirely.

stevehipwell:
@lee2sh the GitHub API and therefore this provider are full of edge cases. This PR just moves the edges around without fixing the source of the issue. An actual improvement would be to lookup the org settings as a data source but the API doesn't exist.

lee2sh:
gotcha makes sense. will close the PR.

PR #3094, opened 2026-01-19
chore: Fix repo allow forking

### LINKED ISSUE #3088: [BUG]: allow_forking option breaks repo creation when the org doesn't allow forking
### Expected Behavior

The repository should be created normally.

### Actual Behavior

I'm getting 422 error when I try to create a repository because my org doesn't allow forking (forking disabled), but it sends the request with `allow_forking` option regardless.

### Terraform Version

opentofu version 1.11.0

### Affected Resource(s)

- github repository

```hcl

```

### Steps to Reproduce

Disable forking at the org level and try to create a repository using the latest github provider

### Debug Output

```shell

```

### Panic Output

```shell

```

### Code of Conduct

stevehipwell:
Resolves #3088

----

### Before the change?

- The `github_repository` resource incorrectly sends `AllowForking` even when not set.

### After the change?

- The `github_repository` resource only sends `AllowForking` when not set.

### Pull request checklist

- [ ] Yes
- [x] No

----

### DIFF
--- a/github/resource_github_repository.go
+++ b/github/resource_github_repository.go
@@ -634,9 +634,10 @@ func resourceGithubRepositoryObject(d *schema.ResourceData) *github.Repository {
 	}
 
 	// only configure allow forking if repository is not public
-	allowForking, ok := d.Get("allow_forking").(bool)
-	if ok && visibility != "public" {
-		repository.AllowForking = github.Ptr(allowForking)
+	if allowForking, ok := d.GetOk("allow_forking"); ok && visibility != "public" {
+		if val, ok := allowForking.(bool); ok {
+			repository.AllowForking = github.Ptr(val)
+		}
 	}
 
 	return repository
--- a/github/resource_github_repository_test.go
+++ b/github/resource_github_repository_test.go
@@ -571,7 +571,7 @@ func TestAccGithubRepository(t *testing.T) {
 
 	t.Run("create_private_with_forking", func(t *testing.T) {
 		randomID := acctest.RandStringFromCharSet(5, acctest.CharSetAlphaNum)
-		repoName := fmt.Sprintf("%svisibility-%s", testResourcePrefix, randomID)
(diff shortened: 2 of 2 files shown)


deiga:
@stevehipwell is the secret drift in here accidentally?

github/resource_github_repository.go
```diff
@@ -634,9 +634,8 @@ func resourceGithubRepositoryObject(d *schema.ResourceData) *github.Repository {
 	}
 
 	// only configure allow forking if repository is not public
-	allowForking, ok := d.Get("allow_forking").(bool)
-	if ok && visibility != "public" {
-		repository.AllowForking = github.Ptr(allowForking)
+	if allowForking, ok := d.GetOk("allow_forking"); ok && visibility != "public" {
+		repository.AllowForking = github.Ptr(allowForking.(bool))
```
deiga: nit: I don't like this pattern. Inlining the check for `ok` I can somewhat live with, but having to parse a condition and the actual action isn't IMO great readability. 

Could the if be separated again?

stevehipwell: This is the correct pattern for checking if a field is set, it blocks the variable leaking out of the scope.

I've actually just updated my prefered pattern to also check the conversion as I think we should always do that even though the type is in the schema.

The alternative to the pattern above would be this where the visibility check gets its own line.

```go
	if allowForking, ok := d.GetOk("allow_forking"); ok {
		if visibility != "public {
			if val, ok := allowForking.(bool); ok {
				repository.AllowForking = github.Ptr(val.(bool))
			}
		}
	}
```

stevehipwell: Actually we could swap the check order.

```go
	if visibility != "public {
		if allowForking, ok := d.GetOk("allow_forking"); ok {
			if val, ok := allowForking.(bool); ok {
				repository.AllowForking = github.Ptr(val.(bool))
			}
		}
	}
```

PR #3095, opened 2026-01-19
feat: add excluded_members support to team settings

### LINKED ISSUE #1972: [FEAT]: Exclude list for Team Code Review settings

janeklb:
Implements support for excluding specific team members from PR review assignments.

Resolves #1972

## Changes:
- Add `excluded_members` field to `review_request_delegation`
  - values are GitHub usernames
  - provider resolves usernames to node IDs in one GraphQL request
- Use the generated GraphQL mutation input to send excluded team member IDs
- Preserve the configured exclusion list in Terraform state because the GitHub Team query does not expose it
- Add acceptance coverage
- Update documentation with the new field

## Usage:
```hcl
resource "github_team_settings" "example" {
  team_id = github_team.example.id
  review_request_delegation {
    algorithm = "ROUND_ROBIN"
    member_count = 2
    excluded_members = ["octocat"]
  }
}
```


github/resource_github_team_settings.go
```diff
@@ -151,6 +159,9 @@ func resourceGithubTeamSettingsRead(d *schema.ResourceData, meta any) error {
 			return err
 		}
 	}
+	// NOTE: The exclusion list is not available via the GraphQL read query yet.
+	// The excluded_team_member_node_ids field can be set but cannot be read back from the GitHub API.
+	// This is because the GraphQL API for team review assignments is currently in preview.
```
janeklb: This comment (alongside the rest of this PR) was generated by copilot. I instructed copilot to take inspiration from https://github.com/integrations/terraform-provider-github/pull/1965/changes so maybe that's why it added it.

github/resource_github_team_settings.go
```diff
@@ -83,6 +105,14 @@ func resourceGithubTeamSettings() *schema.Resource {
 							Default:     false,
 							Description: "whether to notify the entire team when at least one member is also assigned to the pull request.",
 						},
+						"excluded_team_members": {
```
stevehipwell: We're in the context of a team, so I'm not sure why we need team in this name?

github/resource_github_team_settings.go
```diff
@@ -177,12 +210,27 @@ func resourceGithubTeamSettingsUpdate(d *schema.ResourceData, meta any) error {
 				} `graphql:"updateTeamReviewAssignment(input:$input)"`
 			}
 
+			exclusionList := make([]githubv4.ID, 0)
+			if excludedMembers, ok := settings["excluded_team_members"]; ok && excludedMembers != nil {
+				for _, v := range excludedMembers.(*schema.Set).List() {
+					if v != nil {
+						username := v.(string)
+						nodeId, err := getUserNodeId(ctx, meta, username)
```
stevehipwell: This is an expensive operation that you're calling repeatedly, have you looked at making this require fewer API calls?

stevehipwell (changes requested):
@janeklb based on your comment am I correct in thinking that Copilot created this whole PR for you? Are you a subject matter expert in Terraform or the GitHub API?

github/resource_github_team_settings.go
janeklb: Fair -- would you suggest `excluded_members`?

github/resource_github_team_settings.go
janeklb: I haven't looked into it but it did cross my mind to see if there's a "batch" endpoint that accepts a list of usernames instead of going one by one. I will check it out

janeklb:
> @janeklb based on your comment am I correct in thinking that Copilot created this whole PR for you? Are you a subject matter expert in Terraform or the GitHub API?

Correct -- I am not an sme in these areas. I've been wanting this feature so I tried the "code with copilot" button and iterated on the solution a little bit.

github/resource_github_team_settings.go
janeklb: done

github/resource_github_team_settings.go
janeklb: updated -- please take a look

stevehipwell:
@janeklb I don't think we can take this PR while there isn't a way to query the current state, that fundamentally breaks how Terraform works.

janeklb:
> @janeklb I don't think we can take this PR while there isn't a way to query the current state, that fundamentally breaks how Terraform works.

Totally fair

PR #3100, opened 2026-01-20
fix: Restrict allow_forking to org-owned repositories only

### LINKED ISSUE #3088: [BUG]: allow_forking option breaks repo creation when the org doesn't allow forking

smur89:
The allow_forking field should only be set for org-owned repositories.

GitHub API returns 422 error when this field is included in requests for private user-owned repositories.

Resolves #3088

----

### Before the change?

Screenshot 2026-01-20 at 18 27 22

### After the change?

- 

### Pull request checklist

- [ ] Yes
- [ ] No

----

### DIFF
--- a/github/resource_github_repository.go
+++ b/github/resource_github_repository.go
@@ -586,7 +586,7 @@ func calculateSecurityAndAnalysis(d *schema.ResourceData) *github.SecurityAndAna
 	return &securityAndAnalysis
 }
 
-func resourceGithubRepositoryObject(d *schema.ResourceData) *github.Repository {
+func resourceGithubRepositoryObject(d *schema.ResourceData, meta any) *github.Repository {
 	visibility := calculateVisibility(d)
 
 	repository := &github.Repository{
@@ -633,8 +633,9 @@ func resourceGithubRepositoryObject(d *schema.ResourceData) *github.Repository {
 		}
 	}
 
-	// only configure allow forking if repository is not public
-	if allowForking, ok := d.GetOk("allow_forking"); ok && visibility != "public" {
+	// only configure allow forking if repository is org-owned and not public
+	// allow_forking can only be changed on org-owned repositories
+	if allowForking, ok := d.GetOk("allow_forking"); ok && visibility != "public" && meta.(*Owner).IsOrganization {
 		if val, ok := allowForking.(bool); ok {
 			repository.AllowForking = github.Ptr(val)
 		}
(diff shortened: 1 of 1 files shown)


stevehipwell:
@smur89 #3088 is already fixed by #3094.

smur89:
> @smur89 #3088 is already fixed by #3094.

@stevehipwell that fix is not handling private user-owned repositories, which is the issue I ran in to

stevehipwell:
@smur I'm not sure how it's not handling user-owned repositories given that if you don't set the new `allow_forking` attribute it's a no-op over previous versions?

If this is broken for you then please open a new issue with the full details.

pasdam:
I'm not sure it's related, I'm just reporting it here in case it helps others: I'm using the version `6.10.2` and I was getting the same error while setting `allow_auto_merge = true` on my own private repo, not part of any org, and I wasn't setting any forking config.

That feature is not available for private repos, so setting that to `false` solved the issue for me.

I guess the error was coming directly from Github, but it was misleading.

stevehipwell:
@pasdam the `allow_forking` implementation is correct as of the `v6.11.1` release and the [documentation](https://registry.terraform.io/providers/integrations/github/latest/docs/resources/repository#allow_forking-1) covers the fact that this is for organization repositories.

pasdam:
Sorry, I guess I wasn't clear: I was just suggesting that in my case the error was not related to `allow_forking`, which I wasn't using at all, event though the message was pointing me in that direction, but rather the fact I was using `allow_auto_merge = true` on a private personal repo, which is not allowed.

The error is misleading, and I guess it's coming from GitHub API directly, not from the terraform provider, but I might be wrong about that. Anyway if I find the time I'll test the latest version to see if that fixes my issue as well.

I was just posting here in case some other folks would come here having the same issue I was facing.

stevehipwell:
@pasdam as you didn't provide an example of what you were seeing or how it related to forking I don't see how you're adding any value here? Also you're not on the latest version, or the version that fixes `allow_forking`, so your comment is at best stale.

Issue #3105, opened 2026-01-21
Regression: environments cannot have `:` charcaters


acouvreur:
Introduced by: https://github.com/integrations/terraform-provider-github/blame/15fff783674a8c26cf536a3f84ef662520baccb7/github/resource_github_repository_environment.go#L234

Our environments have the `:` character, and since update [v6.10.0](https://github.com/integrations/terraform-provider-github/releases/tag/v6.10.0) we can no longer use the provider to provision our environments.

We have the following error message:

```
Error: unexpected ID format ("some-repository:my:environment")
```

Where we have one environment named `my:environment`.

---

I can probably do a fix, but the issue is that everything relies on the `parseID` method which ensures that it properly split into a given count.

https://github.com/integrations/terraform-provider-github/blob/15fff783674a8c26cf536a3f84ef662520baccb7/github/util.go#L57-L68

---

I would fix this by ensuring that `parseID2` splits in at least 2, and collect all the remaining splits in the second argument. Same for 3. Or by using a different separator but I'm not sure what the implications would be.

I'll do a PR and you'll give me some feedback.

stevehipwell:
@acouvreur is this for an existing environment (e.g. one created before v6.10.x)?

acouvreur:
> @acouvreur is this for an existing environment (e.g. one created before v6.10.x)?

Yes

stevehipwell:
@acouvreur I've opened #3129 to fix this regression. I missed that `github_repository_environment` would work fine without the IDs being escaped as all the other resources were broken by the lack of escaping `:`.

If you want to fix this before the next release you can remove and then re-[import](https://registry.terraform.io/providers/integrations/github/latest/docs/resources/repository_environment#import) the resource using the correctly escaped ID.

PR #3106, opened 2026-01-21
fix(parseID): separator in value

acouvreur:
parseID now ensures that there is at least count parts in the string ID. It then joins the extra together

Resolves https://github.com/integrations/terraform-provider-github/issues/3105
----

### Before the change?

- parseID2 errors with ID: "part1:part2:part3"

### After the change?

- parseID2 returns: "part1" and "part2:part3"

### Pull request checklist

- [ ] Yes
- [x] No

It actually fix a breaking change that some values have the separator.

----

### DIFF
--- a/github/util.go
+++ b/github/util.go
@@ -60,8 +60,15 @@ func parseID(id string, count int) ([]string, error) {
 	}
 
 	parts := strings.Split(id, idSeparator)
-	if len(parts) != count {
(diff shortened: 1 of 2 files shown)


stevehipwell:
@acouvreur this PR is invalid as it breaks the intended functionality. Please could we discuss on #3105 to come up with a solution.

PR #3110, opened 2026-01-22
[FEAT]: Add Support for GitHub Enterprise Rulesets

### LINKED ISSUE #2666: [FEAT]: Add Support for GitHub Enterprise Rulesets
### Describe the need

Currently, the Terraform GitHub provider supports [organization rulesets](https://registry.terraform.io/providers/integrations/github/latest/docs/resources/organization_ruleset) and [repository rulesets](https://registry.terraform.io/providers/integrations/github/latest/docs/resources/repository_ruleset). However, there is no support for rulesets at the enterprise level which GitHub supports on its UI (if you're on the Enterprise plan).

More information around rulesets:
- [About rulesets](https://docs.github.com/en/enterprise-cloud@latest/repositories/configuring-branches-and-merges-in-your-repository/managing-rulesets/about-rulesets)
- [Managing code rulesets for repositories in your enterprise](https://docs.github.com/en/enterprise-cloud@latest/admin/enforcing-policies/enforcing-policies-for-your-enterprise/managing-policies-for-code-governance)

### SDK Version

_No response_

### API Version

_No response_

### Relevant log output

```shell

```

### Code of Conduct

Ravio1i:
Resolves #2666

----

### Before the change?

No possibility to create or fetch [rulesets](https://docs.github.com/en/enterprise-cloud@latest/rest/repos/rules) at enterprise level

### After the change?

- Create rulesets with resource
- Fetch rulesets with datasource

### Pull request checklist

- [ ] Yes
- [x] No

### Notes 

- Tested with own GitHub Enterprise Cloud Instance 
- Currently there is a race condition in the test `testAccGithubEnterpriseRuleset_required_workflows`: `Error running post-test destroy, there may be dangling resources for `vulnerability-alerts`. The test still works but the vulnerabtily alert seems to be not finished in time. Happy to hear feedback about this
----


github/resource_github_enterprise_ruleset.go
```diff
@@ -0,0 +1,860 @@
...
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/validation"
+)
+
+func resourceGithubEnterpriseRuleset() *schema.Resource {
+	return &schema.Resource{
+		CreateContext: resourceGithubEnterpriseRulesetCreate,
+		ReadContext:   resourceGithubEnterpriseRulesetRead,
+		UpdateContext: resourceGithubEnterpriseRulesetUpdate,
+		DeleteContext: resourceGithubEnterpriseRulesetDelete,
+
```
deiga, line 46: Please add `Importer` functionality :)

github/resource_github_enterprise_ruleset.go
```diff
@@ -0,0 +1,860 @@
...
+		SchemaVersion: 1,
+
```
deiga, line 48: Please add a top-level `Description`

github/resource_github_enterprise_ruleset.go
```diff
@@ -0,0 +1,860 @@
...
+)
+
+func resourceGithubEnterpriseRuleset() *schema.Resource {
+	return &schema.Resource{
+		CreateContext: resourceGithubEnterpriseRulesetCreate,
+		ReadContext:   resourceGithubEnterpriseRulesetRead,
+		UpdateContext: resourceGithubEnterpriseRulesetUpdate,
+		DeleteContext: resourceGithubEnterpriseRulesetDelete,
+
+		SchemaVersion: 1,
```
deiga: Please don't set `SchemaVersion` until there is a need for migrating the Schema
```suggestion
```

github/resource_github_enterprise_ruleset.go
```diff
@@ -0,0 +1,860 @@
...
+						"actor_id": {
+							Type:        schema.TypeInt,
+							Optional:    true,
+							Default:     nil,
+							Description: "The ID of the actor that can bypass a ruleset. When `actor_type` is `OrganizationAdmin`, this should be set to `1`. Some resources such as DeployKey do not have an ID and this should be omitted.",
+						},
+						"actor_type": {
+							Type:         schema.TypeString,
+							Required:     true,
+							ValidateFunc: validation.StringInSlice([]string{"Integration", "OrganizationAdmin", "RepositoryRole", "Team", "DeployKey"}, false),
```
deiga: issue: This seems to be missing some types. From API docs: Integration, OrganizationAdmin, RepositoryRole, Team, DeployKey, EnterpriseOwner

github/resource_github_enterprise_ruleset.go
```diff
@@ -0,0 +1,860 @@
...
+							Description:  "The type of actor that can bypass a ruleset. See https://docs.github.com/en/rest/orgs/rules for more information",
```
deiga: issue: The link should target correct docs for Enterprise rules

github/resource_github_enterprise_ruleset.go
```diff
@@ -0,0 +1,860 @@
...
+								},
+							},
+						},
+						"repository_name": {
+							Type:         schema.TypeList,
+							Optional:     true,
+							MaxItems:     1,
+							Description:  "Conditions for repository names that the ruleset targets. Conflicts with `repository_id`.",
+							ExactlyOneOf: []string{"conditions.0.repository_id"},
+							AtLeastOneOf: []string{"conditions.0.repository_id"},
```
deiga: Using `ExactlyOneOf` and/or `AtLeastOneOf` would make these fields required, but there are other options as well.
```suggestion
							ConflictsWith: []string{"conditions.0.repository_id"},
```

See API docs:
```
Conditions for an enterprise ruleset. The conditions object supports either of the following combinations: - organization_id and repository_name - organization_id and repository_property - organization_name and repository_name - organization_name and repository_property - organization_property and repository_name - organization_property and repository_property For branch and tag rulesets, the conditions object should also contain the ref_name property.
```

github/resource_github_enterprise_ruleset.go
```diff
@@ -0,0 +1,860 @@
...
+			"error":           err.Error(),
+		})
+		return diag.FromErr(err)
+	}
+
+	d.SetId(strconv.FormatInt(*ruleset.ID, 10))
+	_ = d.Set("ruleset_id", ruleset.ID)
+	_ = d.Set("node_id", ruleset.GetNodeID())
+	_ = d.Set("etag", resp.Header.Get("ETag"))
+
```
deiga: Please handle any errors that come from here

github/resource_github_enterprise_ruleset.go
```diff
@@ -0,0 +1,860 @@
...
+	_ = d.Set("ruleset_id", ruleset.ID)
+	_ = d.Set("name", ruleset.Name)
+	_ = d.Set("target", ruleset.GetTarget())
+	_ = d.Set("enforcement", ruleset.Enforcement)
+	_ = d.Set("bypass_actors", flattenBypassActors(ruleset.BypassActors))
+	_ = d.Set("conditions", flattenConditions(ruleset.GetConditions(), true))
+	_ = d.Set("rules", flattenRules(ruleset.Rules, true))
+	_ = d.Set("node_id", ruleset.GetNodeID())
+	_ = d.Set("etag", resp.Header.Get("ETag"))
```
deiga: Please handle any errors from these

github/resource_github_enterprise_ruleset.go
```diff
@@ -0,0 +1,860 @@
...
+	if err != nil {
+		tflog.Error(ctx, fmt.Sprintf("Could not convert ruleset ID '%s' to int64", d.Id()), map[string]any{
+			"enterprise_slug": enterpriseSlug,
+			"ruleset_id":      d.Id(),
+			"error":           err.Error(),
+		})
+		return diag.FromErr(unconvertibleIdErr(d.Id(), err))
+	}
+
+	tflog.Debug(ctx, fmt.Sprintf("Updating enterprise ruleset: %s/%d", enterpriseSlug, rulesetID), map[string]any{
```
deiga: Please don't use `fmt.Sprintf` inside `tflog` calls

github/resource_github_enterprise_ruleset_test.go
```diff
@@ -0,0 +1,750 @@
...
+		}
+		`
+	config := fmt.Sprintf(rulesetHCL, testAccConf.enterpriseSlug, rulesetName)
+
+	check := resource.ComposeTestCheckFunc(
+		resource.TestCheckResourceAttr("github_enterprise_ruleset.test", "enterprise_slug", testAccConf.enterpriseSlug),
+		resource.TestCheckResourceAttr("github_enterprise_ruleset.test", "name", rulesetName),
+		resource.TestCheckResourceAttr("github_enterprise_ruleset.test", "target", "branch"),
+		resource.TestCheckResourceAttr("github_enterprise_ruleset.test", "enforcement", "active"),
+	)
```
deiga: Please inline this

deiga (changes requested):
Initial pass. Not a complete review.

Please see if there are other parts of the PR where my comments might be applicable :)

deiga:
You're doing good work!

I'm wondering if there are patterns from https://github.com/integrations/terraform-provider-github/pull/2958 you should be copying here (for example the conditions and rules validation)

Ravio1i:
> You're doing good work!

Thanks! although I wasted a lot of time today being utterly confused of what is actually implemented and exposed in the API/golang sdk. E.g there are parts in the go sdk, which indicate that a ruleset can use the new `repository_properties`, `organization_properties` or  `merge_queue`, but the actual API of the Enterprise Cloud is not supporting it 🔥 

> I'm wondering if there are patterns from #2958 you should be copying here (for example the conditions and rules validation)

I can check it out and see if I can adjust it !

> https://github.com/integrations/terraform-provider-github/pull/3110#discussion_r2717489689

For the importer feature: I left it out intentionally as I was under the impression it is a feature used only in UI. Not sure who actually would use it with terraform/tofu to import json in hcl.

Ravio1i:
> I'm wondering if there are patterns from #2958 you should be copying here (for example the conditions and rules validation)

Similar to the the `CustomizeDiff` in  #2958 I've added the `util_ruleset_enterprise_validation.go` to validate that:
-  `ref_name` is required for branch/tag, but forbidden for push/repository
  - creation, deletion, update, required_linear_history, required_signatures, pull_request, required_status_checks,       
  non_fast_forward, commit_message_pattern, commit_author_email_pattern, committer_email_pattern, branch_name_pattern,    
  tag_name_pattern, required_workflows, required_code_scanning, copilot_code_review rules are only valid for branch/tag   
  targets                                                                                                                 
  - file_path_restriction, max_file_size, max_file_path_length, file_extension_restriction rules are only valid for push  
  target                                                                                                                  
  - repository_creation, repository_deletion, repository_transfer, repository_name, repository_visibility rules are only  
  valid for repository target

deiga:
The Importer is functionality to enable `terraform import` or the `import` block inside the config, so that existing resources can be added to state

Ravio1i:
> The Importer is functionality to enable `terraform import` or the `import` block inside the config, so that existing resources can be added to state

Oh nevermind, yes that I will most definitely add! 

I thought you were referring to the importer functionality in GitHub rulesets. Oops

github/resource_github_enterprise_ruleset.go
Ravio1i: Added in https://github.com/integrations/terraform-provider-github/pull/3110/changes/7dd2b1310302147dd8a21f7264def74ab2e087d0

Ravio1i:
> The Importer is functionality to enable `terraform import` or the `import` block inside the config, so that existing resources can be added to state

https://github.com/integrations/terraform-provider-github/pull/3110/changes/7dd2b1310302147dd8a21f7264def74ab2e087d0

Added the importer functionality + test

```bash
TF_CLI_CONFIG_FILE=../dev.tfrc terraform import github_enterprise_ruleset.imported siemens:440187
github_enterprise_ruleset.imported: Importing from ID "siemens:440187"...
github_enterprise_ruleset.imported: Import prepared!
  Prepared github_enterprise_ruleset for import
github_enterprise_ruleset.imported: Refreshing state... [id=440187]

Import successful!
```

github/resource_github_enterprise_ruleset.go
```diff
@@ -0,0 +1,1002 @@
...
+			"name": {
+				Type:         schema.TypeString,
+				Required:     true,
+				ValidateFunc: validation.StringLenBetween(1, 100),
+				Description:  "The name of the ruleset.",
+			},
+			"target": {
+				Type:         schema.TypeString,
+				Required:     true,
+				ValidateFunc: validation.StringInSlice([]string{"branch", "tag", "push", "repository"}, false),
```
deiga: Please use `github.RulesetTarget` types instead of strings

github/resource_github_enterprise_ruleset.go
```diff
@@ -0,0 +1,1002 @@
...
+										Optional:    true,
+										Description: "Array of ref names or patterns to exclude. The condition will not pass if any of these patterns match.",
+										Elem: &schema.Schema{
+											Type: schema.TypeString,
+										},
+									},
+								},
+							},
+						},
+						"repository_name": {
```
deiga, line 284: Please add `repository_property` while you're at this

github/resource_github_enterprise_ruleset.go
```diff
@@ -0,0 +1,1002 @@
...
+			"enterprise_slug": enterpriseSlug,
+			"ruleset_id":      d.Id(),
+			"error":           err.Error(),
+		})
+		return diag.FromErr(unconvertibleIdErr(d.Id(), err))
+	}
+
+	if !d.IsNewResource() {
+		ctx = context.WithValue(ctx, ctxEtag, d.Get("etag").(string))
+	}
```
deiga, line 869:
```suggestion
```

github/resource_github_enterprise_ruleset.go
```diff
@@ -0,0 +1,1002 @@
...
+
+	_, err = client.Enterprise.DeleteRepositoryRuleset(ctx, enterpriseSlug, rulesetID)
+	if err != nil {
+		tflog.Error(ctx, "Failed to delete enterprise ruleset", map[string]any{
+			"enterprise_slug": enterpriseSlug,
+			"ruleset_id":      rulesetID,
+			"error":           err.Error(),
+		})
+		return diag.FromErr(err)
+	}
```
deiga: Add handling of 404

github/resource_github_enterprise_ruleset.go
```diff
@@ -0,0 +1,1002 @@
...
+	tflog.Info(ctx, "Deleted enterprise ruleset", map[string]any{
+		"enterprise_slug": enterpriseSlug,
+		"ruleset_id":      rulesetID,
+	})
+
+	return nil
+}
+
+func resourceGithubEnterpriseRulesetImport(ctx context.Context, d *schema.ResourceData, meta any) ([]*schema.ResourceData, error) {
+	enterpriseSlug, rulesetIDStr, err := parseTwoPartID(d.Id(), "enterprise_slug", "ruleset_id")
```
deiga:
```suggestion
	enterpriseSlug, rulesetIDStr, err := parseID2(d.Id())
```

github/resource_github_enterprise_ruleset.go
```diff
@@ -0,0 +1,1002 @@
...
+	if err != nil {
+		return []*schema.ResourceData{d}, err
```
deiga: Don't return data on error
```suggestion
		return nil, err
```

deiga: Add error message giving example of import ID

github/resource_github_enterprise_ruleset.go
```diff
@@ -0,0 +1,1002 @@
...
+		"enterprise_slug": enterpriseSlug,
+		"ruleset_id":      rulesetID,
+	})
+
+	client := meta.(*Owner).v3client
+
+	ruleset, _, err := client.Enterprise.GetRepositoryRuleset(ctx, enterpriseSlug, rulesetID)
+	if ruleset == nil || err != nil {
+		return []*schema.ResourceData{d}, err
+	}
```
deiga: `Read` will be called right after `Import`
```suggestion
```

deiga (changes requested):

github/util_ruleset_validation.go
```diff
@@ -173,9 +173,12 @@ func validateConditionsFieldForBranchAndTagTargets(ctx context.Context, target g
...
 	if isOrg {
-		if (conditions["repository_name"] == nil || len(conditions["repository_name"].([]any)) == 0) && (conditions["repository_id"] == nil || len(conditions["repository_id"].([]any)) == 0) {
-			tflog.Debug(ctx, fmt.Sprintf("Missing repository_name or repository_id for %s target", target), map[string]any{"target": target})
-			return fmt.Errorf("either repository_name or repository_id must be set for %s target", target)
+		hasRepoName := conditions["repository_name"] != nil && len(conditions["repository_name"].([]any)) != 0
+		hasRepoID := conditions["repository_id"] != nil && len(conditions["repository_id"].([]any)) != 0
+		hasRepoProp := conditions["repository_property"] != nil && len(conditions["repository_property"].([]any)) != 0
+		if !hasRepoName && !hasRepoID && !hasRepoProp {
+			tflog.Debug(ctx, fmt.Sprintf("Missing repository_name, repository_id, or repository_property for %s target", target), map[string]any{"target": target})
+			return fmt.Errorf("either repository_name, repository_id, or repository_property must be set for %s target", target)
```
deiga: question: Do you think this section would become irrelevant if we'd use `ExactlyOneOf` instead of `ConflictsWith` in the schema? (I might have tried that originally, but can't remember anymore 😬 )

deiga (approved):

examples/enterprise_rulesets/branch_rulesets.tf
```diff
@@ -0,0 +1,174 @@
...
+      include = ["~DEFAULT_BRANCH", "refs/heads/release/*"]
+      exclude = []
+    }
+  }
+
+  rules {
+    deletion         = true
+    non_fast_forward = true
+  }
+}
```
stevehipwell: Please fix missing line endings.

github/data_source_github_enterprise_ruleset.go
```diff
@@ -0,0 +1,419 @@
...
+
+	"github.com/google/go-github/v82/github"
+	"github.com/hashicorp/terraform-plugin-log/tflog"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+)
+
+func dataSourceGithubEnterpriseRuleset() *schema.Resource {
+	return &schema.Resource{
+		ReadContext: dataSourceGithubEnterpriseRulesetRead,
```
stevehipwell, line 16: Please add a description here.

github/data_source_github_enterprise_ruleset_test.go
```diff
@@ -0,0 +1,95 @@
...
+		`, testAccConf.enterpriseSlug, testRulesetName)
+
+		config2 := config + `
+			data "github_enterprise_ruleset" "test" {
+				enterprise_slug = github_enterprise_ruleset.test.enterprise_slug
+				ruleset_id      = github_enterprise_ruleset.test.ruleset_id
+			}
+		`
+
+		check := resource.ComposeTestCheckFunc(
```
stevehipwell: Please inline the checks.

github/data_source_github_enterprise_ruleset_test.go
```diff
@@ -0,0 +1,95 @@
...
+
+		resource.Test(t, resource.TestCase{
+			PreCheck: func() {
+				skipUnlessEnterprise(t)
+			},
+			ProviderFactories: providerFactories,
+			Steps: []resource.TestStep{
+				{
+					Config: config,
+					Check:  resource.ComposeTestCheckFunc(),
```
stevehipwell:
```suggestion
```

github/data_source_github_enterprise_ruleset_test.go
```diff
@@ -0,0 +1,95 @@
...
+				"data.github_enterprise_ruleset.test", "etag",
+			),
+		)
+
+		resource.Test(t, resource.TestCase{
+			PreCheck: func() {
+				skipUnlessEnterprise(t)
+			},
+			ProviderFactories: providerFactories,
+			Steps: []resource.TestStep{
```
stevehipwell, line 34: Why not use `depends_on` instead of two steps?

github/resource_github_enterprise_ruleset.go
```diff
@@ -0,0 +1,1072 @@
...
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/validation"
+)
+
+var supportedEnterpriseRulesetTargetTypes = []string{
+	string(github.RulesetTargetBranch),
+	string(github.RulesetTargetTag),
+	string(github.RulesetTargetPush),
+	string(github.RulesetTargetRepository),
+}
```
stevehipwell, line 35: This probably belongs in the utils file or at a minimum below the actual logic.

github/resource_github_enterprise_ruleset.go
```diff
@@ -0,0 +1,1072 @@
...
+				Type:         schema.TypeString,
+				Required:     true,
+				ValidateFunc: validation.StringLenBetween(1, 100),
+				Description:  "The name of the ruleset.",
+			},
+			"target": {
+				Type:             schema.TypeString,
+				Required:         true,
+				ValidateDiagFunc: validation.ToDiagFunc(validation.StringInSlice(supportedEnterpriseRulesetTargetTypes, false)),
+				Description:      "Possible values are " + strings.Join(supportedEnterpriseRulesetTargetTypes[:len(supportedEnterpriseRulesetTargetTypes)-1], ", ") + " and " + supportedEnterpriseRulesetTargetTypes[len(supportedEnterpriseRulesetTargetTypes)-1] + ". Note: The `repository` target is in preview and is subject to change.",
```
stevehipwell: I know you've copied this pattern but I don't think it's readable, especially when using the collection twice to try and make it human readable. Please just make this a literal string.

@deiga if we want to generate descriptions I think we'd need to use `go generate` as this pattern isn't readable. If you just want to check that nothing is missing you could always add a unit test to check the descrption contains all of the relevant words.

github/resource_github_enterprise_ruleset.go
```diff
@@ -0,0 +1,1072 @@
...
+									"protected": {
+										Type:        schema.TypeBool,
+										Optional:    true,
+										Default:     false,
+										Description: "Whether renaming of target repositories is prevented.",
+									},
+								},
+							},
+						},
+						"repository_id": {
```
stevehipwell: I don't see repository ID in the REST API?

github/resource_github_enterprise_ruleset.go
```diff
@@ -0,0 +1,1072 @@
...
+			"etag": {
+				Type:     schema.TypeString,
+				Computed: true,
+			},
+		},
+	}
+}
+
+// resourceGithubEnterpriseRulesetObject creates a GitHub RepositoryRuleset object for enterprise-level rulesets
+func resourceGithubEnterpriseRulesetObject(d *schema.ResourceData) github.RepositoryRuleset {
```
stevehipwell: I'm not sure this logic is worth splitting out as it impacts readability. At a minimum please move it below the actual resource logic.

github/resource_github_enterprise_ruleset.go
```diff
@@ -0,0 +1,1072 @@
...
+func resourceGithubEnterpriseRulesetRead(ctx context.Context, d *schema.ResourceData, meta any) diag.Diagnostics {
+	client := meta.(*Owner).v3client
+	enterpriseSlug := d.Get("enterprise_slug").(string)
+
+	tflog.Trace(ctx, "Reading enterprise ruleset", map[string]any{
+		"enterprise_slug": enterpriseSlug,
+		"ruleset_id":      d.Id(),
+	})
+
+	rulesetID, err := strconv.ParseInt(d.Id(), 10, 64)
```
stevehipwell: Please use `ruleset_id`.

github/resource_github_enterprise_ruleset.go
```diff
@@ -0,0 +1,1072 @@
...
+
+	return nil
+}
+
+func resourceGithubEnterpriseRulesetUpdate(ctx context.Context, d *schema.ResourceData, meta any) diag.Diagnostics {
+	client := meta.(*Owner).v3client
+	enterpriseSlug := d.Get("enterprise_slug").(string)
+	name := d.Get("name").(string)
+
+	rulesetID, err := strconv.ParseInt(d.Id(), 10, 64)
```
stevehipwell: Same as above.

github/resource_github_enterprise_ruleset.go
```diff
@@ -0,0 +1,1072 @@
...
+			"ruleset_id":      rulesetID,
+			"error":           err.Error(),
+		})
+		return diag.FromErr(err)
+	}
+
+	d.SetId(strconv.FormatInt(*ruleset.ID, 10))
+	if err := d.Set("ruleset_id", ruleset.ID); err != nil {
+		return diag.FromErr(err)
+	}
```
stevehipwell:
```suggestion
```
Neither of these should be able to change.

github/resource_github_enterprise_ruleset.go
```diff
@@ -0,0 +1,1072 @@
...
+			"enterprise_slug": enterpriseSlug,
+			"ruleset_id":      rulesetID,
+			"error":           err.Error(),
+		})
+		return diag.FromErr(err)
+	}
+
+	if err := d.Set("ruleset_id", ruleset.ID); err != nil {
+		return diag.FromErr(err)
+	}
```
stevehipwell:
```suggestion
```
This shouldn't be able to change.

github/resource_github_enterprise_ruleset.go
```diff
@@ -0,0 +1,1072 @@
...
+	})
+
+	return nil
+}
+
+func resourceGithubEnterpriseRulesetDelete(ctx context.Context, d *schema.ResourceData, meta any) diag.Diagnostics {
+	client := meta.(*Owner).v3client
+	enterpriseSlug := d.Get("enterprise_slug").(string)
+
+	rulesetID, err := strconv.ParseInt(d.Id(), 10, 64)
```
stevehipwell: Same as above.

github/resource_github_enterprise_ruleset.go
```diff
@@ -0,0 +1,1072 @@
...
+		return nil, fmt.Errorf("error importing enterprise ruleset (expected format: <enterprise_slug>:<ruleset_id>): ruleset_id must be present")
+	}
+
+	tflog.Debug(ctx, "Importing enterprise ruleset", map[string]any{
+		"enterprise_slug": enterpriseSlug,
+		"ruleset_id":      rulesetID,
+	})
+
+	d.SetId(rulesetIDStr)
+	_ = d.Set("enterprise_slug", enterpriseSlug)
```
stevehipwell: Please don't ignore errors here and you need to set `ruleset_id`.

github/util_rules.go
stevehipwell: Please change the signature of this as the type isn't a bool.

stevehipwell (changes requested):
Thanks for the PR @Ravio1i.

github/resource_github_enterprise_ruleset.go
deiga: @stevehipwell Yeah, I think you're right with that

github/resource_github_enterprise_ruleset.go
stevehipwell: https://docs.github.com/en/enterprise-cloud@latest/rest/enterprise-admin/rules?apiVersion=2022-11-28#create-an-enterprise-repository-ruleset

github/resource_github_enterprise_ruleset.go
Ravio1i: I put this here, because the same way it is done in https://github.com/integrations/terraform-provider-github/blob/main/github/resource_github_organization_ruleset.go#L19

moved it below

github/data_source_github_enterprise_ruleset_test.go
Ravio1i: Merged the two test steps into a single step. The data source config now references the resource directly and includes `depends_on = [github_enterprise_ruleset.test]` to ensure ordering.

github/resource_github_enterprise_ruleset.go
Ravio1i: You're right, `repository_id` is NOT in the official GitHub Enterprise Ruleset API specification. The API docs only document these valid condition combinations for enterprise rulesets.

The go-github library includes a `RepositoryID` field in `RepositoryRulesetConditions`, but it's a shared struct across org and enterprise rulesets. For org-level rulesets `repository_id` IS supported, but for enterprise-level rulesets it's not documented 🔥

github/resource_github_enterprise_ruleset.go
Ravio1i: Moved `resourceGithubEnterpriseRulesetObject` below all CRUD functions

github/resource_github_enterprise_ruleset.go
Ravio1i: That is true, btw https://github.com/integrations/terraform-provider-github/blob/main/github/resource_github_organization_ruleset.go#L781-L783 org also does this

github/resource_github_enterprise_ruleset.go
Ravio1i: But isnt camelCase used everyhwere? https://github.com/integrations/terraform-provider-github/blob/main/github/resource_github_organization_ruleset.go#L739 I also just took reference for organization

github/util_rules.go
Ravio1i: The `org bool` parameter on `expandConditions`/`flattenConditions`/`expandRules`/`flattenRules` is a shared utility across all three resource types (repo, org, enterprise). It controls whether org-level fields (like `organization_name`, `organization_id`, `repository_property`) are included.

I agree, changing this to a more descriptive type (e.g. an enum like `RulesetLevel`) would be a good improvement but affects all three resource files and their tests. Should this be done as a separate refactoring PR to keep the scope of this PR focused?

github/data_source_github_enterprise_ruleset_test.go
```diff
@@ -83,11 +59,29 @@ func TestAccGithubEnterpriseRulesetDataSource(t *testing.T) {
 			Steps: []resource.TestStep{
 				{
 					Config: config,
-					Check:  resource.ComposeTestCheckFunc(),
-				},
-				{
-					Config: config2,
-					Check:  check,
+					Check: resource.ComposeTestCheckFunc(
```
deiga: We've just upgraded the testing library and have now better tools available.
Could you please refactor all the `Check:` parts in your PR to `ConfigStateChecks` or similar applicable? https://developer.hashicorp.com/terraform/plugin/testing/acceptance-tests/state-checks/resource

github/data_source_github_enterprise_ruleset.go
```diff
@@ -0,0 +1,420 @@
...
+								Schema: map[string]*schema.Schema{
+									"do_not_enforce_on_create": {
+										Type:        schema.TypeBool,
+										Computed:    true,
+										Description: "Allow repositories and branches to be created if a check would otherwise prohibit it.",
+									},
+								},
+							},
+						},
+						// Repository target rules (only valid when target = "repository")
```
deiga: issue: This comment seems misplaced. As this is a Data Source these fields are not `valid` or `invalid`. Maybe the wording could be changed to say "only populated when `target = "repository"`
But I don't know if even that would be useful here

github/util_ruleset_validation.go
Ravio1i: you are absolutely right, we can get rid of it!

Ravio1i: see https://github.com/integrations/terraform-provider-github/pull/3110/changes/7fb3bfe98d078cb18db8cd67887bfbf8cf8893a4

github/resource_github_enterprise_ruleset.go
```diff
@@ -0,0 +1,1050 @@
...
+			"enterprise_slug": {
+				Type:        schema.TypeString,
+				Required:    true,
+				ForceNew:    true,
+				Description: "The slug of the enterprise.",
+			},
+			"name": {
+				Type:         schema.TypeString,
+				Required:     true,
+				ValidateFunc: validation.StringLenBetween(1, 100),
```
deiga: Make sure that you replace all usage of `ValidateFunc`
```suggestion
				ValidateDiagFunc: validation.StringLenBetween(1, 100),
```

github/resource_github_enterprise_ruleset.go
```diff
@@ -0,0 +1,1050 @@
...
+				Type:        schema.TypeString,
+				Computed:    true,
+				Description: "GraphQL global node id for use with v4 API.",
+			},
+			"ruleset_id": {
+				Type:        schema.TypeInt,
+				Computed:    true,
+				Description: "GitHub ID for the ruleset.",
+			},
+			"conditions": {
```
deiga, line 119: Discussion: What if we split `conditions` into `target_organizations` and `target_repositories` or some similar naming? Similar to what it is like in the UI.

This could make the code for handling and validating these simpler and it could make the config UX clearer too.

@stevehipwell @nickfloyd thoughts?

Image

github/resource_github_enterprise_ruleset.go
```diff
@@ -0,0 +1,1050 @@
...
+											},
+										},
+									},
+								},
+							},
+						},
+					},
+				},
+			},
+			"rules": {
```
deiga, line 320: Discussion:
On the shoulders of https://github.com/integrations/terraform-provider-github/pull/3110/changes#r2827158611: What if we split the rules into separate lists based on which target they work with?

That would enable us to remove the rules validation for "is this target allowed to set this rule" and replace it with simpler built-in validators.
It would make the UX for our users nicer since they don't have to trial-and-error the rules.

@stevehipwell @nickfloyd thoughts?

github/resource_github_enterprise_ruleset.go
```diff
@@ -0,0 +1,1050 @@
...
+// resourceGithubEnterpriseRulesetObject creates a GitHub RepositoryRuleset object for enterprise-level rulesets
+func resourceGithubEnterpriseRulesetObject(d *schema.ResourceData) github.RepositoryRuleset {
+	enterpriseSlug := d.Get("enterprise_slug").(string)
+	target := github.RulesetTarget(d.Get("target").(string))
+	enforcement := github.RulesetEnforcement(d.Get("enforcement").(string))
+	sourceTypeEnum := github.RulesetSourceType("Enterprise")
+
+	return github.RepositoryRuleset{
+		Name:         d.Get("name").(string),
+		Target:       &target,
```
deiga, line 1053:
```suggestion
		Target:       github.Ptr(target),
```

github/resource_github_enterprise_ruleset.go
```diff
@@ -0,0 +1,1050 @@
...
+		Source:       enterpriseSlug,
+		SourceType:   &sourceTypeEnum,
```
deiga:
```suggestion
		SourceType:   github.Ptr(sourceTypeEnum),
```

deiga (changes requested):

github/resource_github_enterprise_ruleset.go
Ravio1i: https://github.com/integrations/terraform-provider-github/pull/3110/changes/21159433355b0949d6a40ec1c0b2b206a905109a did it here, but needed to do its like `ValidateDiagFunc: validation.ToDiagFunc(validation.StringLenBetween(1, 100)),`which feels a bit off to make it compile. Is this right?

github/resource_github_enterprise_ruleset.go
Ravio1i: https://github.com/integrations/terraform-provider-github/pull/3110/changes/b3dfcdda298c0b734d85d03a90887f4a64aa4b64

github/resource_github_enterprise_ruleset.go
Ravio1i: https://github.com/integrations/terraform-provider-github/pull/3110/changes/b3dfcdda298c0b734d85d03a90887f4a64aa4b64

github/resource_github_enterprise_ruleset.go
deiga: I think they meant that you don't need to parse `d.Id()` and instead use `d.Get("ruleset_id")`

github/data_source_github_enterprise_ruleset.go
```diff
@@ -0,0 +1,420 @@
...
+		return diag.FromErr(err)
+	}
+
+	// Set the ID to the ruleset ID
+	d.SetId(strconv.FormatInt(ruleset.GetID(), 10))
+
+	// Set all computed attributes
+	if err := d.Set("ruleset_id", ruleset.ID); err != nil {
+		return diag.FromErr(err)
+	}
```
deiga: This isn't a computed value and shouldn't change

github/resource_github_enterprise_ruleset.go
Ravio1i: Ah got it! But still isnt that breaking the consistency? In both `resource_github_organization_ruleset.go` and `resource_github_repository_ruleset.go`, all use `d.Id()`. 

I consistently changed it now to `d.Get("ruleset_id")`

github/data_source_github_enterprise_ruleset_test.go
Ravio1i: I refactored the tests here https://github.com/integrations/terraform-provider-github/pull/3110/changes/ac5cbe365b88b9b880a369a23bf42439a2e6873c

github/resource_github_enterprise_ruleset.go
stevehipwell: I'm happy for you to fix it in both places or open an issue to tack the differences.

github/util_rules.go
stevehipwell: It needs fixing in this PR as before this PR a bool can represent the supported states.

github/resource_github_enterprise_ruleset.go
stevehipwell: I'd hold off on this for now.

github/resource_github_enterprise_ruleset.go
stevehipwell: I'd suggest opening an issue to discuss or creating a PR to demonstrate a POC.

github/resource_github_enterprise_ruleset.go
```diff
@@ -0,0 +1,1034 @@
...
+	if err != nil {
+		tflog.Error(ctx, "Failed to create enterprise ruleset", map[string]any{
+			"enterprise_slug": enterpriseSlug,
+			"name":            name,
+			"error":           err.Error(),
+		})
+		return diag.FromErr(err)
+	}
+
+	d.SetId(strconv.FormatInt(*ruleset.ID, 10))
```
deiga:
```suggestion
	d.SetId(strconv.FormatInt(ruleset.GetID(), 10))
```

github/resource_github_enterprise_ruleset.go
```diff
@@ -0,0 +1,1034 @@
...
+		return diag.FromErr(err)
+	}
+	if err := d.Set("etag", resp.Header.Get("ETag")); err != nil {
+		return diag.FromErr(err)
+	}
+
+	tflog.Info(ctx, "Created enterprise ruleset", map[string]any{
+		"enterprise_slug": enterpriseSlug,
+		"name":            name,
+		"ruleset_id":      *ruleset.ID,
```
deiga:
```suggestion
		"ruleset_id":      ruleset.GetID(),
```

github/resource_github_enterprise_ruleset_test.go
```diff
@@ -0,0 +1,897 @@
...
+
+	"github.com/hashicorp/terraform-plugin-testing/helper/acctest"
+	"github.com/hashicorp/terraform-plugin-testing/helper/resource"
+	"github.com/hashicorp/terraform-plugin-testing/knownvalue"
+	"github.com/hashicorp/terraform-plugin-testing/statecheck"
+	"github.com/hashicorp/terraform-plugin-testing/terraform"
+	"github.com/hashicorp/terraform-plugin-testing/tfjsonpath"
+)
+
+func TestAccGithubEnterpriseRuleset_basic(t *testing.T) {
```
deiga:
```suggestion
func TestAccGithubEnterpriseRuleset(t *testing.T) {
```

github/resource_github_enterprise_ruleset_test.go
```diff
@@ -0,0 +1,897 @@
...
+					statecheck.ExpectKnownValue("github_enterprise_ruleset.test", tfjsonpath.New("name"), knownvalue.StringExact(rulesetName)),
+					statecheck.ExpectKnownValue("github_enterprise_ruleset.test", tfjsonpath.New("target"), knownvalue.StringExact("branch")),
+					statecheck.ExpectKnownValue("github_enterprise_ruleset.test", tfjsonpath.New("enforcement"), knownvalue.StringExact("active")),
+				},
+			},
+		},
+	})
+}
+
+func TestAccGithubEnterpriseRuleset_branch_rules(t *testing.T) {
```
deiga: "inline" these tests as subtests with `t.Run` into `TestAccGithubEnterpriseRuleset`

github/resource_github_enterprise_ruleset_test.go
```diff
@@ -0,0 +1,897 @@
...
+		}
+		`
+	config := fmt.Sprintf(rulesetHCL, testAccConf.enterpriseSlug, rulesetName)
+
+	resource.Test(t, resource.TestCase{
+		PreCheck:  func() { skipUnlessEnterprise(t) },
+		ProviderFactories: providerFactories,
+		Steps: []resource.TestStep{
+			{
+				Config: config,
```
deiga:
```suggestion
				Config: fmt.Sprintf(rulesetHCL, testAccConf.enterpriseSlug, rulesetName),
```

github/resource_github_enterprise_ruleset_test.go
```diff
@@ -0,0 +1,897 @@
...
+				ConfigStateChecks: []statecheck.StateCheck{
+					statecheck.ExpectKnownValue("github_enterprise_ruleset.test", tfjsonpath.New("enterprise_slug"), knownvalue.StringExact(testAccConf.enterpriseSlug)),
+					statecheck.ExpectKnownValue("github_enterprise_ruleset.test", tfjsonpath.New("name"), knownvalue.StringExact(rulesetName)),
+					statecheck.ExpectKnownValue("github_enterprise_ruleset.test", tfjsonpath.New("target"), knownvalue.StringExact("branch")),
+					statecheck.ExpectKnownValue("github_enterprise_ruleset.test", tfjsonpath.New("enforcement"), knownvalue.StringExact("active")),
```
deiga: There isn't that much value in checking if the same values we passed in come out. Add checks for the computed values and anything where values go through some mangling before ending in state

github/resource_github_enterprise_ruleset_test.go
```diff
@@ -0,0 +1,897 @@
...
+	target          = "branch"
+	enforcement     = "active"
+
+	bypass_actors {
+		actor_type  = "DeployKey"
+		bypass_mode = "always"
+	}
+
+	bypass_actors {
+		actor_id    = 1
```
deiga: OrganizationAdmin doesn't have an ID anymore
```suggestion
```

github/resource_github_enterprise_ruleset_test.go
```diff
@@ -0,0 +1,897 @@
...
+		},
+	})
+}
+
+func TestAccGithubEnterpriseRuleset_update_name(t *testing.T) {
+	randomID := acctest.RandStringFromCharSet(5, acctest.CharSetAlphaNum)
+	name := fmt.Sprintf("test-enterprise-ruleset-%s", randomID)
+	nameUpdated := fmt.Sprintf("test-enterprise-ruleset-updated-%s", randomID)
+
+	configs := map[string]string{
```
deiga: issue: Use only one template string and pass if different values instead of duplicating

github/resource_github_enterprise_ruleset_test.go
```diff
@@ -0,0 +1,897 @@
...
+	}
+
+	checks := map[string][]statecheck.StateCheck{
+		"before": {
+			statecheck.ExpectKnownValue("github_enterprise_ruleset.test", tfjsonpath.New("name"), knownvalue.StringExact(name)),
+		},
+		"after": {
+			statecheck.ExpectKnownValue("github_enterprise_ruleset.test", tfjsonpath.New("name"), knownvalue.StringExact(nameUpdated)),
+		},
+	}
```
deiga: Don't use his structure, inline these

deiga (changes requested):

github/util_rules.go
Ravio1i: I addded RulesetLevel instad of the bool now

https://github.com/integrations/terraform-provider-github/pull/3110/changes/2a59f50c955ad62be8de5fe1522493118f27e448

Please check if this is how you would do it or you want to handle it differently :)

github/data_source_github_enterprise_ruleset.go
```diff
@@ -0,0 +1,417 @@
...
+		var ghErr *github.ErrorResponse
+		if errors.As(err, &ghErr) {
+			if ghErr.Response.StatusCode == http.StatusNotFound {
+				tflog.Error(ctx, "Enterprise ruleset not found", map[string]any{
+					"enterprise_slug": enterpriseSlug,
+					"ruleset_id":      rulesetID,
+				})
+				return diag.Errorf("enterprise ruleset %d not found in enterprise %s", rulesetID, enterpriseSlug)
+			}
+		}
```
deiga: I don't think this adds anything useful to the data source
```suggestion
```

github/resource_github_enterprise_ruleset.go
```diff
@@ -0,0 +1,1035 @@
...
+		CreateContext: resourceGithubEnterpriseRulesetCreate,
+		ReadContext:   resourceGithubEnterpriseRulesetRead,
+		UpdateContext: resourceGithubEnterpriseRulesetUpdate,
+		DeleteContext: resourceGithubEnterpriseRulesetDelete,
+
+		Importer: &schema.ResourceImporter{
+			StateContext: resourceGithubEnterpriseRulesetImport,
+		},
+
+		CustomizeDiff: resourceGithubEnterpriseRulesetCustomizeDiff,
```
deiga: nit: rename `Diff` function
```suggestion
		CustomizeDiff: resourceGithubEnterpriseRulesetDiff,
```

github/resource_github_enterprise_ruleset.go
```diff
@@ -0,0 +1,1035 @@
...
+			"enterprise_slug": {
+				Type:        schema.TypeString,
+				Required:    true,
+				ForceNew:    true,
+				Description: "The slug of the enterprise.",
+			},
+			"name": {
+				Type:         schema.TypeString,
+				Required:     true,
+				ValidateDiagFunc: validation.ToDiagFunc(validation.StringLenBetween(1, 100)),
```
deiga, line 62: question: Does this have an actual max length of 100? Or should we use `validation.StringIsNotEmpty`?

github/resource_github_enterprise_ruleset.go
```diff
@@ -0,0 +1,1035 @@
...
+				Optional:         true,
+				DiffSuppressFunc: bypassActorsDiffSuppressFunc,
+				Description:      "The actors that can bypass the rules in this ruleset.",
+				Elem: &schema.Resource{
+					Schema: map[string]*schema.Schema{
+						"actor_id": {
+							Type:        schema.TypeInt,
+							Optional:    true,
+							Default:     nil,
+							Description: "The ID of the actor that can bypass a ruleset. When `actor_type` is `OrganizationAdmin`, this should be set to `1`. Some resources such as DeployKey do not have an ID and this should be omitted.",
```
deiga:
```suggestion
							Description: "The ID of the actor that can bypass a ruleset. Required for `Integration`, `RepositoryRole`, and `Team` actor types. If `actor_type` is OrganizationAdmin` or `EnterpriseOwner`, `actor_id` is ignored. If `actor_type` is `DeployKey`, this should be null.",
```

github/resource_github_enterprise_ruleset.go
```diff
@@ -0,0 +1,1035 @@
...
+							Type:         schema.TypeString,
+							Required:     true,
+							ValidateDiagFunc: validation.ToDiagFunc(validation.StringInSlice([]string{"Integration", "OrganizationAdmin", "RepositoryRole", "Team", "DeployKey", "EnterpriseOwner"}, false)),
+							Description:  "The type of actor that can bypass a ruleset. See https://docs.github.com/en/rest/enterprise-admin/rules for more information",
+						},
+						"bypass_mode": {
+							Type:         schema.TypeString,
+							Required:     true,
+							ValidateDiagFunc: validation.ToDiagFunc(validation.StringInSlice([]string{"always", "pull_request", "exempt"}, false)),
+							Description:  "When the specified actor can bypass the ruleset. pull_request means that an actor can only bypass rules on pull requests. Can be one of: `always`, `pull_request`, `exempt`.",
```
deiga:
```suggestion
							Description:  "When the specified actor can bypass the ruleset. `pull_request` means that an actor can only bypass rules on pull requests. `pull_request` is not applicable for the `DeployKey` actor type. Also, `pull_request` is only applicable to branch rulesets. When `bypass_mode` is `exempt`, rules will not be run for that actor and a bypass audit entry will not be created.",
```

website/docs/r/enterprise_ruleset.html.markdown
```diff
@@ -0,0 +1,606 @@
...
+
+- `private` - (Optional) (Boolean) Allow private visibility for repositories. Defaults to `false`.
+
+### bypass_actors
+
+- `actor_id` - (Optional) (Number) The ID of the actor that can bypass a ruleset. When `actor_type` is `OrganizationAdmin`, this should be set to `1`. Some resources such as DeployKey do not have an ID and this should be omitted.
+
+- `actor_type` - (Required) (String) The type of actor that can bypass a ruleset. Can be one of: `Integration`, `OrganizationAdmin`, `RepositoryRole`, `Team`, `DeployKey`.
+
+- `bypass_mode` - (Required) (String) When the specified actor can bypass the ruleset. pull_request means that an actor can only bypass rules on pull requests. Can be one of: `always`, `pull_request`, `exempt`.
```
deiga: Update these to correspond to the new suggestions in the schema

website/docs/r/enterprise_ruleset.html.markdown
```diff
@@ -0,0 +1,606 @@
...
+
+~>Note: at the time of writing this, the following actor types correspond to the following actor IDs:
+
+- `OrganizationAdmin` -> `1`
```
deiga:
```suggestion
```

deiga (changes requested):

github/resource_github_enterprise_ruleset.go
Ravio1i: There is actually no limit set in api, but I tested with UI and it seems to be >256, which then lets the api return a 500

Ravio1i:
So bit of a summary:

- [ ] Is the way https://github.com/integrations/terraform-provider-github/pull/3110#discussion_r2845480876 is implemented okay with level?
- [ ] Can this https://github.com/integrations/terraform-provider-github/pull/3110#discussion_r2827158611 be closed ?
- [ ] https://github.com/integrations/terraform-provider-github/pull/3110#discussion_r2813403904 shall we remove repository_id although go lib has it?
- [ ] https://github.com/integrations/terraform-provider-github/pull/3110#discussion_r2827187209 is there already issue and can the thread be closed?

arealfracker:
Would this PR be able to get some attention? I've been watching this since January so I can escape some click-ops in a GHES instance I manage, but [the conversation stalled out back in March with some unanswered questions](https://github.com/integrations/terraform-provider-github/pull/3110#issuecomment-3991557602). Thanks!

deiga:
@arealfracker that would require the author to rebase and refactor to the current standards first

Ravio1i:
Rebased onto current `main` and squashed down to 5 commits, so the conflicts are gone. Downside is that every inline comment got marked outdated in the process. Sorry about the noise. Here's what happened to each thread so nothing gets lost.

**Two schema changes I made without hearing back, and why:**

I removed `repository_id` from the enterprise conditions. It isn't in the [enterprise ruleset API docs](https://docs.github.com/en/rest/enterprise-admin/rules?apiVersion=2022-11-28), which only list `repository_name` and `repository_property`. It was only there because go-github's `RepositoryRulesetConditions` is shared with org rulesets. @stevehipwell caught this a while back. I'd asked whether to remove it or deprecate it, and since the resource has never shipped I just removed it. Organization rulesets keep theirs, untouched.

I also added `organization_property` conditions. They're in the spec (six legal org x repo combinations) and were just missing. Both selector groups are now `ExactlyOneOf` sets, which is @deiga's suggestion applied literally, and it enforces those six combinations declaratively instead of by hand.

**On #3584:** I'm keeping the single API-shaped `rules` block for now. Splitting would be breaking for the two existing ruleset resources, and adding a third resource with a different shape would lock in the inconsistency whichever way #3584 goes. I did arrange the code so the split stays mechanical: which rules a target accepts lives in one map keyed by `github.RulesetTarget`, and expand/flatten work per rule rather than per target.

**Feedback that's applied:** structured `tflog` fields instead of `fmt.Sprintf`, `SchemaVersion` dropped since it's a new resource with no upgraders, `d.Get("ruleset_id")` throughout with the importer seeding both `enterprise_slug` and `ruleset_id`, `Update` no longer asymmetric with `Create`, `bypass_mode` caveats as worded, and `rulesetLevel` replacing the `org bool`.

**Feedback I deliberately skipped,** because after the rebase it would make things inconsistent rather than better. The sibling resources now do the opposite of what was asked in these three cases: `target` description as a literal, moving `supportedEnterpriseRulesetTargetTypes` below the logic, and dropping the `name` length cap. Happy to revisit if you'd rather change all three resources together.

One thing I kept as-is: the composite `<enterprise_slug>:<ruleset_id>` ID. An enterprise slug has no provider-level source, so import needs it in there.

`make lint`, `make generatedocs` and the unit tests are all clean, and each commit builds on its own. Acceptance tests are unrun since they need an enterprise plan and `admin:enterprise`.

github/data_source_github_enterprise_ruleset_test.go
```diff
@@ -0,0 +1,51 @@
...
+
+		name := fmt.Sprintf("%senterprise-ds-%s", testResourcePrefix, acctest.RandStringFromCharSet(5, acctest.CharSetAlphaNum))
+		config := fmt.Sprintf(`
+			%s
+
+			data "github_enterprise_ruleset" "test" {
+			  enterprise_slug = github_enterprise_ruleset.test.enterprise_slug
+			  ruleset_id      = github_enterprise_ruleset.test.ruleset_id
+			}
+		`, fmt.Sprintf(enterpriseRulesetBasicConfig, testAccConf.enterpriseSlug, name))
```
deiga: issue: please create a mustCreateTestEnterpriseRuleset instead here

deiga (changes requested):
Partial review

github/data_source_github_enterprise_ruleset_test.go
Ravio1i: added

deiga:
@Ravio1i please rebase, resolve conflicts and address review comments

github/util_ruleset_validation.go
```diff
@@ -66,27 +66,46 @@ var pushOnlyRules = []github.RepositoryRuleType{
...
+	github.RulesetRuleTypeRepositoryVisibility,
+}
+
+// rulesAllowedByTarget maps each ruleset target to the rules GitHub accepts for it.
+// Keeping the target/rule relationship in one declarative place means adding a target
+// (or splitting `rules` per target, see integrations/terraform-provider-github#3584)
+// does not require touching the validation logic itself.
+var rulesAllowedByTarget = map[github.RulesetTarget][]github.RepositoryRuleType{
+	github.RulesetTargetBranch:     branchTagOnlyRules,
+	github.RulesetTargetTag:        branchTagOnlyRules,
```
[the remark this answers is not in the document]
Ravio1i: Agreed that `branch` and `tag` don't have identical rule sets, but this predates the PR. Upstream's `validateRulesForBranchTagTarget` already passed `branchTagOnlyRules` for both targets; all this PR did was move that switch into the declarative `rulesAllowedByTarget` map so the new `repository` target could be added without touching the validation logic. The branch/tag list itself is unchanged.

Splitting it also isn't enterprise-scoped: `github_organization_ruleset` and `github_repository_ruleset` both support the `tag` target and route through the same `validateRulesetRules` path, so any split changes their plan-time behaviour too.

As the comments on these lists note, the REST docs don't cleanly separate rules per target, so the lists were derived by trial against the API. A hand-written tag list that is too strict would reject configurations that apply successfully today, which makes this worth verifying empirically rather than inferring from the docs.

There's already an open discussion for exactly this restructuring in #3584, so I'd rather land the split there with proper per-target verification than widen this PR further.

github/util_rules.go
```diff
@@ -882,6 +1026,27 @@ func flattenRules(ctx context.Context, rules *github.RepositoryRulesetRules, org
...
+				"negate":  rules.RepositoryName.Negate,
+				"pattern": rules.RepositoryName.Pattern,
+			}}
+		}
+
+		if rules.RepositoryVisibility != nil {
+			rulesMap["repository_visibility"] = []map[string]any{{
+				"internal": rules.RepositoryVisibility.Internal,
+				"private":  rules.RepositoryVisibility.Private,
+			}}
```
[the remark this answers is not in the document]
Ravio1i: Good catch, fixed in dfcf3738.

`Public` is now set in `expandRules` and read back in `flattenRules`, and a `public` attribute was added to both the resource and the data-source `repository_visibility` schemas. `TestExpandRulesRepositoryTargetRules` now round-trips `public: true` through expand and flatten to cover the state-loss case, the `repository` target acceptance test asserts it, and the generated docs were regenerated.

github/data_source_github_enterprise_ruleset.go
```diff
@@ -0,0 +1,720 @@
...
+			"enterprise_slug": {
+				Type:             schema.TypeString,
+				Required:         true,
+				ValidateDiagFunc: validation.ToDiagFunc(validation.StringIsNotEmpty),
+				Description:      "The slug of the enterprise the ruleset belongs to.",
+			},
+			"ruleset_id": {
+				Type:        schema.TypeInt,
+				Required:    true,
+				Description: "GitHub ID for the ruleset.",
```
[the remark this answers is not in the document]
Ravio1i: Fixed in dfcf3738. `ruleset_id` now uses `validation.ToDiagFunc(validation.IntAtLeast(1))`, matching the existing required ID inputs in this provider such as `github_organization_repository_role` and `github_team_repositories`.

github/resource_github_enterprise_ruleset.go
```diff
@@ -0,0 +1,1103 @@
...
+							Type:             schema.TypeString,
+							Required:         true,
+							ValidateDiagFunc: validation.ToDiagFunc(validation.StringInSlice([]string{"Integration", "OrganizationAdmin", "RepositoryRole", "Team", "DeployKey", "EnterpriseOwner"}, false)),
+							Description:      "The type of actor that can bypass a ruleset. Can be one of: `Integration`, `OrganizationAdmin`, `RepositoryRole`, `Team`, `DeployKey`, or `EnterpriseOwner`.",
+						},
+						"bypass_mode": {
+							Type:             schema.TypeString,
+							Required:         true,
+							ValidateDiagFunc: validation.ToDiagFunc(validation.StringInSlice([]string{"always", "pull_request", "exempt"}, false)),
+							Description:      "When the specified actor can bypass the ruleset. `pull_request` means that an actor can only bypass rules on pull requests. `pull_request` is not applicable for the `DeployKey` actor type. Also, `pull_request` is only applicable to branch rulesets. When `bypass_mode` is `exempt`, rules will not be run for that actor and a bypass audit entry will not be created. Can be one of: `always`, `pull_request`, `exempt`.",
```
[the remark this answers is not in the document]
Ravio1i: Fixed in 37c9620c. `resourceGithubEnterpriseRulesetDiff` now calls `validateRulesetBypassActors`, which rejects `pull_request` on any target other than `branch` and for the `DeployKey` actor type, reporting the offending `bypass_actors.<n>` index. `Test_validateBypassActorsForTarget` covers the valid cases plus each invalid combination.

Scoped to the enterprise resource deliberately: `github_organization_ruleset` and `github_repository_ruleset` have the same enum-only `bypass_mode` today, so validating them here too would change plan-time behaviour for existing configurations. Happy to extend it to all three in a follow-up if you'd prefer that.

Ravio1i:
@deiga done, thanks for the nudge.

### Rebase

Rebased onto `deb27940` (current `main` tip, including the go-github v92 bump). The branch is now zero commits behind and the PR shows as mergeable again.

Two conflicts, both in the shared ruleset helpers:

- `resource_github_organization_ruleset.go` / `resource_github_repository_ruleset.go` — `main` added `diffETag` to the `CustomizeDiff` functions and an `isOrg bool` parameter to `validateRulesetConditions`, while this branch replaced that bool with the `rulesetLevel` type. Resolved by keeping both: `rulesetLevel` plus the new etag suppression. The `isOrg` parameter was only being logged, never branched on, so dropping it is behaviour-preserving.
- The new enterprise files still imported go-github v89, bumped to v92 in f370be24.

### Review comments

| Comment | Resolution |
| --- | --- |
| [`repository_visibility` missing the public flag](https://github.com/integrations/terraform-provider-github/pull/3110#discussion_r3886491220) | Fixed in dfcf3738 — added to expand/flatten, both schemas, tests and generated docs ([reply](https://github.com/integrations/terraform-provider-github/pull/3110#discussion_r4062438823)) |
| [`ruleset_id` accepts zero and negative values](https://github.com/integrations/terraform-provider-github/pull/3110#discussion_r3886491224) | Fixed in dfcf3738 — `validation.IntAtLeast(1)` ([reply](https://github.com/integrations/terraform-provider-github/pull/3110#discussion_r4062440324)) |
| [`pull_request` bypass mode not cross-validated](https://github.com/integrations/terraform-provider-github/pull/3110#discussion_r3886491229) | Fixed in 37c9620c — `CustomizeDiff` now rejects it for non-branch targets and for `DeployKey`, with unit tests ([reply](https://github.com/integrations/terraform-provider-github/pull/3110#discussion_r4062440495)) |
| [Split the branch/tag rule allow-lists](https://github.com/integrations/terraform-provider-github/pull/3110#discussion_r3886491211) | Deferred to #3584 — see [reply](https://github.com/integrations/terraform-provider-github/pull/3110#discussion_r4062422234) for the reasoning |

The two older threads should also be covered now:

- [`repository_id` isn't in the enterprise REST API](https://github.com/integrations/terraform-provider-github/pull/3110#discussion_r2813403904) — the condition-level `repository_id` selector is gated behind `rulesetLevel.targetsRepositoriesByID()`, which is organization-only, with `TestExpandConditionsRepositoryIDIsOrganizationOnly` and `TestFlattenConditionsRepositoryIDIsOrganizationOnly` covering it. The `repository_id` that remains in the enterprise schema is the `required_workflow` source repository, which is a different field and is supported at enterprise level.
- [The `org bool` parameter isn't really a bool](https://github.com/integrations/terraform-provider-github/pull/3110#discussion_r2813432909) — replaced with the `rulesetLevel` type (`rulesetLevelRepository` / `rulesetLevelOrganization` / `rulesetLevelEnterprise`) and the intent-revealing predicates `targetsRepositories()`, `targetsOrganizations()` and `targetsRepositoriesByID()`.

Happy to squash the three follow-up commits into the feature commits if you'd rather review a cleaner history.

PR #3113, opened 2026-01-22
[MAINT] enable nilnesserr linter

deiga:
----

### Before the change?

- Code could be checking for `if err != nil` but return a different error

### After the change?

- Linter will mark that every checked error is being returned correctly

### Pull request checklist

- [ ] Yes
- [x] No

----

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -19,6 +19,7 @@ linters:
     - misspell
     - modernize
     - nilerr
+    - nilnesserr
     - predeclared
     - staticcheck
     - unconvert
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -1,4 +1,4 @@
-version: "2"
+version: '2'
 run:
   modules-download-mode: vendor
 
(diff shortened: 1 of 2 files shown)


.golangci.yml
```diff
@@ -1,4 +1,4 @@
-version: "2"
+version: '2'
```
stevehipwell: Please don't modify the YAML formatting.

.golangci.yml
```diff
@@ -44,7 +45,7 @@ linters:
         text: always receives
       - linters:
           - staticcheck
-        text: "SA1019.*ByID.*deprecated"
+        text: 'SA1019.*ByID.*deprecated'
```
stevehipwell: Same as above.

stevehipwell (changes requested):
I like this idea, minor nits to fix.

FYI we still have a core linter disabled which need the code updating to re-enable it.

deiga:
> I like this idea, minor nits to fix.
> 
> FYI we still have a core linter disabled which need the code updating to re-enable it.

Yeah, I tried enabling it and was overwhelmed 😂

stevehipwell (dismissed):
LGTM

stevehipwell (approved):
LGTM

PR #3118, opened 2026-01-25
[BUG] Ensure `github_emu_group_mapping` behaves correctly if mapping changes upstream

deiga:
----

### Before the change?

- `Read` would not check if the Team and External Group we're still  connected to each other
- Changing `group_id` would not force a new resource, potentially leaving dangling connections
- Resource ID format did not follow current conventions
- `Create` and `Update` shared same functionality, leading to potentially wasted API calls

### After the change?

- Update ID format to `<team_id>:<team_slug>:<group_id>`
- Adds Schema migration to ensure IDs are migrated to new format
- `Read` properly checks that Team ID of `team_slug` and any Team mapped to `group_id` match
- Changing `group_id` forces new resource creation
- Split `Create` and `Update` to allow fine-grained control in `Update`
- Adds check on diff for `team_slug` change to vlaidate if it's a team rename or switch

### Pull request checklist

- [ ] Yes
- [x] No

----


github/resource_github_emu_group_mapping.go
```diff
@@ -51,75 +75,85 @@ func resourceGithubEMUGroupMappingRead(ctx context.Context, d *schema.ResourceDa
...
+	}
+	if err := d.Set("team_id", teamID); err != nil {
+		return diag.FromErr(err)
+	}
+
 	return nil
 }
 
-func resourceGithubEMUGroupMappingCreateOrUpdate(ctx context.Context, d *schema.ResourceData, meta any) diag.Diagnostics {
+func resourceGithubEMUGroupMappingCreate(ctx context.Context, d *schema.ResourceData, meta any) diag.Diagnostics {
```
stevehipwell: Could we please move the create function above read?

github/resource_github_emu_group_mapping.go
```diff
@@ -273,10 +379,75 @@ func resourceGithubEMUGroupMappingImport(ctx context.Context, d *schema.Resource
...
+
+	// Get old team_id from state
+	oldTeamID := toInt64(diff.Get("team_id"))
+	if oldTeamID == 0 {
+		return false
+	}
+
+	// Resolve new team_slug to team ID via API
+	oldTeamSlug, newTeamSlug := diff.GetChange("team_slug")
+	newTeamID, err := getTeamID(newTeamSlug.(string), meta)
```
stevehipwell: Please don't use `getTeamID()` when you know you have a slug as it's for the case where it could be slug or ID. My teams PR should have a helper function you could copy.

github/resource_github_emu_group_mapping.go
```diff
@@ -273,10 +379,75 @@ func resourceGithubEMUGroupMappingImport(ctx context.Context, d *schema.Resource
...
 	})
 	d.SetId(resourceID)
+
 	return []*schema.ResourceData{d}, nil
 }
+
+func matchTeamID(ctx context.Context, meta any, teamSlug string, groupTeams []*github.ExternalGroupTeam) (int64, error) {
+	client := meta.(*Owner).v3client
+	orgName := meta.(*Owner).name
+	configuredTeam, _, err := client.Teams.GetTeamBySlug(ctx, orgName, teamSlug)
```
stevehipwell: Why do we have this when we force a replacement if the team changes?

go.mod
```diff
@@ -1,14 +1,16 @@
 module github.com/integrations/terraform-provider-github/v6
 
-go 1.24.0
+go 1.24.4
```
stevehipwell, line 3:
```suggestion
go 1.24
```

stevehipwell (changes requested):
I've added some review comments, but please could you remove the new deps. I don't think we should be introducing new dependencies with functional changes and I'm not sold on the value of doing something like this especially where we get a whole old copy of go-github in the vendor dir.

github/resource_github_emu_group_mapping.go
deiga: We only force a replacement if teamSlug and teamID change. So AFAIK we need to do this at least in the schema migration and `Create`. 

You're right that we might not need this in `Read` and `Update` though... 🤔

deiga:
> I've added some review comments, but please could you remove the new deps. I don't think we should be introducing new dependencies with functional changes and I'm not sold on the value of doing something like this especially where we get a whole old copy of go-github in the vendor dir.

Regarding `migueleliasweb/go-github-mock`:
- I agree that having a second vendored `go-github` version isn't great. I'm hoping to resolve that as I opened a PR with the package to upgrade it.
- I do think it makes little sense for us to maintain mocking code for the github API if it's readily available as a package for us to use, with direct struct support.

Regarding `go-test/deep`: 
It was incredibly cumbersome to figure out why the migration tests we're failing because the structs in https://github.com/integrations/terraform-provider-github/pull/3118/changes#diff-6e6f4a2962b97dd1b6c8f7283eb71cb6cc2cae860c17d87041a32f0dfeb4cd0cR106 were visually the same, the integer types were just not matching. And I couldn't find a reasonable way to get that information out of reflect or fmt.Sprintf.
If you know of a better way, I'm all ears.

I would propose we keep both package additions as neither should be posing a risk for the functionality of the provider, while trying to make maintenance simpler

stevehipwell:
RE `migueleliasweb/go-github-mock` could you use [google/go-cmp](https://github.com/google/go-cmp) instead? This is the idiomatic library for Go test comparisons.

RE `migueleliasweb/go-github-mock` if you want to use this please open an issue and or seperate PR to discuss.

deiga:
> RE `migueleliasweb/go-github-mock` could you use [google/go-cmp](https://github.com/google/go-cmp?rgh-link-date=2026-01-28T16%3A25%3A09Z) instead? This is the idiomatic library for Go test comparisons.
> 
> RE `migueleliasweb/go-github-mock` if you want to use this please open an issue and or seperate PR to discuss.

@stevehipwell Thanks for the `go-cmp` pointer, implemented that now.

stevehipwell (changes requested):
Sorry, but I'm not following the point of `matchTeamID()`? Given that the team ID can't change without recreating the resource I'd only expect there to be a call to `client.Teams.GetTeamBySlug()` in create, to set it into the schema, and also in the diff if the slug changes. Other than that there shouldn't be any need to call that API during the general lifecycle.

FYI I like the addition of `team_id` to the schema, I'll add that into my team changes.

deiga:
> Sorry, but I'm not following the point of `matchTeamID()`? Given that the team ID can't change without recreating the resource I'd only expect there to be a call to `client.Teams.GetTeamBySlug()` in create, to set it into the schema, and also in the diff if the slug changes. Other than that there shouldn't be any need to call that API during the general lifecycle.
> 
> FYI I like the addition of `team_id` to the schema, I'll add that into my team changes.

@stevehipwell My idea with `matchTeamId` was to check if the Group and Team had been disconnected upstream.

But I just realize that in `Create` we just created the connection, so `matchTeamID` is silly.
And for `Update` we've just checked in `Read` that group IDs from state and upstream match.

You're right. it's unnecessary!

github/resource_github_emu_group_mapping.go
```diff
@@ -29,19 +39,34 @@ func resourceGithubEMUGroupMapping() *schema.Resource {
...
+			},
+		},
 	}
 }
 
-func resourceGithubEMUGroupMappingRead(ctx context.Context, d *schema.ResourceData, meta any) diag.Diagnostics {
-	tflog.Trace(ctx, "Reading EMU group mapping", map[string]any{
-		"resource_id": d.Id(),
+func resourceGithubEMUGroupMappingCreate(ctx context.Context, d *schema.ResourceData, meta any) diag.Diagnostics {
+	resourceID := d.Id()
```
stevehipwell: I'm not sure why this is here given that in most cases this should be empty?

github/resource_github_emu_group_mapping.go
```diff
@@ -50,77 +75,67 @@ func resourceGithubEMUGroupMappingRead(ctx context.Context, d *schema.ResourceDa
...
 	}
 
-	tflog.Debug(ctx, "Successfully retrieved external group from GitHub API", map[string]any{
-		"org_name":   orgName,
-		"group_id":   id64,
-		"team_count": len(group.Teams),
-	})
+	if err := d.Set("team_id", teamID); err != nil {
+		return diag.FromErr(err)
+	}
```
stevehipwell: Should this not be down with the other set calls?

github/resource_github_emu_group_mapping.go
```diff
@@ -130,60 +145,145 @@ func resourceGithubEMUGroupMappingCreateOrUpdate(ctx context.Context, d *schema.
...
+	teamID, err := lookupTeamID(ctx, meta.(*Owner), teamSlug)
 	if err != nil {
 		return diag.FromErr(err)
 	}
+	if teamID != toInt64(d.Get("team_id")) {
+		return diag.Errorf("configured Team does not match with any team mapped to external group: %d != %d", teamID, toInt64(d.Get("team_id")))
+	}
+	if err := d.Set("team_id", teamID); err != nil {
+		return diag.FromErr(err)
+	}
```
stevehipwell:
```suggestion
```
This isn't required in read.

github/resource_github_emu_group_mapping.go
```diff
@@ -130,60 +145,145 @@ func resourceGithubEMUGroupMappingCreateOrUpdate(ctx context.Context, d *schema.
...
+}
 
-	newResourceID := fmt.Sprintf("teams/%s/external-groups", teamSlugStr)
-	tflog.Trace(ctx, "Setting resource ID", map[string]any{
-		"resource_id": newResourceID,
+func resourceGithubEMUGroupMappingUpdate(ctx context.Context, d *schema.ResourceData, meta any) diag.Diagnostics {
+	resourceID := d.Id()
+	tflog.Trace(ctx, "Updating EMU group mapping", map[string]any{
+		"resource_id": resourceID,
 	})
```
stevehipwell:
```suggestion
	tflog.Trace(ctx, "Updating EMU group mapping", map[string]any{
		"resource_id": d.Id(),
})
```

github/resource_github_emu_group_mapping.go
```diff
@@ -224,33 +324,15 @@ func resourceGithubEMUGroupMappingDelete(ctx context.Context, d *schema.Resource
...
 func resourceGithubEMUGroupMappingImport(ctx context.Context, d *schema.ResourceData, meta any) ([]*schema.ResourceData, error) {
 	importID := d.Id()
 	tflog.Trace(ctx, "Importing EMU group mapping with two-part ID", map[string]any{
 		"import_id": importID,
 		"strategy":  "two_part_id",
 	})
 
-	groupIDString, teamSlug, err := parseTwoPartID(d.Id(), "group_id", "team_slug")
+	// <team-slug>:<group-id>
+	teamSlug, groupIDString, err := parseID2(d.Id())
```
stevehipwell: Why have we swapped the ID order?

stevehipwell (changes requested):
This is coming along well, I've added some questions/suggestions but the main one is to remove the unnecessary API call to get the team ID in read.

github/resource_github_emu_group_mapping.go
deiga: IIRC that was based on your suggestion. But I can revert that

stevehipwell: Not intentionally, so yes please to reverting this. My suggestion was meant to be `<group-id>:<team-id>:<team-slug>` before I adopted your pattern of adding the computed ID as an attribute.

github/resource_github_emu_group_mapping.go
deiga: Copy-pasta from the `CreateUpdate` split

github/resource_github_emu_group_mapping.go
```diff
@@ -2,25 +2,35 @@ package github
...
 		ReadContext:   resourceGithubEMUGroupMappingRead,
-		UpdateContext: resourceGithubEMUGroupMappingCreateOrUpdate,
+		UpdateContext: resourceGithubEMUGroupMappingUpdate,
 		DeleteContext: resourceGithubEMUGroupMappingDelete,
 		Importer: &schema.ResourceImporter{
 			StateContext: resourceGithubEMUGroupMappingImport,
 		},
+		CustomizeDiff: customdiff.All(
+			customdiff.ForceNewIf("team_slug", hasNewTeamID),
+		),
```
stevehipwell:
```suggestion
		CustomizeDiff: customdiff.ForceNewIf("team_slug", hasNewTeamID),
```

github/resource_github_emu_group_mapping.go
```diff
@@ -273,10 +349,57 @@ func resourceGithubEMUGroupMappingImport(ctx context.Context, d *schema.Resource
...
+
 	tflog.Trace(ctx, "Setting resource ID", map[string]any{
 		"resource_id": resourceID,
 	})
 	d.SetId(resourceID)
+
 	return []*schema.ResourceData{d}, nil
 }
+
+func hasNewTeamID(ctx context.Context, diff *schema.ResourceDiff, meta any) bool {
```
stevehipwell: Could you mirror the pattern in #3069 for reusable diff functions and create `diffTeam`?

github/resource_github_emu_group_mapping_migration.go
```diff
@@ -0,0 +1,67 @@
...
+import (
+	"context"
+	"net/http"
+	"strconv"
+
+	"github.com/hashicorp/terraform-plugin-log/tflog"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+)
+
+func resourceGithubEMUGroupMappingResourceV0() *schema.Resource {
```
stevehipwell:
```suggestion
func resourceGithubEMUGroupMappingV0() *schema.Resource {
```

github/resource_github_emu_group_mapping_migration.go
```diff
@@ -0,0 +1,67 @@
...
+			},
+			"etag": {
+				Type:     schema.TypeString,
+				Computed: true,
+			},
+		},
+	}
+}
+
+func resourceGithubEMUGroupMappingInstanceStateUpgradeV0(ctx context.Context, rawState map[string]any, meta any) (map[string]any, error) {
```
stevehipwell:
```suggestion
func resourceGithubEMUGroupMappingStateUpgradeV0(ctx context.Context, rawState map[string]any, meta any) (map[string]any, error) {
```

github/resource_github_emu_group_mapping_migration_test.go
```diff
@@ -0,0 +1,127 @@
...
+{
+	"id": %d
+}
+`, testTeamID),
+			StatusCode: 200,
+		},
+	}
+}
+
+func TestGithub_MigrateEMUGroupMappingsState(t *testing.T) {
```
stevehipwell: Please could you use table test for this, see #3069 for an example of a conversion.

github/resource_github_emu_group_mapping_test.go
```diff
@@ -183,3 +213,20 @@ func testAccGithubEMUGroupMappingConfig(teamName string, groupID int) string {
 	}
 	`, teamName, groupID)
 }
+
+func testAccGithubEMUGroupMappingTwoTeamsConfig(teamName1, teamName2 string, groupID int, useTeam string) string {
```
stevehipwell: Please could you inline this, KISS > DRY.

stevehipwell (changes requested):

github/resource_github_emu_group_mapping_migration_test.go
deiga: Could you point me to the exact change you mean? AFAIK I'm already using a test table

stevehipwell: Sorry, missed it due to the additional functions. Will re-review.

stevehipwell:
```suggestion
func Test_resourceGithubEMUGroupMappingStateUpgradeV0(t *testing.T) {
```

We're aiming for a test function per function.

github/resource_github_emu_group_mapping_migration_test.go
```diff
@@ -0,0 +1,127 @@
...
+	currentStateFunc  func() map[string]any
+	expectedStateFunc func(t *testing.T) map[string]any
+)
+
+var (
+	testTeamID  = 432574718
+	testGroupID = 1234567890
+)
+
+func testResourceGithubEMUGroupMappingInstanceStateDataV0() map[string]any {
```
stevehipwell: Please could this be inlined.

github/resource_github_emu_group_mapping_migration_test.go
```diff
@@ -0,0 +1,127 @@
...
+	return map[string]any{
+		"id":        "teams/test-team/external-groups",
+		"team_slug": "test-team",
+		"group_id":  testGroupID,
+	}
+}
+
+func testResourceGithubEMUGroupMappingInstanceStateDataV1(t *testing.T) map[string]any {
```
stevehipwell: Please could this be inlined.

github/resource_github_emu_group_mapping_migration_test.go
```diff
@@ -0,0 +1,127 @@
...
+	v0["team_id"] = int64(testTeamID)
+	resourceID, err := buildID(strconv.Itoa(testTeamID), v0["team_slug"].(string), strconv.Itoa(v0["group_id"].(int)))
+	if err != nil {
+		t.Fatalf("error building resource ID: %s", err)
+	}
+	v0["id"] = resourceID
+	return v0
+}
+
+func buildMockResponsesForMigrationV0toV1() []*mockResponse {
```
stevehipwell: I like the mocking pattern here but the details of the payload should be in the test alongside the data.

github/resource_github_emu_group_mapping_migration_test.go
```diff
@@ -0,0 +1,127 @@
...
+			client := github.NewClient(httpCl)
+			u, _ := url.Parse(ts.URL + "/")
+			client.BaseURL = u
+			meta.v3client = client
+
+			currentState := d.rawState()
+			got, err := d.migrationFunc(t.Context(), currentState, meta)
+			expectedState := d.want(t)
+			if (err != nil) != d.shouldError {
+				t.Fatalf("unexpected error state: %s", err.Error())
```
stevehipwell: `err` may be nil.

stevehipwell (changes requested):
I should have clarified the table test comment, I'm expecting the table tests to be self contained. Otherwise there isn't much point as it's non-trivial to add new cases.

github/resource_github_emu_group_mapping_migration_test.go
deiga: I agree with you, but I can't currently wrap my head around how to implement that nicely

stevehipwell: Can't you put the URL and response payload into the test case and then use a generic helper function in the test body?

stevehipwell:
@deiga this PR has conflicts.

github/resource_github_emu_group_mapping_migration_test.go
deiga: I've come up with one structure, let me know what you think of it

deiga:
@stevehipwell Conflicts resolved

stevehipwell (changes requested):
@deiga could you remove the migration test from this PR and add it to a new PR? I think that pattern needs more discussion and may benefit from being looked at from a simpler perspective, such as actions secret. If this is done I think we're ready to merge.

deiga:
@stevehipwell Done

stevehipwell (approved):
LGTM

stevehipwell:

deiga:
@stevehipwell Done

nickfloyd (approved):

PR #3123, opened 2026-01-26
feat: support internal visibility for repositories created by a template

### LINKED ISSUE #925: Failed setting visibility=internal on a newly created repo from template
Hi there,

### Terraform Version
`1.0.5`

### Affected Resource(s)
- github_repository

```hcl
# create github repository for the domain under domain team
resource "github_repository" "domain" {
  name = lower("${var.domain}-domain")

  visibility = "internal"

  template {
    owner      = "our-enterprise-org-id"
    repository = "tf-domain-template"
  }
}
```

### Debug Output
Gist: https://gist.github.com/damir-dezeljin/e9de28e11aa7de8d7a60fc2e123794f6

**NOTE:** I removed the sensible parts from the logs. Still, I think the logs shows what's the actual issue.  Let me recap it here:

- We're using TF to create a GitHub repo out of a template.
- The template repo visibility is set to `Internal`.
- The newly created repo visibility should be `Internal` as well.
- The repo gets created, but it remains `Private` and the visibility setup phase fails with error 422.

### Panic Output
n/a

### Expected Behavior
I'd expect a new repo with visibility `Internal` being created.

### Actual Behavior
The `terraform apply` fails with error code `422` on visibility change step.

### Steps to Reproduce
1. Set org policy to prevent users from changing repos visibility.
2. Create a template repo with visibility `Internal`
3. Use the provided TF code for creating a new repo with visibility `Internal` out of the template one.
4. Use a GitHub App identity with scopes listed below.
5. Run `terraform apply`.

### Important Factoids
- We're running TF with GitHub Apps credentials with following scopes:
  - Repository access: All repositories
  - Repository permissions:
    - Administration: r/w
    - Contents: r/w
    - Metadata: ro
    - Secrets: r/w
  - Organization permissions:
    - Members: r/w
    - Administration: r/w
- Our Enterprise has an Org.Policy set that prevents users from creating Public repos, as well only Org.Owners are allowed to change repos visibility.

### References
n/a

### LINKED ISSUE #3134: [BUG]: Template-generated repositories default to public visibility in > 6.9.1
### Expected Behavior

1. Creating a GitHub repository from a template with "internal" visibility clones the new repository as private; subsequent update attempts to update the visibility to "internal" (if supported - i.e. when org is part of an enterprise).
2. This is treated as a single atomic operation.  If any constituent part of the operation fails, rollback ensures the new repository is destroyed.

### Actual Behavior

[Cloned repository is created as public, unless visibility is set to private](https://github.com/integrations/terraform-provider-github/blob/v6.10.2/github/resource_github_repository.go#L657-L676).  `internal` is ignored entirely.

If the organization rulesets feature (currently in preview) is enabled with a rule in force to prohibit creation of public repositories, this will prevent the repository from being created.

Otherwise, the repository is created initially as `public`.  In scenarios where illegal settings (for example, explicitly disabling GHAS) for public repositories are applied:
1. The underlying CRUD operation will fail;
2. The failure cascades to the whole apply operation;
3. The repository is not deleted
4. The repository - intended for internal use - is left as public.  If an internal/private template was used, there exists a risk of compromised material (however limited the form or scope).

### Terraform Version

Terraform v1.14.3
on linux_x64
Provider version 6.10.2

### Affected Resource(s)

- `github_repository`

```hcl
# This will fail if organization rulesets are in play
resource "github_repository" "default" {
  name = "my-repo"
  template {
    owner = "my-org"
    repository = "example-template"
  }
  visibility = "internal"
}

# If appropriate organization ruleset rules are not enforced
# this will fail regardless
# leaving a public clone of the template repository
resource "github_repository" "default" {
  name = "my-repo"
  template {
    owner = "my-org"
    repository = "example-template"
  }
  visibility = "internal"

  security_and_analysis {
    advanced_security {
      status = "disabled"
    }
    secret_scanning {
      status = "disabled"
    }
    secret_scanning_push_protection {
      status = "disabled"
    }
  }
}
```

### Steps to Reproduce

Apply one of the scenario configurations shown above (subject to appropriate organization rulesets being in place), then:

```
$ terraform apply
```

### Debug Output

```shell

```

### Panic Output

```shell
# In case of the organisation ruleset
╷
│ Error: POST https://api.github.com/repos/my-org/example-template/generate: 422 Could not clone: Repository Due to policy, repository visibility can not be set to public [{Resource: Field: Code: Message:Could not clone: Repository Due to policy, repository visibility can not be set to public}]
│ 
│   with module.common-repo.github_repository.default,
│   on ../../modules/common-repo/main.tf line 7, in resource "github_repository" "default":
│    7: resource "github_repository" "default" {
│ 
╵

# Otherwise, an HTTP 422 when attempting to disable GHAS
# (I can't reproduce this without creating another compromised repository within my org - which policy that I am not removing now prevents)
```

### Code of Conduct

puneet-arora15:
Resolves #925
Resolves #3134

----

### Before the change?

- When creating a repository from a template with `visibility = "internal"`, the repository was created as **public** due to incorrect handling of the `Private` boolean field
- The code only set `Private=true` when `visibility == "private"`, causing `visibility = "internal"` to default to `Private=false` (public)
- This resulted in repositories being temporarily exposed publicly before being updated to internal visibility
- **Security Impact**: Repositories intended to be internal to an organization were briefly accessible to the public, potentially exposing sensitive data

### After the change?

- Modified the `isPrivate` logic to set `Private=true` for both `"private"` and `"internal"` visibility values
- Repositories with `visibility = "internal"` are now correctly created as private initially
- The visibility is then properly updated to "internal" via the existing PATCH request in `resourceGithubRepositoryUpdate()`
- **Security Fixed**: No public exposure window - repositories remain private/internal throughout the entire creation process
- Added inline comment explaining the API limitation and two-step process

### Pull request checklist

**Note on checklist items:**
- **Schema migrations**: Not needed - no schema changes required
- **Tests**: Added new test case `"sets internal visibility for repositories created by a template"` that verifies repositories created from templates with `visibility = "internal"` are correctly created as internal
- **Docs**: No documentation updates needed - this is a bug fix that makes the existing documented behavior work correctly

- [ ] Yes
- [x] No

This is a **bug fix** that corrects the existing behavior to match the documented and expected functionality. Users who specify `visibility = "internal"` for template-based repositories will now get the correct behavior without any changes to their Terraform configurations.

----

## Technical Details

**Root Cause**: The GitHub template creation API (`POST /repos/{template_owner}/{template_repo}/generate`) only accepts a `private` boolean parameter and does not support the `visibility` parameter directly.

**Code Change** (line ~660 in `github/resource_github_repository.go`):
```go
// Before:
isPrivate := repoReq.GetVisibility() == "private"

// After:
// Template API only supports Private boolean, so treat "internal" as private, then update via PATCH.
isPrivate := repoReq.GetVisibility() == "private" || repoReq.GetVisibility() == "internal"

### DIFF
--- a/github/resource_github_repository_test.go
+++ b/github/resource_github_repository_test.go
@@ -1422,7 +1422,7 @@ resource "github_repository" "test" {
 					Config: config,
 					Check: resource.ComposeTestCheckFunc(
 						resource.TestCheckResourceAttr("github_repository.test", "visibility", "internal"),
-						resource.TestCheckResourceAttr("github_repository.test", "private", "false"),
+						resource.TestCheckResourceAttr("github_repository.test", "private", "true"),
 					),
 				},
 			},

### CODE REVIEW

**Comment by deiga** [comment]
Please mark this as resolving https://github.com/integrations/terraform-provider-github/issues/925

~AFAIK these changes break the `allow_forking` tests, please address those as well~
Apparently those tests are broken on `main`

File: github/resource_github_repository.go
Comment (stevehipwell) [inline github/resource_github_repository.go]:
```suggestion
if template, ok := d.GetOk("template"); ok {
templateConfigBlocks := template.([]any)

for _, templateConfigBlock := range templateConfigBlocks {
templateConfigMap, ok := templateConfigBlock.(map[string]any)
if !ok {
return diag.FromErr(errors.New("failed to unpack template configuration block"))
}

templateRepo := templateConfigMap["repository"].(string)
templateRepoOwner := templateConfigMap["owner"].(string)
includeAllBranches := templateConfigMap["include_all_branches"].(bool)

// Template API only supports Private boolean, so treat "internal" as private, then update via PATCH.
	private := repoReq.GetVisibility() != "public"

templateRepoReq := github.TemplateRepoRequest{
Name:               &repoName,
Owner:              &owner,
Description:        github.Ptr(d.Get("description").(string)),
Private:            github.Ptr(private),
IncludeAllBranches: github.Ptr(includeAllBranches),
}
```

The code modification isn't valid as it's only when creating using a template that we want to create a private repository for internal. While we're making this change we can remove the unnecessary setting of the `Private` struct field as it's ignored if the visibility field is set (as it always is).

github/resource_github_repository_test.go
```diff
@@ -1295,6 +1295,43 @@ resource "github_repository" "test" {
 			},
 		})
 	})
+
+	t.Run("sets internal visibility for repositories created by a template", func(t *testing.T) {
```
stevehipwell:
```suggestion
	t.Run("create_internal_repo_from_template", func(t *testing.T) {
```

github/resource_github_repository_test.go
```diff
@@ -1295,6 +1295,43 @@ resource "github_repository" "test" {
...
+		randomID := acctest.RandStringFromCharSet(5, acctest.CharSetAlphaNum)
+		testRepoName := fmt.Sprintf("%stemplate-visibility-internal-%s", testResourcePrefix, randomID)
```
stevehipwell:
```suggestion
		testRepoName := fmt.Sprintf("%s%s", testResourcePrefix, randomID)
```

github/resource_github_repository_test.go
```diff
@@ -1295,6 +1295,43 @@ resource "github_repository" "test" {
...
+		config := fmt.Sprintf(`
+			resource "github_repository" "internal" {
```
stevehipwell:
```suggestion
			resource "github_repository" "test" {
```

github/resource_github_repository_test.go
```diff
@@ -1295,6 +1295,43 @@ resource "github_repository" "test" {
...
+		check := resource.ComposeTestCheckFunc(
+			resource.TestCheckResourceAttr(
+				"github_repository.internal", "visibility",
+				"internal",
+			),
+			resource.TestCheckResourceAttr(
+				"github_repository.internal", "private",
+				"false",
+			),
+		)
```
stevehipwell:
```suggestion
		check := resource.ComposeTestCheckFunc(
			resource.TestCheckResourceAttr("github_repository.internal", "visibility", "internal"),
			resource.TestCheckResourceAttr("github_repository.internal", "private", "false"),
		)
```
And please inline this, it doesn't need to be declared as a variable.

puneet-arora15:
> Please mark this as resolving #925
> 
> ~AFAIK these changes break the `allow_forking` tests, please address those as well~ Apparently those tests are broken on `main`

Thanks @deiga, I have updated the description for it.

github/resource_github_repository.go
puneet-arora15: Fixed

github/resource_github_repository_test.go
puneet-arora15: Fixed

github/resource_github_repository_test.go
puneet-arora15: Fixed

github/resource_github_repository_test.go
puneet-arora15: Fixed

puneet-arora15:
> Thanks for the PR @puneet-arora15. As well as the suggested changes this PR is going to need some docs changes (they're currently handwritten) to explain the updated behaviour.

Thanks @stevehipwell for detailed updates to the PR, those changes have now been merged.

puneet-arora15:
I have now (screenshot attached), also changed test assertion from `private = "false"` to `private = "true"` for internal repositories.

This matches GitHub API's actual response where internal repositories have the private field set to true, aligning with the fix that treats internal repos as private during template creation.

image

github/resource_github_repository_test.go
```diff
@@ -1295,6 +1295,35 @@ resource "github_repository" "test" {
...
+
+		resource.Test(t, resource.TestCase{
+			PreCheck:          func() { skipUnlessMode(t, enterprise) },
+			ProviderFactories: providerFactories,
+			Steps: []resource.TestStep{
+				{
+					Config: config,
+					Check: resource.ComposeTestCheckFunc(
+						resource.TestCheckResourceAttr("github_repository.test", "visibility", "internal"),
+						resource.TestCheckResourceAttr("github_repository.test", "private", "true"),
```
stevehipwell, line 1425: This would be an invalid state, you likely need to set the private field as part of create/update.

puneet-arora15: I tested this directly with GitHub's API:

Internal repos always return private=true - this is how GitHub represents them

Setting private=false is explicitly rejected by GitHub Enterprise Cloud with: "Public repositories are not permitted for Enterprise Managed Organizations"

The private field is boolean: false = public, true = not-public (includes both private AND internal)

The visibility field distinguishes between private and internal, while private=true for both indicates "not publicly accessible."

stevehipwell: Sorry, you're correct.

stevehipwell (approved):
LGTM

github/resource_github_repository.go
```diff
@@ -654,8 +654,6 @@ func resourceGithubRepositoryCreate(ctx context.Context, d *schema.ResourceData,
 	owner := meta.(*Owner).name
 	repoName := repoReq.GetName()
 
-	isPrivate := repoReq.GetVisibility() == "private"
-	repoReq.Private = github.Ptr(isPrivate)
```
FlaviusRaducu, line 658: This might be a stupid question, sorry :)

Moving this to the template if scope, would't that create a situation where we don't update the repoReq.Private field based on visibility when the else block executes, that is when we create without a template?

I assume the endpoint code does a a similar thing when it gets a request with private = false and visibility = private/internal?

stevehipwell: @FlaviusRaducu the `visibility` field overrides the `private` field; so we don't need to set `private` as we always set `visibility`.

stevehipwell:
@puneet-arora15 could you please rebase this PR?

Issue #3038, opened 2026-01-04
[MAINT]: Add a new resource for `github_repository_vulnerability_alerts`


deiga:
### Describe the need

https://github.com/integrations/terraform-provider-github/pull/3024#discussion_r2657695912

As the vulnerability alerts of a repository are a separate API it makes sense to separate their config from `github_repository`. Especially with the Enterprise and Organization policies that prevent certain configurations for vulnerability alerts it's going to be easier to maintain sensible defaults.

Since this is deprecating existing functionality we need to do the following steps:

- [x] Create a new resource and data source `github_repository_vulnerability_alerts`
- ~~[ ] Add a new field `ignore_vulnerability_alerts` to `github_repository` which disables any interactions with the vulnerability alerts API~~
- [x] Mark the fields `ignore_vulnerability_alerts_during_read` and `vulnerability_alerts` as Deprecated in `github_repository`
- [ ] #3039

### Relevant log output

```shell

```

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

PR #3127, opened 2026-01-27
fix: Correct forking and vulnerability alert logic

stevehipwell:
Resolves #2988
Resolves #3111

----

### Before the change?

- Setting `allow_forking = false` on `github_repository` resulted in incorrect behaviour
- The `vulnerability_alerts` attribute of `github_repository` couldn't be left unset to support default behaviour
- Missing tests for `github_repository` allow forking and vulnerability alerts
- Tests from `github_organization_role_users` data source were broken

### After the change?

- Setting `allow_forking = false` on `github_repository` works as expected
- The `vulnerability_alerts` attribute of `github_repository` works correctly when unset
- Tests for `github_repository` allow forking and vulnerability alerts added
- Tests from `github_organization_role_users` data source fixed

### Pull request checklist

- [x] Schema migrations have been created if needed ([example](https://github.com/F-Secure-web/terraform-provider-github/blob/main/github/migrate_github_actions_organization_secret.go))
- [x] Tests for the changes have been added (for bug fixes / features)
- [x] Docs have been reviewed and added / updated if needed (for bug fixes / features)

### Does this introduce a breaking change?

Please see our docs on [breaking changes](https://github.com/octokit/.github/blob/master/community/breaking_changes.md) to help!

- [ ] Yes
- [x] No

----

PR #3130, opened 2026-01-27
feat: new field `ignore_vulnerability_alerts` to `github_repository`

lens0021:
Resolves the second task of #3038

----

### Before the change?

- `ignore_vulnerability_alerts_during_read` is only for read, so user cannot control of create, read, or update.

### After the change?

- Setting `ignore_vulnerability_alerts` to `true` makes it possible not to call the vulnerability alerts endpoint in any case.

### Pull request checklist

- [ ] Yes
- [x] No

----


stevehipwell:
Thanks for the PR @lens0021, but there is a fix on main (#3127) that stops the endpoint being hit on create/update if the field isn't set. This combined with `ignore_vulnerability_alerts_during_read` already meets the requirements so we don't need another attribute.

Issue #2592, opened 2025-03-13
[FEAT]: Support for Ephemeral Resources


evantissot:
### Describe the need

Hello! 
I'm reaching out to inquire about the potential support for ephemeral resources within the GitHub Terraform provider. 
My specific use case involves creating pull requests (PRs) using this provider. I don't really need to store the PR resource in the state? Currently, I have to apply Terraform a second time to remove these resources from the state. Ephemeral resources would be a great addition, allowing resources to exist temporarily without being stored persistently.
Is there any plan to support this feature in the future? 
Thank you 

### Relevant log output

```shell

```

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

twaslowski:
Seconding this. Specifically providing write-only implementations of `github_actions_secret` and `github_dependabot_secret` would be incredibly helpful.

davidjoliver86:
I would love to be able to manage `github_repository_deploy_key` resources using a write-only argument referencing an ephemeral `tls_private_key` resource.

honi:
> I would love to be able to manage `github_repository_deploy_key` resources using a write-only argument referencing an ephemeral `tls_private_key` resource.

I want the exact same thing.

yannickvr:
github_repository_file and github_repository_pull_request need this too

deiga:
Hey there 👋 

As this provider is still using terraform-plugin-sdk/v2 we don't have direct access to Ephemeral Resources yet. At some point in the future there are plans for migrating to use terraform-plugin-framework instead, but that timeline is as of yet unknown.

kokolight:
Hey @deiga , What about adding `write only` option to make sure the value of the secret is not written into the state?
This way you can leverage tools like [sops](https://github.com/getsops/sops) to store the secret encrypted and use it in the github action secret creation.
You can see in other providers like aws with [ssm_parameter](https://github.com/hashicorp/terraform-provider-aws/blob/main/internal/service/ssm/parameter.go) as a reference, just need to add *_wo at the end and this way you can keep the current behavior as it is and no need to do any breaking changes.

deiga:
> Hey @deiga , What about adding `write only` option to make sure the value of the secret is not written into the state?
> This way you can leverage tools like [sops](https://github.com/getsops/sops) to store the secret encrypted and use it in the github action secret creation.
[...]

PR #3151, opened 2026-02-01
feat: write-only values support to github_actions_environment_secret

kokolight:
Implements write-only values support for `github_actions_environment_secret`

Resolves part of #2592 (only write-only support, not ephemeral resources)

----

## Changes:
- Add `plaintext_value_wo` and `plaintext_value_wo_version` fields to `github_actions_environment_secret`
- Added tests to `github_actions_environment_secret` resource
Update documentation with usage examples

## Usage:
```hcl
resource "github_actions_environment_secret" "example" {
  repository                 = "my-repo"
  secret_name                = "my-secret"
  environment                = "dev"
  plaintext_value_wo         = "my-secret-value"
  plaintext_value_wo_version = 1
}
```
### Pull request checklist

- [ ] Yes
- [x] No

----


stevehipwell:
Thanks for the PR @kokolight, but this whole area is under active development in #3069; I actually had write only support coded before I removed it due to the additional complexity it introduced. Once #3069 has been merged we will need to carefully consider the design for write only before we implement it.

kokolight:
Ok @stevehipwell , hopefully this is going to be solved soon so we can get this great and needed feature 😄

kokolight:
Hey @stevehipwell, any news regarding this PR?
I remember I've seen some plans for write only variables in v6.12.0 but I see the release is lacking them.
Are you planning on merging any of these features for future releases?
Thanks!

stevehipwell:
@kokolight if you want to rebase your PR, I think we could take a look at adding write only support to the refactored secrets (actions and dependabot).

VladAdGad:
@kokolight I'm happy to do the rebase for you if that helps move this forward.

Two things worth flagging first, because this is not a mechanical rebase after #3069 landed:

- `resourceGithubActionsEnvironmentSecretCreateOrUpdate` no longer exists on `main`. The resource is now split into separate `Create` / `Update` / `Import` funcs, plus `SchemaVersion: 1` with a state upgrader and a `CustomizeDiff` (`diffRepository`, `diffSecret`). The write-only branch in this PR has to be reimplemented against that shape rather than merged.
- `plaintext_value` and `encrypted_value` are now deprecated in favour of `value` and `value_encrypted` (+ `key_id`). So the new fields should very likely be `value_wo` / `value_wo_version` to match the new naming, not `plaintext_value_wo`.

Both conflicting files (`resource_github_actions_environment_secret.go` and its test) need hand resolution; `git rebase origin/main` stops on both.

To push the rebased branch I'd need write access to `kokolight/terraform-provider-github` - `maintainer_can_modify` on this PR only lets base-repo maintainers push, and I'm not one. Could you add me (`VladAdGad`) as a collaborator on your fork? Then I can force-push `feat/write-only-github-environment-secret` and this PR updates in place, keeping your commits and the review history here.

If you'd rather not hand out access, I can instead open a PR against your fork's branch and you just merge it - same result, no permissions needed. Whichever you prefer.

kokolight:
> VladAdGad

Absolutely!
Feel free to do the rebase, added you to collaborators on the fork.

VladAdGad:
Rebased onto current `main` and force-pushed (`98d24e2`). Thanks for the access, @kokolight.

As flagged above, #3069 meant this could not be replayed as-is, so the feature is reimplemented on the new resource shape. What changed relative to the original two commits:

**Renamed to match the refactored schema.** `plaintext_value_wo` / `plaintext_value_wo_version` are now `value_wo` / `value_wo_version`, following `value` and `value_encrypted`. `value_wo` joins the existing `ExactlyOneOf` group and is added to `key_id`'s `ConflictsWith`.

**Fixed a bug that would have written empty secrets.** The original code read the write-only value with `d.GetOk("plaintext_value_wo")`. A write-only argument is absent from both state and the diff, so `d.Get*` always returns the zero value and GitHub would have received a secret encrypted from `""`. The value must come from the raw config via `d.GetRawConfigAt(cty.GetAttrPath("value_wo"))`, which is also the pattern already used in `resource_github_repository_collaborators.go`. I verified this with a throwaway harness driving a real `terraform apply` against a mock GitHub API and decrypting what the provider pushed: with the original approach both writes arrive as `""`; with the raw-config read they arrive as the configured values.

**Dropped the manual version-comparison branch.** It existed because the old resource had a single `CreateOrUpdate` that ran on every apply. With separate `Create`/`Update`, `value_wo_version` is simply the only attribute that can produce a diff for a write-only secret, so Terraform already gates the write. `ForceNew` is also removed from the version attribute, since a changed version should update in place, not replace the secret. Unknown-at-apply values now raise an error rather than silently writing an empty secret.

**Docs are generated now.** `website/docs/r/*.html.markdown` no longer exists; docs come from `tfplugindocs`. I added `examples/resources/github_actions_environment_secret/resource_4.tf`, a paragraph in the template, and regenerated `docs/`. The original docs commit was superseded, so it is not in the rebased history; both remaining commits are still authored by you.

**Tests.** Added a `with_value_wo` subtest in the current file's style, covering the contract that matters: changing `value_wo` without touching `value_wo_version` must plan as a no-op, and bumping the version must plan an update. It is gated with `tfversion.SkipBelow(tfversion.Version1_11_0)`.

Locally green on the rebased branch: `go build ./...`, `go vet`, `gofmt -l`, `golangci-lint run ./github/...` (0 issues), the unit suite (`go test ./... -skip '^TestAcc'`), and `tfplugindocs` regeneration leaves no diff. Acceptance tests need credentials I do not have, so `with_value_wo` has not been run against real GitHub.

@stevehipwell this is ready for a look whenever you have time. If you would rather `value_wo` also land on the other refactored secrets in the same PR, say so and it can be extended.

stevehipwell:
@VladAdGad at a high level the maintainers still need to agree that we're happy to add write only inputs as technically they require Terraform v1.11 (although they should be fine on older versions if not used). This would require some level of testing and a potential maintenance burden on top of what we already have.

RE this PR, I'd want to see all of the secrets support this at the same time. I was planning on picking up the codespace secrets/variables to align them to the other types before starting the write only work, so in an ideal world this happens first but I wouldn't consider not including codespaces a blocker.

CC @deiga what do you think about supporting write-only inputs?

deiga:
@stevehipwell I think TF 1.11 is old enough that creating a breaking change is fine to consider. Especially, since customers can just pin to a previous version (or upgrade TF). 

I support having write-only attributes!

stevehipwell:
> @stevehipwell I think TF 1.11 is old enough that creating a breaking change is fine to consider. Especially, since customers can just pin to a previous version (or upgrade TF).

@deiga I don't think they're breaking if not used, which leaves us still able to be run by OSS TF.

VladAdGad:
Hi, @stevehipwell. Should I rebase again now, or should I wait until you ping me?

PR #3162, opened 2026-02-05
feat: Fixup repository environment logic

stevehipwell:
Resolves #ISSUE_NUMBER

----

### Before the change?

- `github_repository_environment` resource logic for reviewers had incorrect constraints
- `github_repository_environment_deployment_policy` resource missing policy ID attribute

### After the change?

- `github_repository_environment` has correct constraints for reviewers, including validation for the sum of users and teams
- `github_repository_environment_deployment_policy` has new `policy_id` attribute
- Improved tests
- Improved repository diff logic to support gradual adoption of `repository_id`

### Pull request checklist

- [ ] Yes
- [x] No

----

### DIFF
--- a/github/resource_github_repository_environment.go
+++ b/github/resource_github_repository_environment.go
@@ -3,31 +3,39 @@ package github
 import (
 	"context"
 	"errors"
-	"log"
+	"fmt"
 	"net/http"
 	"net/url"
 
 	"github.com/google/go-github/v83/github"
+	"github.com/hashicorp/terraform-plugin-log/tflog"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/customdiff"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/validation"
 )
 
 func resourceGithubRepositoryEnvironment() *schema.Resource {
 	return &schema.Resource{
-		CreateContext: resourceGithubRepositoryEnvironmentCreate,
-		ReadContext:   resourceGithubRepositoryEnvironmentRead,
-		UpdateContext: resourceGithubRepositoryEnvironmentUpdate,
-		DeleteContext: resourceGithubRepositoryEnvironmentDelete,
-		Importer: &schema.ResourceImporter{
-			StateContext: resourceGithubRepositoryEnvironmentImport,
+		SchemaVersion: 1,
+		StateUpgraders: []schema.StateUpgrader{
+			{
+				Type:    resourceGithubRepositoryEnvironmentV0().CoreConfigSchema().ImpliedType(),
+				Upgrade: resourceGithubRepositoryEnvironmentStateUpgradeV0,
+				Version: 0,
+			},
 		},
+
 		Schema: map[string]*schema.Schema{
 			"repository": {
 				Type:        schema.TypeString,
 				Required:    true,
-				ForceNew:    true,
 				Description: "The repository of the environment.",
+			},
+			"repository_id": {
+				Description: "The ID of the GitHub repository.",
+				Type:        schema.TypeInt,
+				Computed:    true,
 			},
 			"environment": {
 				Type:        schema.TypeString,
@@ -56,20 +64,22 @@ func resourceGithubRepositoryEnvironment() *schema.Resource {
 			"reviewers": {
 				Type:        schema.TypeList,
 				Optional:    true,
-				MaxItems:    6,
+				MaxItems:    1,
 				Description: "The environment reviewers configuration.",
 				Elem: &schema.Resource{
 					Schema: map[string]*schema.Schema{
 						"teams": {
 							Type:        schema.TypeSet,
-							Optional:    true,
 							Elem:        &schema.Schema{Type: schema.TypeInt},
+							Optional:    true,
+							MaxItems:    6,
 							Description: "Up to 6 IDs for teams who may review jobs that reference the environment. Reviewers must have at least read access to the repository. Only one of the required reviewers needs to approve the job for it to proceed.",
 						},
 						"users": {
 							Type:        schema.TypeSet,
-							Optional:    true,
 							Elem:        &schema.Schema{Type: schema.TypeInt},
+							Optional:    true,
+							MaxItems:    6,
 							Description: "Up to 6 IDs for users who may review jobs that reference the environment. Reviewers must have at least read access to the repository. Only one of the required reviewers needs to approve the job for it to proceed.",
 						},
 					},
@@ -96,12 +106,50 @@ func resourceGithubRepositoryEnvironment() *schema.Resource {
 				},
 			},
 		},
+
+		CustomizeDiff: customdiff.All(
+			diffRepository,
+			resourceGithubRepositoryEnvironmentDiff,
+		),
+
+		CreateContext: resourceGithubRepositoryEnvironmentCreate,
+		ReadContext:   resourceGithubRepositoryEnvironmentRead,
+		UpdateContext: resourceGithubRepositoryEnvironmentUpdate,
+		DeleteContext: resourceGithubRepositoryEnvironmentDelete,
+		Importer: &schema.ResourceImporter{
+			StateContext: resourceGithubRepositoryEnvironmentImport,
+		},
 	}
 }
 
-func resourceGithubRepositoryEnvironmentCreate(ctx context.Context, d *schema.ResourceData, meta any) diag.Diagnostics {
-	client := meta.(*Owner).v3client
-	owner := meta.(*Owner).name
+func resourceGithubRepositoryEnvironmentDiff(_ context.Context, d *schema.ResourceDiff, _ any) error {
+	if d.Id() == "" {
+		return nil
+	}
+
+	if v, ok := d.GetOk("reviewers"); ok {
+		count := 0
+		o := v.([]any)[0]
+		if t, ok := o.(map[string]any)["teams"]; ok {
+			count += t.(*schema.Set).Len()
+		}
+
+		if t, ok := o.(map[string]any)["users"]; ok {
+			count += t.(*schema.Set).Len()
+		}
+
+		if count > 6 {
+			return fmt.Errorf("a maximum of 6 reviewers (users and teams combined) can be set for an environment")
+		}
+	}
+
+	return nil
+}
+
+func resourceGithubRepositoryEnvironmentCreate(ctx context.Context, d *schema.ResourceData, m any) diag.Diagnostics {
+	meta := m.(*Owner)
+	client := meta.v3client
+	owner := meta.name
 
 	repoName := d.Get("repository").(string)
 	envName := d.Get("environment").(string)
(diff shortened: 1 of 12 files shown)


github/resource_github_repository_environment.go
```diff
@@ -3,30 +3,24 @@ package github
...
-		UpdateContext: resourceGithubRepositoryEnvironmentUpdate,
-		DeleteContext: resourceGithubRepositoryEnvironmentDelete,
-		Importer: &schema.ResourceImporter{
-			StateContext: resourceGithubRepositoryEnvironmentImport,
-		},
 		Schema: map[string]*schema.Schema{
 			"repository": {
 				Type:        schema.TypeString,
 				Required:    true,
-				ForceNew:    true,
```
deiga, line 29: question: Should we add `repository_id` as a Computed field?

github/resource_github_repository_environment.go
```diff
@@ -121,38 +155,36 @@ func resourceGithubRepositoryEnvironmentCreate(ctx context.Context, d *schema.Re
...
+	envName := d.Get("environment").(string)
 
 	env, _, err := client.Repositories.GetEnvironment(ctx, owner, repoName, url.PathEscape(envName))
 	if err != nil {
 		var ghErr *github.ErrorResponse
 		if errors.As(err, &ghErr) {
 			if ghErr.Response.StatusCode == http.StatusNotFound {
-				log.Printf("[INFO] Removing repository environment %s from state because it no longer exists in GitHub",
-					d.Id())
+				log.Printf("[INFO] Removing repository environment %s from state because it no longer exists in GitHub", d.Id())
```
deiga: question: Would you be open to changing to `tflog` in this PR already?

github/resource_github_repository_environment_deployment_policy.go
```diff
@@ -3,33 +3,35 @@ package github
...
+				Version: 0,
+			},
 		},
+
 		Schema: map[string]*schema.Schema{
 			"repository": {
 				Description: "The name of the GitHub repository.",
 				Type:        schema.TypeString,
 				Required:    true,
-				ForceNew:    true,
```
deiga, line 32: question: Should we be adding the `repository_id` Computed field here as well?

github/resource_github_repository_environment_deployment_policy.go
```diff
@@ -38,148 +40,147 @@ func resourceGithubRepositoryEnvironmentDeploymentPolicy() *schema.Resource {
...
-func resourceGithubRepositoryEnvironmentDeploymentPolicyDiff(_ context.Context, diff *schema.ResourceDiff, v any) error {
-	if diff.Id() == "" {
+func resourceGithubRepositoryEnvironmentDeploymentPolicyDiff(_ context.Context, d *schema.ResourceDiff, _ any) error {
+	if d.Id() == "" {
 		return nil
 	}
 
-	if diff.HasChange("branch_pattern") && diff.HasChange("tag_pattern") {
-		if err := diff.ForceNew("branch_pattern"); err != nil {
+	if d.HasChange("branch_pattern") && d.HasChange("tag_pattern") {
```
deiga, line 90: question: While you're touching this, would it make sense to use `d.HasChanges` here?

github/util_check.go
```diff
@@ -0,0 +1,25 @@
...
+
+		attr, ok := res.Primary.Attributes[key]
+		if !ok {
+			return fmt.Errorf("attribute not found: %s", key)
+		}
+
+		*target = attr
+		return nil
+	}
+}
```
deiga: Looking forward to getting the `terraform-plugin-testing` included, so that we can replace this func and similar ones with https://developer.hashicorp.com/terraform/plugin/testing/acceptance-tests/value-comparers#values-same

Could you add a TODO comment mentioning that link?

github/util_diff.go
```diff
@@ -32,18 +40,8 @@ func diffRepository(ctx context.Context, diff *schema.ResourceDiff, m any) error
...
-		var repoName string
-		if o, ok := diff.GetOk("repository"); ok {
-			if v, ok := o.(string); ok {
-				repoName = v
-			} else {
-				return fmt.Errorf("repository is not a string")
-			}
-		} else {
-			return fmt.Errorf("repository is not set")
+			log.Printf("[INFO] No repository_id in schema so changing repository from %s to %s forces the resource to be recreated", old.(string), repoName)
```
deiga: question: Should we be using `tflog` here since `diffRepository` always has correct context?

github/resource_github_repository_environment.go
stevehipwell: I've not needed a migration for anything else so I've left the behaviour as-is but using the custom diff. I'm open to adding a migration and `repository_id` though.

github/resource_github_repository_environment_deployment_policy.go
stevehipwell: Same as above.

github/resource_github_repository_environment.go
stevehipwell: Yes, I think the scope of this PR is small enough.

github/resource_github_repository_environment_deployment_policy.go
stevehipwell: ~~Yes it would.~~

github/util_diff.go
stevehipwell: Yes.

github/util_check.go
stevehipwell: Me too, I'm used to working on the framework and using testing!

github/resource_github_repository_environment_deployment_policy.go
stevehipwell: Sorry, I misread my own code. This change would break the logic as it's `&&` not `||`, the current code is correct.

github/util_check.go
stevehipwell: This whole pattern isn't required when we can check the resource action instead of having to work it out based on other signals.

github/resource_github_repository_environment.go
stevehipwell: I've added this as I did have a migration for the deployment policy so I made them match.

stevehipwell:
@deiga this is ready for a re-review. I've made the most of your testing improvements. I also have thoughts around the use of `tflog.SetField()`; in as much as we should use `tflog.SetField(d.GetId())` by default and then not use it unless there is a specific reason to do so.

github/resource_github_repository_environment.go
```diff
@@ -118,41 +166,50 @@ func resourceGithubRepositoryEnvironmentCreate(ctx context.Context, d *schema.Re
...
+	}
+
 	return nil
 }
 
-func resourceGithubRepositoryEnvironmentRead(ctx context.Context, d *schema.ResourceData, meta any) diag.Diagnostics {
-	client := meta.(*Owner).v3client
-	owner := meta.(*Owner).name
+func resourceGithubRepositoryEnvironmentRead(ctx context.Context, d *schema.ResourceData, m any) diag.Diagnostics {
+	tflog.SetField(ctx, "id", d.Id())
```
deiga: I discovered that the usage pattern is the following
```suggestion
	ctx = tflog.SetField(ctx, "id", d.Id())
```

github/resource_github_repository_environment_deployment_policy.go
```diff
@@ -38,148 +45,157 @@ func resourceGithubRepositoryEnvironmentDeploymentPolicy() *schema.Resource {
...
 	}
+	if err := d.Set("policy_id", policy.GetID()); err != nil {
+		return diag.FromErr(err)
+	}
+
+	return nil
+}
+
+func resourceGithubRepositoryEnvironmentDeploymentPolicyRead(ctx context.Context, d *schema.ResourceData, m any) diag.Diagnostics {
+	tflog.SetField(ctx, "id", d.Id())
```
deiga:
```suggestion
	ctx = tflog.SetField(ctx, "id", d.Id())
```

github/util_diff.go
```diff
@@ -19,11 +18,21 @@ func diffRepository(ctx context.Context, diff *schema.ResourceDiff, m any) error
 		return nil
 	}
 
+	tflog.SetField(ctx, "id", diff.Id())
```
deiga:
```suggestion
	ctx = tflog.SetField(ctx, "id", diff.Id())
```

github/util_diff.go
```diff
@@ -118,6 +126,8 @@ func diffTeam(ctx context.Context, diff *schema.ResourceDiff, m any) error {
...
```
deiga:
```suggestion
	ctx = tflog.SetField(ctx, "id", diff.Id())
```

deiga (changes requested):

stevehipwell:
@deiga thanks for the `tflog.SetField()` correction. FYI there are still some instances of this pattern in the EMU group mapping resource.

github/resource_github_repository_environment.go
```diff
@@ -96,12 +106,50 @@ func resourceGithubRepositoryEnvironment() *schema.Resource {
...
+		}
+
+		if t, ok := o.(map[string]any)["users"]; ok {
+			r += t.(*schema.Set).Len()
+		}
+
+		if r > 6 {
+			return fmt.Errorf("a maximum of 6 reviewers (users and teams combined) can be set for an environment")
+		}
+	}
```
deiga, line 144: Could we use more descriptive variable names? IMO this is context is already too large and has too many variables, which makes readability degrade

github/resource_github_repository_environment_deployment_policy.go
```diff
@@ -204,42 +220,55 @@ func resourceGithubRepositoryEnvironmentDeploymentPolicyUpdate(ctx context.Conte
...
+func resourceGithubRepositoryEnvironmentDeploymentPolicyImport(ctx context.Context, d *schema.ResourceData, m any) ([]*schema.ResourceData, error) {
+	meta := m.(*Owner)
+	client := meta.v3client
+	owner := meta.name
 
-	branchPolicyId, err := strconv.ParseInt(branchPolicyIdString, 10, 64)
+	repoName, envNamePart, policyIDStr, err := parseID3(d.Id())
 	if err != nil {
-		return diag.FromErr(err)
+		return nil, err
```
deiga: Let's use a more descriptive error here :)
```suggestion
		return nil, fmt.Errorf("invalid import specified: supplied import must be written as <repository>:<environment>:<policyID>. Original error: %w", err)
```

github/resource_github_repository_environment.go
stevehipwell: I don't agree the context is large, it's pretty simple but verbose. I've made a change to name the important `count` variable.

github/resource_github_repository_environment.go
```diff
@@ -96,12 +106,50 @@ func resourceGithubRepositoryEnvironment() *schema.Resource {
...
+		}
+
+		if t, ok := o.(map[string]any)["users"]; ok {
+			count += t.(*schema.Set).Len()
+		}
+
+		if count > 6 {
+			return fmt.Errorf("a maximum of 6 reviewers (users and teams combined) can be set for an environment")
+		}
 	}
```
nickfloyd, line 144: Can we add some safe slice access checks here for a little more durability and `ok` evals to prevent panics? Also I noticed that we only iterate on the first reviewer in the slice and not all - was this intended (I might be reading this wrong).

```suggestion
func resourceGithubRepositoryEnvironmentDiff(_ context.Context, d *schema.ResourceDiff, _ any) error {
    if d.Id() == "" {
        return nil
    }

    if v, ok := d.GetOk("reviewers"); ok {
        count := 0
        reviewers, ok := v.([]any)
        if !ok || len(reviewers) == 0 {
            return nil
        }
        for _, o := range reviewers {
            obj, ok := o.(map[string]any)
            if !ok {
                continue
            }
            if t, ok := obj["teams"]; ok {
                if set, ok := t.(*schema.Set); ok {
                    count += set.Len()
                }
            }
            if t, ok := obj["users"]; ok {
                if set, ok := t.(*schema.Set); ok {
                    count += set.Len()
                }
            }
        }
        if count > 6 {
            return fmt.Errorf("a maximum of 6 reviewers (users and teams combined) can be set for an environment")
        }
    }
    return nil
}
```

stevehipwell: @nickfloyd this function is in the context of a TF schema, so this code is all "safe". For example calling `d.GetOk("reviewers")` will only return true when the slice has `1` item in it.

PR #3166, opened 2026-02-08
[FEAT] Add new resource for repository vulnerability alerts

### LINKED ISSUE #3038: [MAINT]: Add a new resource for `github_repository_vulnerability_alerts`

deiga:
Resolves #3038

----

### Before the change?

- Vulnerability alerts for a repository were handled inside the `github_repository` resource.

### After the change?

- Vulnerability alerts for a repository can now be managed in a separate resource
- `vulnerability_alerts` in the `github_repository` resource are marked as deprecated

### Pull request checklist

- [ ] Yes
- [x] No

----

### DIFF
--- a/github/resource_github_repository_vulnerability_alerts.go
+++ b/github/resource_github_repository_vulnerability_alerts.go
@@ -2,6 +2,7 @@ package github
 
 import (
 	"context"
+	"net/http"
 	"strconv"
 
 	"github.com/hashicorp/terraform-plugin-log/tflog"
@@ -51,6 +52,7 @@ func resourceGithubRepositoryVulnerabilityAlerts() *schema.Resource {
 }
 
 func resourceGithubRepositoryVulnerabilityAlertsCreate(ctx context.Context, d *schema.ResourceData, m any) diag.Diagnostics {
+	tflog.Info(ctx, "Creating repository vulnerability alerts", map[string]any{"id": d.Id()})
 	meta := m.(*Owner)
 	client := meta.v3client
 
@@ -87,15 +89,31 @@ func resourceGithubRepositoryVulnerabilityAlertsCreate(ctx context.Context, d *s
 }
 
 func resourceGithubRepositoryVulnerabilityAlertsRead(ctx context.Context, d *schema.ResourceData, m any) diag.Diagnostics {
+	tflog.Info(ctx, "Reading repository vulnerability alerts", map[string]any{"id": d.Id()})
 	meta := m.(*Owner)
 	client := meta.v3client
 
 	owner := meta.name // TODO: Add owner support // d.Get("owner").(string)
 	repoName := d.Get("repository").(string)
-	vulnerabilityAlertsEnabled, _, err := client.Repositories.GetVulnerabilityAlerts(ctx, owner, repoName)
+	vulnerabilityAlertsEnabled, resp, err := client.Repositories.GetVulnerabilityAlerts(ctx, owner, repoName)
 	if err != nil {
+		if resp.StatusCode == http.StatusNotFound {
+			d.SetId("")
+			return diag.Errorf("vulnerability alerts don't exist for repository %s/%s, removing resource from state", owner, repoName)
+		}
 		return diag.Errorf("error reading repository vulnerability alerts: %s", err.Error())
 	}
+	// If no error, but the response status code is 404, we need to check if the repository is accessible.
+	if resp.StatusCode == http.StatusNotFound {
+		repo, _, err := client.Repositories.Get(ctx, owner, repoName)
+		if err != nil {
+			return diag.Errorf("repository doesn't exist anymore, please remove the resource from your configuration: %s", err.Error())
+		}
+		if repo.GetArchived() {
+			return diag.Errorf("repository %s/%s is archived, please remove the resource from your configuration", owner, repoName)
+		}
+	}
+	tflog.Debug(ctx, "Setting vulnerability alerts enabled state", map[string]any{"owner": owner, "repo_name": repoName, "vulnerability_alerts_enabled": vulnerabilityAlertsEnabled, "response_status": resp.StatusCode})
 	if err = d.Set("enabled", vulnerabilityAlertsEnabled); err != nil {
 		return diag.FromErr(err)
 	}
@@ -104,6 +122,7 @@ func resourceGithubRepositoryVulnerabilityAlertsRead(ctx context.Context, d *sch
 }
 
 func resourceGithubRepositoryVulnerabilityAlertsUpdate(ctx context.Context, d *schema.ResourceData, m any) diag.Diagnostics {
+	tflog.Info(ctx, "Updating repository vulnerability alerts", map[string]any{"id": d.Id()})
 	meta := m.(*Owner)
 	client := meta.v3client
 
@@ -127,6 +146,7 @@ func resourceGithubRepositoryVulnerabilityAlertsUpdate(ctx context.Context, d *s
 }
 
 func resourceGithubRepositoryVulnerabilityAlertsDelete(ctx context.Context, d *schema.ResourceData, m any) diag.Diagnostics {
+	tflog.Info(ctx, "Deleting repository vulnerability alerts", map[string]any{"id": d.Id()})
 	meta := m.(*Owner)
 	client := meta.v3client
 
--- a/github/resource_github_repository_vulnerability_alerts_test.go
+++ b/github/resource_github_repository_vulnerability_alerts_test.go
@@ -9,7 +9,6 @@ import (
 	"github.com/hashicorp/terraform-plugin-testing/helper/acctest"
 	"github.com/hashicorp/terraform-plugin-testing/helper/resource"
 	"github.com/hashicorp/terraform-plugin-testing/knownvalue"
-	"github.com/hashicorp/terraform-plugin-testing/plancheck"
 	"github.com/hashicorp/terraform-plugin-testing/statecheck"
 	"github.com/hashicorp/terraform-plugin-testing/tfjsonpath"
 )
@@ -191,7 +190,7 @@ func TestAccGithubRepositoryVulnerabilityAlerts(t *testing.T) {
 		})
 	})
 
-	t.Run("destroys_from_archived_repository_without_error", func(t *testing.T) {
+	t.Run("errors_when_reading_archived_repository", func(t *testing.T) {
 		randomID := acctest.RandString(5)
 		repoName := fmt.Sprintf("%svuln-alerts-%s", testResourcePrefix, randomID)
 
@@ -209,15 +208,6 @@ func TestAccGithubRepositoryVulnerabilityAlerts(t *testing.T) {
 			}
 		`
 
-		archivedOnlyConfig := fmt.Sprintf(`
-			resource "github_repository" "test" {
-				name       = "%s"
-				visibility = "private"
-				auto_init  = true
-				archived   = true
-			}
-		`, repoName)
-
 		resource.Test(t, resource.TestCase{
 			PreCheck:          func() { skipUnauthenticated(t) },
 			ProviderFactories: providerFactories,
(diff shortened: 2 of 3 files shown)


github/github_repository_vulnerability_alerts.go
```diff
@@ -0,0 +1,146 @@
...
+		CreateContext: resourceGithubRepositoryVulnerabilityAlertsCreate,
+		ReadContext:   resourceGithubRepositoryVulnerabilityAlertsRead,
+		UpdateContext: resourceGithubRepositoryVulnerabilityAlertsUpdate,
+		DeleteContext: resourceGithubRepositoryVulnerabilityAlertsDelete,
+		Importer: &schema.ResourceImporter{
+			StateContext: resourceGithubRepositoryVulnerabilityAlertsImport,
+		},
+
+		Schema: map[string]*schema.Schema{
+			"repository_name": {
```
stevehipwell: Could we add the `repository_id` pattern here with the custom diff logic?

github/github_repository_vulnerability_alerts.go
```diff
@@ -0,0 +1,146 @@
...
+				Type:        schema.TypeString,
+				Required:    true,
+				ForceNew:    true,
+				Description: "The repository name to configure vulnerability alerts for.",
+			},
+			"repository_owner": {
```
stevehipwell: This is different to the pattern for pages, let's come up with an idiomatic pattern given that there are currently multiple patterns.

For the ability to re-use logic I'd suggest that we go with `owner` & `repository` as the idiomatic pattern?

github/github_repository_vulnerability_alerts.go
```diff
@@ -0,0 +1,146 @@
...
+			return diag.FromErr(err)
+		}
+	} else {
+		_, err := client.Repositories.DisableVulnerabilityAlerts(ctx, owner, repoName)
+		if err != nil {
+			return diag.FromErr(err)
+		}
+	}
+
+	id, err := buildID(owner, repoName)
```
stevehipwell: Let's just use the repo ID here.

stevehipwell (changes requested):
This is looking good.

github/resource_github_repository_vulnerability_alerts.go
```diff
@@ -0,0 +1,175 @@
...
+				Type:        schema.TypeInt,
+				Computed:    true,
+				Description: "The ID of the repository to configure vulnerability alerts for.",
+			},
+			"owner": {
+				Type:        schema.TypeString,
+				Required:    true,
+				ForceNew:    true,
+				Description: "The owner of the repository to configure vulnerability alerts for.",
+			},
```
stevehipwell:
```suggestion
			// "owner": { // TODO: Add owner support
			// 	Type:        schema.TypeString,
			// 	Required:    true,
			// 	ForceNew:    true,
			// 	Description: "The owner of the repository to configure vulnerability alerts for.",
			// },
```
I'm not 100% sure about adding this in now? The logic is complicated (for an app owner can only be the installation org, for a user it can be any org thay have access to) and it's not consistent. I'd be happier with an issue tracking the resources to add as that way we can make sure the client is well positioned to support it. Longer term I'd like to remove the single app installation limitation as part of the pother client work, which would make owner much more useful.

Could we leave the code in but commented out with a TODO, then as part of implementing owner for a group of resources (repository in this case) we could have the implementations ready to go.

github/resource_github_repository_vulnerability_alerts.go
```diff
@@ -0,0 +1,175 @@
...
+
+		CustomizeDiff: diffRepository,
+	}
+}
+
+func resourceGithubRepositoryVulnerabilityAlertsCreate(ctx context.Context, d *schema.ResourceData, m any) diag.Diagnostics {
+	meta := m.(*Owner)
+	client := meta.v3client
+
+	owner := d.Get("owner").(string)
```
stevehipwell:
```suggestion
	owner := meta.name // TODO: Add owner support d.Get("owner").(string)
```

github/resource_github_repository_vulnerability_alerts.go
```diff
@@ -0,0 +1,175 @@
...
+	d.SetId(id)
+
+	return nil
+}
+
+func resourceGithubRepositoryVulnerabilityAlertsRead(ctx context.Context, d *schema.ResourceData, m any) diag.Diagnostics {
+	meta := m.(*Owner)
+	client := meta.v3client
+
+	owner := d.Get("owner").(string)
```
stevehipwell:
```suggestion
	owner := meta.name // TODO: Add owner support d.Get("owner").(string)
```

github/resource_github_repository_vulnerability_alerts.go
```diff
@@ -0,0 +1,175 @@
...
+	}
+
+	return nil
+}
+
+func resourceGithubRepositoryVulnerabilityAlertsUpdate(ctx context.Context, d *schema.ResourceData, m any) diag.Diagnostics {
+	meta := m.(*Owner)
+	client := meta.v3client
+
+	owner := d.Get("owner").(string)
```
stevehipwell:
```suggestion
	owner := meta.name // TODO: Add owner support d.Get("owner").(string)
```

github/resource_github_repository_vulnerability_alerts.go
```diff
@@ -0,0 +1,175 @@
...
+	}
+
+	return nil
+}
+
+func resourceGithubRepositoryVulnerabilityAlertsDelete(ctx context.Context, d *schema.ResourceData, m any) diag.Diagnostics {
+	meta := m.(*Owner)
+	client := meta.v3client
+
+	owner := d.Get("owner").(string)
```
stevehipwell:
```suggestion
	owner := meta.name // TODO: Add owner support d.Get("owner").(string)
```

github/resource_github_repository_vulnerability_alerts.go
```diff
@@ -0,0 +1,175 @@
...
+
+func resourceGithubRepositoryVulnerabilityAlertsImport(ctx context.Context, d *schema.ResourceData, m any) ([]*schema.ResourceData, error) {
+	tflog.Debug(ctx, "Importing repository vulnerability alerts", map[string]any{"id": d.Id()})
+	repoOwner, repoName, err := parseID2(d.Id())
+	if err != nil {
+		return nil, fmt.Errorf("invalid ID specified: supplied ID must be written as <repository_owner>:<repository_name>. Original error: %w", err)
+	}
+	if err := d.Set("owner", repoOwner); err != nil {
+		return nil, err
+	}
```
stevehipwell:
```suggestion
	// if err := d.Set("owner", repoOwner); err != nil { // TODO: Add owner support 
	// 	return nil, err
	// }
```

github/resource_github_repository_vulnerability_alerts.go
```diff
@@ -0,0 +1,175 @@
...
+	if err = d.Set("repository_id", repo.GetID()); err != nil {
+		return diag.FromErr(err)
+	}
+
+	id, err := buildID(owner, strconv.Itoa(int(repo.GetID())))
+	if err != nil {
+		return diag.FromErr(err)
+	}
+	d.SetId(id)
```
stevehipwell: ID is globally unique, no need to add owner.

github/resource_github_repository_vulnerability_alerts.go
```diff
@@ -0,0 +1,175 @@
...
+	if err != nil {
+		return diag.FromErr(handleArchivedRepoDelete(err, "repository vulnerability alerts", d.Id(), owner, repoName))
+	}
+
+	return nil
+}
+
+func resourceGithubRepositoryVulnerabilityAlertsImport(ctx context.Context, d *schema.ResourceData, m any) ([]*schema.ResourceData, error) {
+	tflog.Debug(ctx, "Importing repository vulnerability alerts", map[string]any{"id": d.Id()})
+	repoOwner, repoName, err := parseID2(d.Id())
```
stevehipwell: We only need to use the repo ID here as we can use the `GetByID()` function that will also return the owner which we can check against `meta.name` until we support defining the owner.

deiga: Yes, but I don't like asking folks to use repo ID as the import ID, since it's not findable in the GH UI AFAIK

stevehipwell: Isn't it the pattern currently used in the repository resource?

deiga: In repository we use only `repoName`

stevehipwell:
@deiga could you please rebase?

github/resource_github_repository_vulnerability_alerts.go
```diff
@@ -0,0 +1,167 @@
...
```
stevehipwell: Please make this just the repo name, same as for the new pages resource.

github/resource_github_repository_vulnerability_alerts.go
```diff
@@ -0,0 +1,180 @@
...
+
+func resourceGithubRepositoryVulnerabilityAlertsImport(ctx context.Context, d *schema.ResourceData, m any) ([]*schema.ResourceData, error) {
+	tflog.Debug(ctx, "Importing repository vulnerability alerts", map[string]any{"id": d.Id()})
+	repoName := d.Id()
+	// if err := d.Set("owner", repoOwner); err != nil { // TODO: Add owner support
+	// 	return nil, err
+	// }
+	if err := d.Set("repository", repoName); err != nil {
+		return nil, err
+	}
```
stevehipwell, line 171: By convention, this should be below `d.SetId()`.

github/resource_github_repository_vulnerability_alerts.go
```diff
@@ -0,0 +1,180 @@
...
+	owner := meta.name
+	client := meta.v3client
+
+	repo, _, err := client.Repositories.Get(ctx, owner, repoName)
+	if err != nil {
+		return nil, err
+	}
+	if err = d.Set("repository_id", repo.GetID()); err != nil {
+		return nil, err
+	}
```
stevehipwell: By convention, this should be below `d.SetId()`.

github/resource_github_repository_vulnerability_alerts.go
```diff
@@ -0,0 +1,181 @@
...
+
+	d.SetId(strconv.Itoa(int(repo.GetID())))
+
+	if err = d.Set("repository_id", repo.GetID()); err != nil {
+		return nil, err
+	}
+	return []*schema.ResourceData{d}, nil
+}
+
+func handleVulnerabilityAlertsErrorOnArchivedRepository(err error, resp *github.Response, repoName string) error {
```
stevehipwell: If we want to do this, would it not make more sense to use the client to fetch the repo to see if it is archived instead of string matching an error message?

deiga: I guess it would. I just didn't want to add another API call 😬  But that would be better than trusting a weird error message.

And the error was weird enough that I think it's better we offer something tangible to the user as there is no documentation about a 422 in the API docs 😬

deiga: Yeah, that got a lot cleaner. Thanks for pointing that out!

github/resource_github_repository_vulnerability_alerts.go
```diff
@@ -0,0 +1,168 @@
...
+func resourceGithubRepositoryVulnerabilityAlertsRead(ctx context.Context, d *schema.ResourceData, m any) diag.Diagnostics {
+	meta := m.(*Owner)
+	client := meta.v3client
+
+	owner := meta.name // TODO: Add owner support // d.Get("owner").(string)
+	repoName := d.Get("repository").(string)
+	vulnerabilityAlertsEnabled, _, err := client.Repositories.GetVulnerabilityAlerts(ctx, owner, repoName)
+	if err != nil {
+		return diag.Errorf("error reading repository vulnerability alerts: %s", err.Error())
+	}
```
robert-crandall, line 105: If the repository is deleted outside Terraform, `GetVulnerabilityAlerts` will return a 404. The Read function does not check for this and will return an error instead of calling `d.SetId("")` to remove the resource from state. This breaks `terraform plan` and `terraform apply` when the backing repository no longer exists.


```suggestion
    if err != nil {
        if resp != nil && resp.StatusCode == http.StatusNotFound {
            tflog.Warn(ctx, "Repository not found, removing from state", map[string]any{"repository": repoName})
            d.SetId("")
            return nil
        }
        return diag.Errorf("error reading repository vulnerability alerts: %s", err.Error())
    }
```

stevehipwell (approved):
LGTM

PR #3168, opened 2026-02-08
[FEAT] Add `github_repository_pages` Resource and Data Source

### LINKED ISSUE #1045: feature request: Add support for 'public' attribute on github pages configuration

### LINKED ISSUE #2087: [BUG]: Updating GitHub Pages Fails in GHES
### Expected Behavior

Hello,

I have been encountering an issue making modifications to the `pages` block of `github_repository` resources in GitHub Enterprise Server (3.10.3).

I am able to initially create a `github_repository` resource with a `pages` block, such as this:

```terraform
resource "github_repository" "main" {
  name = "terraform-provider-github-pages-test"

  pages {
    build_type = "legacy"
    source {
      branch = "main"
      path   = "/"
    }
  }
}
```

I expect that changing an argument, such as `build_type`, would result in the GitHub repository's configuration successfully being modified in GHES. An example valid, new Terraform configuration could be

```terraform
resource "github_repository" "main" {
  name = "terraform-provider-github-pages-test"

  pages {
    build_type = "workflow"
    source {
      branch = "main"
      path   = "/"
    }
  }
}
```

### Actual Behavior

The Terraform provider attempts to update the GitHub Pages configuration, but fails with a 400 error:

```
│ Error: PUT https://<base_url>/api/v3/repos/SStrullmyer/terraform-provider-github-pages-test/pages: 400 Custom domains are not available for GitHub Pages []
│
│   with github_repository.main,
│   on main.tf line 1, in resource "github_repository" "main":
│    1: resource "github_repository" "main" {
```

This results in the desired configuration not being effected. This is doubly problematic, as the tfstate associated with this block appears to be prematurely updated, anticipating a successful PUT; coupled with #1913, this can lead to additional challenges.

I've tested calling the `/api/v3/repos/.../pages` endpoint directly (outside of Terraform contexts), and can confirm that a PUT request that includes any `cname` entry, including empty strings and `null`, results in the same 400 error message.

I _suspect_ all that would be required here would be to not provide the cname argument to the `*/pages` endpoint if the `GITHUB_BASE_URL` indicates the target is a GHES instance. If this path (or another) is amenable, I'd be happy to pursue a PR for it.

### Terraform Version

Terraform v1.6.6 on windows_amd64 + provider registry.terraform.io/integrations/github v5.43.0

### Affected Resource(s)

- github_repository

```hcl
resource "github_repository" "main" {
  name = "terraform-provider-github-pages-test"

  pages {
    build_type = "legacy"
    source {
      branch = "main"
      path   = "/"
    }
  }
}
```

### Steps to Reproduce

- Run `terraform apply` on the above Terraform configuration file targeting a GitHub Enterprise Server instance via GITHUB_BASE_URL (or equivalent provider configuration)
- Change `build_type` to `"workflow"`
- Run `terraform apply`

### Debug Output

```shell
[DEBUG] provider.terraform-provider-github_v5.43.0.exe: PUT /api/v3/repos/SStrullmyer/terraform-provider-github-pages-test/pages HTTP/1.1
[DEBUG] provider.terraform-provider-github_v5.43.0.exe: User-Agent: go-github/v55.0.0
[DEBUG] provider.terraform-provider-github_v5.43.0.exe: Content-Length: 39
[DEBUG] provider.terraform-provider-github_v5.43.0.exe: Accept: application/vnd.github.v3+json,application/vnd.github.stone-crop-preview+json
[DEBUG] provider.terraform-provider-github_v5.43.0.exe: Content-Type: application/json
[DEBUG] provider.terraform-provider-github_v5.43.0.exe: X-Github-Api-Version: 2022-11-28
[DEBUG] provider.terraform-provider-github_v5.43.0.exe: Accept-Encoding: gzip
[DEBUG] provider.terraform-provider-github_v5.43.0.exe
[DEBUG] provider.terraform-provider-github_v5.43.0.exe: {
[DEBUG] provider.terraform-provider-github_v5.43.0.exe:  "cname": null,
[DEBUG] provider.terraform-provider-github_v5.43.0.exe:  "build_type": "workflow"
[DEBUG] provider.terraform-provider-github_v5.43.0.exe: }
```

### Panic Output

_No response_

### Code of Conduct

### LINKED ISSUE #2335: [BUG]: `github_repository` gets 403 error on Pages API even though there's no page
### Expected Behavior

Terraform plan to succeed.

### Actual Behavior

```tf
  # github_repository.this["mfe-school"] will be updated in-place
  ~ resource "github_repository" "this" {
        id                          = "my-repo"
        name                        = "my-repo"
        # (36 unchanged attributes hidden)

      + template {
          + include_all_branches = false
          + owner                = "foobar"
          + repository           = "repo-template"
        }
    }

Plan: 0 to add, 4 to change, 0 to destroy.

Error: GET https://api.github.com/repos/foobar/repo-template/pages: 403 Resource not accessible by integration []

  with github_repository.this["repo-template"],
  on config.tf line 7, in resource "github_repository" "this":
   7: resource "github_repository" "this" {
```

### Terraform Version

1.9.3

### Affected Resource(s)

- `github_repository`

_No response_

### Steps to Reproduce

_No response_

### Debug Output

_No response_

### Panic Output

_No response_

### Code of Conduct

### LINKED ISSUE #2653: [FEAT]: Support the https_enforced setting for GitHub Pages
### Describe the need

Reopening: https://github.com/integrations/terraform-provider-github/issues/2276

Allow setting https_enforced when setting up GitHub Pages. It would be a nice addition to cname and build_type that are already possible.

### SDK Version

_No response_

### API Version

_No response_

### Relevant log output

```shell

```

### Code of Conduct

### LINKED ISSUE #2671: [BUG]: github_repository: Pages is enabled before gh-pages branch is created – API returns 422
### Expected Behavior

A single terraform apply should
	1.	create the repository,
	2.	create the gh-pages branch,
	3.	enable GitHub Pages that serves from that branch, and
	4.	exit successfully (plan: 2 to add, 0 to change, 0 to destroy).

### Actual Behavior

During the same apply the provider sends the POST /repos/{owner}/{repo}/pages request before the gh-pages branch exists, so GitHub responds

```422 Validation Failed – The gh-pages branch must exist before GitHub Pages can be built.```

### Terraform Version

```
Terraform v1.5.0
on linux_amd64
 provider registry.terraform.io/integrations/github v6.5.2

### Affected Resource(s)

 ```
Affected Resource(s)
	•	github_repository
	•	github_branch

```hcl
The smallest reproducible case

terraform {
  required_providers {
    github = {
      version = "~> 6.5"
    }
  }
}

provider "github" {
  owner = var.github_owner          # e.g. "my-org"
  token = var.github_token          # PAT with repo + pages:write
}

# --- repository ------------------------------------------------------
resource "github_repository" "pages_site" {
  name        = "pages-test-single-apply"
  visibility  = "public"
  auto_init   = true                 # seeds the default branch (main)

  pages {                            # <-- fails because branch not yet there
    source {
      branch = "gh-pages"
      path   = "/"
    }
  }

  delete_branch_on_merge = true
}

# --- branch ----------------------------------------------------------
resource "github_branch" "gh_pages" {
  repository    = github_repository.pages_site.name
  branch        = "gh-pages"
  source_branch = github_repository.pages_site.default_branch
}
```

### Steps to Reproduce

Steps to Reproduce
	1.	terraform init
	2.	terraform apply → observe the 422 error above.

### Debug Output

```shell
github_repository.pages_site: Creating...
│ Error: POST https://api.github.com/repos/<org>/pages-test-single-apply/pages: 422 The gh-pages branch must exist before GitHub Pages can be built. []
│
│   with github_repository.pages_site,
│   on main.tf line 18, in resource "github_repository" "pages_site":
│   18: resource "github_repository" "pages_site" {
```

### Panic Output

```shell

```

### Code of Conduct

### LINKED ISSUE #3142: [BUG]: 403 failure results in mismatched state
### Expected Behavior

Expected GitHub pages to be appropriately applied per the terraform plan. 

Image

### Actual Behavior

On an initial deployment for a GitHub Page, the terraform apply failed due to our GitHub app not having the appropriate permissions. 

Image

After fixing the permissions and rerunning the same workflow, the terraform plan no longer found any changes that needed to be applied. The screenshot above shows the original plan, and the second there were no changes for the github_repository.module. It's as if the first run somehow modified the state to show that a page was actually created, even though it failed. Visiting the GitHub repo settings showed that a page had not been created.

Even after manually creating the page, the terraform plan was still not picking up any changes. 

It seems like there's some motivation to refactor pages into its own resource on #3133, #2087, and #1913.

### Terraform Version

>= 1.7

### Affected Resource(s)

- github_repository.module.pages

```hcl

```

### Steps to Reproduce

_No response_

### Debug Output

```shell

```

### Panic Output

```shell

```

### Code of Conduct

### LINKED ISSUE #3167: [FEAT]: Add new resource for `github_repository_pages`
### Describe the need

We want to extract the management of repository pages out of the `github_repository` resource

### SDK Version

_No response_

### API Version

_No response_

### Relevant log output

```shell

```

### Code of Conduct

deiga:
Resolves #3167
Resolves #2671 
Resolves #3142
Resolves #1045 
Resolves #2653
Resolves #2335 
Resolves #2087 

----

### Before the change?

- A repository's Pages resource was only manageable through `github_repository`

### After the change?

- There is now a `github_repository_pages` resource to manage the pages
-  There is now a `github_repository_pages` data source to fetch information about the pages
- The pages fields are now deprecated in `github_repository` Resource and Data Source
- It's possible to configure the `public` field of `github_repository_pages`
- It's possible to configure the `https_enforced` field of `github_repository_pages`

### Pull request checklist

- [ ] Yes
- [x] No

----


github/data_source_github_repository_pages_test.go
```diff
@@ -0,0 +1,62 @@
...
+			}
+		`, repoName, baseRepoVisibility, testAccConf.owner, testAccConf.owner)
+
+		resource.Test(t, resource.TestCase{
+			PreCheck:          func() { skipUnauthenticated(t) },
+			ProviderFactories: providerFactories,
+			Steps: []resource.TestStep{
+				{
+					Config: config,
+					Check: resource.ComposeTestCheckFunc(
```
stevehipwell: You probably want to make use of the new `ConfigStateChecks`.

github/resource_github_repository_pages.go
```diff
@@ -0,0 +1,331 @@
...
+		CreateContext: resourceGithubRepositoryPagesCreate,
+		ReadContext:   resourceGithubRepositoryPagesRead,
+		UpdateContext: resourceGithubRepositoryPagesUpdate,
+		DeleteContext: resourceGithubRepositoryPagesDelete,
+		Importer: &schema.ResourceImporter{
+			StateContext: resourceGithubRepositoryPagesImport,
+		},
+
+		Schema: map[string]*schema.Schema{
+			"repository_name": {
```
stevehipwell: Can we add `repository_id` support with the repo diff logic?

github/resource_github_repository_pages.go
```diff
@@ -0,0 +1,331 @@
...
+				Type:        schema.TypeString,
+				Required:    true,
+				ForceNew:    true,
+				Description: "The repository name to configure GitHub Pages for.",
+			},
+			"owner": {
```
stevehipwell: Do we support `owner` in other similar resources?

github/resource_github_repository_pages_test.go
```diff
@@ -0,0 +1,178 @@
...
+			}
+		`, repoName, baseRepoVisibility, testAccConf.owner)
+
+		resource.Test(t, resource.TestCase{
+			PreCheck:          func() { skipUnauthenticated(t) },
+			ProviderFactories: providerFactories,
+			Steps: []resource.TestStep{
+				{
+					Config: config,
+					Check: resource.ComposeTestCheckFunc(
```
stevehipwell: As per above.

stevehipwell (changes requested):
This is looking really good.

github/data_source_github_repository_pages.go
```diff
@@ -0,0 +1,138 @@
...
+			return diag.Errorf("GitHub Pages not found for repository %s/%s", owner, repoName)
+		}
+		return diag.Errorf("error reading repository pages: %s", err.Error())
+	}
+
+	id, err := buildID(owner, repoName)
+	if err != nil {
+		return diag.FromErr(err)
+	}
+	d.SetId(id)
```
stevehipwell: This can just be the repository ID.

github/resource_github_repository_pages.go
```diff
@@ -0,0 +1,331 @@
...
+	pages, _, err := client.Repositories.EnablePages(ctx, owner, repoName, pages)
+	if err != nil {
+		return diag.FromErr(err)
+	}
+
+	id, err := buildID(owner, repoName)
+	if err != nil {
+		return diag.FromErr(err)
+	}
+	d.SetId(id)
```
stevehipwell: This can just be the repository ID.

github/resource_github_repository_pages.go
deiga: We don't, but IMO we might as well start supporting it. I think it fits nicely with the thought of having more explicit configs. 
And it has been a requested feature to be able to target other `owner` than the one set in provider config

stevehipwell: See vulnerability alerts PR.

github/data_source_github_repository_pages.go
```diff
@@ -0,0 +1,140 @@
...
+			"repository": {
+				Type:        schema.TypeString,
+				Required:    true,
+				Description: "The repository name to get GitHub Pages information for.",
+			},
+			"owner": {
+				Type:        schema.TypeString,
+				Required:    true,
+				Description: "The owner of the repository.",
+			},
```
stevehipwell: I think this also needs commenting out?

github/resource_github_repository_pages.go
```diff
@@ -0,0 +1,357 @@
...
+	_, err := client.Repositories.DisablePages(ctx, owner, repoName)
+	if err != nil {
+		return diag.FromErr(handleArchivedRepoDelete(err, "repository pages", d.Id(), owner, repoName))
+	}
+
+	return nil
+}
+
+func resourceGithubRepositoryPagesImport(ctx context.Context, d *schema.ResourceData, m any) ([]*schema.ResourceData, error) {
+	owner, repoName, err := parseID2(d.Id())
```
stevehipwell: We only need the repo ID as the owner is in the provider config.

stevehipwell (changes requested):

healiha:
Would it be possible to add `visibility` to `github_repository_pages`?
image

The documentation doesn't seem to cover this, but after testing, a `PUT` request to `https://api.github.com/repos/<org>/<repo>/pages` with `{"public": true/false}` works.

deiga:
@healiha I'll check that out

stevehipwell (dismissed):
LGTM

stevehipwell:

github/resource_github_repository_pages.go
```diff
@@ -0,0 +1,424 @@
...
+			return diag.FromErr(err)
+		}
+	}
+
+	if shouldUpdatePage {
+		_, err = client.Repositories.UpdatePages(ctx, owner, repoName, update)
+		if err != nil {
+			return diag.FromErr(err)
+		}
+	}
```
stevehipwell:
```suggestion
	// Check if we have values set that can't be configured as part of the create logic
	cname, cnameOK := d.GetOk("cname")
	public := d.get("public").(bool)
	httpsEnforced := d.Get("https_enforced").(bool)
	
	if cnameOK || public || httpsEnforced {
		update := &github.PagesUpdate{}
		
		if cnameOK {
			update.CNAME = github.Ptr(cname.(string))		
		}
		
		if public {
			update.Public = github.Ptr(public)
		}
		
		if httpsEnforced {
			update.HTTPSEnforced = github.Ptr(httpsEnforced)
		}
		
		_, err = client.Repositories.UpdatePages(ctx, owner, repoName, update)
		if err != nil {
			return diag.FromErr(err)
		}
	}
```
The logic for this can be significantly simplified as we only care about `true` values and this code is only run for the create.

github/resource_github_repository_pages.go
```diff
@@ -0,0 +1,424 @@
...
+func resourceGithubRepositoryPagesUpdate(ctx context.Context, d *schema.ResourceData, m any) diag.Diagnostics {
+	meta := m.(*Owner)
+	client := meta.v3client
+
+	owner := meta.name // TODO: Add owner support // d.Get("owner").(string)
+	repoName := d.Get("repository").(string)
+
+	update := &github.PagesUpdate{}
+
+	if d.HasChange("cname") {
```
stevehipwell, line 320: We generally don't use this pattern, I'd expect to see `d.get()` & `d.GetOk()` here.

github/resource_github_repository_pages.go
```diff
@@ -0,0 +1,424 @@
...
+	repoName := d.Id()
+	if strings.Contains(repoName, " ") {
+		return nil, fmt.Errorf("invalid ID specified: supplied ID must be the slug of the repository name")
+	}
+	// if err := d.Set("owner", owner); err != nil { // TODO: Add owner support
+	// 	return nil, err
+	// }
+	if err := d.Set("repository", repoName); err != nil {
+		return nil, err
+	}
```
stevehipwell: Idiomatically thee want calling after we've made the remote call and set the ID.

github/resource_github_repository_pages.go
```diff
@@ -0,0 +1,424 @@
...
+	return []*schema.ResourceData{d}, nil
+}
+
+func resourceGithubRepositoryPagesDiff(ctx context.Context, d *schema.ResourceDiff, _ any) error {
+	if d.Id() == "" {
+		return nil
+	}
+
+	buildType := d.Get("build_type").(string)
+	switch buildType {
```
stevehipwell: Wouldn't a boolean `if` work better here?

stevehipwell (changes requested):

github/resource_github_repository_pages.go
deiga: Thanks! We do care are about the `false` value for `public` too, since a non-EMU GHEC can set either. 
But I'll adopt your suggestion with that modification!

stevehipwell: What I mean is that the unset value matches the default value. We only need to use the hacky method if we care about differentiating between `false` and unset.

In update, I'd expect the value to be set as we usually do rather than checking for a diff. Unless there is a specific reason not to?

github/resource_github_repository_pages.go
deiga: I'm not sure I understand what you mean. `d.HasChange` has been used ~36 times in the codebase. 
`d.HasChange()` tells us if there is a value change in the plan, whereas `d.Get()` and `d.GetOk()` would cause us to send fields even if they are not changed

stevehipwell: I don't think 36 usages, some of which are definitely in diff functions, comes close to the occurrences of the get functions. The only reason you're in the update function is that there are changes to apply, so building the struct using the get functions (`d.Get()` for bools & `d.GetOk()` for strings where that aren't always set) is going to be the most efficient and readable pattern.

github/resource_github_repository_pages.go
deiga: I restructured it, but had to re-add the `d.Set` calls as otherwise the computed values won't be set

github/resource_github_repository_pages.go
deiga: Since the API supports partial updates, I don't fully understand why we wouldn't want to utilize that.

For example these `Optional` & `Computed` fields are things that I wouldn't want to send unless they are actually changed in the plan and have been modified by the user. 
Though I guess `d.GetOk()` isn't able to determine the latter once they have been set by `Create` or `Read`?

Could you elaborate the issue you're seeing with using `d.HasChange()`?

stevehipwell: The API supports partial updates, but Terraform doesn't. The idiomatic pattern in this provider is to set all of the fields being managed by Terraform before making a request; this makes the code as readable as possible and makes it less likely that we are impacted by edge case issues (our code path is as small and non-branching as possible). By using `d.hasChanges()` you're adding unnecessary complexity.

github/resource_github_repository_pages.go
```diff
@@ -0,0 +1,432 @@
...
+			},
+			"api_url": {
+				Type:        schema.TypeString,
+				Computed:    true,
+				Description: "The API URL of the GitHub Pages resource.",
+			},
+			"public": {
+				Type:        schema.TypeBool,
+				Optional:    true,
+				Computed:    true,
```
stevehipwell, line 104:
```suggestion
				Default:    false,
```

github/resource_github_repository_pages.go
```diff
@@ -0,0 +1,432 @@
...
+				Description: "Whether the GitHub Pages site is publicly visible. If set to `true`, the site is accessible to anyone on the internet. If set to `false`, the site will only be accessible to users who have at least `read` access to the repository that published the site.",
+			},
+			"https_enforced": {
+				Type:         schema.TypeBool,
+				Optional:     true,
+				Computed:     true,
```
stevehipwell, line 110:
```suggestion
				Default:     false,
```

github/resource_github_repository_pages.go
```diff
@@ -0,0 +1,432 @@
...
+	if !publicOKExists {
+		if err := d.Set("public", pages.GetPublic()); err != nil {
+			return diag.FromErr(err)
+		}
+	}
+	if !httpsEnforcedExists {
+		if err := d.Set("https_enforced", pages.GetHTTPSEnforced()); err != nil {
+			return diag.FromErr(err)
+		}
+	}
```
stevehipwell:
```suggestion
	// Check if we have values set that can't be configured as part of the create logic
	cname, cnameOK := d.GetOk("cname")
	public := d.Get("public").(bool)
	httpsEnforced := d.Get("https_enforced").(bool)

	if cnameOK || public || httpsEnforced {
		tflog.Debug(ctx, "values set that require an update after create", map[string]any{
			"public":                public,
			"https_enforced":        httpsEnforced,
			"cname":                 cname,
		})
		
		update := &github.PagesUpdate{}

		if cnameOK {
			update.CNAME = github.Ptr(cname.(string))
		}

		if public {
			update.Public = github.Ptr(public)
		}

		if httpsEnforcedExists {
			update.HTTPSEnforced = github.Ptr(httpsEnforced)
		}

		_, err = client.Repositories.UpdatePages(ctx, owner, repoName, update)
		if err != nil {
			return diag.FromErr(err)
		}
	}
```
We only care about the non-default values so we can short circuit the behaviour in the create function. Note the boolean values now have defaults so they don't need setting.

stevehipwell (changes requested):

github/resource_github_repository_pages.go
deiga: I don't think it's a good idea to set a default, since the API default is different depending on which type of Owner you have. Only GHEC (and GHES?) are able to have private pages AFAIK, so for most users we'd be forcing them to configure this explicitly. We could default to true, but GHEC with EMU can't even have public pages and it'd be annoying for every other Enterprise owner to have to remember to set it to `false`.

What is the issue with using `Computed` and trust the API default?

stevehipwell: The logic is simpler with default for booleans. If the behaviour is dynamic we can keep it computed and but we then need to use `d.GetOkExists()` and know if we're in the create or update phase. The enforce https isn't dynamic so that should keep a default for simplicity.

github/resource_github_repository_pages.go
deiga: I'm not sure this should have a default value, as it's default value isn't straighforward IMO. In GHEC with EMU the API set's that value as `true` by default and apparently so does a Free Org as well.

So we could maybe set the default as `true`

stevehipwell: Yeah, sorry my suggestion was incorrect I meant to make it `true`.

github/resource_github_repository_pages.go
deiga: See above why I think default values for those booleans are problematic

github/resource_github_repository_pages.go
deiga: I'm not sure what you mean with "Terraform doesn't support partial updates" as that is exactly what it supports.

But I do understand your point about less conditionals == less complexity

deiga: After removing them I encounter that at least `cname` requires special handling, because we don't want to send the null value we get from the SDK ever. 
Do you think it would be better to us `if d.HasChanges("cname")` there or something like `if cname != ""`?

deiga: Actually, that turned out not be relevant. I already check `if cname != ""` I think this stems from sending the  same value for `https_enforced` which the API returns 😬 

I think it's not allowed to send `https_enforced = true` if `cname = null` . Although that is the default API response when creating a new page 😬

github/resource_github_repository_pages.go
deiga: I'll verify, but I think it's not possible to send https_enforced = true with no name set 😬

github/resource_github_repository_pages.go
stevehipwell: OK, configure it for computed and then I'll re-review (need to get my head in the game for these bools).

github/resource_github_repository_pages.go
deiga: Yeah, I added the test `errors_when_https_enforced_without_cname` which failed without `cname` set. Solved with `RequiredWith`, but it shows that we can't be sending it, unless it's been explicitly configured or changed

github/resource_github_repository_pages.go
deiga: Unfortunately not the case for `https_enforced` 😬 
https://github.com/integrations/terraform-provider-github/pull/3168#discussion_r2879890422

github/resource_github_repository_pages.go
robert-crandall: If this value is cleared instead of changed, will the GitHub resource be updated? Would using `update.CNAME = &cname ` handle this?

github/resource_github_repository_pages.go
```diff
@@ -0,0 +1,432 @@
...
+		return nil, err
+	}
+
+	return []*schema.ResourceData{d}, nil
+}
+
+func resourceGithubRepositoryPagesDiff(ctx context.Context, d *schema.ResourceDiff, _ any) error {
+	if d.Id() == "" {
+		return nil
+	}
```
robert-crandall: Copilot is saying this should be removed:

The `if d.Id() == ""` early return skips all validation for new resources. This means a config with `build_type = "legacy"` and no `source` block passes planning, and `expandPagesForCreate` silently defaults to `branch = "main"` without writing it back to state.

On every subsequent plan the validation runs (resource now has an ID), sees `legacy` + no `source` in state, and errors permanently — leaving the resource stuck until state is manually edited.

Suggest removing the early return so the validation fires on create too.

github/resource_github_repository_pages.go
stevehipwell: As my comment above, please remove these checks and just set the object directly. There is no benefit to only sending a partial update here.

github/resource_github_repository_pages.go
```diff
@@ -0,0 +1,433 @@
...
+		return diag.FromErr(err)
+	}
+	if err := d.Set("api_url", pages.GetURL()); err != nil {
+		return diag.FromErr(err)
+	}
+
+	// Check if we have values set that can't be configured as part of the create logic
+	cname, cnameOK := d.GetOk("cname")
+	public, publicOKExists := d.GetOkExists("public")                     //nolint:staticcheck // SA1019: d.GetOkExists is deprecated but necessary for bool fields
+	httpsEnforced, httpsEnforcedExists := d.GetOkExists("https_enforced") //nolint:staticcheck // SA1019: d.GetOkExists is deprecated but necessary for bool fields
```
stevehipwell: I don't think you need this pattern as at this point in the logic there is no difference between unset and `false`. So please can we just use `d.Get()`.

stevehipwell (changes requested):

github/resource_github_repository_pages.go
deiga: @stevehipwell I already tried that. The API returns an error if these fields are being sent if they are not supposed to be set. Sending the computed value doesn't work 🙃

github/resource_github_repository_pages.go
deiga: I'm fairly certain that I already tried that, but I'll do another try an record in comments if this structure is necessary

deiga: Yup :(
`https_enforced` can't be sent to the API unless `cname` is set. Otherwise the API will return "404 The certificate does not exist yet"

deiga:
@stevehipwell Last one! Do you still want to iterate on the code and postpone this?

eyalgal:
@deiga let's move it to 6.13.0 if possible.

stevehipwell:
@deiga @eyalgal I was planning on reviewing this a part of v6.12.0.

deiga:
@stevehipwell please do!

I guess Eyal was hoping to cut the release ASAP and waiting for this to be merged didn't sound enticing 😂

I would of course prefer to get this out quickly as well ☺️

stevehipwell:
@deiga if you can rebase, I'll get this reviewed either later on this evening or tomorrow morning (UK).

stevehipwell:
@deiga it looks like you still got a reference to go-github v84.

github/resource_github_repository_pages.go
```diff
@@ -0,0 +1,433 @@
...
+	}
+
+	// Sending the `public` value on updates will return an error if the repository doesn't have public pages enabled.
+	// Hence we make sure to only send the value if it's changed.
+	// Error: "400 Private pages is not enabled for this repository. All Pages will be public."
+	if d.HasChange("public") {
+		public, ok := d.Get("public").(bool)
+		if ok {
+			update.Public = new(public)
+		}
```
stevehipwell:
```suggestion
		if public, ok := d.Get("public").(bool); ok {
			update.Public = new(public)
		}
```
This would be consistent with our coding pattern, but `ok` is unnecessary here given that the type is defined in the schema and will always be a boolean.

github/resource_github_repository_pages.go
```diff
@@ -0,0 +1,433 @@
...
+	}
+
+	// `https_enforced` can't be sent to the API unless `cname` is set. Otherwise the API will return "404 The certificate does not exist yet".
+	if d.HasChange("https_enforced") {
+		httpsEnforced, ok := d.Get("https_enforced").(bool)
+		if ok {
+			update.HTTPSEnforced = new(httpsEnforced)
+		}
```
stevehipwell:
```suggestion
		if httpsEnforced, ok := d.Get("https_enforced").(bool); ok {
			update.HTTPSEnforced = new(httpsEnforced)
		}
```
As above.

github/resource_github_repository_pages.go
```diff
@@ -0,0 +1,433 @@
...
+		return fmt.Errorf("'source' is not supported for workflow build type")
+	}
+	if buildType == "legacy" && !ok {
+		return fmt.Errorf("'source' is required for legacy build type")
+	}
+
+	return nil
+}
+
+func expandPagesForCreate(d *schema.ResourceData) *github.Pages {
```
stevehipwell: This function is only called once, I think it would actually be easier to understand the (necessary complexity) code if this was inlined.

stevehipwell (changes requested):

deiga:
@stevehipwell last time? 😬

stevehipwell (approved):

stevehipwell:
LGTM

robert-crandall (approved):

PR #3175, opened 2026-02-09
[MAINT] Fixup `github_repository_file`

### LINKED ISSUE #3161: [BUG]: Unable to import a github_repository_file from default branch with provider 6.11.0
### Expected Behavior

Running the following, should import a file from the default branch:
```
terraform import 'github_repository_file.this' <REPO>/<FILE_PATH>
```

### Actual Behavior

Returns the follwing:
```
Error: invalid ID specified. Supplied ID must be written as <repository>/<file path>:<branch>. unexpected ID format ("<REPO>/<FILE_PATH>"); expected 2 parts separated by ":"
```

### Terraform Version

Terraform v1.14.4
on linux_amd64
+ provider registry.terraform.io/integrations/github v6.11.0

### Affected Resource(s)

- github-repository-file

```hcl
resource "github_repository_file" "this" {
  repository = <REPO>
  file       = <FILE_PATH>
  content = ""
  commit_message      = "chore: update file"
  overwrite_on_create = true
}
```

### Steps to Reproduce

```
terraform import 'github_repository_file.this' <REPO>/<FILE_PATH>
```

### Debug Output

```shell

```

### Panic Output

```shell

```

### Code of Conduct

deiga:
Resolves #3161

----

### Before the change?

- There was no test for importing `github_repository_file`
- Import format: repo/file or repo/file:branch; default branch was imported without an explicit branch segment.
- Branch field: branch was optional and not Computed; state could omit it for default-branch resources.
- State: No migration for old ID formats; repository_id was not part of the resource’s state.
- autocreate_branch: Used in Read, Update, and Delete; no deprecation notice.

### After the change?

- We have tests for importing `github_repository_file`
- Adds Computed: true to `branch`
- Adds computed `repository_id` field and enables repo renaming without resource recreation
- Computes `branch` on create or import if not specified
- Changes ID schema from `<repo>/<file>` to `<repo>:<file>:<branch>` to enable safer usage of multiple files from same repo but different branches
- Adds State migration for computing `branch`, `repository_id` and new `ID` format
- Import format: repo:file:branch (colon-separated); default branch uses repo:file: (trailing colon). Matches the internal ID format.
- autocreate_branch: Deprecated and no longer used in Read, Update, or Delete.

### Pull request checklist

- [ ] Yes
- [x] No

----

### DIFF
--- a/website/docs/r/repository_file.html.markdown
+++ b/website/docs/r/repository_file.html.markdown
@@ -80,17 +80,19 @@ The following arguments are supported:
 
 - `overwrite_on_create` - (Optional) Enable overwriting existing files. If set to `true` it will overwrite an existing file with the same name. If set to `false` it will fail if there is an existing file with the same name.
 
-- `autocreate_branch` - (Optional) Automatically create the branch if it could not be found. Defaults to false. Subsequent reads if the branch is deleted will occur from 'autocreate_branch_source_branch'.
+- `autocreate_branch` - (Optional) **Deprecated** Automatically create the branch if it could not be found. Defaults to false. Subsequent reads if the branch is deleted will occur from 'autocreate_branch_source_branch'. Use the `github_branch` resource instead.
 
-- `autocreate_branch_source_branch` - (Optional) The branch name to start from, if 'autocreate_branch' is set. Defaults to 'main'.
+- `autocreate_branch_source_branch` - (Optional) **Deprecated** The branch name to start from, if 'autocreate_branch' is set. Defaults to 'main'. Use the `github_branch` resource instead.
 
-- `autocreate_branch_source_sha` - (Optional) The commit hash to start from, if 'autocreate_branch' is set. Defaults to the tip of 'autocreate_branch_source_branch'. If provided, 'autocreate_branch_source_branch' is ignored.
+- `autocreate_branch_source_sha` - (Optional) **Deprecated** The commit hash to start from, if 'autocreate_branch' is set. Defaults to the tip of 'autocreate_branch_source_branch'. If provided, 'autocreate_branch_source_branch' is ignored. Use the `github_branch` resource instead.
 
 ## Attributes Reference
 
 The following additional attributes are exported:
 
 - `commit_sha` - The SHA of the commit that modified the file.
+
+- `repository_id` - The ID of the repository.
 
 - `sha` - The SHA blob of the file.
 


deiga:
image

github/resource_github_repository_file.go
```diff
@@ -436,10 +436,20 @@ func resourceGithubRepositoryFileImport(ctx context.Context, d *schema.ResourceD
 	owner := meta.(*Owner).name
 	repo, file := splitRepoFilePath(repoFilePath)
 
-	opts := &github.RepositoryContentGetOptions{Ref: branch}
-	if err := d.Set("branch", branch); err != nil {
+	repoInfo, _, err := client.Repositories.Get(ctx, owner, repo)
```
stevehipwell, line 446: Why do we need to know if the branch is the default if it's been passed in?

nickfloyd (dismissed):

stevehipwell (changes requested):
I missed the fact that the branch is optional, I think this should be required in the next major version. As it's optional then the import ID should allow it to be not set.

stevehipwell:
So what I'm saying is the following changes are needed to fix this resource.

- Add `Computed: true` to `branch`
- Compute branch on create if not specified
- Support importing with ID missing branch (compute on import only if missing)
- Add migration to fix defect with ID not containing branch and missing `branch` computed value

github/resource_github_repository_file.go
```diff
@@ -215,7 +234,9 @@ func resourceGithubRepositoryFileCreate(ctx context.Context, d *schema.ResourceD
 		return diag.FromErr(err)
 	}
 
-	d.SetId(fmt.Sprintf("%s/%s", repo, file))
+	branch := d.Get("branch").(string)
+
+	d.SetId(fmt.Sprintf("%s/%s:%s", repo, file, branch))
```
stevehipwell: This should use `buildID()`.

github/resource_github_repository_file.go
```diff
@@ -427,27 +448,34 @@ func autoBranchDiffSuppressFunc(k, _, _ string, d *schema.ResourceData) bool {
 }
 
 func resourceGithubRepositoryFileImport(ctx context.Context, d *schema.ResourceData, meta any) ([]*schema.ResourceData, error) {
-	repoFilePath, branch, err := parseID2(d.Id())
-	if err != nil {
-		return nil, fmt.Errorf("invalid ID specified. Supplied ID must be written as <repository>/<file path>:<branch>. %w", err)
+	importIDParts := strings.Split(d.Id(), idSeparator)
```
stevehipwell: This should use `parseID2()` and the ID spec should be `<repo>/<file>:` for the default branch.

deiga:
@stevehipwell Done. But I wonder if I should just refactor the import ID now as well? Sure, it's technically a breaking change, but imports are "one-off" usually

stevehipwell:
> @stevehipwell Done. But I wonder if I should just refactor the import ID now as well? Sure, it's technically a breaking change, but imports are "one-off" usually

@deiga I think this makes sense. @nickfloyd thoughts?

We should also add `repository_id` to the schema as part of the migration, add the `diffRepository`, and remove `ForceNew` from `repository`; this will stop files being destroyed if the repo name changes.

deiga:
@stevehipwell Addressed your comments. And re-introduced the API Mock pattern here, since it's the only way to verify the migration.

Would you still prefer that we discuss before we can add this pattern?

stevehipwell:
> Would you still prefer that we discuss before we can add this pattern?

Yes, I don't think this is the time to add this; but I think it's something we need to explore further.

deiga:
@stevehipwell Done

github/resource_github_repository_file_migration_test.go
```diff
@@ -0,0 +1,155 @@
+// TODO: Enable this test once we have a way to mock the GitHub API
+
+// package github
```
stevehipwell:
```suggestion
package github

// TODO: Enable this test once we have a way to mock the GitHub API
```

github/resource_github_repository_file.go
```diff
@@ -161,6 +176,8 @@ func resourceGithubRepositoryFileCreate(ctx context.Context, d *schema.ResourceD
 
 	checkOpt := github.RepositoryContentGetOptions{}
 
+	repoInfo, _, err := client.Repositories.Get(ctx, owner, repo)
```
stevehipwell, line 181: This is missing an error handler.

github/resource_github_repository_file.go
```diff
@@ -175,6 +192,17 @@ func resourceGithubRepositoryFileCreate(ctx context.Context, d *schema.ResourceD
 			}
 		}
 		checkOpt.Ref = branch.(string)
+	} else {
+		if err != nil {
+			return diag.FromErr(err)
+		}
```
stevehipwell: I think this is the missing error handler from above?

github/resource_github_repository_file.go
```diff
@@ -227,7 +261,11 @@ func resourceGithubRepositoryFileRead(ctx context.Context, d *schema.ResourceDat
 	client := meta.(*Owner).v3client
 	owner := meta.(*Owner).name
 
-	repo, file := splitRepoFilePath(d.Id())
+	repo, file, _, err := parseID3(d.Id())
+	if err != nil {
+		return diag.FromErr(err)
+	}
```
stevehipwell: We should use `d.Get()` instead of parsing the ID.

github/resource_github_repository_file.go
```diff
@@ -175,6 +192,17 @@ func resourceGithubRepositoryFileCreate(ctx context.Context, d *schema.ResourceD
 			}
 		}
 		checkOpt.Ref = branch.(string)
+	} else {
+		if err != nil {
+			return diag.FromErr(err)
+		}
+		if err := d.Set("branch", repoInfo.GetDefaultBranch()); err != nil {
+			return diag.FromErr(err)
+		}
```
stevehipwell: Why not set `branch` here and then do `d.Set()` after the file has been created successfully?

github/resource_github_repository_file.go
```diff
@@ -215,7 +243,13 @@ func resourceGithubRepositoryFileCreate(ctx context.Context, d *schema.ResourceD
 		return diag.FromErr(err)
 	}
 
-	d.SetId(fmt.Sprintf("%s/%s", repo, file))
+	branch := d.Get("branch").(string)
```
stevehipwell: See above.

github/resource_github_repository_file.go
stevehipwell: The branch will always be set so this logic should be simplified.

github/resource_github_repository_file.go
```diff
@@ -427,28 +465,50 @@ func autoBranchDiffSuppressFunc(k, _, _ string, d *schema.ResourceData) bool {
...
 
 	client := meta.(*Owner).v3client
 	owner := meta.(*Owner).name
-	repo, file := splitRepoFilePath(repoFilePath)
 
-	opts := &github.RepositoryContentGetOptions{Ref: branch}
-	if err := d.Set("branch", branch); err != nil {
+	opts := &github.RepositoryContentGetOptions{}
+
+	repoInfo, _, err := client.Repositories.Get(ctx, owner, repo)
```
stevehipwell, line 446: The error handler for this is misisng.

github/resource_github_repository_file.go
stevehipwell: Same as above.

deiga:
@stevehipwell 

> you'll also need to add an ID set to update as the ID needs to be updated if the repo name has changed.

Doesn't ForceNew run `Create`, so that shouldn't be needed in `Update`?

stevehipwell:
> Doesn't ForceNew run Create, so that shouldn't be needed in Update?

The diff function supports changing the repo name without forcing new, so we have to set the ID in update.

deiga:
@stevehipwell If the `repository_id` changes, then we set `ForceNew` on `repository`. Which triggers `Create` instead if `Update`, or do I have that mixed up? 
Is there any case where `repository_id` would change and need to be set in `Update`?

stevehipwell:
> @stevehipwell If the repository_id changes, then we set ForceNew on repository. Which triggers Create instead if Update, or do I have that mixed up?
Is there any case where repository_id would change and need to be set in Update?

The resource ID uses repo name so we need to call `d.SetId()`, not update the `repository_id` as that is static.

github/resource_github_repository_file.go
```diff
@@ -174,7 +196,13 @@ func resourceGithubRepositoryFileCreate(ctx context.Context, d *schema.ResourceD
...
 		}
-		checkOpt.Ref = branch.(string)
+		checkOpt.Ref = branch
+	} else {
+		branch = repoInfo.GetDefaultBranch()
+	}
+
+	if err := d.Set("repository_id", int(repoInfo.GetID())); err != nil {
+		return diag.FromErr(err)
 	}
```
stevehipwell: I'd move this below `d.SetId()`.

github/resource_github_repository_file.go
```diff
@@ -215,7 +243,18 @@ func resourceGithubRepositoryFileCreate(ctx context.Context, d *schema.ResourceD
 		return diag.FromErr(err)
 	}
 
-	d.SetId(fmt.Sprintf("%s/%s", repo, file))
+	if err := d.Set("branch", branch); err != nil {
+		return diag.FromErr(err)
+	}
```
stevehipwell: I'd move this below `d.SetId()`.

github/resource_github_repository_file.go
```diff
@@ -227,30 +266,34 @@ func resourceGithubRepositoryFileRead(ctx context.Context, d *schema.ResourceDat
...
-	repo, file := splitRepoFilePath(d.Id())
-	ctx = tflog.SetField(ctx, "repository", repo)
+	repoName, ok := d.Get("repository").(string)
+	if !ok {
+		return diag.FromErr(fmt.Errorf("repository not found or is not a string"))
+	}
+	file, ok := d.Get("file").(string)
+	if !ok {
+		return diag.FromErr(fmt.Errorf("file not found or is not a string"))
+	}
```
stevehipwell:
```suggestion
	repoName := d.Get("repository").(string)
	file := d.Get("file").(string)
	branch := d.Get("branch").(string
```

I don't think you need the OK check here as the types a defined in the schema and `d.Get()` returns an empty type if not found; so the outcome is always valid even if nothing is found.

Also branch should be fetched here.

github/resource_github_repository_file.go
```diff
@@ -349,18 +387,15 @@ func resourceGithubRepositoryFileUpdate(ctx context.Context, d *schema.ResourceD
...
-		tflog.Debug(ctx, "Using explicitly set branch", map[string]any{
-			"branch": branch.(string),
-		})
-		if err := checkRepositoryBranchExists(client, owner, repo, branch.(string)); err != nil {
-			if d.Get("autocreate_branch").(bool) {
-				if err := resourceGithubRepositoryFileCreateBranch(ctx, d, meta); err != nil {
-					return diag.FromErr(err)
-				}
-			} else {
+	branch := d.Get("branch").(string)
```
stevehipwell: This should be up with repo and file.

github/resource_github_repository_file.go
```diff
@@ -227,30 +266,34 @@ func resourceGithubRepositoryFileRead(ctx context.Context, d *schema.ResourceDat
...
-		if err := checkRepositoryBranchExists(client, owner, repo, branch.(string)); err != nil {
-			if d.Get("autocreate_branch").(bool) {
-				branch = d.Get("autocreate_branch_source_branch").(string)
-			} else {
-				tflog.Info(ctx, "Removing repository path from state because the branch no longer exists in GitHub")
-				d.SetId("")
-				return nil
-			}
+	branch := d.Get("branch").(string)
+	if err := checkRepositoryBranchExists(ctx, client, owner, repoName, branch); err != nil {
```
stevehipwell: This looks wrong to me, read is only called after the resource is created so if the branch no longer exists it should error or have the ID set to empty.

IMHO longer term the ability for this resource to create a branch needs to be removed as it doesn't fit with the TF model, if you need a branch there is a resource for that which will follow the correct TF lifecycle.

github/resource_github_repository_file.go
```diff
@@ -227,30 +266,34 @@ func resourceGithubRepositoryFileRead(ctx context.Context, d *schema.ResourceDat
...
+		if d.Get("autocreate_branch").(bool) {
+			branch = d.Get("autocreate_branch_source_branch").(string)
```
stevehipwell: OK, this is fully incorrect. Read is meant to reflect the state of the system and if the branch doesn't exist then the content is empty.

github/resource_github_repository_file.go
```diff
@@ -427,28 +458,60 @@ func autoBranchDiffSuppressFunc(k, _, _ string, d *schema.ResourceData) bool {
...
+			return nil, err
+		}
+	} else {
+
+		defaultBranch := repoInfo.GetDefaultBranch()
+		branch = defaultBranch
+		if err := d.Set("branch", branch); err != nil {
+			return nil, err
+		}
+	}
```
stevehipwell:
```suggestion
	if branch == "" {
		branch = repoInfo.GetDefaultBranch()		
	}
	
	opts.Ref = branch
	
	if err := d.Set("branch", branch); err != nil {
		return nil, err
	}
```
But move the two sets down with the other sets for consistency.

stevehipwell (changes requested):

deiga:
@stevehipwell Doh! 🤦 
Sorry, I kept misunderstanding your point

deiga:
@stevehipwell I went and deprecated the whole `autocreate_branch` stuff, let me know if I went too far :D

And, could you update the title of this PR to something like "resource_github_repository_file: fix import format and add state migrations"

github/resource_github_repository_file.go
```diff
@@ -174,7 +199,9 @@ func resourceGithubRepositoryFileCreate(ctx context.Context, d *schema.ResourceD
 				return diag.FromErr(err)
 			}
 		}
-		checkOpt.Ref = branch.(string)
+		checkOpt.Ref = branch
```
stevehipwell: Can we move this out of the if/else block so we're always being explicit about the branch? This would also be consistent with the rest of the code.

github/resource_github_repository_file.go
```diff
@@ -215,10 +242,25 @@ func resourceGithubRepositoryFileCreate(ctx context.Context, d *schema.ResourceD
...
 		return diag.FromErr(err)
 	}
+	if err := d.Set("branch", branch); err != nil {
+		return diag.FromErr(err)
+	}
+	if err := d.Set("repository_id", int(repoInfo.GetID())); err != nil {
+		return diag.FromErr(err)
+	}
 
 	return resourceGithubRepositoryFileRead(ctx, d, meta)
```
stevehipwell, line 265: We shouldn't need this as long as all of the computed values are being set above.

github/resource_github_repository_file.go
```diff
@@ -227,35 +269,26 @@ func resourceGithubRepositoryFileRead(ctx context.Context, d *schema.ResourceDat
...
-		}
-		opts.Ref = branch.(string)
-	}
+	opts.Ref = branch
 
-	fc, _, _, err := client.Repositories.GetContents(ctx, owner, repo, file, opts)
+	fc, _, _, err := client.Repositories.GetContents(ctx, owner, repoName, file, opts)
 	if err != nil {
 		var errorResponse *github.ErrorResponse
 		if errors.As(err, &errorResponse) && errorResponse.Response.StatusCode == http.StatusTooManyRequests {
```
stevehipwell: Is this correct?

github/resource_github_repository_file.go
```diff
@@ -345,31 +373,20 @@ func resourceGithubRepositoryFileUpdate(ctx context.Context, d *schema.ResourceD
...
+	ctx = tflog.SetField(ctx, "branch", branch)
 
 	opts := resourceGithubRepositoryFileOptions(d)
 
 	if *opts.Message == fmt.Sprintf("Add %s", file) {
 		opts.Message = github.Ptr(fmt.Sprintf("Update %s", file))
 	}
 
+	// TODO: Use UpdateFile if the file already exists
 	create, _, err := client.Repositories.CreateFile(ctx, owner, repo, file, opts)
```
stevehipwell: Why not swap this over now? The `sha` value should be empty if the file doesn't already exist.

github/resource_github_repository_file.go
deiga: I know, but I've been dreading removing it as there is so much computed stuff going in `Read` which doesn't exist in `Create`.

Honestly, it's a can of worms I'd like to keep out of this scope

github/resource_github_repository_file.go
deiga: It was in there. I doubt it actually does anything 🤷

deiga: It was introduced in https://github.com/integrations/terraform-provider-github/pull/2345/changes 
Considering that it was supposed to protect from rate-limiting, I think it's no longer applicable

github/resource_github_repository_file.go
deiga: I'm scared 😂

github/resource_github_repository_file.go
stevehipwell: I don't think any of that is actually relevant after create has run, but I guess we could pick that up in a follow up PR?

github/resource_github_repository_file.go
```diff
@@ -227,35 +269,22 @@ func resourceGithubRepositoryFileRead(ctx context.Context, d *schema.ResourceDat
...
+	opts.Ref = branch
 
-	fc, _, _, err := client.Repositories.GetContents(ctx, owner, repo, file, opts)
+	fc, _, _, err := client.Repositories.GetContents(ctx, owner, repoName, file, opts)
 	if err != nil {
-		var errorResponse *github.ErrorResponse
-		if errors.As(err, &errorResponse) && errorResponse.Response.StatusCode == http.StatusTooManyRequests {
-			return diag.FromErr(err)
-		}
+		return diag.FromErr(deleteResourceOn404AndSwallow304OtherwiseReturnError(err, d, "repository file %s/%s:%s:%s", owner, repoName, file, branch))
```
stevehipwell, line 287: I don't think we're using etag here so do we think this is safe or would we be better to just use the inline 404 pattern?

github/resource_github_repository_file.go
```diff
@@ -345,40 +369,36 @@ func resourceGithubRepositoryFileUpdate(ctx context.Context, d *schema.ResourceD
...
 	if err != nil {
 		return diag.FromErr(err)
 	}
 
-	if err = d.Set("commit_sha", create.GetSHA()); err != nil {
+	if err = d.Set("commit_sha", update.GetSHA()); err != nil {
 		return diag.FromErr(err)
 	}
 
+	if d.HasChanges("repository", "file", "branch") {
```
stevehipwell, line 394: Is this worth the check or are we best just setting the value? If it's worth the check we ought to use it elsewhere.

website/docs/r/repository_file.html.markdown
```diff
@@ -57,55 +59,53 @@ resource "github_repository_file" "foo" {
...
 
-* `autocreate_branch_source_sha` - (Optional) The commit hash to start from, if 'autocreate_branch' is set. Defaults to the tip of 'autocreate_branch_source_branch'. If provided, 'autocreate_branch_source_branch' is ignored.
+- `autocreate_branch_source_sha` - (Optional) The commit hash to start from, if 'autocreate_branch' is set. Defaults to the tip of 'autocreate_branch_source_branch'. If provided, 'autocreate_branch_source_branch' is ignored.
 
 ## Attributes Reference
 
 The following additional attributes are exported:
 
-* `commit_sha` - The SHA of the commit that modified the file.
+- `commit_sha` - The SHA of the commit that modified the file.
```
stevehipwell, line 93: I think we need to add `repository_id` here?

github/resource_github_repository_file.go
deiga: It looks safe to me, since it just compares status codes. The GH API doesn't return a 304 unless stuff didn't change

github/resource_github_repository_file.go
deiga: Probably not _really_ worth it. But I like the thought of only doing any changes if there are changes. Especially if it's API things (which this isn't, sure)

No strong feelings on this specific change, but I think the pattern is something we should discuss adopting

nickfloyd (approved):

PR #3179, opened 2026-02-11
[MAINT] Add enterprise EMU testing config

deiga:
----

### Before the change?

- Contributors running tests against GHEC EMU would need to manually skip lot's of tests

### After the change?

- Tests which are not applicable to GHEC EMU (anything using public content) are skipped if EMU is configured
- Tests which create repos use new `testAcc` config for repo visibility to enable them to be run in GHEC EMU

### Pull request checklist

- [ ] Yes
- [x] No

----

### DIFF
--- a/github/acc_test.go
+++ b/github/acc_test.go
@@ -77,6 +77,9 @@ type testAccConfig struct {
 
 	// Test options
 	testAdvancedSecurity bool
+
+	// Test repository configuration
+	testRepositoryVisibility string
 }
 
 var testAccConf *testAccConfig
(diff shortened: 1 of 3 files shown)


github/acc_test.go
```diff
@@ -165,6 +171,10 @@ func TestMain(m *testing.M) {
 		if err == nil {
 			config.testEnterpriseEMUGroupId = i
 		}
+
+		if config.enterpriseIsEMU {
+			config.testRepositoryVisibility = "private"
```
stevehipwell, line 176: Would internal not be better?

deiga: Hmm, I don't think there is any benefit in using `internal`. Any functionality that relies on `internal` should specifically use it anyway. And I don't really want to "bleed" any testing repos in my GH Enterprise if they fail to cleanup 😬

stevehipwell: There definitely used to be functionality that was disabled for private repos, and IIRC it was available for internal repos. I'm not sure if this is still the case, and I agree that if you're using an actual enterprise for testing (which 100% shouldn't be required) you don't want anything "bleeding".

PR #3186, opened 2026-02-12
fix(data_source_github_release_asset): Fix #2514 remove accept header…

felipebraga:
Resolves #2514 and  #2513

Fixes #2514 remove accept header that's cause http error `406 Not Acceptable` when trying to download the asset.

The new data source `github_release_asset` introduced in provider version 6.11, which adds the functionality to download the file asset, but it wasn't working due to the `req.Header.Set("Accept", "application/octet-stream")` which was causing the GitHub API to return a `406 Not Acceptable`.

----

### Before the change?

- The data source `github_release_asset` returns `406 Not Acceptable` when trying to download the asset.
- Doc has a mismatch attribute for the data source

### After the change?

- The data source `github_release_asset` is able to download the specified asset.
- Doc is updated

### Pull request checklist

- [ ] Yes
- [x] No

----

### DIFF
--- a/website/docs/d/release_asset.html.markdown
+++ b/website/docs/d/release_asset.html.markdown
@@ -25,10 +25,10 @@ into a `file` attribute on the data source:
 
 ```hcl
 data "github_release_asset" "example" {
-    repository    = "example-repository"
-    owner         = "example-owner"
-    asset_id      = 12345
-    download_file = true
+    repository             = "example-repository"
+    owner                  = "example-owner"
+    asset_id               = 12345
+    download_file_contents = true
 }
 ```
 
--- a/github/data_source_github_release_asset.go
+++ b/github/data_source_github_release_asset.go
@@ -147,7 +147,6 @@ func dataSourceGithubReleaseAssetRead(ctx context.Context, d *schema.ResourceDat
 		return diag.FromErr(err)
 	}
 
-	req.Header.Set("Accept", "application/octet-stream")
 	resp, err := clientCopy.Do(req)
 	if err != nil {
 		return diag.FromErr(err)


stevehipwell:
CC @mdb

stevehipwell:
@felipebraga what are your thoughts on the other changes introduced in #2514 related to setting the header? I would think that they should also be reverted?

website/docs/d/release_asset.html.markdown
```diff
@@ -25,10 +25,10 @@ into a `file` attribute on the data source:
 
 ```hcl
 data "github_release_asset" "example" {
-    repository    = "example-repository"
-    owner         = "example-owner"
-    asset_id      = 12345
-    download_file = true
```
mdb, line 31: Oh, good catch!

mdb:
> The new data source github_release_asset introduced in provider version 6.11, which adds the functionality to download the file asset, but it wasn't working due to the req.Header.Set("Accept", "application/octet-stream") which was **causing the GitHub API to return a 406 Not Acceptable.**

@felipebraga Are you able to provide a specific release asset example where this can be reproduced? You may be onto something, though acceptance tests seemingly continue to pass both with and without the header handling changes introduced in https://github.com/integrations/terraform-provider-github/pull/2514. So, it'd be good to have a more clear sense of the 406's cause.

```
$ TF_ACC=1 go test -v ./... -run TestAccGithubReleaseAssetDataSource
?   	github.com/integrations/terraform-provider-github/v6	[no test files]
=== RUN   TestAccGithubReleaseAssetDataSource
=== RUN   TestAccGithubReleaseAssetDataSource/queries_and_downloads_specified_asset_ID
=== RUN   TestAccGithubReleaseAssetDataSource/queries_without_downloading_the_specified_asset_ID
--- PASS: TestAccGithubReleaseAssetDataSource (5.79s)
    --- PASS: TestAccGithubReleaseAssetDataSource/queries_and_downloads_specified_asset_ID (3.42s)
    --- PASS: TestAccGithubReleaseAssetDataSource/queries_without_downloading_the_specified_asset_ID (2.38s)
PASS
ok  	github.com/integrations/terraform-provider-github/v6/github	6.913s
```

(Also, given https://github.com/integrations/terraform-provider-github/pull/2514 was originally opened in 2024, it's possible a subsequent upgrade to https://github.com/google/go-github has made the header handling intro'd by https://github.com/integrations/terraform-provider-github/pull/2514 unnecessary altogether; I just want to be sure)

PR #3193, opened 2026-02-14
[MAINT] Fix test of Org Repo Roles

deiga:
----

### Before the change?

- Computed fields were not being set correctly in `resource_github_organization_repository_role`
- Tests for `resource_github_organization_repository_role` were failing due to configuration issues
- TODO comments for `github_organization_repository_role` about upgrading to v68+

### After the change?

- Corrected behaviour of Computed fields
- Fixed tests
- Added validation for the `permissions` field in. `resource_github_organization_repository_role`
- Update to use new SDK functions

### Pull request checklist

- [ ] Yes
- [x] No

----

### DIFF
--- a/github/resource_github_organization_repository_role.go
+++ b/github/resource_github_organization_repository_role.go
@@ -2,8 +2,9 @@ package github
 
 import (
 	"context"
+	"errors"
 	"fmt"
-	"log"
+	"net/http"
 	"slices"
 	"strconv"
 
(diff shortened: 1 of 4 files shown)


github/resource_github_organization_repository_role.go
```diff
@@ -177,11 +170,15 @@ func resourceGithubOrganizationRepositoryRoleUpdate(ctx context.Context, d *sche
 		Permissions: permissionsStr,
 	}
 
-	_, _, err = client.Organizations.UpdateCustomRepoRole(ctx, orgName, roleId, update)
+	role, _, err := client.Organizations.UpdateCustomRepoRole(ctx, orgName, roleId, update)
 	if err != nil {
 		return diag.FromErr(fmt.Errorf("error updating GitHub organization repository role (%s/%s): %w", orgName, d.Get("name").(string), err))
 	}
 
+	if err = d.Set("role_id", role.GetID()); err != nil {
```
stevehipwell: Does this need doing here?

deiga: Hmm, this looks wrong. I wonder if I was trying to add it to `Create`

stevehipwell (approved):
LGTM

@deiga thanks for fixing these, I'm not sure how I managed to miss setting the `repo_role`. I'll get the API for listing the roles into the SDK, but with the current provider implementation I think we probably want to keep it static and use `go generate` to keep it up-to-date (my docs PR uses `go generate` so it should be straightforward to extend it for this an other potential uses.

nickfloyd (approved):

PR #3215, opened 2026-02-21
[MAINT] Consolidate ID format of `github_emu_group_mapping`

### LINKED ISSUE #3200: [MAINT] Consolidate ID in EMU group mapping resource
@deiga I've only just spotted this, but why hasn't the ID be corrected to match the input ID? The resource is scoped to the EMU group, so I'd expect the ID to be `<group-id>:<team-id>` as it is in the import? Also why is the ID using 3 parts instead of the 2 parts used for this pattern elsewhere in the re-written provider code?

_Originally posted by @stevehipwell in https://github.com/integrations/terraform-provider-github/issues/3163#issuecomment-3907424373_

deiga:
Resolves #3200

----

### Before the change?

- Import ID and Resource ID had different format
- `tflog.SetField` was being used incorrectly
- Tests didn't use `terraform-plugin-testing`
- Some tests didn't fully work
- Docs were missing computed fields

### After the change?

- Import ID and Resource ID have the same format
- Fixes `tflog.SetField` usage
- Updates tests to use `terraform-plugin-testing`
- Fixed tests
- Updates docs

### Pull request checklist

- [ ] Yes
- [x] No

----

### DIFF
--- a/github/resource_github_emu_group_mapping.go
+++ b/github/resource_github_emu_group_mapping.go
@@ -2,6 +2,7 @@ package github
 
 import (
 	"context"
+	"fmt"
 	"net/http"
 	"strconv"
 
@@ -302,11 +303,11 @@ func resourceGithubEMUGroupMappingImport(ctx context.Context, d *schema.Resource
 	// <group-id>:<team-slug>
 	groupIDString, teamSlug, err := parseID2(d.Id())
 	if err != nil {
-		return nil, err
+		return nil, fmt.Errorf("could not parse import ID (%s), expected format: <group-id>:<team-slug>. Parse error: %w", importID, err)
 	}
 	groupID, err := strconv.Atoi(groupIDString)
 	if err != nil {
-		return nil, err
+		return nil, unconvertibleIdErr(groupIDString, err)
 	}
 
 	tflog.Debug(ctx, "Parsed two-part import ID", map[string]any{


github/resource_github_emu_group_mapping.go
```diff
@@ -69,13 +74,13 @@ func resourceGithubEMUGroupMappingCreate(ctx context.Context, d *schema.Resource
 	}
 	client := meta.(*Owner).v3client
 	orgName := meta.(*Owner).name
-	tflog.SetField(ctx, "org_name", orgName)
+	ctx = tflog.SetField(ctx, "org_name", orgName)
```
stevehipwell: I don't think we should be creating a large context for the logger by default. I think ID should be set, but everything else should only be set if it has an actual value in the context of the log being written.

deiga: AFAIK the log context is closed at the end of the function, right? My main thought here was, that I'd want to set these fields in all the log calls.

But now that I think of it that's not too useful. I'd rather just log these values once and then have a clear context.
Good call!

github/resource_github_emu_group_mapping.go
```diff
@@ -142,36 +153,29 @@ func resourceGithubEMUGroupMappingRead(ctx context.Context, d *schema.ResourceDa
...
 
-	tflog.SetField(ctx, "group_id", groupID)
-	tflog.SetField(ctx, "team_slug", teamSlug)
-	tflog.SetField(ctx, "org_name", orgName)
-
-	tflog.Debug(ctx, "Querying external groups linked to team from GitHub API")
+	tflog.Debug(ctx, "Querying external groups linked to team from GitHub API", map[string]any{
+		"org_name":  orgName,
+		"team_slug": teamSlug,
+	})
```
stevehipwell: I don't think you need these fields as you've already got the ID and this just duplicates the ID parts.

github/resource_github_emu_group_mapping.go
```diff
@@ -180,8 +184,9 @@ func resourceGithubEMUGroupMappingRead(ctx context.Context, d *schema.ResourceDa
 	group := groupsList.Groups[0]
 
 	tflog.Debug(ctx, "Successfully retrieved external group from GitHub API", map[string]any{
-		"group_id":   group.GetGroupID(),
-		"group_name": group.GetGroupName(),
+		"configured_group_id": groupID,
+		"upstream_group_id":   group.GetGroupID(),
+		"group_name":          group.GetGroupName(),
```
stevehipwell: I don't think these are required here, ID is already in context and if something doesn't match there is a log line below to show this.

github/resource_github_emu_group_mapping.go
```diff
@@ -205,30 +210,30 @@ func resourceGithubEMUGroupMappingRead(ctx context.Context, d *schema.ResourceDa
...
 	}
 
-	if d.HasChanges("group_id", "team_slug") {
+	if d.HasChange("team_slug") {
 
-		tflog.Debug(ctx, "Updating connected external group via GitHub API")
+		tflog.Debug(ctx, "Updating connected external group via GitHub API", map[string]any{
+			"org_name":  orgName,
+			"team_slug": teamSlug,
+			"group_id":  groupID,
```
stevehipwell: As above.

stevehipwell:

stevehipwell:
@robert-crandall please could you re-approve this PR?

PR #3216, opened 2026-02-21
[MAINT] Improve Import behaviour of `github_default_branch`

### LINKED ISSUE #2613: [BUG]: github_branch_default attempts to rename to the same name on import
### Expected Behavior

When you create a repo and assign the default branch, the code should check to see if it needs to do work.
If the default branch supplied is the same as the current branch, then nothing needs to be done.

This is visible when importing a repo settings.

### Actual Behavior

I have a repo called x, x repo has a branch called main that is the primary branch. I created x repo outside of terraform.

I import the repo and assign the default branch using a github_branch_default resource which is also imported.

The import of the repo is done as expected, but the github_branch_default is imported and the rename config is set to additive and true. This then triggers a rename to happen, which then fails because the source and destination name are the same.

This is even more difficult to deal with since I happen to use the github_branch_default within a module that manages the repos for me.

### Terraform Version

Terraform 1.11.3
Provider version: 6.6.0

### Affected Resource(s)

- github_branch_default

```hcl
import {
  to = module.generic_repos["junk-1"].github_repo.current
  id = "randy-junk-1"
}

import {
  to = module.generic_repos["junk-1"].github_branch_default.main
  id = "randy-junk-1"
}

# -- Within the module 'generic_repos' ---

variable "name" {
  description = "Repo name"
  type = string
}

variable "description" {
  description = "Repo description"
  type = string
}

variable "visibility" {
  description = "Repo visibility"
  type = string
  default = "internal"
}
variable "primary_branch" {
  description = "Default branch"
  type = string
  default = "main"
}

resource github_repository "current" {
  name        = var.name
  description = var.description
  visibility  = var.visibility
  auto_init   = true
  # Failsafe to prevent accidental deletion of the repo.
  archive_on_destroy = true
  
  vulnerability_alerts = true
}

resource "github_branch_default" "main" {
  repository = resource.github_repository.current.name
  branch     = var.primary_branch
  rename     = true
}
```

### Steps to Reproduce

Create a repo outside of terraform.
Set anything you want as the primary branch.

Create a repo using code above, making sure to set the primary branch the same as default branch in your repo. Before you apply do the imports like below.

Import both the repo and the default branch like in the code snip it.

When the apply happens you'll see that it tries to rename the primary branch to the same name that you imported.

### Debug Output

```shell

```

### Panic Output

```shell

```

### Code of Conduct

deiga:
Resolves #2613

----

### Before the change?

- Resource didn't use Context-aware functions
- Resource didn't use `tflog` for logging
- Missing tests for  `Update` and `Import` functionality

### After the change?

- Changes to use Context-aware functions
- Changes to use `tflog` for logging
- Adds support for repository renaming, by add `repository_id` field and removing `ForceNew` from `repository`

### Pull request checklist

- [ ] Yes
- [x] No

----


github/resource_github_branch_default.go
```diff
@@ -101,70 +159,141 @@ func resourceGithubBranchDefaultRead(d *schema.ResourceData, meta any) error {
...
+		repository, resp, err := client.Repositories.Get(ctx, owner, repoName)
 		if err != nil {
-			return err
+			return diag.FromErr(err)
 		}
-		if _, _, err := client.Repositories.RenameBranch(ctx, owner, repoName, *repository.DefaultBranch, defaultBranch); err != nil {
-			return err
+		etag = resp.Header.Get("ETag")
+		if _, _, err := client.Repositories.RenameBranch(ctx, owner, repoName, repository.GetDefaultBranch(), defaultBranch); err != nil {
+			return diag.FromErr(err)
```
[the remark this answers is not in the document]
deiga: Good spot. Will add regression test and implement skipping behaviour

github/resource_github_branch_default.go
```diff
@@ -51,47 +58,98 @@ func resourceGithubBranchDefault() *schema.Resource {
...
-	client := meta.(*Owner).v3client
-	owner := meta.(*Owner).name
-	repoName := d.Get("repository").(string)
-	defaultBranch := d.Get("branch").(string)
-	rename := d.Get("rename").(bool)
+func resourceGithubBranchDefaultCreate(ctx context.Context, d *schema.ResourceData, m any) diag.Diagnostics {
+	meta, ok := m.(*Owner)
+	if !ok {
+		return diag.FromErr(errors.New("unexpected meta type, expected *Owner"))
+	}
```
stevehipwell:
```suggestion
	meta, _ := m.(*Owner)
```
For meta this would be fine as we'd have a bigger problem if the type changed.

github/resource_github_branch_default.go
```diff
@@ -51,47 +58,98 @@ func resourceGithubBranchDefault() *schema.Resource {
...
+	client := meta.v3client
+	owner := meta.name
+
+	repoName, ok := d.Get("repository").(string)
+	if !ok {
+		return diag.FromErr(errors.New("repository must be a string"))
+	}
```
stevehipwell: I think we need a formal decision about this pattern, the schema says it's a string and the schema mapping here is 1-2-1 so this shouldn't need to be checked. Is it cleaner to keep the checks as in some other contexts the schema mapping isn't absolute?

github/resource_github_branch_default.go
```diff
@@ -51,47 +58,98 @@ func resourceGithubBranchDefault() *schema.Resource {
...
+		return diag.FromErr(errors.New("rename must be a boolean"))
+	}
 
-	ctx := context.Background()
+	tflog.Trace(ctx, "Creating default branch resource", map[string]any{
+		"owner":      owner,
+		"repository": repoName,
+		"branch":     defaultBranch,
+		"rename":     rename,
+	})
```
stevehipwell: I think we agreed to make these a single line?

stevehipwell (changes requested):
Just a quick pre-pass to look at the coding patterns for a re-work. The multi line tflog statements make reviewing harder; as does the type checking but I think we need to keep that (other than for `meta` in the resource functions).

github/resource_github_branch_default.go
```diff
@@ -51,47 +58,98 @@ func resourceGithubBranchDefault() *schema.Resource {
...
-				return err
+			tflog.Debug(ctx, "Renaming branch to new default")
+			if _, _, err := client.Repositories.RenameBranch(ctx, owner, repoName, repository.GetDefaultBranch(), defaultBranch); err != nil {
+				return diag.FromErr(err)
 			}
 		} else {
+			tflog.Debug(ctx, "Setting new default branch")
 			repository := &github.Repository{
-				DefaultBranch: &defaultBranch,
+				DefaultBranch: new(defaultBranch),
```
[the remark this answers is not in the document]
stevehipwell: I'm pretty sure the code is compiling correctly as this is a Go 1.26 project where `new` has improved capabilities.

github/resource_github_branch_default.go
```diff
@@ -101,70 +159,141 @@ func resourceGithubBranchDefaultRead(d *schema.ResourceData, meta any) error {
...
-			return err
+		etag = resp.Header.Get("ETag")
+		if _, _, err := client.Repositories.RenameBranch(ctx, owner, repoName, repository.GetDefaultBranch(), defaultBranch); err != nil {
+			return diag.FromErr(err)
 		}
 	} else {
+		tflog.Debug(ctx, "Setting new default branch")
 		repository := &github.Repository{
-			DefaultBranch: &defaultBranch,
+			DefaultBranch: new(defaultBranch),
```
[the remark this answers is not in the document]
stevehipwell: I'm pretty sure the code is compiling correctly as this is a Go 1.26 project where `new` has improved capabilities.

github/resource_github_branch_default.go
```diff
@@ -51,47 +58,80 @@ func resourceGithubBranchDefault() *schema.Resource {
...
+				return diag.FromErr(err)
 			}
 		} else {
+			tflog.Debug(ctx, "Setting new default branch")
 			repository := &github.Repository{
-				DefaultBranch: &defaultBranch,
+				DefaultBranch: new(defaultBranch),
 			}
 
 			if _, _, err := client.Repositories.Edit(ctx, owner, repoName, repository); err != nil {
```
stevehipwell: I think we need to set the etag on this path?

github/resource_github_branch_default.go
```diff
@@ -101,70 +141,134 @@ func resourceGithubBranchDefaultRead(d *schema.ResourceData, meta any) error {
 		var ghErr *github.ErrorResponse
 		if errors.As(err, &ghErr) {
```
stevehipwell: It might be worth switching to `errors.AsType` here?

stevehipwell (changes requested):
It looks like the code for the `repository_id` addition is missing (e.g. migration, import, update ID modification). I'd also suggest that the docs be updated to the new pattern as part of this change given the attributes are being modified.

github/resource_github_branch_default.go
deiga: Good catch!

deiga:
> It looks like the code for the `repository_id` addition is missing (e.g. migration, import, update ID modification). I'd also suggest that the docs be updated to the new pattern as part of this change given the attributes are being modified.

🤦 Doh!
Good catch, adding that now

github/resource_github_branch_default.go
deiga: I followed your lead and just swallowed the `ok`

templates/resources/branch_default.md.tmpl
```diff
@@ -1,39 +1,28 @@
...
 
-- `repository` - (Required) The GitHub repository
-- `branch` - (Required) The branch (e.g. `main`)
-- `rename` - (Optional) Indicate if it should rename the branch rather than use an existing branch. Defaults to `false`.
+{{ .SchemaMarkdown | trimspace }}
 
 ## Import
 
-GitHub Branch Defaults can be imported using an ID made up of `repository`, e.g.
+GitHub Branch Defaults can be imported using the repository name with the [`terraform import` command](https://developer.hashicorp.com/terraform/cli/commands/import), for example:
```
stevehipwell: This doesn't look right, it should be a direct copy of the generic template with any additional text added below `## Import`. You will then also want to add the _import-by-string-id.tf_ file.

templates/resources/branch_default.md.tmpl
```diff
@@ -1,39 +1,28 @@
...
 
 # {{.Name}} ({{.Type}})
 
-Provides a GitHub branch default resource.
+{{ .Description | trimspace }}
 
-This resource allows you to set the default branch for a given repository.
-
-Note that use of this resource is incompatible with the `default_branch` option of the `github_repository` resource. Using both will result in plans always showing a diff.
+~> **Note:** This resource is incompatible with the `default_branch` option of the `github_repository` resource. Using both will result in plans always showing a diff.
```
stevehipwell:
```suggestion
~> This resource is incompatible with the `default_branch` option of the [`github_repository`](github_repository) resource. Using both will result in plans always showing a diff.
```

github/resource_github_branch_default.go
```diff
@@ -51,120 +70,202 @@ func resourceGithubBranchDefault() *schema.Resource {
...
 }
 
-func resourceGithubBranchDefaultUpdate(d *schema.ResourceData, meta any) error {
-	client := meta.(*Owner).v3client
-	owner := meta.(*Owner).name
-	repoName := d.Id()
-	defaultBranch := d.Get("branch").(string)
-	rename := d.Get("rename").(bool)
+func resourceGithubBranchDefaultUpdate(ctx context.Context, d *schema.ResourceData, m any) diag.Diagnostics {
+	ctx = tflog.SetField(ctx, "resource_id", d.Id())
```
stevehipwell:
```suggestion
```

github/resource_github_branch_default.go
```diff
@@ -51,120 +70,202 @@ func resourceGithubBranchDefault() *schema.Resource {
...
+	etag := resp.Header.Get("ETag")
+
+	tflog.Debug(ctx, "Fetched repository", map[string]any{"current_default_branch": repository.GetDefaultBranch()})
+
+	if repository.GetDefaultBranch() != defaultBranch {
 		if rename {
-			if _, _, err := client.Repositories.RenameBranch(ctx, owner, repoName, *repository.DefaultBranch, defaultBranch); err != nil {
-				return err
+			tflog.Debug(ctx, "Renaming branch to new default")
+			if _, _, err := client.Repositories.RenameBranch(ctx, owner, repoName, repository.GetDefaultBranch(), defaultBranch); err != nil {
```
stevehipwell, line 141: Should we clear the etag here to make it explicit that we don't expect it to work on this code path (not that it ever works with non-static tokens)?

github/resource_github_branch_default.go
```diff
@@ -51,120 +70,202 @@ func resourceGithubBranchDefault() *schema.Resource {
...
 		if err != nil {
-			return err
+			return diag.FromErr(err)
 		}
-		if _, _, err := client.Repositories.RenameBranch(ctx, owner, repoName, *repository.DefaultBranch, defaultBranch); err != nil {
-			return err
+		if repository.GetDefaultBranch() != defaultBranch {
+			tflog.Debug(ctx, "Renaming branch to new default")
+			etag = resp.Header.Get("ETag")
+			if _, _, err := client.Repositories.RenameBranch(ctx, owner, repoName, repository.GetDefaultBranch(), defaultBranch); err != nil {
```
stevehipwell, line 255: Same etag question as above.

stevehipwell (changes requested):

github/resource_github_branch_default.go
deiga: What do you mean? It's only used for the reading the repo

stevehipwell: Yes, but if you rename the branch you're explicitly modifying the repo which means the repository read is stale and requires a new etag. By manually clearing it you're making this clear to anyone reading the code.

github/resource_github_branch_default.go
```diff
@@ -51,120 +70,201 @@ func resourceGithubBranchDefault() *schema.Resource {
...
+}
+
+func resourceGithubBranchDefaultImport(ctx context.Context, d *schema.ResourceData, m any) ([]*schema.ResourceData, error) {
+	repoName := d.Id()
+
+	if err := d.Set("repository", repoName); err != nil {
+		return nil, err
+	}
+
+	return []*schema.ResourceData{d}, nil
```
stevehipwell, line 326: We need to fetch all of the required fields as part of the import (e.g. repo ID & branch).

stevehipwell (changes requested):
If you wait for #3476 to be merged you can reuse the repo ID migration logic.

github/resource_github_branch_default_migration.go
```diff
@@ -0,0 +1,70 @@
...
+			if s, ok := v.(string); ok && s != "" {
+				repoName = s
+			}
+		}
+	}
+	if repoName == "" {
+		return nil, fmt.Errorf("repository is not a string or not set")
+	}
+
+	rawState["repository"] = repoName
```
stevehipwell:
```suggestion
```

Given that `repository` is required for v0 and the generic migration logic errors if it's not found in the state I don't see why we need to do this? It also makes the code much harder to read.

github/resource_github_branch_default_migration_test.go
```diff
@@ -0,0 +1,111 @@
...
+
+import (
+	"regexp"
+	"testing"
+
+	"github.com/google/go-cmp/cmp"
+	"github.com/google/go-github/v88/github"
+)
+
+func Test_resourceGithubBranchDefaultStateUpgradeV0toV1(t *testing.T) {
```
stevehipwell, line 11: Based on my suggestion to remove the unnecessary code from for the implementation of the migration this test should only have a success and failure case as the other logic is already being tested in the child function test.

github/resource_github_branch_default.go
```diff
@@ -3,45 +3,64 @@ package github
...
+				Type:        schema.TypeInt,
+				Computed:    true,
+				Description: "The ID of the GitHub repository.",
 			},
 			"rename": {
 				Type:        schema.TypeBool,
 				Optional:    true,
 				Default:     false,
-				Description: "Indicate if it should rename the branch rather than use an existing branch. Defaults to 'false'.",
+				Description: "Indicate if the current default branch should be renamed rather than switching to an existing branch. Defaults to 'false'.",
```
stevehipwell:
```suggestion
				Description: "If `true` rename the existing branch when the `branch` input is changed. Defaults to 'false'.",
```
Something along these lines might be clearer?

github/resource_github_branch_default.go
```diff
@@ -51,120 +70,219 @@ func resourceGithubBranchDefault() *schema.Resource {
...
+
+	return nil
 }
 
-func resourceGithubBranchDefaultRead(d *schema.ResourceData, meta any) error {
-	client := meta.(*Owner).v3client
-	owner := meta.(*Owner).name
-	repoName := d.Id()
+func resourceGithubBranchDefaultRead(ctx context.Context, d *schema.ResourceData, m any) diag.Diagnostics {
+	ctx = tflog.SetField(ctx, "resource_id", d.Id())
```
stevehipwell: We use `ctx = tflog.SetField(ctx, "id", d.Id())` in more places than this pattern, but personally I think it's probably better not to use it and to add the fields directly through the map opt. Either way we should be consistent.

github/resource_github_branch_default.go
```diff
@@ -51,120 +70,219 @@ func resourceGithubBranchDefault() *schema.Resource {
...
+	meta, _ := m.(*Owner)
+	client := meta.v3client
+	owner := meta.name
+	repoName, _ := d.Get("repository").(string)
+
+	tflog.Trace(ctx, "Deleting default branch resource", map[string]any{"owner": owner, "repository": repoName})
+
+	repository := &github.Repository{
+		DefaultBranch: nil,
```
stevehipwell: This does absolutely nothing as the field is `omitempty`, this function can just be a no-op.

stevehipwell (commented):
This is looking good, just a couple of minor things to discuss.

github/resource_github_branch_default_migration.go
deiga: Yeah, you're right. With the repo being set in Import now this isn't needed anymore

github/resource_github_branch_default.go
```diff
@@ -51,120 +109,210 @@ func resourceGithubBranchDefault() *schema.Resource {
...
-			if _, _, err := client.Repositories.RenameBranch(ctx, owner, repoName, *repository.DefaultBranch, defaultBranch); err != nil {
-				return err
+			tflog.Debug(ctx, "Renaming branch to new default")
+			if _, _, err := client.Repositories.RenameBranch(ctx, owner, repoName, repository.GetDefaultBranch(), defaultBranch); err != nil {
+				return diag.FromErr(err)
+			}
+			err := waitForDefaultBranch(ctx, client, owner, repoName, defaultBranch, defaultBranchRenameTimeout)
+			if err != nil {
+				return diag.FromErr(err)
 			}
```
stevehipwell:
```suggestion
			if err := waitForDefaultBranch(ctx, client, owner, repoName, defaultBranch, defaultBranchRenameTimeout); err != nil {
				return diag.FromErr(err)
}
```
Should this behaviour not be optional? I'm guessing that this was catching you during testing?

github/resource_github_branch_default.go
```diff
@@ -51,120 +109,210 @@ func resourceGithubBranchDefault() *schema.Resource {
...
+		etag = resp.Header.Get("ETag")
+		if repository.GetDefaultBranch() != defaultBranch {
+			tflog.Debug(ctx, "Renaming branch to new default", map[string]any{"resource_id": d.Id()})
+			if _, _, err := client.Repositories.RenameBranch(ctx, owner, repoName, repository.GetDefaultBranch(), defaultBranch); err != nil {
+				return diag.FromErr(err)
+			}
+			err := waitForDefaultBranch(ctx, client, owner, repoName, defaultBranch, defaultBranchRenameTimeout)
+			if err != nil {
+				return diag.FromErr(err)
+			}
```
stevehipwell:
```suggestion
			if err := waitForDefaultBranch(ctx, client, owner, repoName, defaultBranch, defaultBranchRenameTimeout); err != nil {
				return diag.FromErr(err)
			}
```
Same as above?

stevehipwell (changes requested):
This looks great, just my questions about if we actually need to wait and if we do should be an option?

github/resource_github_branch_default.go
deiga: @stevehipwell if we don't wait there is possibly State drift after a rename. What would be the reason for making it optional? 
I don't want to get people complaining that the resource is broken and causes state drift 😂

stevehipwell: @deiga state drift? This is a documented eventual consistency path. I can see how this could be a problem in a tight loop such as acceptance testing, but is it an actual real-world problem?

stevehipwell (dismissed):
LGTM

robert-crandall (dismissed):
Thanks for the thorough work on this, @deiga! The rename-safe `repository_id` pattern plus the v0→v1 state upgrader is exactly the right approach, and I appreciate the regression test that pins the import-without-rename behavior. I pulled it locally: it builds, vets clean, the upgrade unit test passes, and `make generatedocs` produces no drift. Looks great to me. 🙏

stevehipwell:
@deiga if you can rebase this PR it should auto-merge.

stevehipwell:
@deiga any idea which check hasn't passed? It all looks good to me.

stevehipwell (approved):
LGTM

PR #3220, opened 2026-02-23
[BUG] Enable setting review notifications without delegation

### LINKED ISSUE #2273: [BUG]: Allow setting review notifications without delegation
### Expected Behavior

It should be possible to set `github_team_settings.review_request_delegation.notify` without enabling auto assignment.

These are independent settings:
image

### Actual Behavior

When only setting `notify`, `terraform apply` fails with
```
Error: Review request delegation member count must be greater than 0
```

### Terraform Version

Terraform v1.8.3
on linux_amd64
+ provider registry.terraform.io/integrations/github v5.39.0

### Affected Resource(s)

- `github_team_settings`

_No response_

### Steps to Reproduce

_No response_

### Debug Output

_No response_

### Panic Output

_No response_

### Code of Conduct

deiga:
Resolves #2273

----

### Before the change?

- Docs for `github_team_settings` we're missing computed attributes
- Tests didn't cover all behaviour
- Tests used legacy `Check:` structure
- Resource didn't use Context-aware CRUD functions
- Resource CRUD-functions we're tightly coupled
- It was not possible to set an empty `review_request_delegation` block
- It wasn't possible to set `notify` without enabling delegation

### After the change?

- Docs for `github_team_settings` are complete
- Tests cover all behaviour
- Tests use new check structures from `terraform-plugin-testing`
- Resource uses Context-aware CRUD functions
- Resource CRUD-functions are independent
- It is possible to set an empty `review_request_delegation` block
- It is possible to set `notify` without enabling delegation

### Pull request checklist

- [ ] Yes
- [x] No

----


github/resource_github_team_settings.go
```diff
@@ -201,7 +305,7 @@ func resourceGithubTeamSettingsImport(d *schema.ResourceData, meta any) ([]*sche
 	if err = d.Set("team_uid", nodeId); err != nil {
 		return nil, err
 	}
-	return []*schema.ResourceData{d}, resourceGithubTeamSettingsRead(d, meta)
+	return []*schema.ResourceData{d}, nil
 }
 
 func resolveTeamIDs(idOrSlug string, meta *Owner, ctx context.Context) (nodeId, slug string, err error) {
```
stevehipwell:
```suggestion
func resolveTeamIDs(ctx context.Context, meta *Owner, idOrSlug string) (nodeId, slug string, err error) {
```
Do you want to fix this signature while we're here?

deiga: Sure!

stevehipwell: Other than this the PR looks good, I'll give it a final review after you push this.

stevehipwell (dismissed):
LGTM

robert-crandall (dismissed):

stevehipwell:

deiga:
@stevehipwell @robert-crandall Rebased, please review.
I'll review the others as well, but will wait for this to be merged before asking for reviews there

robert-crandall (dismissed):

github/resource_github_team_settings.go
```diff
@@ -201,37 +303,39 @@ func resourceGithubTeamSettingsImport(d *schema.ResourceData, meta any) ([]*sche
...
-	} else {
-		// The given id is an integer, assume it is a team id
-		team, _, teamIdErr := client.Teams.GetTeamByID(ctx, orgId, teamId)
-		if teamIdErr != nil {
-			// There isn't a team with the given ID, assume it is a teamslug
-			team, _, slugErr := client.Teams.GetTeamBySlug(ctx, orgName, idOrSlug)
-			if slugErr != nil {
-				return "", "", errors.New(teamIdErr.Error() + slugErr.Error())
-			}
+func resolveTeamIDs(ctx context.Context, meta *Owner, idOrSlug string) (nodeId, slug string, err error) {
```
stevehipwell:
```suggestion
func resolveTeamIDs(ctx context.Context, meta *Owner, idOrSlug string) (string, string, error) {
```
Could you please remove the named returns?

github/resource_github_team_settings.go
```diff
@@ -201,37 +303,39 @@ func resourceGithubTeamSettingsImport(d *schema.ResourceData, meta any) ([]*sche
...
 
-			return team.GetNodeID(), team.GetSlug(), nil
-		}
+// resolveNotify returns the notify value from the top-level attribute or the
+// deprecated nested attribute inside review_request_delegation. The top-level
+// attribute takes precedence. Since ConflictsWith prevents both from being set,
+// only one source can be active at a time.
+func resolveNotify(ctx context.Context, d *schema.ResourceData) bool {
+	// Check if top-level notify is explicitly configured.
+	if v, ok := d.GetOk("notify"); ok {
```
stevehipwell: I'm not sure there is any value to using `GetOk()` here given the way it works and the fact that you're using an explicit schema?

stevehipwell (changes requested):
Just a couple of things to look at.

github/resource_github_team_settings.go
deiga: The way I understand the code is that this is exactly where `GetOk()` should work for `TypeBool`. 
It should check if `notify` has been set in the config, right? Of course the tests don't have that scenario at all 🙃

deiga: Or how else would I be checking which one of the fields is actually set? Should I just read both and see if either is true? That sounds simpler 🤔

github/resource_github_team_settings.go
```diff
@@ -115,66 +184,95 @@ func resourceGithubTeamSettingsRead(d *schema.ResourceData, meta any) error {
...
+			settings := reviewRequestDelegation[0].(map[string]any)
 
 			teamReviewAlgorithm := githubv4.TeamReviewAssignmentAlgorithm(settings["algorithm"].(string))
-			return graphql.Mutate(ctx, &mutation, githubv4.UpdateTeamReviewAssignmentInput{
+			updateTeamReviewAssignmentInput := githubv4.UpdateTeamReviewAssignmentInput{
 				ID:              d.Id(),
 				Enabled:         githubv4.Boolean(true),
 				Algorithm:       &teamReviewAlgorithm,
 				TeamMemberCount: new(githubv4.Int(settings["member_count"].(int))),
 				NotifyTeam:      new(githubv4.Boolean(settings["notify"].(bool))),
```
robert-crandall: Should this use the new `resolveNotify` instead of the legacy field?

github/resource_github_team_settings.go
stevehipwell: Ideally you read them and look for `true`, if you can't do that then you need to fallback to using a hack like we had to use on some of the repo fields.

robert-crandall (dismissed):

github/resource_github_team_settings.go
deiga: Yeah, I'll check that out. Too long since this discussion was had :D

deiga:
@stevehipwell Thanks. Simplified the bool checking a bit now

stevehipwell (approved):
LGTM

stevehipwell:
@robert-crandall please could you approve this.

robert-crandall (approved):

PR #3224, opened 2026-02-25
fix: handle `sha_pinning_required=false`

### LINKED ISSUE #3223: [BUG]: sha_pinning_required = false is silently ignored due to d.GetOk() zero-value bug
### Expected Behavior

Setting `sha_pinning_required = false` on `github_actions_organization_permissions` (or `github_actions_repository_permissions`) should send `false` to the GitHub API and disable SHA pinning enforcement.

After `terraform apply`, the state should converge and subsequent `terraform plan` should show no changes.

### Actual Behavior

Setting `sha_pinning_required = false` is silently ignored. The `if v, ok := d.GetOk("sha_pinning_required"); ok` guard in `resourceGithubActionsOrganizationPermissionsCreateOrUpdate` returns `ok = false` when the value is `false` (the zero value for `bool`), so `SHAPinningRequired` is never set on the API request payload. The GitHub API receives no value and leaves the existing setting unchanged.

This causes **perpetual drift**: every `terraform plan` shows `sha_pinning_required = true -> false`, but `terraform apply` never actually changes it.

The same bug exists in `resource_github_actions_repository_permissions.go` with the identical `d.GetOk` pattern.

### Terraform Version

```
Terraform v1.14.3
on darwin_arm64
+ provider registry.terraform.io/integrations/github v6.11.0
```

### Affected Resource(s)

- `github_actions_organization_permissions`
- `github_actions_repository_permissions`

```hcl
resource "github_actions_organization_permissions" "actions_permissions" {
  allowed_actions      = "all"
  enabled_repositories = "all"
  sha_pinning_required = false
}
```

### Steps to Reproduce

1. Have an organization where `sha_pinning_required` is currently `true` (e.g. set via the UI or a previous apply with `true`).
2. Set `sha_pinning_required = false` in the Terraform configuration.
3. Run `terraform plan` — it correctly shows `sha_pinning_required = true -> false`.
4. Run `terraform apply` — it reports success.
5. Run `terraform plan` again — it shows the same `true -> false` diff again (perpetual drift).

### Debug Output

The root cause is in [`resource_github_actions_organization_permissions.go`](https://github.com/integrations/terraform-provider-github/blob/v6.11.0/github/resource_github_actions_organization_permissions.go#L156-L158) (introduced in #2870):

```go
// Bug: d.GetOk() returns ok=false for zero-value bools (false),
// so sha_pinning_required=false is never sent to the API.
if v, ok := d.GetOk("sha_pinning_required"); ok {
    actionsPermissions.SHAPinningRequired = github.Ptr(v.(bool))
}
```

This is the well-known Terraform SDK `GetOk` + zero-value footgun. A fix would be:

```go
if d.HasChange("sha_pinning_required") || d.IsNewResource() {
    actionsPermissions.SHAPinningRequired = github.Ptr(d.Get("sha_pinning_required").(bool))
}
```

The identical pattern exists in [`resource_github_actions_repository_permissions.go`](https://github.com/integrations/terraform-provider-github/blob/v6.11.0/github/resource_github_actions_repository_permissions.go#L134-L136) and needs the same fix.

### Code of Conduct

sheeeng:
Split `CreateOrUpdate` into separate `Create` and `Update` functions for both `github_actions_organization_permissions` and `github_actions_repository_permissions` resources.

In `Create`, replace `d.GetOk()` with `d.GetOkExists()` for `sha_pinning_required`. The field is Optional and Computed, so `GetOk` returns `(false, false)` when a user explicitly writes `sha_pinning_required = false`, which is  indistinguishable from omitting the field. `GetOkExists` returns `(false, true)`, correctly detecting the explicit value.

In `Update`, use `d.HasChange()` + `d.Get()`, which avoids the zero-value problem entirely by checking the state diff.

Add two-step acceptance tests (true → false) to verify the fix.

Fix https://github.com/integrations/terraform-provider-github/issues/3223.

----

### Before the change?

- Setting `sha_pinning_required = false` applied successfully in Terraform, but not actually changed in GitHub resources itself.

### After the change?

- Setting `sha_pinning_required = false` applied successfully in Terraform, and actually changed in GitHub resources itself.

### Pull request checklist

- [ ] Yes
- [x] No

----


github/resource_github_actions_organization_permissions.go
```diff
@@ -158,8 +158,8 @@ func resourceGithubActionsOrganizationPermissionsCreateOrUpdate(d *schema.Resour
 		EnabledRepositories: &enabledRepositories,
 	}
 
-	if v, ok := d.GetOk("sha_pinning_required"); ok {
-		actionsPermissions.SHAPinningRequired = github.Ptr(v.(bool))
+	if d.HasChange("sha_pinning_required") || d.IsNewResource() {
```
deiga: AFAIK this would set an unconfigured value as false instead of relying on the API default. I wonder if it'd be simpler to change the field to have Default: false and change GetOk to just Get without conditional. 

I guess there shouldn't be any downside to sending this field on every call?

github/resource_github_actions_organization_permissions_test.go
```diff
@@ -104,6 +104,76 @@ func TestAccGithubActionsOrganizationPermissions(t *testing.T) {
...
+			}
+		`, enabledRepositories)
+
+		resource.Test(t, resource.TestCase{
+			PreCheck:          func() { skipUnlessHasOrgs(t) },
+			ProviderFactories: providerFactories,
+			Steps: []resource.TestStep{
+				{
+					Config: config,
+					Check: resource.ComposeTestCheckFunc(
```
deiga: Please use `ConfigStateChecks` or `ConfigPlanChecks` instead of `Check` https://developer.hashicorp.com/terraform/plugin/testing/acceptance-tests/plan-checks/resource

github/resource_github_actions_organization_permissions_test.go
```diff
@@ -104,6 +104,76 @@ func TestAccGithubActionsOrganizationPermissions(t *testing.T) {
...
+			}
+		`, enabledRepositories)
+
+		configFalse := fmt.Sprintf(`
+			resource "github_actions_organization_permissions" "test" {
+				allowed_actions      = "all"
+				enabled_repositories = "%s"
+				sha_pinning_required = false
+			}
+		`, enabledRepositories)
```
deiga, line 165: Please use one template string and multiple `fmt.Sprintf` calls to build config

deiga (changes requested):

github/resource_github_actions_organization_permissions.go
sheeeng: Thanks for the review. Implemented with [fd7ac91](https://github.com/integrations/terraform-provider-github/commit/fd7ac9183d5787c5c5f27b87df0c6d2323bfd487).

github/resource_github_actions_organization_permissions_test.go
sheeeng: Thanks for the review. Implemented with [fd7ac91](https://github.com/integrations/terraform-provider-github/commit/fd7ac9183d5787c5c5f27b87df0c6d2323bfd487).

github/resource_github_actions_organization_permissions_test.go
sheeeng: Thanks for the review. Implemented with [fd7ac91](https://github.com/integrations/terraform-provider-github/commit/fd7ac9183d5787c5c5f27b87df0c6d2323bfd487).

deiga (approved):
LGTM! @stevehipwell WDYT?

stevehipwell (changes requested):
@sheeeng I thought the whole point here was that SHA pinning was dependent on organization configuration? This change would not fix the issue, but would break this resource.

I'm not sure how I missed this in review, but I'll repeat my suggestion to look at how `allow_forking` works in the repository resource. To fix this you should split the create and update functions in the resource to simplify the logic. Then in create you can use `d.GetOkExists()` to see if it should be sent. In update you need to check `d.Changed()` and if it has changed you need to use `d.Get()` and then send that value.

github/resource_github_actions_organization_permissions.go
```diff
@@ -12,9 +12,9 @@ import (
 
 func resourceGithubActionsOrganizationPermissions() *schema.Resource {
 	return &schema.Resource{
-		Create: resourceGithubActionsOrganizationPermissionsCreateOrUpdate,
+		Create: resourceGithubActionsOrganizationPermissionsCreate,
 		Read:   resourceGithubActionsOrganizationPermissionsRead,
-		Update: resourceGithubActionsOrganizationPermissionsCreateOrUpdate,
+		Update: resourceGithubActionsOrganizationPermissionsUpdate,
 		Delete: resourceGithubActionsOrganizationPermissionsDelete,
```
deiga, line 18: While you're in here, please refactor these to use the Context-aware functions

github/resource_github_actions_organization_permissions.go
```diff
@@ -201,6 +198,67 @@ func resourceGithubActionsOrganizationPermissionsCreateOrUpdate(d *schema.Resour
...
+		orgName,
+		actionsPermissions)
+	if err != nil {
+		return err
+	}
+
+	if allowedActions == "selected" {
+		actionsAllowedData := resourceGithubActionsOrganizationAllowedObject(d)
+		if actionsAllowedData != nil {
+			log.Printf("[DEBUG] Update `actionsAllowedData` configuration.")
```
deiga: Please use tflog.Debug instead

github/resource_github_actions_organization_permissions.go
```diff
@@ -201,6 +198,67 @@ func resourceGithubActionsOrganizationPermissionsCreateOrUpdate(d *schema.Resour
...
+		}
+		_, err = client.Actions.SetEnabledReposInOrg(ctx,
+			orgName,
+			enabledReposData)
+		if err != nil {
+			return err
+		}
+	}
+
+	d.SetId(orgName)
```
deiga: Update should not be updating resource Id

github/resource_github_actions_organization_permissions.go
```diff
@@ -201,6 +198,67 @@ func resourceGithubActionsOrganizationPermissionsCreateOrUpdate(d *schema.Resour
...
+	return resourceGithubActionsOrganizationPermissionsRead(d, meta)
```
deiga, line 272: Please set any computed fields in Update, instead of calling Read

```suggestion
	return nil
```

github/resource_github_actions_repository_permissions.go
```diff
@@ -11,9 +11,9 @@ import (
 
 func resourceGithubActionsRepositoryPermissions() *schema.Resource {
 	return &schema.Resource{
-		Create: resourceGithubActionsRepositoryPermissionsCreateOrUpdate,
+		Create: resourceGithubActionsRepositoryPermissionsCreate,
 		Read:   resourceGithubActionsRepositoryPermissionsRead,
-		Update: resourceGithubActionsRepositoryPermissionsCreateOrUpdate,
+		Update: resourceGithubActionsRepositoryPermissionsUpdate,
 		Delete: resourceGithubActionsRepositoryPermissionsDelete,
```
deiga, line 17: Please refactor to use the Context-aware functions

deiga (changes requested):
Partial review. Apply comments suggestions to all applicable sections

github/resource_github_actions_organization_permissions.go
sheeeng: Thanks for the review! Include the suggestion now.

github/resource_github_actions_organization_permissions.go
sheeeng: Thanks for the review! Include the suggestion now.

github/resource_github_actions_repository_permissions.go
sheeeng: Thanks for the review! Include the suggestion now.

github/resource_github_actions_organization_permissions.go
sheeeng: Thanks for the review! Include the suggestion now.

github/resource_github_actions_organization_permissions.go
sheeeng: Good catch! Thanks for the review! Include the suggestion now.

github/resource_github_actions_organization_permissions.go
```diff
@@ -158,61 +155,125 @@ func resourceGithubActionsOrganizationPermissionsCreateOrUpdate(d *schema.Resour
...
 			enabledReposData)
 		if err != nil {
-			return err
+			return diag.FromErr(err)
 		}
 	}
 
 	d.SetId(orgName)
-	return resourceGithubActionsOrganizationPermissionsRead(d, meta)
+	return resourceGithubActionsOrganizationPermissionsRead(ctx, d, meta)
```
deiga:
```suggestion
	return nil
```

github/resource_github_actions_repository_permissions.go
```diff
@@ -141,39 +138,95 @@ func resourceGithubActionsRepositoryPermissionsCreateOrUpdate(d *schema.Resource
...
 			}
 		} else {
-			log.Printf("[DEBUG] Allowed actions config not set, skipping")
+			tflog.Debug(ctx, "Skip setting allowed actions configuration because none is specified.")
 		}
 	}
 
 	d.SetId(repoName)
-	return resourceGithubActionsRepositoryPermissionsRead(d, meta)
+	return resourceGithubActionsRepositoryPermissionsRead(ctx, d, meta)
```
deiga:
```suggestion
	return nil
```

deiga (changes requested):
The code looks like you're using an AI to generate it and it's lost all notion of what is supposed to happen. 

Please go through the code yourself and understand for yourself what it should be doing.

If the next time I come to review this and you're still showing disregard for the comments that have been left earlier, then I'll close the PR.

It's not worth anybody's time for us to play Broken Telephone from us to your AI.

github/resource_github_actions_repository_permissions.go
sheeeng: Thank you for the review. I have included this suggestion in [7ca2562](https://github.com/integrations/terraform-provider-github/pull/3224/commits/7ca2562e7698e5462cdea2b3b287aa1b0492258f).

github/resource_github_actions_organization_permissions.go
sheeeng: Thank you. I have included this suggestion in [7ca2562](https://github.com/integrations/terraform-provider-github/pull/3224/commits/7ca2562e7698e5462cdea2b3b287aa1b0492258f).

sheeeng:
I did a rebase from default branch. The pull request is updated with a force-push.

PR #3225, opened 2026-02-25
feat: Refactor secret values

stevehipwell:
Resolves #ISSUE_NUMBER

----

### Before the change?

- Secret value naming was confusing
- Encrypted secret value didn't require `key_id` to be passed in

### After the change?

- Secret value naming is consistent (`value` & `value_encrypted`)
- Secret value names ready for `value_wo` pattern
- `value_encrypted` requires `key_id` to be passed in

### Pull request checklist

- [ ] Yes
- [x] No

----


github/resource_github_actions_organization_secret_migration_test.go
```diff
@@ -18,20 +18,20 @@ func Test_resourceGithubActionsOrganizationSecretStateUpgradeV0(t *testing.T) {
...
-				"visibility":      "private",
-				"created_at":      "2023-01-01T00:00:00Z",
-				"updated_at":      "2023-01-01T00:00:00Z",
-				"plaintext_value": "secret-value",
+				"id":          "test-secret",
+				"secret_name": "test-secret",
+				"visibility":  "private",
+				"created_at":  "2023-01-01T00:00:00Z",
+				"updated_at":  "2023-01-01T00:00:00Z",
+				"value":       "secret-value",
```
deiga: question: Isn't rawstate the "input" value? That should be following schema v0 in this test, right?

Then this change would be problematic as v0 doesn't have a "value" field

stevehipwell: Good spot, thanks.

stevehipwell:
@robert-crandall please could you give this a review?

stevehipwell:
@robert-crandall the analysis you've got above is muddled, Copilot I assume? I added the `key_id` and made it conflict with `plaintext_value` when I fixed the secrets. I wanted to make it required for `encrypted_value`, as technically it is and without it the value could be un-decryptable, but that would have been a breaking change. This PR add the **NEW** `value_encrypted` and made `key_id` required when this is used. No changes outside of the deprecation for the old fields.

PR #3229, opened 2026-02-26
feat: add org invite resource

### LINKED ISSUE #2197: [FEAT]: github_membership invite using the email address
### Describe the need

Would it be possible to have a membership invite to an organisation for someone that does not have a github account yet? It will let people run the tf provider without waiting for the end user to create their github account.

https://registry.terraform.io/providers/integrations/github/latest/docs/resources/membership

### SDK Version

_No response_

### API Version

_No response_

### Relevant log output

_No response_

### Code of Conduct

Gowiem:
Resolves #2197

This is in response to @deiga's request. 

Fair warning: This was mostly letting Claude Code rip on that issue thread. I did however test that this works in a local project and using the resource with an email did show me 2 validation errors (no seats + user already in org) and then with a new email, it was able to add that user and I was able to get an invite email to that user's email address. 

Happy to address any feedback!

----

### Before the change?

There was no means by which to invite a user to the organization via email. You had to invite them via their github handle. 

- 

### After the change?

You can now invite users to the organization via an email. 

- 

### Pull request checklist

  - Not sure on this one, but I don't imagine migrations are needed. 

- [ ] Yes
- [x] No

----

### DIFF
--- a/github/resource_github_organization_invitation.go
+++ b/github/resource_github_organization_invitation.go
@@ -1,0 +1,199 @@
+package github
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"net/http"
+	"strconv"
+
+	"github.com/google/go-github/v83/github"
+	"github.com/hashicorp/terraform-plugin-log/tflog"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+)
+
+func resourceGithubOrganizationInvitation() *schema.Resource {
+	return &schema.Resource{
+		Description: "Invite a user to a GitHub organization by email address or GitHub user ID.",
+
+		CreateContext: resourceGithubOrganizationInvitationCreate,
+		ReadContext:   resourceGithubOrganizationInvitationRead,
+		DeleteContext: resourceGithubOrganizationInvitationDelete,
+
+		Schema: map[string]*schema.Schema{
+			"email": {
+				Description:  "The email address of the person to invite to the organization. Exactly one of `email` or `invitee_id` must be set.",
+				Type:         schema.TypeString,
+				Optional:     true,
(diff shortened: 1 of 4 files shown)


github/resource_github_organization_invitation.go
```diff
@@ -0,0 +1,199 @@
...
+	if err != nil {
+		return diag.FromErr(err)
+	}
+
+	invitationID := invitation.GetID()
+	d.SetId(strconv.FormatInt(invitationID, 10))
+
+	if err = d.Set("invitation_id", int(invitationID)); err != nil {
+		return diag.FromErr(err)
+	}
```
deiga, line 92: This seems superfluous, since we are setting the same as the resource ID alrady

github/resource_github_organization_invitation.go
```diff
@@ -0,0 +1,199 @@
...
+
+	client := meta.(*Owner).v3client
+	orgName := meta.(*Owner).name
+
+	invitationID, err := strconv.ParseInt(d.Id(), 10, 64)
+	if err != nil {
+		return diag.FromErr(err)
+	}
+
+	ctx = context.WithValue(ctx, ctxId, d.Id())
```
deiga, line 114: This isn't needed
```suggestion
```

github/resource_github_organization_invitation.go
```diff
@@ -0,0 +1,199 @@
...
+	}
+
+	if invitation == nil {
+		// Invitation was accepted, cancelled, or expired — remove from state
+		tflog.Info(ctx, fmt.Sprintf("Removing organization invitation %s from state because it is no longer pending", d.Id()), map[string]any{
+			"invitation_id": d.Id(),
+		})
+		d.SetId("")
+		return nil
+	}
```
deiga, line 149: This seems problematic as this breaks idempotency. The resource would automatically be removed and next plan would try to re-add it again

PR #3233, opened 2026-02-26
feat: Refactor repository collaborators

### LINKED ISSUE #2874: [BUG]: `github_repository_collaborators` error on repository rename
### Expected Behavior

The `github_repository_collaborators` has the property `repository` which is used as the resource ID.

But changing a the `repository` property on a `github_repository_collaborators` can mean two things:
- the collaborators are no longer assigned to the previous repository, and are now assigned to the new repository
- the repository has been renamed

The ID should be set to the repository ID to properly delete the collaborators only of the ID changed.

### Actual Behavior

The first case will work, the second will fail because it can no longer remove that repository by name from the collaborators.

### Terraform Version

Terraform v1.13.3
on darwin_arm64
+ provider registry.terraform.io/integrations/github v6.6.0

### Affected Resource(s)

- github_repository_collaborators

```hcl

```

### Steps to Reproduce

_No response_

### Debug Output

```shell

```

### Panic Output

```shell

```

### Code of Conduct

stevehipwell:
Resolves #2874

----

### Before the change?

- `github_repository_collaborators` errored if the repo was renamed
- `github_repository_collaborators` prefered team ID and made extra calls if slug was used
- `github_repository_collaborators` used unnecessary API calls to re-read
- Team logic was inconsistent
- Team resources didn't use native context

### After the change?

- `github_repository_collaborators` supports the repository being renamed
- `github_repository_collaborators` uses the minimum number of API calls
- Team logic is consistent and context is being used in more places

### Pull request checklist

- [ ] Yes
- [x] No

----

### DIFF
--- a/github/resource_github_repository_collaborators.go
+++ b/github/resource_github_repository_collaborators.go
@@ -2,32 +2,49 @@ package github
 
 import (
 	"context"
+	"errors"
 	"fmt"
-	"log"
 	"slices"
-	"sort"
 	"strconv"
+	"strings"
 
 	"github.com/google/go-github/v84/github"
+	"github.com/hashicorp/go-cty/cty"
+	"github.com/hashicorp/terraform-plugin-log/tflog"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/customdiff"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
 )
 
 func resourceGithubRepositoryCollaborators() *schema.Resource {
 	return &schema.Resource{
-		Create: resourceGithubRepositoryCollaboratorsCreate,
-		Read:   resourceGithubRepositoryCollaboratorsRead,
-		Update: resourceGithubRepositoryCollaboratorsUpdate,
-		Delete: resourceGithubRepositoryCollaboratorsDelete,
+		SchemaVersion: 1,
+		StateUpgraders: []schema.StateUpgrader{
+			{
+				Type:    resourceGithubRepositoryCollaboratorsV0().CoreConfigSchema().ImpliedType(),
+				Upgrade: resourceGithubRepositoryCollaboratorsStateUpgradeV0,
+				Version: 0,
+			},
+		},
+
+		CreateContext: resourceGithubRepositoryCollaboratorsCreate,
+		ReadContext:   resourceGithubRepositoryCollaboratorsRead,
+		UpdateContext: resourceGithubRepositoryCollaboratorsUpdate,
+		DeleteContext: resourceGithubRepositoryCollaboratorsDelete,
 		Importer: &schema.ResourceImporter{
-			StateContext: schema.ImportStatePassthroughContext,
+			StateContext: resourceGithubRepositoryCollaboratorsImport,
 		},
 
 		Schema: map[string]*schema.Schema{
 			"repository": {
-				Type:     schema.TypeString,
-				Required: true,
-				ForceNew: true,
+				Type:        schema.TypeString,
+				Required:    true,
+				Description: "Name of the repository.",
+			},
+			"repository_id": {
+				Type:        schema.TypeInt,
+				Computed:    true,
+				Description: "ID of the repository.",
 			},
 			"user": {
 				Type:        schema.TypeSet,
@@ -35,16 +52,16 @@ func resourceGithubRepositoryCollaborators() *schema.Resource {
 				Description: "List of users.",
 				Elem: &schema.Resource{
 					Schema: map[string]*schema.Schema{
-						"permission": {
-							Type:     schema.TypeString,
-							Optional: true,
-							Default:  "push",
-						},
 						"username": {
 							Type:             schema.TypeString,
 							Description:      "(Required) The user to add to the repository as a collaborator.",
 							Required:         true,
 							DiffSuppressFunc: caseInsensitive(),
+						},
+						"permission": {
+							Type:     schema.TypeString,
+							Optional: true,
+							Default:  "push",
 						},
 					},
 				},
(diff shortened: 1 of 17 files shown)


github/resource_github_repository_collaborators.go
```diff
@@ -93,132 +108,405 @@ func resourceGithubRepositoryCollaborators() *schema.Resource {
...
+		users := d.Get("user").(*schema.Set).List()
+		seen := make(map[string]any)
+
+		for _, u := range users {
+			user := u.(map[string]any)
+			username := user["username"].(string)
+
+			if _, ok := seen[username]; ok {
+				return fmt.Errorf("duplicate username %s found in user collaborators", username)
+			}
```
deiga, line 130: question: Since "user" is `TypeSet` is it even possible to add duplicates? Oh, only if their hashes match

We could use the `Set` schema field to use only the `username` to calculate the hash, then this check wouldn't be necessary. 

NOTE: Though, it doesn't error or complain, it just silently adds the first(?) of the duplicates

```
Set: func(v interface{}) int {
    m := v.(map[string]interface{})
    return schema.HashString(m["username"].(string))
  },
```

github/resource_github_repository_collaborators.go
```diff
@@ -93,132 +108,405 @@ func resourceGithubRepositoryCollaborators() *schema.Resource {
...
+				}
+
+				teamID := val.AsString()
+				if _, ok := seen[teamID]; ok {
+					return fmt.Errorf("duplicate team %s found in team collaborators", teamID)
+				}
+				seen[teamID] = nil
+			}
+		}
+	}
```
deiga, line 158: question: This seems way too complex for the benefit. Would it make sense to simplify by using the `Set:` field to change the Hash calculation?

github/util_team.go
```diff
@@ -0,0 +1,152 @@
...
+	if id, ok := parseTeamID(teamID); ok {
+		t.id = &id
+	} else {
+		t.slug = &teamID
+	}
+
+	return t
+}
+
+// gtID returns the team ID as an int64 if it's available, or 0 if it's not.
```
deiga:
```suggestion
// getID returns the team ID as an int64 if it's available, or 0 if it's not.
```

github/resource_github_repository_collaborators.go
stevehipwell: We can't use `Set` as it errors if the values aren't known at plan time. I did spend a bit of time testing this out and seeing if I could work around it, but this behaviour (and better validation) needs the framework.

github/resource_github_repository_collaborators.go
stevehipwell: As above, this isn't possible with the plugin SDK.

stevehipwell:
@robert-crandall please could you review this PR?

github/resource_github_repository_collaborators.go
```diff
@@ -284,359 +609,155 @@ func listTeams(client *github.Client, isOrg bool, ctx context.Context, owner, re
...
-	}
-
-	ignoreTeamIds, err := getIgnoreTeamIds(d, meta)
-	if err != nil {
-		return err
-	}
-
-	userCollaborators, invitations, teamCollaborators, err := listAllCollaborators(client, isOrg, ctx, owner, repoName, ignoreTeamIds)
-	if err != nil {
-		return deleteResourceOn404AndSwallow304OtherwiseReturnError(err, d, "repository collaborators (%s/%s)", owner, repoName)
```
robert-crandall, line 522: The old code removed repos from state when that repository was deleted. Is that still handled anywhere?

github/resource_github_repository_collaborators.go
```diff
@@ -93,132 +108,405 @@ func resourceGithubRepositoryCollaborators() *schema.Resource {
...
-			collaborators, resp, err := client.Repositories.ListCollaborators(ctx,
-				owner, repoName, opt)
+			collaborators, resp, err := client.Repositories.ListCollaborators(ctx, owner, repoName, opt)
 			if err != nil {
 				return nil, err
 			}
 
 			for _, c := range collaborators {
-				// owners are listed in the collaborators list even though they don't have direct permissions
-				if !isOrg && c.GetLogin() == owner {
```
robert-crandall, line 216: This check is now removed. So, on personal repositories, the old state would not include owners in the list of collaborators, and the new state will. Just making sure this is intentional.

github/resource_github_repository_collaborators.go
```diff
@@ -93,132 +108,405 @@ func resourceGithubRepositoryCollaborators() *schema.Resource {
...
+	users := d.Get("user").(*schema.Set).List()
+	teams := d.Get("team").(*schema.Set).List()
+	ignoreTeams := d.Get("ignore_team").(*schema.Set).List()
+
+	inUsers, err := getUserCollaborators(users)
+	if err != nil {
+		return diag.FromErr(err)
 	}
 
-	for _, invite := range invites {
```
robert-crandall, line 136: From what I understand, the old state included unaccepted invitations, and the new state does not. This seems like an improvement to me! However, just making sure this is intentional, since it's a state change.

stevehipwell: This change only adds to the state, nothing is removed (see line 316 in the new code). The code has just been improved. I did look at including the invitation ID in the user and deprecating the invitations but I think this looked like more trouble than it was worth at this time.

github/resource_github_repository_collaborators.go
stevehipwell: The old code isn't right as it would cause on perpetual diff if you added yourself in as a collaborator, but I agree that the new code does change the behaviour. The whole `isOrg` and `owner` pattern is fundamentally flawed, but that's a problem for another day.

I'll add some logic in the user reconciliation to skip removing the owner of a non-org repo.

github/resource_github_repository_collaborators.go
stevehipwell: No, intentionally so here as the only place it's valid to not error is in the read function (in all other places in the lifecycle it should error as TF is misconfigured). I'll add the code to remove the resource from state in the read.

stevehipwell:
@robert-crandall this should be good for a re-review.

Issue #2936, opened 2025-11-23
[MAINT]: Naming of organization custom properties needs the change (from plural to singular)


ViacheslavKudinov:
### Describe the need

Currently data and resource use `properties` (plural) in the naming, but in reality both work only with a single property.

To follow the naming of other resources and properties naming should be changed to `property` (singular).

Resource [organization_custom_properties](https://registry.terraform.io/providers/integrations/github/latest/docs/resources/organization_custom_properties ) 
Data [organization_custom_properties](https://registry.terraform.io/providers/integrations/github/latest/docs/data-sources/organization_custom_properties)
both use attribute `property_name ` which is only a single one and both can't work with more than a single custom property.

It probable the use case when data and resource should be renamed to **organization_custom_property** (singular),
plurals need to be marked as deprecated within version `v6` of provider and  probably removed in `v7`. 

### Relevant log output

```shell

```

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

mroyme:
**Minor suggestion:**
We could also consider changing `property_name` to `name`.

I know that `property_name` was probably used because it maps 1:1 with the parameters in the API, but it just seems a bit wonky to have it since it is already implied.

mkushakov:
#3234 covers most of this, but not all of it, so flagging what's in and what isn't before anyone closes it.

Addressed:

- A singular resource and data source, `github_organization_repository_custom_property`, written against `ARCHITECTURE.md`.
- The plural `github_organization_custom_properties` pair deprecated with a `DeprecationMessage` pointing at the new name, CRUD otherwise untouched — the deprecate-in-v6 half of the plan in the description.

Not addressed:

- The new name is `organization_repository_custom_property` rather than `organization_custom_property`, per @stevehipwell's [review feedback](https://github.com/integrations/terraform-provider-github/pull/3234#issuecomment-4629880469): these definitions apply to repositories, and the extra segment keeps them clearly related to `github_repository_custom_property`.
- `property_name` is kept rather than shortened to `name` (@mroyme's suggestion above). It matches the API field, and changing it would mean carrying a second rename through the deprecated pair as well.
- Removal of the plural pair, which the description itself defers to v7.

No strong preference on whether this closes when #3234 merges or stays open to track the v7 removal — it's currently linked as a plain reference, not a closing keyword, so it won't auto-close either way.

deiga:
Resolving this is fine, we have "remove all deprecations" issue for v7 somewhere

[...]

PR #3234, opened 2026-02-27
fix: improve github_organization_custom_properties resource

### LINKED ISSUE #2806: [BUG]: GitHub_organization_custom_properties
### Expected Behavior

If I want to create a custom property with mutli_select (organization wide) I cannot set mutiple default values. 

resource "github_organization_custom_properties" "repo_props" {
  property_name       = "temp1"
  value_type = "multi_select"
  required      = true
  description = "BlaBla"
  allowed_values = [
    "b1",
    "b2",
    "b3"                 
  ]
  default_value = ["b1", "b2"] # does not work because terraform expects a string here but would work with the api
  default_value = "b1,b2" # does not work as GitHub expects a array here. 
}

It's therefore not possible to set 2 default values as far as I see as there is a mismatch between terraform and the GitHub API. 

### Actual Behavior

I think I should be able to set a list or a string here... string for single select and list for multi select

### Terraform Version

Version 6.7

### Affected Resource(s)

github_organization_custom_properties

```hcl

```

### Steps to Reproduce

_No response_

### Debug Output

```shell

```

### Panic Output

```shell

```

### Code of Conduct

### LINKED ISSUE #3191: Proposal: Separate `resource_github_organization_custom_properties` into separate resources per `value_type`
### Describe the need

The schema doesn't currently map properly to the different `value_type`s and their possibilities (for example: we can't set default values that are arrays https://github.com/integrations/terraform-provider-github/issues/2806) 

Instead of making the single resource very complex with conditional logic and some weird schema fields, I'd propose we create separate Org Custom Property resource for each `value_type` 

### SDK Version

_No response_

### API Version

_No response_

### Relevant log output

```shell

```

### Code of Conduct

### LINKED ISSUE #3580: [BUG]: github_organization_custom_properties of type `true_false` detects drift on `default_value` indefinitely
### Expected Behavior

With following configuration planned and applied. 
```
resource "github_organization_custom_properties" "foo" {
  property_name = "foo"
  value_type    = "true_false"
  required      = true
 
  default_value = "true"
}
```

any additional plan results in no drift detected. 

### Actual Behavior

An additional plan will result in following drift: 

```
  ~ update in-place

  # github_organization_custom_properties.stedisentry_enabled will be updated in-place
  ~ resource "github_organization_custom_properties" "foo" {
      + default_value      = "true"
        id                 = "foo"
    }

```

### Terraform Version

Terraform: `1.15.7`
GitHub provider: `6.12.1`

### GitHub Installation Type

- [ ] I don't know

### Affected Resource(s)

- github_organization_custom_properties

```hcl
resource "github_organization_custom_properties" "foo" {
  property_name = "foo"
  value_type    = "true_false"
  required      = true
 
  default_value = "true"
}
```

### Steps to Reproduce

_No response_

### Debug Output

```shell

```

### Code of Conduct

### LINKED ISSUE #3641: github_organization_custom_properties Read doesn't handle 404, breaking recovery after deletion
## Description

`resourceGithubCustomPropertiesRead` in [`resource_github_organization_custom_properties.go`](https://github.com/integrations/terraform-provider-github/blob/v6.13.0/github/resource_github_organization_custom_properties.go#L113-L135) propagates any error from `client.Organizations.GetCustomProperty` verbatim, without checking for a 404 (not found) response:

```go
func resourceGithubCustomPropertiesRead(d *schema.ResourceData, meta any) error {
	ctx := context.Background()
	ownerName := meta.(*Owner).name

	customProperty, _, err := client.Organizations.GetCustomProperty(ctx, ownerName, d.Get("property_name").(string))
	if err != nil {
		return err   // <-- no 404 handling; should clear state (d.SetId("")) and return nil
	}
	...
}
```

This violates the standard Terraform provider idiom: when `Read` gets a 404 for a resource that no longer exists (e.g., because it was deleted out-of-band, or because a prior `Delete` succeeded but something is re-verifying), it should call `d.SetId("")` and return `nil`, signaling "no longer exists" rather than a hard error. The closely related **repo-scoped** sibling, [`resource_github_repository_custom_property.go`](https://github.com/integrations/terraform-provider-github/blob/v6.13.0/github/resource_github_repository_custom_property.go#L154-L166), correctly does:

```go
    // clears state
}
```

## Impact

Any tool that relies on this resource's Read returning a clean "not found" for a deleted `github_organization_custom_properties` resource gets stuck: after a successful `Delete`, the next `Read`/refresh gets a 404 from GitHub (correctly, since the delete succeeded), but the 404 is returned as an error instead of "resource gone."

## Steps to reproduce

1. `terraform import` or create a `github_organization_custom_properties` resource.
2. Delete the underlying property directly via the GitHub API (or via `terraform destroy`, then attempt a subsequent plan/refresh).
3. Observe that `Read` returns a hard error (`404 Not Found`) instead of clearing resource state.

## Expected behavior

`Read` should detect a 404 from `GetCustomProperty` and call `d.SetId("")`, returning `nil`, consistent with `resource_github_repository_custom_property.go`'s handling.

## Suggested fix

```go
customProperty, resp, err := client.Organizations.GetCustomProperty(ctx, ownerName, d.Get("property_name").(string))
if err != nil {
		d.SetId("")
		return nil
	}
	return err
}
```

## Environment

- `terraform-provider-github` v6.13.0

mkushakov:
Resolves #2806
Resolves #3191
Resolves #3580
Resolves https://github.com/integrations/terraform-provider-github/issues/3641
Refs #2936

----

### Before the change?

`github_organization_custom_properties` was the only way to manage org-level custom property *definitions*, and it had problems that could not be fixed in place without breaking existing state:

- The plural name did not match the behaviour — it manages a single property at a time (#2936).
- `default_value` was a plain string, so a `multi_select` property could not express its list-valued default at all (#2806), and the schema did not map cleanly onto the five `value_type`s (#3191). It also never round-tripped for `true_false` or `multi_select`: `DefaultValueString()` returns `("", false)` for those types and the read path discarded the `ok`, writing `""` into state and producing a permanent `+ default_value` diff on every plan (#3580).
- `value_type` was optional even though the API requires it, and `property_name`/`value_type` were not `ForceNew` despite being immutable server-side.
- No 404 handling in `Read`, so drift outside Terraform errored instead of removing the resource from state; a `CustomizeDiff` referencing a non-existent `slug` field; a redundant double-read in `Update`; and no validation of `allowed_values` against `value_type`.

Per @stevehipwell's [review feedback](https://github.com/integrations/terraform-provider-github/pull/3234#issuecomment-4629880469), this PR adds a correctly-named replacement rather than patching the old resource.

### After the change?

**New resource and data source `github_organization_repository_custom_property`** (singular), written from scratch against `ARCHITECTURE.md` and the `github_repository_custom_property` refactor in #3476:

- `*Context` CRUD returning `diag.Diagnostics`, `tflog` structured logging, and `checkOrganizationOK` on every entry point.
- `errors.AsType[*github.ErrorResponse]` 404 handling: `Read` removes the property from state, `Delete` treats it as success.
- `ForceNew` on the immutable `property_name` and `value_type`.
- Plan-time validation, so misconfigurations fail before reaching the API: `value_type` and `values_editable_by` against their allowed sets, non-empty `property_name` and non-empty list elements, `allowed_values` required for `single_select`/`multi_select` and rejected for the other types, `default_value` limited to one element for everything but `multi_select`, and `true_false` defaults constrained to exactly `"true"`/`"false"`.
- **`default_value` is a list of strings**, which is what makes #2806 and #3191 work. `github.CustomProperty.DefaultValue` is polymorphic — `multi_select` carries a JSON array, `true_false` a stringified bool, the rest plain strings — so requests send an array for `multi_select` and a bare string otherwise, and responses are normalised through the matching go-github accessor per `value_type`, raising an error rather than silently yielding `""`. Same "value is a collection, cardinality depends on the type" shape #3476 uses for `property_value`.
- `Timeouts` block, and an import function that seeds `property_name` from the import ID.
- `Description` on every schema attribute.

**Deprecation:** the existing `github_organization_custom_properties` resource and data source each gain a `DeprecationMessage` pointing at the new name. Their CRUD is otherwise untouched, so existing state and configurations keep working with only a warning.

**Shared helpers:** the polymorphic-value handling lives in a new `github/util_custom_property.go`. `parseRepositoryCustomPropertyValueToStringSlice` moves there too — it was already shared between `github_repository_custom_property` and its plural data source, so this is the only change to a pre-existing Go file beyond the deprecation notices and the provider registrations.

**Tests:** 18 `t.Run` subtests using `ConfigStateChecks`/`ConfigPlanChecks`, covering all five `value_type`s (including scalar and multi-element defaults), create/update/import, both `ForceNew` paths, every validation rule, out-of-band deletion — both that a subsequent plan recreates rather than errors, and that destroying an already-deleted property succeeds — and the `values_editable_by` scenarios. Fixtures are created through the API with randomized names so the suite is safe to run in parallel against an org that already has properties.

**Docs:** no per-resource doc templates are needed — both pages render from the repo-level `templates/{resources,data-sources}.md.tmpl` fallbacks plus the schema `Description`s. Examples are in the auto-discovered layout including `import.sh` and `import-by-string-id.tf`; `RESOURCES.md` and `ARCHITECTURE.md` list the new resource/data source and flag the deprecated pair.

> [!NOTE]
> On #3191: rather than splitting into one resource per `value_type` as originally proposed, the single resource now models every type correctly — including the array defaults that issue cites as its concrete example. Happy to revisit the split if you'd still prefer it.

> [!NOTE]
> One known limitation: a `default_value` can be changed but not removed. `CustomProperty.DefaultValue` is an `any` field with `omitempty`, so a nil value is omitted from the request rather than sent as JSON `null` and GitHub keeps the previous default. The attribute is `Computed`, so the old value is absorbed back into state without erroring. This is documented in the attribute description.

> [!NOTE]
> On #3580: the perpetual `true_false` drift is fixed in the new resource, which reads every default back through the accessor matching its `value_type`. The deprecated resource and data source keep their existing read path, so they still exhibit the drift until they are removed — anyone hitting that bug needs to move to the new name. Happy to patch the deprecated pair in place too, here or as a follow-up, if you'd rather not make migration a prerequisite for the fix.

### Pull request checklist

- [x] Schema migrations have been created if needed ([example](https://github.com/F-Secure-web/terraform-provider-github/blob/main/github/migrate_github_actions_organization_secret.go)) — _N/A: new resource, no existing state to migrate; the deprecated resource keeps its current schema._

- [ ] Yes
- [x] No

The old resource and data source remain fully functional and only emit a deprecation warning.

Edit: @deiga added new "Resolves" item


github/resource_github_organization_custom_properties.go
```diff
@@ -2,39 +2,38 @@ package github
...
-		Read:   resourceGithubCustomPropertiesRead,
-		Update: resourceGithubCustomPropertiesUpdate,
-		Delete: resourceGithubCustomPropertiesDelete,
+		Description: "Creates and manages a custom property for a GitHub Organization.",
+		Create:      resourceGithubCustomPropertiesCreate,
+		Read:        resourceGithubCustomPropertiesRead,
+		Update:      resourceGithubCustomPropertiesUpdate,
+		Delete:      resourceGithubCustomPropertiesDelete,
 		Importer: &schema.ResourceImporter{
 			State: resourceGithubCustomPropertiesImport,
```
deiga: Please refactor to use the Context functions

github/resource_github_organization_custom_properties.go
```diff
@@ -137,19 +167,22 @@ func resourceGithubCustomPropertiesRead(d *schema.ResourceData, meta any) error
 }
 
 func resourceGithubCustomPropertiesUpdate(d *schema.ResourceData, meta any) error {
-	if err := resourceGithubCustomPropertiesCreate(d, meta); err != nil {
-		return err
-	}
-	return resourceGithubCustomPropertiesRead(d, meta)
+	// Create uses the same upsert API, and already calls Read at the end
+	return resourceGithubCustomPropertiesCreate(d, meta)
```
deiga: issue: Never call any CRUD function

deiga (changes requested):

stevehipwell:
@mkushakov if you're still interested in contributing this change, I think it may be better to create a new resource named correctly (#2936) and then deprecate this resource. Off the top of my head I think the resource should be `github_organization_repository_custom_property` so as to not collide with the organization level custom properties (which should probably be `github_enterprise_organization_custom_property` & `github_organization_custom_property` respectively).

I've done a full refactor on the `github_repository_custom_property` resource in #3476, which you could use as a basis. Please note that we've updated the [contributing guidelines](https://github.com/integrations/terraform-provider-github?tab=contributing-ov-file#readme) and added an [architecture doc](https://github.com/integrations/terraform-provider-github/blob/main/ARCHITECTURE.md) to help contributors.

mkushakov:
> @mkushakov if you're still interested in contributing this change, I think it may be better to create a new resource named correctly (#2936) and then deprecate this resource. Off the top of my head I think the resource should be `github_organization_repository_custom_property` so as to not collide with the organization level custom properties (which should probably be `github_enterprise_organization_custom_property` & `github_organization_custom_property` respectively).
> 
> I've done a full refactor on the `github_repository_custom_property` resource in #3476, which you could use as a basis. Please note that we've updated the [contributing guidelines](https://github.com/integrations/terraform-provider-github?tab=contributing-ov-file#readme) and added an [architecture doc](https://github.com/integrations/terraform-provider-github/blob/main/ARCHITECTURE.md) to help contributors.

hello @stevehipwell thanks for suggestions, as you asked i have following in my PR:
- created a new resource and data source `github_organization_custom_property` (singular)
- marked existing `github_organization_custom_properties` as deprecated
Please let me know if you see any issues with my PR and I am open for suggestions. I have also updated PR body to reflect those changes

github/data_source_github_organization_repository_custom_property.go
```diff
@@ -0,0 +1,92 @@
...
+}
+
+func dataSourceGithubOrganizationRepositoryCustomPropertyRead(ctx context.Context, d *schema.ResourceData, meta any) diag.Diagnostics {
+	if err := checkOrganization(meta); err != nil {
+		return diag.FromErr(err)
+	}
+
+	client := meta.(*Owner).v3client
+	orgName := meta.(*Owner).name
+	propertyName := d.Get("property_name").(string)
```
stevehipwell:
```suggestion
func dataSourceGithubOrganizationRepositoryCustomPropertyRead(ctx context.Context, d *schema.ResourceData, m any) diag.Diagnostics {
	meta, _ := m.(*Owner)
	client := meta.v3client
	owner := meta.name

	if !meta.IsOrganization {
		return return diag.FromErr(fmt.Errorf("repository custom properties are only supported for organizations, %q is a user", owner))
	}

	propertyName := d.Get("property_name").(string)
```

github/data_source_github_organization_repository_custom_property.go
```diff
@@ -0,0 +1,92 @@
...
+		}
+		return diag.FromErr(fmt.Errorf("error reading organization custom property %q: %w", propertyName, err))
+	}
+
+	if !slices.Contains([]github.PropertyValueType{
+		github.PropertyValueTypeSingleSelect,
+		github.PropertyValueTypeMultiSelect,
+	}, cp.ValueType) {
+		cp.AllowedValues = nil
+	}
```
stevehipwell: I think a switch statement would work better here.

github/data_source_github_organization_repository_custom_property_test.go
```diff
@@ -0,0 +1,46 @@
...
+	"github.com/hashicorp/terraform-plugin-testing/helper/resource"
+	"github.com/hashicorp/terraform-plugin-testing/knownvalue"
+	"github.com/hashicorp/terraform-plugin-testing/statecheck"
+	"github.com/hashicorp/terraform-plugin-testing/tfjsonpath"
+)
+
+func TestAccGithubOrganizationRepositoryCustomPropertyDataSource(t *testing.T) {
+	const dataAddr = "data.github_organization_repository_custom_property.test"
+
+	t.Run("reads a property created by the resource", func(t *testing.T) {
```
stevehipwell: Can we make this test work correctly in an organization where there are already existing properties?

github/data_source_github_organization_repository_custom_property_test.go
```diff
@@ -0,0 +1,46 @@
...
+		config := `
+		resource "github_organization_repository_custom_property" "test" {
```
stevehipwell: Could you look at #3476 and copy the pattern for configuring the test dependencies outside of TF code?

github/data_source_github_organization_repository_custom_property.go
```diff
@@ -0,0 +1,92 @@
...
+
+	if err := setOrganizationRepositoryCustomPropertyState(d, cp); err != nil {
```
stevehipwell: I think for now I'd rather see the code duplicated in the method bodies, as if we move the logic we ought to have a single documented pattern for how to do it.

stevehipwell (changes requested):
Thanks for the changes @mkushakov, this looks like a great addition. As well as the inline comments I've got the following high level feedback.

- We need a `github_organization_repository_custom_properties` data source to replace `github_organization_custom_properties`
- Could you align your changes/patterns to #3476 (this should be "correct" in terms of the contribution guide)

abdonrd:
@mkushakov are you still going ahead with this PR? Thanks! 🙏

stevehipwell:
@mkushakov if you're unable to continue with this work, please let me know and I'll pick it up.

github/resource_github_organization_custom_properties.go
mkushakov: This comment targeted the in-place bug-fix approach on the old resource. Per @stevehipwell's suggestion below, that approach was replaced with a new resource (`github_organization_repository_custom_property`) instead, which uses `*Context` CRUD throughout. The old resource is now deprecated and left otherwise as-is so existing state keeps working — happy to revisit if you'd rather see it migrated too.

github/resource_github_organization_custom_properties.go
mkushakov: Same as above — this was flagging `Update` calling `Create` then `Read` on the old resource. That code path is unchanged now that the fix lives in the new `github_organization_repository_custom_property` resource, whose `Update` doesn't call another CRUD function.

github/data_source_github_organization_repository_custom_property.go
mkushakov: Done, with one deviation: I used the existing `checkOrganizationOK` helper (used by ~15 other data sources) instead of the inline `fmt.Errorf` check, since it returns `diag.Diagnostics` directly. Applied the same `m any` / `meta, _ := m.(*Owner)` / `owner := meta.name` shape to all 4 resource CRUD functions too, to match `ARCHITECTURE.md` and the #3476 pattern.

github/data_source_github_organization_repository_custom_property.go
mkushakov: Done — replaced `slices.Contains` with a `switch` in both the data source and the resource's `Read`.

github/data_source_github_organization_repository_custom_property_test.go
mkushakov: Fixed by creating the fixture via `mustCreateTestOrganizationRepositoryCustomProperty` with a randomized name instead of a hardcoded one in the TF config. Also applied the same fix to the resource's own acceptance test (`resource_github_organization_repository_custom_property_test.go`), which had the identical problem across all 10 subtests plus no `t.Parallel()`.

github/data_source_github_organization_repository_custom_property_test.go
mkushakov: Done — switched to `mustCreateTestOrganizationRepositoryCustomProperty(t, ...)`, matching the pattern in #3476.

github/data_source_github_organization_repository_custom_property.go
mkushakov: Done — removed `setOrganizationRepositoryCustomPropertyState` and inlined the `d.Set` calls in Create/Read/Update and the data source's Read.

mkushakov:
@stevehipwell / @abdonrd , sorry for delay. I have time to work on it and try to address all comments. Feel free to let me know if there is some missing function or issues.

github/resource_github_organization_repository_custom_property.go
```diff
@@ -0,0 +1,300 @@
...
+		CreateContext: resourceGithubOrganizationRepositoryCustomPropertyCreate,
+		ReadContext:   resourceGithubOrganizationRepositoryCustomPropertyRead,
+		UpdateContext: resourceGithubOrganizationRepositoryCustomPropertyUpdate,
+		DeleteContext: resourceGithubOrganizationRepositoryCustomPropertyDelete,
+		Importer: &schema.ResourceImporter{
+			StateContext: schema.ImportStatePassthroughContext,
+		},
+
+		CustomizeDiff: customdiff.All(resourceGithubOrganizationRepositoryCustomPropertyDiff),
+
```
deiga, line 40: Please add Timeouts block to enable user configured timeouts

github/resource_github_organization_repository_custom_property.go
```diff
@@ -0,0 +1,300 @@
...
+func resourceGithubOrganizationRepositoryCustomPropertyRead(ctx context.Context, d *schema.ResourceData, m any) diag.Diagnostics {
+	meta, _ := m.(*Owner)
+	if ok, diags := checkOrganizationOK(meta); !ok {
+		return diags
+	}
+
+	client := meta.v3client
+	owner := meta.name
+	propertyName := d.Id()
```
deiga: Please use the attribute getter here

github/resource_github_organization_repository_custom_property.go
```diff
@@ -0,0 +1,300 @@
...
+
+	cp, _, err := client.Organizations.GetCustomProperty(ctx, owner, propertyName)
+	if err != nil {
+		if ghErr, ok := errors.AsType[*github.ErrorResponse](err); ok && ghErr.Response.StatusCode == 404 {
+			tflog.Info(ctx, "Removing organization custom property from state because it no longer exists", map[string]any{"org": owner, "property": propertyName})
+			d.SetId("")
+			return nil
+		}
+		return diag.FromErr(fmt.Errorf("error reading organization custom property %q: %w", propertyName, err))
```
deiga: Let's use diag.Errorf

github/resource_github_organization_repository_custom_property.go
```diff
@@ -0,0 +1,300 @@
...
+	}
+
+	switch cp.ValueType {
+	case github.PropertyValueTypeSingleSelect, github.PropertyValueTypeMultiSelect:
+	default:
+		cp.AllowedValues = nil
+	}
+
+	defaultValue, _ := cp.DefaultValueString()
+	d.SetId(cp.GetPropertyName())
```
deiga, line 276: In theory it's possible for GetPropertyName to return an empty string. Would it make sense to check for that and produce a debugable error message?

github/resource_github_organization_repository_custom_property.go
```diff
@@ -0,0 +1,300 @@
...
+func resourceGithubOrganizationRepositoryCustomProperty() *schema.Resource {
+	return &schema.Resource{
+		Description: "Manages a GitHub organization custom property definition. Custom properties defined here can later be assigned values on individual repositories.",
+
+		CreateContext: resourceGithubOrganizationRepositoryCustomPropertyCreate,
+		ReadContext:   resourceGithubOrganizationRepositoryCustomPropertyRead,
+		UpdateContext: resourceGithubOrganizationRepositoryCustomPropertyUpdate,
+		DeleteContext: resourceGithubOrganizationRepositoryCustomPropertyDelete,
+		Importer: &schema.ResourceImporter{
+			StateContext: schema.ImportStatePassthroughContext,
```
deiga: Please create an import function

github/resource_github_organization_repository_custom_property_test.go
```diff
@@ -0,0 +1,327 @@
...
+			description    = "tf-acc-test single_select property"
+			allowed_values = ["one"]
+		}`, name)
+		configAfter := fmt.Sprintf(`
+		resource "github_organization_repository_custom_property" "test" {
+			property_name  = %[1]q
+			value_type     = "single_select"
+			description    = "tf-acc-test single_select property updated"
+			allowed_values = ["one", "two"]
+		}`, name)
```
deiga: Please use a single template string with modified values passed in

github/resource_github_organization_repository_custom_property_test.go
```diff
@@ -0,0 +1,327 @@
...
+						})),
+					},
+				},
+				{
+					Config: configAfter,
+					ConfigStateChecks: []statecheck.StateCheck{
+						statecheck.ExpectKnownValue(resourceAddr, tfjsonpath.New("allowed_values"), knownvalue.ListExact([]knownvalue.Check{
+							knownvalue.StringExact("one"),
+							knownvalue.StringExact("two"),
+						})),
```
deiga, line 185: Please add a configplancheck for the update

github/resource_github_organization_repository_custom_property_test.go
```diff
@@ -0,0 +1,327 @@
...
+		before := fmt.Sprintf(`
+		resource "github_organization_repository_custom_property" "test" {
+			property_name = %[1]q
+			value_type    = "string"
+		}`, nameBefore)
+		after := fmt.Sprintf(`
+		resource "github_organization_repository_custom_property" "test" {
+			property_name = %[1]q
+			value_type    = "string"
+		}`, nameAfter)
```
deiga: Please use a single template string

deiga (changes requested):
Partial review

github/resource_github_organization_repository_custom_property.go
```diff
@@ -0,0 +1,300 @@
...
+			"required": {
+				Type:        schema.TypeBool,
+				Optional:    true,
+				Description: "Whether the custom property must be set on every repository. When true, `default_value` must be provided.",
+			},
+			"default_value": {
+				Type:        schema.TypeString,
+				Optional:    true,
+				Computed:    true,
+				Description: "Default value applied to repositories that do not explicitly set the property.",
```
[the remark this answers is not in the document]
mkushakov: Good catch, and it was worse than described: `DefaultValueString()` returns (`"", false`) for both `multi_select` and `true_false` (it only handles `string`/`single_select`/`url`), and we discarded the `ok` bool. So `true_false` with a `default_value` didn't just lose data — it failed apply outright with "Provider produced inconsistent result after apply". Our own `resource_1.tf` example hit this.

Fixed in 729d750: `default_value` is now a `TypeList` of strings in both the resource and the data source, mirroring how #3476 models `property_value`. Requests send an array for `multi_select` and a bare string otherwise; responses go through a helper that picks the matching go-github accessor per `value_type`. Added `true_false` and `multi_select` acceptance subtests — both would have failed before this change.

github/data_source_github_organization_repository_custom_property.go
```diff
@@ -0,0 +1,111 @@
...
+			"required": {
+				Type:        schema.TypeBool,
+				Computed:    true,
+				Description: "Whether the custom property must be set on every repository.",
+			},
+			"default_value": {
+				Type:        schema.TypeString,
+				Computed:    true,
+				Description: "Default value applied to repositories that do not explicitly set the property.",
+			},
```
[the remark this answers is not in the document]
mkushakov: Fixed in the same commit. The data source uses the same helper, which returns an error instead of silently yielding `""` when a non-nil `default_value` can't be parsed. Also added a state check asserting the list is empty (not `[""]`) when a property has no default.

github/resource_github_organization_repository_custom_property.go
```diff
@@ -0,0 +1,300 @@
...
+	allowedValues, _ := d.Get("allowed_values").([]any)
+
+	selectType := valueType == github.PropertyValueTypeSingleSelect || valueType == github.PropertyValueTypeMultiSelect
+
+	if selectType && len(allowedValues) == 0 {
+		return fmt.Errorf("allowed_values is required when value_type is %q", valueType)
+	}
+	if !selectType && len(allowedValues) > 0 {
+		return fmt.Errorf("allowed_values must not be set when value_type is %q", valueType)
+	}
```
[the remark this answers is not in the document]
mkushakov: You're right that the description and the validation disagreed. I checked GitHub's REST docs though, and the `required` ⇒ `default_value` relationship isn't documented anywhere. Enforcing it at plan time would block configs that the API might well accept, which seems worse than the current behaviour of surfacing a 422 at apply. So I fixed the mismatch in the other direction — the description now says GitHub may reject `required = true` without a default, and there's no new `CustomizeDiff` rule. Happy to add the validation if maintainer think we should actually enforces it.

github/resource_github_organization_repository_custom_property.go
```diff
@@ -0,0 +1,300 @@
...
+	client := meta.v3client
+	owner := meta.name
+	propertyName := d.Id()
+
+	cp, _, err := client.Organizations.GetCustomProperty(ctx, owner, propertyName)
+	if err != nil {
+		if ghErr, ok := errors.AsType[*github.ErrorResponse](err); ok && ghErr.Response.StatusCode == 404 {
+			tflog.Info(ctx, "Removing organization custom property from state because it no longer exists", map[string]any{"org": owner, "property": propertyName})
+			d.SetId("")
+			return nil
```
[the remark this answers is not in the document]
mkushakov: Added in 1550dae — a subtest deletes the property via the API in `PreConfig`, then asserts the next plan is a `Create`, which only holds if `Read` cleared the ID instead of erroring.

github/resource_github_organization_repository_custom_property.go
mkushakov: Added, with all four of create/read/update/delete — the SDK rejects any key that isn't declared non-nil.

github/resource_github_organization_repository_custom_property.go
mkushakov: [Both](https://github.com/integrations/terraform-provider-github/pull/3234#discussion_r3764151510) done together, since they depend on each other: `Read` used `d.Id()` precisely because `ImportStatePassthroughContext` leaves `property_name` unset. There's now a real import function that seeds `property_name` from the ID, which lets `Read` use `d.Get("property_name")` like the other three CRUD functions.

github/resource_github_organization_repository_custom_property.go
mkushakov: Done throughout the resource and data source.

github/resource_github_organization_repository_custom_property.go
mkushakov: Added guards at all three `d.SetId(cp.GetPropertyName())` sites; they now fail with a message naming the org and the requested property instead of setting an empty ID.

github/resource_github_organization_repository_custom_property_test.go
mkushakov: Done — and applied to the other before/after subtests too (`value_type` ForceNew, `values_editable_by` update, and the out-of-band retain case), so the whole file uses one parameterised template per subtest.

github/resource_github_organization_repository_custom_property_test.go
mkushakov: Added `ExpectResourceAction(..., ResourceActionUpdate)` to that step, and to the other in-place update steps.

templates/data-sources/organization_repository_custom_property.md.tmpl
deiga, line 1: Since this doesn't have any custom contents, let's delete it and use the default template as fallback

templates/resources/organization_repository_custom_property.md.tmpl
deiga, line 1: I don't think the custom note brings value here, I'd rather add that in an example of we want to keep it. 

Since this doesn't have any custom contents, let's delete it and use the default template as fallback

github/resource_github_organization_repository_custom_property.go
```diff
@@ -0,0 +1,391 @@
...
+	}
+
+	return cp
+}
+
+// flattenOrganizationRepositoryCustomPropertyDefaultValue normalises the
+// polymorphic default_value returned by the API into a list of strings. The
+// wire type depends on value_type: multi_select is an array, true_false is a
+// stringified bool and the rest are plain strings.
+func flattenOrganizationRepositoryCustomPropertyDefaultValue(cp *github.CustomProperty) ([]string, error) {
```
deiga: Since the flatten function is used in multiple files, let's move it to a util file for this domain

deiga (changes requested):

github/resource_github_organization_repository_custom_property.go
mkushakov: Done in efb93dd — moved to a new `github/util_custom_property.go`. I left `buildOrganizationRepositoryCustomProperty` and the two value-type var blocks in the resource file since they're only used there.

Worth noting `parseRepositoryCustomPropertyValueToStringSlice` from #3476 is in the same situation — shared by two files but currently living in `data_source_github_repository_custom_properties.go`. Happy to move it into the same util file, but I've left it alone to keep this PR's diff scoped. Let me know if you'd prefer it folded in.

templates/resources/organization_repository_custom_property.md.tmpl
mkushakov: Split this into the two halves:

The note is gone from the template — the pointer to `github_repository_custom_property` and the API link now live in a comment at the top of `resource_1.tf` (1cf2a52), which is where `examples.instructions.md` says extra context belongs. Agreed it didn't earn its place in the template.

On deleting the template though — I checked what the fallback actually renders, and there's a wrinkle I don't think either of us accounted for. The tfplugindocs default template uses `page_title: "{{.Name}} {{.Type}} - ..."` **without** parentheses, whereas all other templates in this repo use `({{.Type}})`. It also prepends `# generated by https://github.com/hashicorp/terraform-plugin-docs` to the front matter, which currently appears in all other docs — because every page here has a template, so nothing has ever fallen back to the default.

So deleting these two would publish the only pages in the provider titled  `github_organization_repository_custom_property Resource - GitHub` rather than `… (Resource) - GitHub`. Everything else is preserved — subcategory, both examples, and both import sections all still render.

Both templates now contain nothing but that front matter, so the trade is purely "title consistency vs. two fewer files". Happy to delete them if you're fine with the title difference, or if you'd rather normalise it repo-wide as a follow-up — just say which and I'll push it. I've left them in place for now rather than guess.

deiga: You clearly didn't look at our default template, which is the one that would be used if you delete the added one: https://github.com/integrations/terraform-provider-github/blob/main/templates/resources.md.tmpl

github/resource_github_organization_repository_custom_property.go
deiga: Yeah, that's a good idea to move as well

templates/resources/organization_repository_custom_property.md.tmpl
mkushakov: You're right — I checked tfplugindocs' built-in default rather than `templates/resources.md.tmpl`. Both templates are deleted in d4427c0.

github/resource_github_organization_repository_custom_property.go
mkushakov: Done in 0c05a02 — `parseRepositoryCustomPropertyValueToStringSlice` now lives in `github/util_custom_property.go` next to the org-level flatten helper, and fmt is no longer needed in `data_source_github_repository_custom_properties.go`.

abdonrd:
> @stevehipwell / @abdonrd , sorry for delay. I have time to work on it and try to address all comments. Feel free to let me know if there is some missing function or issues.

Thanks for working on this, @mkushakov!

stevehipwell:
@mkushakov thanks for updating the PR, I'll try and find some time to take a look later on in the week.

github/resource_github_organization_repository_custom_property.go
```diff
@@ -0,0 +1,388 @@
...
+	if err := d.Set("required", cp.GetRequired()); err != nil {
+		return diag.FromErr(err)
+	}
+	if err := d.Set("default_value", defaultValue); err != nil {
+		return diag.FromErr(err)
+	}
+	if err := d.Set("description", cp.GetDescription()); err != nil {
+		return diag.FromErr(err)
+	}
+	if err := d.Set("allowed_values", cp.AllowedValues); err != nil {
```
abdonrd, line 231: `Read` clears `AllowedValues` for non-select types (lines 265-269) before setting state, but `Create`/`Update` don't. Since `allowed_values` isn't `Computed`, a non-empty value here for a non-select type could cause an "inconsistent result after apply" error. Consider reusing Read's sanitization.

github/resource_github_organization_repository_custom_property.go
```diff
@@ -0,0 +1,388 @@
...
+				Optional:         true,
+				Computed:         true,
+				Description:      fmt.Sprintf("Who can edit values of this property on repositories. One of: %v. Defaults to `org_actors` server-side.", organizationCustomPropertyValuesEditableBy),
+				ValidateDiagFunc: validation.ToDiagFunc(validation.StringInSlice(organizationCustomPropertyValuesEditableBy, false)),
+			},
+		},
+	}
+}
+
+func resourceGithubOrganizationRepositoryCustomPropertyDiff(ctx context.Context, d *schema.ResourceDiff, _ any) error {
```
abdonrd, line 109: This doesn't validate that `required = true` has a `default_value`, though the schema description (line 66) says GitHub may reject that combo. Right now it'd only fail at `apply`, not `plan`, unlike the other checks here.

github/resource_github_organization_repository_custom_property.go
```diff
@@ -0,0 +1,388 @@
...
+
+	if v, ok := d.GetOk("values_editable_by"); ok {
+		s := v.(string)
+		cp.ValuesEditableBy = &s
+	}
+
+	return cp
+}
+
+func resourceGithubOrganizationRepositoryCustomPropertyCreate(ctx context.Context, d *schema.ResourceData, m any) diag.Diagnostics {
```
abdonrd, line 189: nit: The API-call > validate > flatten > `SetId` > `d.Set(...)` block is duplicated in `Create`/`Read`/`Update` (and the data source). A shared helper could help.

github/resource_github_organization_repository_custom_property.go
```diff
@@ -0,0 +1,388 @@
...
+
+		CreateContext: resourceGithubOrganizationRepositoryCustomPropertyCreate,
+		ReadContext:   resourceGithubOrganizationRepositoryCustomPropertyRead,
+		UpdateContext: resourceGithubOrganizationRepositoryCustomPropertyUpdate,
+		DeleteContext: resourceGithubOrganizationRepositoryCustomPropertyDelete,
+		Importer: &schema.ResourceImporter{
+			StateContext: resourceGithubOrganizationRepositoryCustomPropertyImport,
+		},
+
+		CustomizeDiff: customdiff.All(resourceGithubOrganizationRepositoryCustomPropertyDiff),
```
abdonrd, line 39: nit: `customdiff.All(...)` around one function is unnecessary; could just be `CustomizeDiff: resourceGithubOrganizationRepositoryCustomPropertyDiff`.

Issue #1226, opened 2022-07-19
Provider hangs, without giving any logging or feedback


nfx:
when hit by a rate limit, github provider simply hangs, without reporting anything to debug logs or user message... at least it should tell it's rate limited...

nfx:
Update: will this be fixed?..

stevehipwell:
@kfcampbell this is still very much an issue.

pmartindev:
The provider does log if there is a rate limit, but only when [running in debug mode](https://github.com/integrations/terraform-provider-github/blob/66b1d20a2656b3dfadced4648a8afc837dcd94b7/github/transport.go#L104). I think this should be as simple as replacing the [DEBUG] with info if we want to provide this output all of the time?

deiga:
Relates to #2925

Issue #849, opened 2021-07-05
RateLimitTransport incompatible with v4 api


lijok:
v4, graphql api, uses POST for all operations, including read
RateLimitTransport sleeps for 1s between each POST/PATCH/PUT/DELETE operation
Say you have 100 repos with 3 branch protection rules each, the provider will spend 5 minutes sleeping

ayk33:
Anything that can be done about this?

lijok:
the provider needs to either not use v4 api, which is probably not wise long-term
or, rethink the RateLimitTransport

majormoses:
I think its important to keep the `RateLimitTransport` for any `v3` api interactions for larger organizations where we need to rely on this to get through plans in our largest states. In the graphql api I agree we need to think about rate limits very differently.

Here  is the docs from Github: https://docs.github.com/en/graphql/overview/resource-limitations#rate-limit

lijok:
@jcudit thanks for that
set `write_delay_ms` to 200, which has on average sped up our plan times 2x
<img width="154" alt="image" src="https://user-images.githubusercontent.com/28689084/136117341-09d15ff8-669c-402a-a523-512071444031.png">
Lowering this value further yields marginal gains (100ms = ~6m30s)

However, while the PR that introduces `write_delay_ms` option helps this problem, I think long term it's not the correct solution for it. I think the provider should have separate transports for v4 and v3 api's, keeping the rate limit logic as is on the v3 api, but modifying the v4 api transport to only invoke this write delay on mutations, rather than queries

Any thoughts?

majormoses:
> I think its important to keep the `RateLimitTransport` for any `v3` api interactions for larger organizations where we need to rely on this to get through plans in our largest states. In the graphql api I agree we need to think about rate limits very differently.
> 
> Here is the docs from Github: https://docs.github.com/en/graphql/overview/resource-limitations#rate-limit
> 
> What I think we want to do is create a generic rate limit transport and then let each rest and gql apis use the appropriate transport based on the calls.

If we go this route we can start with adding a `no-op` rate limit transport for GQL apis and use the existing rate limiter for REST.

shoddyguard:
This definitely still needs addressing

kfcampbell:
Perhaps we could use separate configuration for GraphQL reads and writes, like: 

```hcl
provider "github" {
  owner = "example-org"
  rest_write_delay_ms = 500 # sleep for 500ms in between requests to the GitHub REST API
[...]

Issue #2925, opened 2025-11-19
[MAINT]: Update transport implementation


stevehipwell:
### Describe the need

I'd like to look at swapping out the custom rate limit code in this provider with [gofri/go-github-ratelimit](https://github.com/gofri/go-github-ratelimit).

### Relevant log output

```shell

```

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

stevehipwell:
I've got a well tested GitHub client pattern that combines modern rate limiting with app token refresh, I'll be opening a PR so we can evaluate what this would look like in practice.

This client would allow the provider to wait on a primary rate limit, but as the primary rate limit is 60 mins and the provider context is canceled after 20 mins we'd need to add custom timeout configuration to all resources to make use of this.

deiga:
@stevehipwell should we create a similar tracking issue for the timeout addition or just a few big PRs?

stevehipwell:
@deiga I expect to see issues to track feature requests.

For this in particular I'd be looking at the other future work requiring changes to most/all resources (framework, owner, etc) and coming up with a plan for how to get there. I also think we need to get stability, consistency and testability dialled in first.

PR #3236, opened 2026-02-27
feat: add modern rate limiter option using go-github-ratelimit

### LINKED ISSUE #2709: [FEAT]: Bypass preemptive rate limit check  by go-github

mkushakov:
## Summary

Add a new `rate_limiter` provider configuration option that allows users to choose between two rate limiting strategies for GitHub API calls.

Resolves #2709
References #849, #1226

## Motivation

The current provider uses a custom rate limiting implementation with configurable `read_delay_ms` and `write_delay_ms` delays. While functional, this approach:

- Requires users to manually tune delay values
- Does not automatically respond to GitHub's rate limit headers (`Retry-After`, `X-RateLimit-*`)
- Can be either too aggressive (causing 429 errors) or too conservative (slowing down operations unnecessarily)

The [`go-github-ratelimit`](https://github.com/gofri/go-github-ratelimit) library provides automatic, header-based rate limit handling that respects GitHub's primary and secondary rate limits without manual configuration.

### Config (`config.go`)

- Added `RateLimiter` field to `Config` struct (accepts `"legacy"` or `"modern"`)
- Added `ModernRateLimitedHTTPClient` function that uses `go-github-ratelimit/v2` to create an HTTP client with automatic rate limit handling
- Updated `AuthenticatedHTTPClient` and `AnonymousHTTPClient` to dispatch to the appropriate rate limiter based on the `RateLimiter` setting

### Provider (`provider.go`)

- Added `rate_limiter` schema field with validation (`"legacy"` or `"modern"`, defaults to `"legacy"`)
- Wired `rate_limiter` value into the `Config` struct in `providerConfigure`

### Dependencies

- Added `github.com/gofri/go-github-ratelimit/v2 v2.0.2`

## Example Usage

```hcl
provider "github" {
  token        = var.github_token
  rate_limiter = "modern"  # Use automatic rate limit handling
}
```

When using `"modern"`, the `read_delay_ms`, `write_delay_ms`, and `parallel_requests` settings are ignored — the library automatically detects and waits for rate limit resets using response headers.

## Testing

- `go build ./...` — passes
- `go vet ./...` — passes
- The default is `"legacy"` to maintain backward compatibility — no existing configurations are affected


github/provider.go
```diff
@@ -93,6 +93,19 @@ func Provider() *schema.Provider {
...
+				Optional:    true,
+				Default:     "legacy",
+				Description: descriptions["rate_limiter"],
+				ValidateFunc: func(val any, key string) (warns []string, errs []error) {
+					v := val.(string)
+					if v != "legacy" && v != "modern" {
+						errs = append(errs, fmt.Errorf("%q must be either 'legacy' or 'modern', got: %s", key, v))
+					}
+					return
+				},
```
deiga:
```suggestion
				ValidateDiagFunc: validation.ToDiagFunc(validation.StringInSlice([]string{"legacy", "modern"})),
```

github/config.go
```diff
@@ -72,13 +74,31 @@ func RateLimitedHTTPClient(client *http.Client, writeDelay, readDelay, retryDela
 	return client
 }
 
+// ModernRateLimitedHTTPClient creates an HTTP client that uses go-github-ratelimit
+// for automatic GitHub API rate limit handling. When using this client, the
+// read_delay_ms, write_delay_ms, and parallel_requests settings are ignored
+// because rate limiting is handled automatically by the library.
+func ModernRateLimitedHTTPClient(client *http.Client, retryDelay time.Duration, retryableErrors map[int]bool, maxRetries int) *http.Client {
+	client.Transport = NewEtagTransport(client.Transport)
```
deiga, line 82:
```suggestion
	client.Transport = NewEtagTransport(client.Transport)
	client.Transport = logging.NewLoggingHTTPTransport(client.Transport)
```

deiga (changes requested):
This is great! Left a few comments :)

github/provider.go
mkushakov: oh, i have missed it, thanks

github/config.go
mkushakov: Added, also i have missed Stone Crop preview header which is added as well.

deiga (approved):
LGTM! @stevehipwell I know you've wanted to use `go-github-ratelimit`, what do you think about this?

stevehipwell (changes requested):
Thanks for the PR @mkushakov but we have #2925 open to discuss these changes before actually making them. There are a number of decision to be made before this should be implemented.

mkushakov:
> Thanks for the PR @mkushakov but we have #2925 open to discuss these changes before actually making them. There are a number of decision to be made before this should be implemented.

thanks for feedback, unfortunately i see no open discussion in the linked issue and this change is backward compatible since default is `legacy`
on top of it we are using this implementation on daily basis [in our fork](https://github.com/nexthink-oss/terraform-provider-github) for few month with no issue where we are managing around 20'000 objects in the state for GitHub provider
we just want to move out of our fork, this is why i have opened [series of PRs to backport all important changes](https://github.com/integrations/terraform-provider-github/pulls/mkushakov) we made since then

deiga:
@stevehipwell What's stopping us from having a discussion here in the PR?

stevehipwell:
@deiga thre are a number of considerations that need discussing around this. What was appropriate for a fork isn't necessarily the right choice moving forwards. For example we have a hard 20 min limit when using the context methods, so an implementation should take this into consideration. Also we have two clients with different rate limits requirements, which should be discussed further. When I quickly looked at this PR the SDK flag to disable it's rate limits logic hasn't been set and it didn't look like the callbacks where being used.

The issue is the placeholder for this discussion, I've got plenty to add but it hasn't seemed like a priority so I've been focusing on other things.

ktella:
@mkushakov  What happens when the modern rate limiting is reached. Do we have any fallback plans ? I was testing with a forked provider and ran into this issue. In our org we are managing 3000+ TF resources in Github. GH allowed 15K per hour im not sure if thats enought or not. We follow CI/CD process for this.

 Error: Get "https://api.github.com/users/xxxxxx-ccccccc": primary rate limit reached on request to https://api.github.com/users/xxxxxx-ccccccc with category: core. wait until 2026-03-30 18:21:55 +0000 UTC before sending more requests.
│

isometry:
> @mkushakov What happens when the modern rate limiting is reached.

With the current configuration of the modern rate limiter, hitting a secondary rate limit will trigger retries, whilst hitting primary rate limits is fatal.
This is *probably* sane, as GitHub App derived installation tokens are only valid for an hour anyway, so the token will likely have expired by the time rate limits reset, but that's debatable.

The upstream module, `github.com/gofri/go-github-ratelimit/v2`, includes a `github_primary_ratelimit.WithSleepUntilReset()` helper with which a sleeper *could* be implemented, but I'm not convinced that this is wise. Perhaps hidden behind another provider option, default disabled?

ktella:
> > Thanks for the PR @mkushakov but we have #2925 open to discuss these changes before actually making them. There are a number of decision to be made before this should be implemented.
> 
> thanks for feedback, unfortunately i see no open discussion in the linked issue and this change is backward compatible since default is `legacy` on top of it we are using this implementation on daily basis [in our fork](https://github.com/nexthink-oss/terraform-provider-github) for few month with no issue where we are managing around 20'000 objects in the state for GitHub provider we just want to move out of our fork, this is why i have opened [series of PRs to backport all important changes](https://github.com/integrations/terraform-provider-github/pulls/mkushakov) we made since then

@mkushakov Happy to hear that you are able to manage 20K+ resources in one state file. Somehow i got hit with ratelimit. Could you elaborate on your testing and setup ?

stevehipwell:
@ktella please use #2925 to discuss.

deiga:
Superseded by #3448

PR #3237, opened 2026-02-27
feat: add github_repository_custom_properties resource for batch property management

### LINKED ISSUE #3240: [FEAT]: add github_repository_custom_properties resource for batch property management
## Summary

Add a new `github_repository_custom_properties` resource (plural) that manages **multiple** custom property values for a single repository in one resource block, with in-place updates when values change.

## Motivation

The existing `github_repository_custom_property` (singular) resource manages a single property per resource instance. While this works well for independent lifecycle management, there are common use cases where managing all properties for a repository as a unit is preferred:

- **Atomic updates**: Set multiple property values in a single API call rather than one resource per property
- **Simplified configuration**: Manage all properties for a repo in a single resource block instead of N separate resources
- **In-place updates**: When a property value changes, the resource updates in-place rather than needing destroy/recreate
- **Clean imports**: Import all current property values from a repository in one `terraform import` command

## Proposed Schema

```hcl
resource "github_repository_custom_properties" "example" {
  repository_name = "my-repo"

  property {
    name  = "environment"
    value = ["production"]
  }

  property {
    name  = "languages"
    value = ["go", "typescript"]  # multi_select
  }
}
```

## Implementation Notes

- Uses the batch `CreateOrUpdateCustomProperties` API endpoint for efficient updates
- Auto-detects property types from org-level property definitions (no `property_type` field needed)
- Import format: `terraform import github_repository_custom_properties.example owner/repo-name`
- Complements (does not replace) the existing singular `github_repository_custom_property` resource

## Related

- Existing singular resource: `github_repository_custom_property`
- Organization property definitions: `github_organization_custom_properties`
- Issue #2936 (naming discussion for org custom properties plural vs singular)

mkushakov:
Resolves #3240

## Summary

Add a new `github_repository_custom_properties` resource (plural) that manages multiple custom property values on a single repository in one resource block, with in-place updates when values change.

This complements the existing `github_repository_custom_property` resource (singular), which manages one property per resource instance and requires destroy+recreate on any change.

## Motivation

The existing singular `github_repository_custom_property` resource has limitations:

- **All fields are ForceNew** — changing a property value destroys and recreates the resource, even though the GitHub API supports `PATCH`
- **One resource per property** — managing 10 properties on a repo requires 10 resource blocks
- **Requires `property_type`** — the user must manually specify the property type, even though it's already defined at the org level
- **One API call per property** — no batching

The new plural resource addresses all of these:

| Capability | Singular (`github_repository_custom_property`) | Plural (`github_repository_custom_properties`) |
|---|---|---|
| Properties per resource | 1 | N |
| Value change behavior | Destroy + recreate | In-place update |
| Requires `property_type` | Yes | No (auto-detected from org definitions) |
| API calls | 1 per property | 1 for all properties (batch) |
| Import | `owner/repo/property_name` | `owner/repo` (imports ALL properties) |

### New Resource (`resource_github_repository_custom_properties.go`)

- Schema with `repository_name` (ForceNew) and `property` (TypeSet of name/value blocks)
- `Create` and `Update` share the same function — looks up org property definitions to auto-detect types, then calls `CreateOrUpdateCustomProperties` with all properties in a single API call
- `Read` filters to only managed properties (from state), or imports ALL properties when state is empty (import case)
- `Delete` sets all managed properties to `nil` in a single API call
- Custom `StateContext` importer that accepts `owner/repo` format and imports all properties
- Hash function based on property name only (not values), so value changes are detected as in-place modifications
- `checkOrganization` guard on all CRUD operations
- Proper error wrapping with `fmt.Errorf` and `%w`

### Provider Registration (`provider.go`)

- Registered as `github_repository_custom_properties`

### Tests (`resource_github_repository_custom_properties_test.go`)

- **Creates and reads multiple properties** — sets two properties (single_select + string) and verifies
- **Updates property value in place** — changes a single_select value from "production" to "staging" without destroy
- **Imports all properties** — creates properties, then imports via `owner/repo`
- **Creates multi_select property** — verifies multi-value property handling

All tests use `providerFactories` (modern pattern) and `skipUnlessHasOrgs`.

## Example Usage

```hcl
resource "github_repository_custom_properties" "example" {
  repository_name = github_repository.example.name

  property {
    name  = "environment"
    value = ["production"]
  }

  property {
    name  = "team"
    value = ["platform"]
  }

  property {
    name  = "languages"
    value = ["go", "typescript"]  # multi_select
  }
}
```

## Relationship to Existing Resources

- **`github_repository_custom_property`** (singular) — unchanged, still available. Users can choose whichever pattern fits their workflow.
- **`github_organization_custom_properties`** — defines properties at the org level. The new resource reads these definitions to auto-detect property types.
- **`data.github_repository_custom_properties`** (data source) — already exists, reads all properties for a repo.

> **Note:** Using both singular and plural resources on the same repository and property is not recommended, as they would conflict. Users should choose one pattern per repository.

## Testing

- `go build ./...` — passes
- `go vet ./...` — passes

### DIFF
--- a/github/resource_github_repository_custom_properties.go
+++ b/github/resource_github_repository_custom_properties.go
@@ -1,0 +1,252 @@
+package github
+
+import (
+	"context"
+	"fmt"
+	"log"
+	"strings"
+
+	"github.com/google/go-github/v83/github"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+)
+
+func resourceGithubRepositoryCustomProperties() *schema.Resource {
+	return &schema.Resource{
+		Description: "Manages custom properties for a GitHub repository. This resource allows you to set multiple custom property values on a single repository in a single resource block, with in-place updates when values change.",
+		Create:      resourceGithubRepositoryCustomPropertiesCreateOrUpdate,
+		Read:        resourceGithubRepositoryCustomPropertiesRead,
+		Update:      resourceGithubRepositoryCustomPropertiesCreateOrUpdate,
+		Delete:      resourceGithubRepositoryCustomPropertiesDelete,
+		Importer: &schema.ResourceImporter{
+			StateContext: resourceGithubRepositoryCustomPropertiesImport,
+		},
+
+		Schema: map[string]*schema.Schema{
+			"repository_name": {
+				Type:        schema.TypeString,
+				Required:    true,
+				ForceNew:    true,
+				Description: "Name of the repository.",
+			},
+			"property": {
+				Type:        schema.TypeSet,
+				Required:    true,
+				MinItems:    1,
+				Description: "Set of custom property values for this repository.",
+				Elem: &schema.Resource{
+					Schema: map[string]*schema.Schema{
+						"name": {
+							Type:        schema.TypeString,
+							Required:    true,
+							Description: "Name of the custom property (must be defined at the organization level).",
+						},
+						"value": {
+							Type:        schema.TypeSet,
+							Required:    true,
+							MinItems:    1,
+							Description: "Value(s) of the custom property. For multi_select properties, multiple values can be specified.",
+							Elem: &schema.Schema{
+								Type: schema.TypeString,
+							},
+						},
+					},
+				},
+				Set: resourceGithubRepositoryCustomPropertiesHash,
+			},
+		},
+	}
+}
+
+// resourceGithubRepositoryCustomPropertiesHash creates a hash for a property block
+// using only the property name, so that value changes are detected as in-place
+// updates rather than remove+add within the set.
+func resourceGithubRepositoryCustomPropertiesHash(v any) int {
+	raw := v.(map[string]any)
+	name := raw["name"].(string)
+	return schema.HashString(name)
+}
+
+func resourceGithubRepositoryCustomPropertiesCreateOrUpdate(d *schema.ResourceData, meta any) error {
+	if err := checkOrganization(meta); err != nil {
+		return err
+	}
+
+	client := meta.(*Owner).v3client
+	ctx := context.Background()
+	owner := meta.(*Owner).name
+	repoName := d.Get("repository_name").(string)
+	properties := d.Get("property").(*schema.Set).List()
+
+	// Get all organization custom property definitions to determine types
+	orgProperties, _, err := client.Organizations.GetAllCustomProperties(ctx, owner)
+	if err != nil {
+		return fmt.Errorf("error reading organization custom property definitions: %w", err)
+	}
+
+	// Create a map of property names to their types
+	propertyTypes := make(map[string]github.PropertyValueType)
+	for _, prop := range orgProperties {
+		if prop.PropertyName != nil {
+			propertyTypes[*prop.PropertyName] = prop.ValueType
+		}
+	}
+
+	// Build custom property values for this repository
+	customProperties := make([]*github.CustomPropertyValue, 0, len(properties))
+
+	for _, propBlock := range properties {
+		propMap := propBlock.(map[string]any)
+		propertyName := propMap["name"].(string)
+		propertyValues := expandStringList(propMap["value"].(*schema.Set).List())
+
+		propertyType, ok := propertyTypes[propertyName]
+		if !ok {
+			return fmt.Errorf("custom property %q is not defined at the organization level", propertyName)
+		}
+
+		customProperty := &github.CustomPropertyValue{
+			PropertyName: propertyName,
+		}
+
+		switch propertyType {
+		case github.PropertyValueTypeMultiSelect:
+			customProperty.Value = propertyValues
+		case github.PropertyValueTypeString, github.PropertyValueTypeSingleSelect,
+			github.PropertyValueTypeTrueFalse, github.PropertyValueTypeURL:
+			if len(propertyValues) > 0 {
+				customProperty.Value = propertyValues[0]
+			}
+		default:
+			return fmt.Errorf("unsupported property type %q for property %q", propertyType, propertyName)
+		}
+
+		customProperties = append(customProperties, customProperty)
+	}
+
+	_, err = client.Repositories.CreateOrUpdateCustomProperties(ctx, owner, repoName, customProperties)
+	if err != nil {
+		return fmt.Errorf("error setting custom properties for repository %s/%s: %w", owner, repoName, err)
+	}
+
+	d.SetId(buildTwoPartID(owner, repoName))
+
+	return resourceGithubRepositoryCustomPropertiesRead(d, meta)
+}
+
+func resourceGithubRepositoryCustomPropertiesRead(d *schema.ResourceData, meta any) error {
+	if err := checkOrganization(meta); err != nil {
+		return err
+	}
+
+	client := meta.(*Owner).v3client
+	ctx := context.Background()
+
+	owner, repoName, err := parseTwoPartID(d.Id(), "owner", "repository")
+	if err != nil {
+		return err
+	}
+
+	// Get current properties from state to know which ones we're managing.
+	// On import this will be empty, which is handled below.
+	propertiesFromState := d.Get("property").(*schema.Set).List()
+	managedPropertyNames := make(map[string]bool)
+	for _, propBlock := range propertiesFromState {
+		propMap := propBlock.(map[string]any)
+		managedPropertyNames[propMap["name"].(string)] = true
+	}
+
+	isImport := len(managedPropertyNames) == 0
+
+	// Read actual properties from GitHub
+	allCustomProperties, _, err := client.Repositories.GetAllCustomPropertyValues(ctx, owner, repoName)
+	if err != nil {
+		return fmt.Errorf("error reading custom properties for repository %s/%s: %w", owner, repoName, err)
+	}
+
+	// Build the property set — either all properties (import) or only managed ones
+	managedProperties := make([]any, 0)
+	for _, prop := range allCustomProperties {
+		if !isImport && !managedPropertyNames[prop.PropertyName] {
+			continue
+		}
+
+		// Skip properties with nil/null values (unset)
+		if prop.Value == nil {
+			continue
+		}
+
+		propertyValue, err := parseRepositoryCustomPropertyValueToStringSlice(prop)
+		if err != nil {
+			return fmt.Errorf("error parsing property %q for repository %s/%s: %w", prop.PropertyName, owner, repoName, err)
+		}
+
+		if len(propertyValue) == 0 {
+			continue
+		}
+
+		managedProperties = append(managedProperties, map[string]any{
+			"name":  prop.PropertyName,
+			"value": propertyValue,
+		})
+	}
+
+	// If no properties exist, remove resource from state
+	if len(managedProperties) == 0 {
+		log.Printf("[WARN] No custom properties found for %s/%s, removing from state", owner, repoName)
+		d.SetId("")
+		return nil
+	}
+
+	_ = d.Set("repository_name", repoName)
+	_ = d.Set("property", managedProperties)
+
+	return nil
+}
+
+func resourceGithubRepositoryCustomPropertiesDelete(d *schema.ResourceData, meta any) error {
+	if err := checkOrganization(meta); err != nil {
+		return err
+	}
+
+	client := meta.(*Owner).v3client
+	ctx := context.Background()
+
+	owner, repoName, err := parseTwoPartID(d.Id(), "owner", "repository")
+	if err != nil {
+		return err
+	}
+
+	properties := d.Get("property").(*schema.Set).List()
+	if len(properties) == 0 {
+		return nil
+	}
+
+	// Set all managed properties to nil (removes them)
+	customProperties := make([]*github.CustomPropertyValue, 0, len(properties))
+	for _, propBlock := range properties {
+		propMap := propBlock.(map[string]any)
+		customProperties = append(customProperties, &github.CustomPropertyValue{
+			PropertyName: propMap["name"].(string),
+			Value:        nil,
+		})
+	}
+
+	_, err = client.Repositories.CreateOrUpdateCustomProperties(ctx, owner, repoName, customProperties)
+	if err != nil {
+		return fmt.Errorf("error deleting custom properties for repository %s/%s: %w", owner, repoName, err)
+	}
+
+	return nil
+}
+
+func resourceGithubRepositoryCustomPropertiesImport(ctx context.Context, d *schema.ResourceData, meta any) ([]*schema.ResourceData, error) {
+	// Import ID format: owner/repo (using standard two-part ID)
+	// On import, Read will detect empty state and import ALL properties
+	parts := strings.SplitN(d.Id(), "/", 2)
+	if len(parts) != 2 || parts[0] == "" || parts[1] == "" {
+		return nil, fmt.Errorf("invalid import ID %q, expected format: owner/repository", d.Id())
+	}
+
+	d.SetId(buildTwoPartID(parts[0], parts[1]))
+	return []*schema.ResourceData{d}, nil
+}
--- a/github/resource_github_repository_custom_properties.go
+++ b/github/resource_github_repository_custom_properties.go
@@ -3,30 +3,40 @@ package github
 import (
 	"context"
 	"fmt"
-	"log"
-	"strings"
 
-	"github.com/google/go-github/v83/github"
+	"github.com/google/go-github/v84/github"
+	"github.com/hashicorp/terraform-plugin-log/tflog"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/customdiff"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
 )
 
 func resourceGithubRepositoryCustomProperties() *schema.Resource {
 	return &schema.Resource{
 		Description: "Manages custom properties for a GitHub repository. This resource allows you to set multiple custom property values on a single repository in a single resource block, with in-place updates when values change.",
-		Create:      resourceGithubRepositoryCustomPropertiesCreateOrUpdate,
-		Read:        resourceGithubRepositoryCustomPropertiesRead,
-		Update:      resourceGithubRepositoryCustomPropertiesCreateOrUpdate,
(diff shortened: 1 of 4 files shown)


mkushakov:
@deiga / @stevehipwell , any chances for reviewing this PR?

deiga:
@mkushakov please don't ping us multiple times. 
We have our own lives and work, OSS is usually a best effort situation when there is capacity for it. 

If you currently have capacity for these PRs maybe you could review the last few merged PRs to look at what are things we usually ask for?

mkushakov:
> @mkushakov please don't ping us multiple times. We have our own lives and work, OSS is usually a best effort situation when there is capacity for it.
> 
> If you currently have capacity for these PRs maybe you could review the last few merged PRs to look at what are things we usually ask for?

sorry, don't want to be PITA, just want to see what is a way forward to backport most important changes from our fork so we can move back to this upstream
also i know there are a many people who also will benefit from this fixes/features

github/resource_github_repository_custom_properties.go
```diff
@@ -0,0 +1,252 @@
...
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+)
+
+func resourceGithubRepositoryCustomProperties() *schema.Resource {
+	return &schema.Resource{
+		Description: "Manages custom properties for a GitHub repository. This resource allows you to set multiple custom property values on a single repository in a single resource block, with in-place updates when values change.",
+		Create:      resourceGithubRepositoryCustomPropertiesCreateOrUpdate,
+		Read:        resourceGithubRepositoryCustomPropertiesRead,
+		Update:      resourceGithubRepositoryCustomPropertiesCreateOrUpdate,
+		Delete:      resourceGithubRepositoryCustomPropertiesDelete,
```
deiga: Please use Context-aware functions

github/resource_github_repository_custom_properties.go
```diff
@@ -0,0 +1,252 @@
...
+			StateContext: resourceGithubRepositoryCustomPropertiesImport,
+		},
+
+		Schema: map[string]*schema.Schema{
+			"repository_name": {
+				Type:        schema.TypeString,
+				Required:    true,
+				ForceNew:    true,
+				Description: "Name of the repository.",
+			},
```
deiga: Please change to "repository", add "repository_id" field and add diffRepository customizediff to support repo renames

github/resource_github_repository_custom_properties.go
```diff
@@ -0,0 +1,252 @@
...
+// resourceGithubRepositoryCustomPropertiesHash creates a hash for a property block
+// using only the property name, so that value changes are detected as in-place
+// updates rather than remove+add within the set.
+func resourceGithubRepositoryCustomPropertiesHash(v any) int {
+	raw := v.(map[string]any)
+	name := raw["name"].(string)
+	return schema.HashString(name)
+}
+
+func resourceGithubRepositoryCustomPropertiesCreateOrUpdate(d *schema.ResourceData, meta any) error {
```
deiga: Please separate Create and Update functions

github/resource_github_repository_custom_properties.go
```diff
@@ -0,0 +1,252 @@
...
+
+		customProperties = append(customProperties, customProperty)
+	}
+
+	_, err = client.Repositories.CreateOrUpdateCustomProperties(ctx, owner, repoName, customProperties)
+	if err != nil {
+		return fmt.Errorf("error setting custom properties for repository %s/%s: %w", owner, repoName, err)
+	}
+
+	d.SetId(buildTwoPartID(owner, repoName))
```
deiga: Please use buildID

github/resource_github_repository_custom_properties.go
```diff
@@ -0,0 +1,252 @@
...
+
+	return resourceGithubRepositoryCustomPropertiesRead(d, meta)
```
deiga: Please don't call CRUD functions directly. If there are computed values to be set, set them here

github/resource_github_repository_custom_properties.go
```diff
@@ -0,0 +1,252 @@
...
+
+func resourceGithubRepositoryCustomPropertiesRead(d *schema.ResourceData, meta any) error {
+	if err := checkOrganization(meta); err != nil {
+		return err
+	}
+
+	client := meta.(*Owner).v3client
+	ctx := context.Background()
+
+	owner, repoName, err := parseTwoPartID(d.Id(), "owner", "repository")
```
deiga: Please use parseID2

github/resource_github_repository_custom_properties.go
```diff
@@ -0,0 +1,252 @@
...
+
+		managedProperties = append(managedProperties, map[string]any{
+			"name":  prop.PropertyName,
+			"value": propertyValue,
+		})
+	}
+
+	// If no properties exist, remove resource from state
+	if len(managedProperties) == 0 {
+		log.Printf("[WARN] No custom properties found for %s/%s, removing from state", owner, repoName)
```
deiga: Please use tflog

github/resource_github_repository_custom_properties.go
```diff
@@ -0,0 +1,252 @@
...
+		d.SetId("")
+		return nil
+	}
+
+	_ = d.Set("repository_name", repoName)
+	_ = d.Set("property", managedProperties)
```
deiga: Please don't swallow errors

github/resource_github_repository_custom_properties.go
```diff
@@ -0,0 +1,252 @@
...
+
+func resourceGithubRepositoryCustomPropertiesDelete(d *schema.ResourceData, meta any) error {
+	if err := checkOrganization(meta); err != nil {
+		return err
+	}
+
+	client := meta.(*Owner).v3client
+	ctx := context.Background()
+
+	owner, repoName, err := parseTwoPartID(d.Id(), "owner", "repository")
```
deiga: Pleas use parseID2

github/resource_github_repository_custom_properties.go
```diff
@@ -0,0 +1,252 @@
...
+	return nil
+}
+
+func resourceGithubRepositoryCustomPropertiesImport(ctx context.Context, d *schema.ResourceData, meta any) ([]*schema.ResourceData, error) {
+	// Import ID format: owner/repo (using standard two-part ID)
+	// On import, Read will detect empty state and import ALL properties
+	parts := strings.SplitN(d.Id(), "/", 2)
+	if len(parts) != 2 || parts[0] == "" || parts[1] == "" {
+		return nil, fmt.Errorf("invalid import ID %q, expected format: owner/repository", d.Id())
+	}
```
deiga: Please change import format to use : separator. Then you can use parseID2

github/resource_github_repository_custom_properties_test.go
```diff
@@ -0,0 +1,243 @@
...
+			}),
+		)
+
+		resource.Test(t, resource.TestCase{
+			PreCheck:          func() { skipUnlessHasOrgs(t) },
+			ProviderFactories: providerFactories,
+			Steps: []resource.TestStep{
+				{
+					Config: config,
+					Check:  check,
```
deiga: Please refactor to use ConfigStsteChecks and inline the variable

deiga (changes requested):
Partial review

github/resource_github_repository_custom_properties.go
```diff
@@ -0,0 +1,334 @@
...
+}
+
+// resourceGithubRepositoryCustomPropertiesHash creates a hash for a property block
+// using only the property name, so that value changes are detected as in-place
+// updates rather than remove+add within the set.
+func resourceGithubRepositoryCustomPropertiesHash(v any) int {
+	raw := v.(map[string]any)
+	name := raw["name"].(string)
+	return schema.HashString(name)
+}
```
deiga, line 79: question: Could you explain further why this is needed?

github/resource_github_repository_custom_properties.go
```diff
@@ -0,0 +1,334 @@
...
+	if err != nil {
+		return diag.FromErr(err)
+	}
+
+	ctx = tflog.SetField(ctx, "id", d.Id())
+
+	client := meta.(*Owner).v3client
+	owner := meta.(*Owner).name
+
+	_, repoName, err := parseID2(d.Id())
```
deiga: suggestion: Instead of parsing the repo name from Id it's recommended to use d.Get the field since it exists

github/resource_github_repository_custom_properties.go
```diff
@@ -0,0 +1,334 @@
...
+
+	isImport := len(managedPropertyNames) == 0
+
+	// Read actual properties from GitHub
+	allCustomProperties, _, err := client.Repositories.GetAllCustomPropertyValues(ctx, owner, repoName)
+	if err != nil {
+		return diag.FromErr(fmt.Errorf("error reading custom properties for repository %s/%s: %w", owner, repoName, err))
+	}
+
+	managedProperties, err := filterManagedCustomProperties(allCustomProperties, managedPropertyNames, isImport)
```
deiga, line 221: issue: this makes the resource non-authoritative which can encourage patterns that cause constant drift and confusion. 
What do you think about this?

github/resource_github_repository_custom_properties.go
```diff
@@ -0,0 +1,334 @@
...
+		propMap := propBlock.(map[string]any)
+		managedPropertyNames[propMap["name"].(string)] = true
+	}
+
+	isImport := len(managedPropertyNames) == 0
+
+	// Read actual properties from GitHub
+	allCustomProperties, _, err := client.Repositories.GetAllCustomPropertyValues(ctx, owner, repoName)
+	if err != nil {
+		return diag.FromErr(fmt.Errorf("error reading custom properties for repository %s/%s: %w", owner, repoName, err))
```
deiga: You can use diag.Errorf

github/resource_github_repository_custom_properties.go
```diff
@@ -0,0 +1,334 @@
...
+	if err != nil {
+		return diag.FromErr(fmt.Errorf("error processing custom properties for repository %s/%s: %w", owner, repoName, err))
+	}
+
+	// If no properties exist, remove resource from state
+	if len(managedProperties) == 0 {
+		tflog.Warn(ctx, "No custom properties found, removing from state", map[string]any{"owner": owner, "repository": repoName})
+		d.SetId("")
+		return nil
+	}
```
deiga, line 231: question: I'm wondering if this makes sense as it will cause the resource to be recreated unnecessarily?
If this was an authorative resource I could fathom the purpose. 
Could you elaborate on the use-case here?

github/resource_github_repository_custom_properties.go
```diff
@@ -0,0 +1,334 @@
...
+	managedPropertyNames := make(map[string]bool)
+	for _, propBlock := range propertiesFromState {
+		propMap := propBlock.(map[string]any)
+		managedPropertyNames[propMap["name"].(string)] = true
+	}
+
+	isImport := len(managedPropertyNames) == 0
+
+	// Read actual properties from GitHub
+	allCustomProperties, _, err := client.Repositories.GetAllCustomPropertyValues(ctx, owner, repoName)
```
deiga, line 210: issue: How does the API react when the repo is gone? If it returns an error then we need to remove this resource from state instead of returning the error to the user

github/resource_github_repository_custom_properties.go
```diff
@@ -0,0 +1,334 @@
...
+func resourceGithubRepositoryCustomPropertiesDelete(ctx context.Context, d *schema.ResourceData, meta any) diag.Diagnostics {
+	err := checkOrganization(meta)
+	if err != nil {
+		return diag.FromErr(err)
+	}
+
+	client := meta.(*Owner).v3client
+	owner := meta.(*Owner).name
+
+	_, repoName, err := parseID2(d.Id())
```
deiga: suggestion: it's advisable to use state values instead of parsing the Id

website/docs/r/repository_custom_properties.html.markdown
```diff
@@ -0,0 +1,90 @@
...
+}
+```
+
+## Argument Reference
+
+The following arguments are supported:
+
+* `repository` - (Required) The name of the repository.
+
+* `repository_id` - The ID of the GitHub repository (computed).
```
deiga: issue: This is not an argument, please move it to attributes

github/resource_github_repository_custom_properties_test.go
```diff
@@ -0,0 +1,233 @@
...
+				repository = github_repository.test.name
+
+				property {
+					name  = github_organization_custom_properties.test.property_name
+					value = ["production"]
+				}
+			}
+		`, propName, repoName)
+
+		configUpdate := fmt.Sprintf(`
```
deiga: issues: Instead of having 2 almost identical `fmt.Sprintf` calls, create one template and use `fmt.Sprintf` to format it.
And don't store it in separate variables

github/resource_github_repository_custom_properties_test.go
deiga, line 1: issue: Using `SetSizeExact` in the tests is hazardous as the resource is non-authorative. It's possible that there are other properties configured in the Test Org

github/resource_github_repository_custom_properties_test.go
```diff
@@ -0,0 +1,233 @@
...
+			Steps: []resource.TestStep{
+				{
+					Config: configCreate,
+					ConfigStateChecks: []statecheck.StateCheck{
+						statecheck.ExpectKnownValue(testCustomPropsResourceAddr, tfjsonpath.New("property"), knownvalue.SetSizeExact(1)),
+					},
+				},
+				{
+					Config: configUpdate,
+					ConfigStateChecks: []statecheck.StateCheck{
```
deiga, line 123: issue: Please add Plan checks to verify that it does in-place updates https://developer.hashicorp.com/terraform/plugin/testing/acceptance-tests/plan-checks/resource

github/resource_github_repository_custom_properties_test.go
```diff
@@ -0,0 +1,233 @@
...
+			}
+		`, propName, repoName)
+
+		resource.Test(t, resource.TestCase{
+			PreCheck:          func() { skipUnlessHasOrgs(t) },
+			ProviderFactories: providerFactories,
+			Steps: []resource.TestStep{
+				{
+					Config: config,
+					ConfigStateChecks: []statecheck.StateCheck{
```
deiga, line 209: issue: The checks don't verify that it's a multi-select property, should they?

github/resource_github_repository_custom_properties.go
```diff
@@ -0,0 +1,334 @@
...
+		customProperties = append(customProperties, &github.CustomPropertyValue{
+			PropertyName: propMap["name"].(string),
+			Value:        nil,
+		})
+	}
+
+	_, err = client.Repositories.CreateOrUpdateCustomProperties(ctx, owner, repoName, customProperties)
+	if err != nil {
+		return diag.FromErr(fmt.Errorf("error deleting custom properties for repository %s/%s: %w", owner, repoName, err))
+	}
```
deiga, line 305: issue: This also needs to check if the repo is deleted or archived, in which case the error is not an error

github/resource_github_repository_custom_properties.go
```diff
@@ -0,0 +1,334 @@
...
+
+	_, repoName, err := parseID2(d.Id())
+	if err != nil {
+		return diag.FromErr(err)
+	}
+
+	properties := d.Get("property").(*schema.Set).List()
+	if len(properties) == 0 {
+		return nil
+	}
```
deiga: question: What use case does this guard against? The Schema defines `property` as min 1 item, so we should never have 0 properties, right?

deiga (changes requested):

github/resource_github_repository_custom_properties.go
mkushakov: By default, Terraform computes the hash of a set element using all its fields (`name` + `value`). This means if you change a property's value, the hash changes, and Terraform treats it as remove the old element + add a new element — effectively a destroy+recreate at the set-element level.

By providing a custom `Set` function that hashes only on `name`, two elements with the same property name but different values produce the same hash. Terraform then sees the element as still present (same identity) but with modified attributes, which triggers an in-place update via the resource's Update function rather than a remove+add cycle.

This is the mechanism that enables the resource's core design goal: changing a property value (e.g., `environment` = `"production"` → `"staging"`) results in an update API call rather than delete+recreate. Without this custom hash, the `TypeSet` would lose the "same element, different value" semantics and the in-place update behavior wouldn't work.

github/resource_github_repository_custom_properties.go
mkushakov: This guard handles the edge case where all managed properties have been removed externally (e.g., the org-level property definitions were deleted). Since the resource has nothing left to manage, removing it from state makes Terraform surface the discrepancy on the next plan — the user sees the resource will be recreated, which prompts them to either fix the org definitions or remove the resource from their config.

Without this, the Read would silently succeed with an empty property set, which would conflict with the `MinItems: 1` constraint and could cause confusing plan output.

That said, I'm open to alternative approaches — e.g., returning a warning diagnostic instead of removing from state, if you think that would be less surprising.

github/resource_github_repository_custom_properties.go
mkushakov: The resource is intentionally non-authoritative: it only manages the properties explicitly declared in the `property {}` blocks and ignores everything else on the repository.

Why non-authoritative:
- Organizations often have custom properties managed by multiple teams, tools, or projects. Authoritative behavior would force all properties into a single resource block, which is impractical and fragile.
- An authoritative resource that deletes unmanaged properties on apply can be destructive, especially during gradual adoption.
- This matches how the existing singular `github_repository_custom_property` already works — it doesn't touch other properties either.
- Many mature providers (e.g. GCP IAM bindings, AWS) offer non-authoritative variants for exactly this reason.

Regarding drift: only the declared properties are tracked, so there's no "phantom drift" from externally managed properties. If someone changes a managed property outside Terraform, the next plan correctly detects and fixes it.

I've updated the documentation and the resource description to explicitly call out the non-authoritative behavior so it's clear to users. See [this](https://github.com/integrations/terraform-provider-github/pull/3237/changes/18c09b525e844f48d91822f2c725ee0621e4289a) commit.

github/resource_github_repository_custom_properties.go
mkushakov: fixed in https://github.com/integrations/terraform-provider-github/pull/3237/changes/03d7c07ee055a9ab11ca41ba3e1410c9e23fd2cc

Issue #3116, opened 2026-01-24
[MAINT]: Rework provider auth configuration


deiga:
### Describe the need

Currently there is some amount of hidden magic involved when figuring out auth for a terraform execution.

1. Get token from ENV
2. Get app_auth from configuration 
3. If app_auth block, try to get app installation token 
4. If for any reason token would be empty try to use gh CLI to fetch token, but ignore any errors 
5. Set token to Meta, no matter if it has a value or not

This makes it quite difficult to actually run anonymously if there is a GITHUB_TOKEN in ENV or gh CLI has any valid auth token. 

Additionally there is magic logic for how, at least repo names are being parsed based on auth mode. Token based auth will fetch username from API and ~always use it as owner for any resource, which means that repo names are not full names (<owner>/<reponame>). And for Organizations/Enterprises Owner needs to be set in provider config and repo names etc. need to be full names (<owner>/<reponame>). This leads to very confusing behaviour

### Proposed rework:

1. Make every provider config settable from environment, to allow switching between token and app_auth with the same configuration. 
2. Change auth configuration to have all options as explicit options
	- auth_mode: anonymous, token, app_auth
  	- environment_variable_prefix: to enable having multiple providers with env vars configured 
 	- remove app_auth block an move app_auth variables to top-level or enable setting app_auth via env variable 
  	- auth_type: user, organization, enterprise
3. Ensure that token and app_auth always use the same logic for determining owner

### Related issues & PRs:
- We currently have no way of informing the user when no token is set, but they want to do authenticated actions: https://github.com/integrations/terraform-provider-github/issues/598
- There used to be a provider configuration option `anonymous: true`, but it's been removed at some point https://github.com/integrations/terraform-provider-github/issues/354
- Sometimes the used auth isn't what the user/API is expecting: https://github.com/integrations/terraform-provider-github/issues/769

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

stevehipwell:
I'd suggest removing the CLI from the scope of this and also deprecate it. This would align with the other changes.

deiga:
> I'd suggest removing the CLI from the scope of this and also deprecate it. This would align with the other changes.

Totally fine with that, it should be trivial for users to use the CLI to populate the token env var if they want to use it. 

That would be an additional simplification.

Issue #949, opened 2021-10-18
[Feature Request] Allow app_auth using path to PEM file (ex: GITHUB_APP_PEM_FILE_PATH) with precedence over GITHUB_APP_PEM_FILE


dgteixeira:
Hi there,

### Software Versions

Terraform:
```
Terraform v1.0.7
on linux_amd64
```

[GitHub Provider:](https://registry.terraform.io/providers/integrations/github/latest/docs#argument-reference)
```
    github = {
      source  = "integrations/github"
      version = "~> 4.14.0"
    }
```

### Affected Resource(s)
Please list the resources as a list, for example:
- [github_provider](https://registry.terraform.io/providers/integrations/github/latest/docs#argument-reference)

### Terraform Configuration Files
```
provider "github" {
  owner = var.org_name
  app_auth {} # When using `GITHUB_APP_XXX` environment variables - GITHUB_APP_ID + GITHUB_APP_INSTALLATION_ID + GITHUB_APP_PEM_FILE
}
```

### Expected Behavior
GitHub provider should accept an environment variable like `GITHUB_APP_PEM_FILE_PATH `that should take precedence over `GITHUB_APP_PEM_FILE`, if it is set as an environment variable.

### Actual Behavior
Whenever I try to insert a path in GITHUB_APP_PEM_FILE instead of the contents, I get the following error:
```
╷
│ Error: No decodeable PEM data found
│
│ with provider["registry.terraform.io/integrations/github"],
│ on main.tf line 17, in provider "github":
│ 17: provider "github" {
│
╵
```

### Steps to Reproduce
1. Create a root module with github provider with an empty app_auth{} configuration;
2. Create env variables for GITHUB_APP_ID + GITHUB_APP_INSTALLATION_ID + GITHUB_APP_PEM_FILE (this one using a path to a file):

### Important Factoids
We wanted this feature because we have [atlantis](https://www.runatlantis.io/docs/using-atlantis.html) deployed on a Kubernetes cluster and we want it to manage our github organization. On K8s, it is preferable to mount the certificate on a file (with restrict access) than importing its contents as a env var.

### References
- GitHub Issue #861

dmitry-mightydevops:
Same thing, version 5.9.0 still has no support for this

```
│ Error: no decodeable PEM data found
│ 
│   with provider["registry.terraform.io/integrations/github"],
│   on config.tf line 83, in provider "github":
│   83: provider "github" {
│ 
╵
[...]

Issue #598, opened 2020-11-11
Misleading error message when GITHUB_TOKEN not set


lukiffer:
When configuring the provider with only the `organization` property and relying on the `GITHUB_TOKEN` environment variable for authentication, if the environment variable hasn't been set, `apply` and `import` operations display the following error message:

> Error: This resource can only be used in the context of an organization, "" is a user.

This is misleading as the actual error is a missing credential.

After setting the environment variable, everything works as expected.

### Terraform Version
0.13.5

### Affected Resource(s)
Have only reproduced this with `github_membership` and `github_team`, but it probably applies to more.

### Terraform Configuration Files
```hcl
terraform {
  required_version = "=0.13.5"

  required_providers {
    github = {
      source  = "hashicorp/github"
      version = "=4.0.0"
    }
  }
}

provider "github" {
  version = "=4.0.0"
  organization = "example-org"
}

resource "github_membership" "member" {
  username = "lukiffer"
  role     = "admin"
}

```

### Expected Behavior
A more accurate error message should be displayed, like `Error: Missing GITHUB_TOKEN environment variable.`

### Actual Behavior
I received the error message `Error: This resource can only be used in the context of an organization, "" is a user.` and subsequently spent an hour debugging the provider config trying to figure out why when I supply `organization` it's being interpreted as an empty string.

### Steps to Reproduce
- `terraform apply`; or
- `terraform import github_membership.member "example-org:lukiffer"`

JeroenRijks:
Has anyone got a workaround for this issue?

jcudit:
Does this relate to the fix that was shipped in [`v4.3.1`](https://github.com/integrations/terraform-provider-github/blob/master/CHANGELOG.md#431-january-22-2021)?  Can anyone try out this release and confirm?

dgteixeira:
Hello everyone, I hope you are all well :)
@jcudit, I have tried and I have a similar problem when my GITHUB_TOKEN is expired or already inactive.

**Terraform version:** 1.0.7;
**Github Provider version:** v4.14.0

When the following terraform code, I get a `Error: Plugin did not respond` error

```
terraform {
  required_version = ">= 1.0.0"

  required_providers {
    github = {
      source  = "integrations/github"
      version = "~> 4.14.0"
    }
  }
}

provider "github" {
  owner = var.org_name
}

data "github_organization" "org_name" {
  name = var.org_name
}

resource "github_membership" "members" {
  username = foobar
  role          = "member"
[...]

PR #3246, opened 2026-03-02
feat: add auth_mode for explicit auth configuration

### LINKED ISSUE #1877: [FEAT]: Switching Between PAT and GitHub App Authentication Without Modifying Terraform Code

### LINKED ISSUE #3117: [MAINT]: Add provider configuration to set `auth_mode: 'anonymous'`
### Describe the need

We want to ensure that users can explicitly disable auth. Then `terraform` doesn't need to do extra work to check all other auth types.

### SDK Version

_No response_

### API Version

_No response_

### Relevant log output

```shell

```

### Code of Conduct

laughedelic:
This PR 
- Resolves https://github.com/integrations/terraform-provider-github/issues/1877
  - Supersedes https://github.com/integrations/terraform-provider-github/pull/2174
- Resolves https://github.com/integrations/terraform-provider-github/issues/3117
- Partially addresses https://github.com/integrations/terraform-provider-github/issues/3116 
  - see the note at the bottom on why it doesn't resolve that issue
- Also, partially addresses https://github.com/integrations/terraform-provider-github/issues/598 (improved error when `auth_mode` is set explicitly)

----

### Before the change?

- `app_auth {}` block is the only way to configure GitHub App authentication; switching between token and app auth requires modifying provider configuration code
- When using environment variables for app auth, an empty `app_auth {}` block is required due to a Terraform SDK limitation (#1877)
- Authentication mode is implicitly auto-detected with no way to opt out: if `GITHUB_TOKEN` is in the environment or `gh` CLI has a valid session, anonymous mode is impossible (#3116)

### After the change?

Based on the proposal in #3116:

- New `auth_mode` argument enables explicit control over authentication:
  - `auth_mode = "anonymous"` ignores any credentials
  - `auth_mode = "token"` errors if no token is provided instead of silently falling back
  - `auth_mode = "app"` errors if app credentials are missing or incomplete
  - When `auth_mode` is not set, the existing auto-detection behavior is preserved (backward compatible)
- New top-level `app_id`, `app_installation_id`, `app_private_key` arguments allow app auth without the `app_auth` block, making it possible to switch between token and app auth purely via environment variables (addressing #1877)
- `app_auth` block is deprecated with a mention of the new top-level arguments
- `gh auth token` CLI fallback is deprecated (as proposed in #3116)

Note, I took the liberty to name the new top-level argument `app_private_key` instead of `app_pem_file` because I believe it is more clear that the expected value is a private key content, not a file reference. I remember there was this [old comment](https://github.com/integrations/terraform-provider-github/issues/949#issuecomment-1743755325):

> I feel like the naming of the `pem_file` argument in `app_auth` is a little confusing because it expects the content of the file, not its path. Maybe calling it `private_key` would have been clearer

---

### Pull request checklist

- [x] No

----

I didn't implement `environment_variable_prefix` and `auth_type` from the #3116's proposal to keep this focused, and implementing those seemed like touching many more places. But I can add those separately.

- Follow-up: https://github.com/laughedelic/terraform-provider-github/pull/1


github/provider.go
```diff
@@ -419,11 +456,42 @@ func providerConfigure(p *schema.Provider) schema.ConfigureContextFunc {
...
+				log.Printf("[INFO] No token found, trying GitHub CLI to get token from hostname %s", baseURL.Host)
+				ghToken := tokenFromGHCLI(baseURL)
+				if ghToken != "" {
+					token = ghToken
+					diags = append(diags, diag.Diagnostic{
+						Severity: diag.Warning,
+						Summary:  "GitHub CLI token fallback is deprecated",
+						Detail: "Automatic token detection from `gh auth token` is deprecated and will be removed in a future major release. " +
+							"Please set the `token` provider argument or `GITHUB_TOKEN` environment variable explicitly. " +
+							"You can use `export GITHUB_TOKEN=$(gh auth token)` as a replacement.",
```
laughedelic, line 510: I've added a deprecation for gh CLI as proposed in #3116, but another option could be adding an explicit `auth_mode = cli` instead

laughedelic:
Thinking some more about the `auth_mode` and the empty default with some auto-resolution for compatibility: what if it was a priority list instead of a single value? This would allow users to set their own fallback logic by saying `auth_mode = ["app", "cli", "token"]` and depending on the environment, it would use different options in that order. WDYT?

UPD: Maybe I'm just overcomplicating it. Feel free to ignore this.

github/provider.go
laughedelic: Never mind. It's probably not worth it.

github/provider.go
```diff
@@ -378,34 +412,37 @@ func providerConfigure(p *schema.Provider) schema.ConfigureContextFunc {
...
 		}
 
-		if appAuth, ok := d.Get("app_auth").([]any); ok && len(appAuth) > 0 && appAuth[0] != nil {
-			appAuthAttr := appAuth[0].(map[string]any)
+		var token string
 
-			var appID, appInstallationID, appPemFile string
+		switch authMode {
+		case "anonymous":
+			log.Printf("[INFO] Auth mode: anonymous")
```
deiga: Could you test if we can use `tflog` inside this func?

github/provider.go
```diff
@@ -378,34 +412,37 @@ func providerConfigure(p *schema.Provider) schema.ConfigureContextFunc {
...
 
-			if v, ok := appAuthAttr["id"].(string); ok && v != "" {
-				appID = v
-			} else {
-				return nil, wrapErrors([]error{fmt.Errorf("app_auth.id must be set and contain a non-empty value")})
+		case "token":
+			token = d.Get("token").(string)
+			if token == "" {
+				return nil, diag.FromErr(fmt.Errorf(
```
deiga:
```suggestion
				return nil, diag.Errorf(
```

github/provider.go
```diff
@@ -123,6 +130,25 @@ func Provider() *schema.Provider {
...
+				DefaultFunc: schema.EnvDefaultFunc("GITHUB_APP_INSTALLATION_ID", nil),
+				Description: descriptions["app_installation_id"],
+			},
+			"app_private_key": {
+				Type:        schema.TypeString,
+				Optional:    true,
+				Sensitive:   true,
+				DefaultFunc: schema.EnvDefaultFunc("GITHUB_APP_PRIVATE_KEY", nil),
+				Description: descriptions["app_private_key"],
+			},
```
deiga, line 156: Could you add `ConflictsWith` fields so that it's not possible to set both App auth configs at the same time? Or does that interfere with the `DefaultFunc` being the same?

github/provider_test.go
```diff
@@ -232,6 +232,135 @@ data "github_ip_ranges" "test" {}
...
+			Steps: []resource.TestStep{
+				{
+					Config:      config,
+					ExpectError: regexp.MustCompile(`auth_mode is set to "app" but the following app credentials are missing`),
+				},
+			},
+		})
+	})
+
+	t.Run("can be configured with top-level app fields without auth_mode", func(t *testing.T) {
```
deiga: Hmm, I would prefer if this wouldn't be possible. Can we force setting of `auth_mode` for top-level app fields? Or are we limited by backwards compatibility?

deiga (changes requested):
I think this is looking great!

I've been wanting to get to this for a while now 😬 

Left a few minor comments

github/provider.go
laughedelic: Seems to be fine. Done in e07604b.

github/provider.go
laughedelic: Done in c33ccb5

github/provider.go
laughedelic: I don't think it interferes with the default value, but it probably doesn't raise the conflict when everything is set through env vars. In any case, I added mutually exclusive ConflictsWith in e3264c4. And further validation is done in `providerConfigure`.

github/provider_test.go
laughedelic: This is a bit more involved than just requiring `auth_mode = app` if the top-lvel app fields are set. I added an extra validation in the "default" case, to make sure that if the top-level fields are set and there is no `app_auth` block, we get an error about setting `auth_mode`.
There are cases when user can set both the token and the app values, but then choose which set of credentials to use via `auth_mode`. Any or all of them can be set via env vars. So I think this shouldn't be too strict. 
See af6fb59. I added some extra tests for the edge cases. Let me know if we're on the same page.

deiga (approved):
Whoo! This seems so much simpler than what I had thought it would be 😬 

I hope we're not missing something

laughedelic:
@stevehipwell could you give some feedback here?

stevehipwell:
Thanks for the PR @laughedelic, this whole area changed significantly in #3448 and we're waiting until v7 to cleanup the unexpected behaviours. That all said I'd be interested in getting in the `auth_mode` configuration option in (maybe with `legacy` as the default option).

laughedelic:
Closing. `auth_mode` landed in #3578. For the record, this PR proposed the same design back in March and was approved here (two years after the first attempt in #2174). Not a great contributing experience.

Glad the feature is in either way, and #1877 seems to be actually fixed.

deiga:
@laughedelic I'm sorry about this. It's hard to keep track of all the contributed work, there is so much. 
And we're just a few people who maintain this for a hobby, there isn't a lot of capacity to make sure that the contributing experience is great.

laughedelic:
@deiga I understand. Thanks for the response and your previous reviews!

PR #3248, opened 2026-03-02
feat: add prefix filter to team sync groups data source

### LINKED ISSUE #1809: [FEAT]: Support filter parameter on github_organization_team_sync_groups data source
### Describe the need

Currently, the `github_organization_team_sync_groups` data source fetches all groups (in fixed page sizes of 100). This causes performance issues if a large number of groups are synced into GitHub (we have about 3000 groups being pushed to GitHub, so the refresh of `github_organization_team_sync_groups` takes 20+ seconds).

As a feature request, please expose [the `q` query parameter](https://docs.github.com/en/enterprise-cloud@latest/rest/teams/team-sync?apiVersion=2022-11-28#list-idp-groups-for-an-organization), to allow the data source to be scoped to a subset of all groups:

> Filters the results to return only those that begin with the value specified by this parameter. For example, a value of ab will return results that begin with "ab".

For example,

```terraform
data "github_organization_team_sync_groups" "filtered" {
  q = "myprefix_"
}
```

### SDK Version

_No response_

### API Version

REST API 2022-11-28 (Enterprise Cloud)

### Relevant log output

_No response_

### Code of Conduct

laughedelic:
- Resolves https://github.com/integrations/terraform-provider-github/issues/1809

----

### Before the change?

> Currently, the github_organization_team_sync_groups data source fetches all groups (in fixed page sizes of 100). This causes performance issues if a large number of groups are synced into GitHub (we have about 3000 groups being pushed to GitHub, so the refresh of github_organization_team_sync_groups takes 20+ seconds).

### After the change?

Now it's possible to filter groups by prefix:

```terraform
data "github_organization_team_sync_groups" "filtered" {
  prefix_filter = "myprefix_"
}
```

### Pull request checklist

- [x] No

---

~~I followed the API and the SDK in naming it `q`, but I'm open to something more descriptive, e.g. `prefix` or `prefix_filter`~~


github/data_source_github_organization_team_sync_groups.go
```diff
@@ -65,7 +73,7 @@ func dataSourceGithubOrganizationTeamSyncGroupsRead(ctx context.Context, d *sche
 		options.Page = resp.NextPageToken
 	}
 
-	d.SetId(fmt.Sprintf("%s/github-org-team-sync-groups", orgName))
+	d.SetId(fmt.Sprintf("%s/github-org-team-sync-groups/%s", orgName, query))
```
deiga: issue: If we're changing the ID pattern we need a Schema migration

github/data_source_github_organization_team_sync_groups.go
```diff
@@ -14,6 +14,12 @@ func dataSourceGithubOrganizationTeamSyncGroups() *schema.Resource {
 		ReadContext: dataSourceGithubOrganizationTeamSyncGroupsRead,
 
 		Schema: map[string]*schema.Schema{
+			"q": {
```
deiga: I think something like `prefix_filter` could be nice and descriptive, WDYT?

website/docs/d/organization_team_sync_groups.html.markdown
```diff
@@ -12,9 +12,15 @@ Use this data source to retrieve the identity provider (IdP) groups for an organ
 ## Example Usage
 
 ```hcl
-data "github_organization_team_sync_groups" "test" {}
+data "github_organization_team_sync_groups" "test" {
+  q = "myprefix_"
+}
```
deiga: issue: Please show both examples

github/data_source_github_organization_team_sync_groups.go
```diff
@@ -14,6 +14,12 @@ func dataSourceGithubOrganizationTeamSyncGroups() *schema.Resource {
...
+				Type:        schema.TypeString,
+				Optional:    true,
+				Default:     "",
```
deiga: I'm not sure the `Default` makes a lot of sense. I would rathe remove it and use `d.GetOk()` to determine if we want to add `query` to the list options

deiga (changes requested):
Thanks for the PR. I left some comments that need addressing

github/data_source_github_organization_team_sync_groups.go
laughedelic: Sounds good. Done in b477069

github/data_source_github_organization_team_sync_groups.go
laughedelic: Done in 50ddc7e

github/data_source_github_organization_team_sync_groups.go
laughedelic: Is it ok to change the ID only when the new parameter is used: d276ec2?

website/docs/d/organization_team_sync_groups.html.markdown
laughedelic: Done in ad7206e

github/data_source_github_organization_team_sync_groups.go
```diff
@@ -65,7 +74,11 @@ func dataSourceGithubOrganizationTeamSyncGroupsRead(ctx context.Context, d *sche
 		options.Page = resp.NextPageToken
 	}
 
-	d.SetId(fmt.Sprintf("%s/github-org-team-sync-groups", orgName))
+	if options.Query != "" {
+		d.SetId(fmt.Sprintf("%s/github-org-team-sync-groups/%s", orgName, options.Query))
+	} else {
+		d.SetId(fmt.Sprintf("%s/github-org-team-sync-groups", orgName))
+	}
```
deiga: Okay, looking at this makes my skin crawl a bit 😬 

While we're at it, let's change to `buildID` to conform to our standards :)
```suggestion
	id, err := buildID(orgName, "team-sync-groups", options.Query)
	if err != nil {
		return diag.FromErr(err)
	}
	d.SetId(id)
```

With this change we'll need you to make a `StateUpgrader` to migrate the old IDs to new IDs. Look at any of the existing ones to see how to go about it :)

github/data_source_github_organization_team_sync_groups_test.go
```diff
@@ -19,6 +19,13 @@ func TestAccGithubOrganizationTeamSyncGroupsDataSource_existing(t *testing.T) {
 					resource.TestCheckResourceAttrSet("data.github_organization_team_sync_groups.test", "groups.0.group_name"),
 				),
 			},
+			{
+				Config: `data "github_organization_team_sync_groups" "test" { prefix_filter = "nonexistent_prefix_" }`,
+				Check: resource.ComposeTestCheckFunc(
```
deiga: Please use `ConfigStateChecks` instead of `Check`

deiga (changes requested):
Sorry, the ID will need to change and thus we need a schema migration

github/data_source_github_organization_team_sync_groups.go
laughedelic: Oh, I see what you mean 😅 I have to admit, I'm moving in the dark here, trying to make the minimal necessary changes, and I don't have a good overview of the whole codebase (yet). Thanks for the guidance.

I didn't know about `buildID`. But it uses `:` as a separator. I wonder why the original ID used `/` 🤔

laughedelic: If the query is empty, the id is going to look like `orgname:team-sync-groups:` with a trailing `:`. Isn't this odd? Should it be `orgname:team-sync-groups` instead?

laughedelic: So how about 

```suggestion
	idParts := []string{orgName, "team-sync-groups"}
	if options.Query != "" {
		idParts = append(idParts, options.Query)
	}
	id, err := buildID(idParts...)
	if err != nil {
		return diag.FromErr(err)
	}
	d.SetId(id)
```
?

github/data_source_github_organization_team_sync_groups_test.go
laughedelic: Done in ad18a986947fc936e64e7f3e144e84f7175f6d2e + 49d1e01

github/data_source_github_organization_team_sync_groups.go
laughedelic: The ID change is done in 0064745 and the migration added in a6e33d8. I hope I'm getting it right, but let me know if I still misunderstood.

github/data_source_github_organization_team_sync_groups.go
```diff
@@ -12,6 +12,15 @@ func dataSourceGithubOrganizationTeamSyncGroups() *schema.Resource {
...
 		ReadContext: dataSourceGithubOrganizationTeamSyncGroupsRead,
 
+		SchemaVersion: 1,
+		StateUpgraders: []schema.StateUpgrader{
+			{
+				Type:    dataSourceGithubOrganizationTeamSyncGroupsV0().CoreConfigSchema().ImpliedType(),
+				Upgrade: dataSourceGithubOrganizationTeamSyncGroupsStateUpgradeV0,
+				Version: 0,
+			},
+		},
```
deiga: 🤦 Silly me. Do data sources even need Schema Migrations? @stevehipwell

github/data_source_github_organization_team_sync_groups.go
deiga: Thanks. I don't think that conditional IDs are great, since that requires then custom logic when parsing them (if necessary). We already have other examples of IDs where they just end with `:` if the last part is empty

laughedelic: Ok. Back to 3-part IDs: e21fe5d

github/data_source_github_organization_team_sync_groups.go
laughedelic: @stevehipwell any input on this?

github/data_source_github_organization_team_sync_groups.go
```diff
@@ -65,7 +82,11 @@ func dataSourceGithubOrganizationTeamSyncGroupsRead(ctx context.Context, d *sche
 		options.Page = resp.NextPageToken
 	}
 
-	d.SetId(fmt.Sprintf("%s/github-org-team-sync-groups", orgName))
+	id, err := buildID(orgName, "team-sync-groups", options.Query)
```
stevehipwell:
```suggestion
	id, err := buildID(orgName, options.Query)
```
There is no need for this part in the ID.

github/data_source_github_organization_team_sync_groups.go
stevehipwell: You don't need to migrate the schema for a data source (can you even do it?) as it'll be refetched.

stevehipwell (changes requested):
@laughadelic please could you rebase this PR, please note that we've changed how the docs work and ideally want to generate the docs for this data source via the automation.

github/data_source_github_organization_team_sync_groups.go
```diff
@@ -48,6 +52,10 @@ func dataSourceGithubOrganizationTeamSyncGroupsRead(ctx context.Context, d *sche
 		},
 	}
 
+	if v, ok := d.GetOk("prefix_filter"); ok {
+		options.Query = v.(string)
```
stevehipwell:
```suggestion
		q, _ := v.(string)
		options.Query = q
```
Otherwise you'll get a lint failure.

github/data_source_github_organization_team_sync_groups_test.go
```diff
@@ -4,6 +4,9 @@ import (
 	"testing"
 
 	"github.com/hashicorp/terraform-plugin-testing/helper/resource"
+	"github.com/hashicorp/terraform-plugin-testing/knownvalue"
+	"github.com/hashicorp/terraform-plugin-testing/statecheck"
+	"github.com/hashicorp/terraform-plugin-testing/tfjsonpath"
 )
 
 func TestAccGithubOrganizationTeamSyncGroupsDataSource_existing(t *testing.T) {
```
stevehipwell:
```suggestion
func TestAccGithubOrganizationTeamSyncGroupsDataSource(t *testing.T) {
```

Please nest sub-tests using `t.Run` instead of using steps for separate tests. I suggest the following sub tests.

- `all`
  - Single step
- `filtered`
  - Step 1 a filter that matches
  - Step 2 change the filter to match nothing

templates/data-sources/organization_team_sync_groups.md.tmpl
```diff
@@ -10,8 +10,18 @@ Use this data source to retrieve the identity provider (IdP) groups for an organ
 
 ## Example Usage
 
+Retrieve all IdP groups:
```
stevehipwell: Please can you update the docs to use the template pattern of automatically inserting the examples and using the schema for the attribute descriptions, meaning this description should be a comment in the example file.

github/data_source_github_organization_team_sync_groups.go
```diff
@@ -65,7 +73,11 @@ func dataSourceGithubOrganizationTeamSyncGroupsRead(ctx context.Context, d *sche
 		options.Page = resp.NextPageToken
 	}
 
-	d.SetId(fmt.Sprintf("%s/github-org-team-sync-groups", orgName))
+	id, err := buildID(orgName, options.Query)
```
stevehipwell: This might be clearer if we use a placeholder when query isn't set.

github/data_source_github_organization_team_sync_groups.go
```diff
@@ -14,6 +13,11 @@ func dataSourceGithubOrganizationTeamSyncGroups() *schema.Resource {
 		ReadContext: dataSourceGithubOrganizationTeamSyncGroupsRead,
 
 		Schema: map[string]*schema.Schema{
+			"prefix_filter": {
+				Type:        schema.TypeString,
+				Optional:    true,
+				Description: "Filters the results to return only those that begin with the specified value.",
```
stevehipwell, line 20: I'm assuming there are rules on the naming of team sync groups? If so we need to validate here.

stevehipwell (changes requested):
Thanks for updating @laughedelic, I've added some review comments.

github/data_source_github_organization_team_sync_groups.go
laughedelic: Done in de55491

github/data_source_github_organization_team_sync_groups_test.go
laughedelic: Done in de55491

templates/data-sources/organization_team_sync_groups.md.tmpl
laughedelic: Done in de55491 and d2915b15

github/data_source_github_organization_team_sync_groups.go
laughedelic: Done in de55491. The ID is now `<org>:*` when no filter is set, and `<org>:<prefix>` otherwise

github/data_source_github_organization_team_sync_groups.go
laughedelic: I don't think there's anything to validate here. These names come from the external identity provider rather than from GitHub, so the permitted characters are whatever the IdP allows. Entra ID group names, Okta group names, etc. From my experience with Okta, there isn't any expected format.

github/data_source_github_organization_team_sync_groups_test.go
```diff
@@ -18,11 +21,43 @@ func TestAccGithubOrganizationTeamSyncGroupsDataSource_existing(t *testing.T) {
...
+			PreCheck:          func() { skipUnlessEnterprise(t) },
+			ProviderFactories: providerFactories,
+			Steps: []resource.TestStep{
+				{
+					Config: `data "github_organization_team_sync_groups" "test" { prefix_filter = "acctest-github-provider" }`,
+					ConfigStateChecks: []statecheck.StateCheck{
+						statecheck.ExpectKnownValue("data.github_organization_team_sync_groups.test",
+							tfjsonpath.New("prefix_filter"), knownvalue.StringExact("acctest-github-provider")),
+						statecheck.ExpectKnownValue("data.github_organization_team_sync_groups.test",
+							tfjsonpath.New("groups"), knownvalue.NotNull()),
```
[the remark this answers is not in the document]
deiga: Since the filtering functionality is part of the upstream API verifying details doesn't make sense here

deiga (commented):
I think in this case it could make sense to have a unit test which mocks the GH API enough to verify that the filter prefix is being sent could make sense here 🤔

stevehipwell:
> I think in this case it could make sense to have a unit test which mocks the GH API enough to verify that the filter prefix is being sent could make sense here 🤔

@deiga the whole unit test problem has repercussions. Technically we should have a large number of unit tests to cover all code paths and then a smaller number of integration tests to validate that the system works as a whole. The problem with this code base and the GitHub API is that there isn't currently a strategy to make this work, meaning that unit tests just increase the overhead (cognitive and test duration) without decreasing the integration test overhead as the unit tests can't be trusted to actually represent the GitHub API.

I think we need a separate issue to define how we could structure the tests and check the API to support this, but in the meantime I'd be very cautious about accepting unit tests that mock the GitHub API surface (or even worse undocumented API behaviour).

github/data_source_github_organization_team_sync_groups.go
```diff
@@ -12,8 +11,14 @@ import (
 func dataSourceGithubOrganizationTeamSyncGroups() *schema.Resource {
 	return &schema.Resource{
 		ReadContext: dataSourceGithubOrganizationTeamSyncGroupsRead,
+		Description: "Get the external identity provider (IdP) groups for an organization.",
```
stevehipwell, line 14:
```suggestion
		Description: "Data source to get the external identity provider (IdP) groups for an organization.",
```

stevehipwell (changes requested):
This looks good. Please add a named documentation template for the data source for consistency with all existing resources/data-sources using the auto generated documentation.

deiga:
@stevehipwell I agree about the unit testing and mocking problems. That's why I proposed to mock the API to verify that the provider is sending the payload we think it should instead of verifying the API itself

stevehipwell:
> @stevehipwell I agree about the unit testing and mocking problems. That's why I proposed to mock the API to verify that the provider is sending the payload we think it should instead of verifying the API itself

The problem is that this doesn't prove anything other than a unit test was written at the same time as the code so they match. Is there a reason why an acceptance test wouldn't work?

deiga:
> > @stevehipwell I agree about the unit testing and mocking problems. That's why I proposed to mock the API to verify that the provider is sending the payload we think it should instead of verifying the API itself
> 
> 
> 
> The problem is that this doesn't prove anything other than a unit test was written at the same time as the code so they match. Is there a reason why an acceptance test wouldn't work?

AFAIK we can't guarantee that a result will be found with an acceptance test since that would require a external IdP connection?

stevehipwell:
> AFAIK we can't guarantee that a result will be found with an acceptance test since that would require a external IdP connection?

Yeah, the test would need a comment explaining how to configure it to function correctly but it's really the only way to check that it actually works with the real API.

PR #3196, opened 2026-02-15
[MAINT] Proposal for adding `ARCHITECTURE.md` and `MAINTERS.md`

deiga:
This PR proposes adding two new documentation files and updating the existing contributing guide to better serve contributors and maintainers:

 ### ARCHITECTURE.md

 A comprehensive implementation guide for contributors covering:

 - **Module map** — project layout and file organization
 - **Core principles** — one resource per API entity, minimize API calls, API-only operations
 - **Resource design** — file naming, schema field guidelines, composite ID patterns (`buildID`/`parseID2`/`parseID3`/`parseID4`)
 - **Implementation patterns** — context-aware CRUD signatures, API client access, error handling, state migrations (`StateUpgraders`), structured logging (`tflog`)
 - **Testing** — test structure, test modes, running/debugging tests
 - **Gotchas & known issues** — API preview headers, deprecated resources, known limitations, pending go-github updates
 - **Appendix** — common utilities reference, naming conventions
 - **Decision log** — records key architectural decisions with rationale (StateUpgraders, explicit auth, transport layer rework, no local git CLI, tflog migration, SDK v2 finalization)

 ### MAINTAINERS.md

 Lists current and emeritus maintainers with GitHub profile links.

 ### CONTRIBUTING.md updates

 - Added step for updating documentation when making changes
 - Added step for ensuring formatting/linting passes
 - Fixed markdown indentation for nested lists
 - Replaced raw `go test` commands with `make testacc T=...` for consistency
 - Updated debugging section to reference `tflog` migration pattern
 - Removed hardcoded line-number GitHub permalink in favor of stable function name reference
 - Added `GITHUB_ENTERPRISE_SLUG` environment variable

 ### Code change

 - Marked `parseTwoPartID` and `parseThreePartID` as `@deprecated` in `github/util.go`, pointing to `parseID2`/`parseID3` replacements

 ## Motivation

 The provider currently lacks an architecture guide, making it harder for new contributors to understand conventions, patterns, and past decisions. These documents aim to:

 1. Reduce onboarding friction for new contributors
 2. Codify existing patterns so they're discoverable (not just tribal knowledge)
 3. Record architectural decisions and their rationale
 4. Establish a clear maintainer list

 ## Discussion points

 This is a proposal — feedback is welcome on:

 - Content accuracy and completeness
 - Whether any sections should be expanded, condensed, or reorganized
 - Additional gotchas or known issues worth documenting
 - Decision log entries that should be added or revised

PR #3254, opened 2026-03-04
[MAINT] Create custom `golangci-lint` to integrate `tfproviderlint`

deiga:
### Before the change?

- Not able to use `tfproviderlint` as part of `golangci-lint`

### After the change?

- Capable to use `tfproviderlint` as part of `golangci-lint`
- Capable of adding new checks for the terraform provider schema
	- Example: tools/tfproviderlint/checks/L001/L001.go
- Add 3 custom linter rules
	- L001: Instructs to use `ValidateDiagFunc` instead of `ValidateFunc`
	- L002: Instructs to use Context-aware CRUD functions
	- L003: Instructs to not call any CRUD functions directly
- Enable all `tfproviderlint` checks, which seem sensible and don't require fixes right now
- Add `make lint-new` to run stricter linters on only code since `main`
- Refactor `ci.yaml` workflow, to run linters in parallel using `golangci-lint-action`
	- This enables inline linter complaints in PRs

### Pull request checklist

- [ ] Yes
- [x] No

----


stevehipwell (changes requested):
I'm sorry but I'm not currently onboard with this change as it stands. It's a long way away from the principle of least surprise and changes standard patterns I'd expect to see in a TF provider codebase. Is there a reason why we can't run `tfproviderlint` as a separate job so at least the complexity is kept away from the other more standard linting?

deiga:
@stevehipwell I don't think I understand your objection. The PR forces no changes on existing code and provides a stable way, with oversight, to have all new changes linted with a stricter ruleset. 

It basically already does exactly what you're asking for: running the new linter rules in a separate job. While also providing seamless integration into the linting landscape I use. 

These changes would reduce a big part of review comments we currently have to type out manually. 

Could you elaborate what the problem with these changes is, so that I could understand better?

stevehipwell:
@deiga you've changed how CI and the local default linting works, neither of which is desirable. I'm keen to have advanced checks but not at the cost of slowdown for local development or changing the fast feedback loop and understandability of the CI job.

Why not run `tfproviderlint` as a standalone?

deiga:
> Why not run `tfproviderlint` as a standalone?
- We'd need to use a custom golangci-lint or revive (not investigated) to enable adding our own linter rules. 
	- And to get the GitHub Actions tie-in
	- And IMO the custom rules here are the most important part of this change


> you've changed how CI and the local default linting works

In some ways, yes. The linter is now fully dependent on `make`, or at least for the necessary targets to have been run once. After that, one could call the `./bin/custom-gcl` directly. BUT it doesn't really change how default linting works IMO. It still uses the same logic for linting and the same ruleset. It just runs a few more rules now.

> but not at the cost of slowdown for local development or changing the fast feedback loop and understandability of the CI job.

I haven't measured it, but there should not be any major slowdown locally. Only the first run after each of these forces a recompile: updated golangci-lint version or updated go (non-test) files in `./tools/`. Neither of which happen too often, unless you're developing new linters.

If there is an impact on the understandability of the CI job, that has not been intentional and needs to be addressed. Please point to this degradation

.github/workflows/ci-linter.yaml
```diff
@@ -0,0 +1,45 @@
+name: CI
```
stevehipwell: This duplicates the actual CI workflow.

.github/workflows/ci-linter.yaml
```diff
@@ -0,0 +1,45 @@
...
+        with:
+          go-version-file: go.mod
+          cache: true
+          cache-dependency-path: |
+            go.sum
+            tools/tfproviderlint/**/*.sum
+      - name: Test Custom Linters
+        run: make test-tools --always-make
+        env:
+          MAKEFLAGS: '--debug=v'
```
stevehipwell: There seem to be a number of missing line endings.

.golangci.yml
```diff
@@ -12,7 +12,6 @@ linters:
     - errcheck
     - errname
     - errorlint
-    # - forcetypeassert TODO: Re-enable when we can fix the issues
```
stevehipwell, line 15: Please don't remove this, we want to re-enable it.

.golangci.yml
```diff
@@ -27,6 +26,64 @@ linters:
     - unparam
     - unused
     - usetesting
+    - tfproviderlint
```
stevehipwell: I think the custom linter should be in a separate configuration.

stevehipwell (changes requested):
I've added some comments, but I'm still of the opinion that the increase in complexity here isn't worth it for the custom rules. I'd be much happier with running the [tfproviderlint/tfproviderlintx](https://github.com/bflad/tfproviderlint) binary in addition to `golangci-lint`.

Once we've got the checks in place; if we really think that the additional complexity of custom linters would be worth it then that's a different discussion.

.golangci.yml
deiga: This is not removed, it's moved to `.golangci.strict.yml`, which is run for all new code. That way we don't accrue any new issue from having this disabled for everything.

deiga:
> I've added some comments, but I'm still of the opinion that the increase in complexity here isn't worth it for the custom rules. I'd be much happier with running the [tfproviderlint/tfproviderlintx](https://github.com/bflad/tfproviderlint) binary in addition to `golangci-lint`.
> 
> Once we've got the checks in place; if we really think that the additional complexity of custom linters would be worth it then that's a different discussion.

@stevehipwell 
I hear you. Added complexity isn't great. And a few of the solutions I've chosen here are too clever (like the Makefile magic)
But at the same time, the frustration of asking ~every PR to do the exact same changes (which can be linted) is just killing my motivation to even do PR reviews anymore.

The problem with vanilla `tfproviderlint/tfproviderlintx` is, that it doesn't have the really impactful rules that we need.
And I'd rather not fork the project to add the rules as that's an even bigger undertaking than what is now added here.

Now that I found `go-ruleguard` I'm interested if these could be built there. I just don't understand these linters so well (yet) that it's going to be easy for me to accomplish.

.golangci.yml
deiga: That's fair. I'll try something out.

deiga: I've added more Make Magic 🙃 to keep the `tfproviderlint` configs separate => the `.golangci.yml` keeps working with any install of `golangci-lint` and "normal" linting works without make

.golangci.yml
stevehipwell: Leaving this here shows we still intend to enable it in the future. I guess an issue to make the default strict in the future would also work, but I'm just worried about forgetting to get this fixed.

stevehipwell:
@deiga I feel your pain, but at the same time my major concern is the project complexity and bus factor gap with GitHub. For now I'd rather keep the complexity lower and be less responsive to incoming PRs to fight the burnout.

The real question here is how much effort are we saving in PR review vs how much effort we are adding to both understanding the codebase and ongoing maintenance? If this was our commercial product and we wanted a great contributor experience I'd be 100% behind this. But it's not our product and GitHub are not putting in the required resources to support such an approach (or us). As it currently stands all of our work on this provider has left it in a better state than we found it, with an increasingly lower cognitive load required for a new contributor. Adding a custom linter would actually increase the cognitive load as well as the barrier to contribution, especially if the primary maintainer of it burned out or was otherwise unavailable.

Looking at the problem "PRs often make the same mistakes and they are repetitive to review" from first principles I suspect that the root cause of the code quality is the quality of the existing codebase and contribution documentation. What if we focus on updating the contribution documentation and make the existing codebase consistent and follow best practices (including automated documentation)? Once we have strong contribution documentation and code following best practices PR review should be easier and just a reference back to the docs and some example code. It's not as "smart" as custom linters, but it might be a better initial approach?

deiga:
@stevehipwell I agree that the first step would be to improve documentation. I've been trying to engage the maintainers in a discussion with proposals since beginning of February 😬 

You can see my PR on the matter [here](https://github.com/integrations/terraform-provider-github/pull/3196)
(Yes, it's gotten a bit stale already, since there have been new decisions made)

I understand your point about complexity and I don't like it either that I need to modify the `golangci-lint` executable to add custom rules. I don't fully agree with the idea, but let's park the custom linter for now. I can keep looking if I can find some other way. Having better docs and a configured Copilot Reviewer to nag about the same issues, will address a lot of the same problems (hopefully)

What if I open a new PR with only the Current Code / New Code linter structure in here? That would give us at least the benefit of new addition following stricter linting requirements than existing code. WDYT?

deiga:
Superseded by https://github.com/integrations/terraform-provider-github/pull/3430

PR #3271, opened 2026-03-12
[BUG] Ensure autolink reference doesn't error when repo is renamed

### LINKED ISSUE #3470: resource_github_repository_autolink_reference.go — github_repository_autolink_reference (field: repository)

deiga:
Resolves #3470

----

### Before the change?

- Renaming a repository caused an existing `github_repository_autolink_reference` to be re-created and throw an error

### After the change?

- Renaming a repository does not cause an error in `github_repository_autolink_reference`

### Pull request checklist

- [ ] Yes
- [x] No

----


github/resource_github_repository_autolink_reference.go
```diff
@@ -4,22 +4,24 @@ import (
...
 	return &schema.Resource{
-		Create: resourceGithubRepositoryAutolinkReferenceCreate,
-		Read:   resourceGithubRepositoryAutolinkReferenceRead,
-		Delete: resourceGithubRepositoryAutolinkReferenceDelete,
+		CreateContext: resourceGithubRepositoryAutolinkReferenceCreate,
+		UpdateContext: resourceGithubRepositoryAutolinkReferenceUpdate,
+		ReadContext:   resourceGithubRepositoryAutolinkReferenceRead,
+		DeleteContext: resourceGithubRepositoryAutolinkReferenceDelete,
 
 		Importer: &schema.ResourceImporter{
```
stevehipwell, line 25: It's probably worth moving this to a func to follow the latest pattern?

stevehipwell (commented):
Do we need a migration to add the repository ID and then rely on it or we just supporting it as a best effort?

RESOURCES.md
```diff
@@ -180,7 +180,7 @@ The overall status of each resource or data source is captured in this document
 | `github_project_column` (🚫) | ❓ | ❓ | ❓ | ❓ | ❓ | ❓ | ❓ |
 | `github_release` | ❓ | ❓ | ❓ | ❓ | ❓ | ❓ | ❓ |
 | `github_repository` | ❓ | ❓ | ❓ | ❓ | ❓ | ❓ | ❓ |
-| `github_repository_autolink_reference` | ❓ | ❓ | ❓ | ❓ | ❓ | ❓ | ❓ |
+| `github_repository_autolink_reference` | ⚠️ | ✅ | ✅ | ✅ | ✅ | ❓ | ❓ |
```
stevehipwell:
```suggestion
| `github_repository_autolink_reference` | ⚠️ | ✅ | ✅ | ✅ | ✅ | ❌ | ❌ |
```

github/resource_github_repository_autolink_reference.go
```diff
@@ -4,65 +4,49 @@ import (
...
-					id = strconv.FormatInt(*autolink.ID, 10)
-				}
-
-				if err = d.Set("repository", repository); err != nil {
-					return nil, err
-				}
-				d.SetId(id)
-				return []*schema.ResourceData{d}, nil
-			},
+			StateContext: resourceGithubRepositoryAutolinkReferenceImport,
```
stevehipwell, line 26:
```suggestion
		DeleteContext: resourceGithubRepositoryAutolinkReferenceDelete,
		Importer: &schema.ResourceImporter{
			StateContext: resourceGithubRepositoryAutolinkReferenceImport,
```

github/resource_github_repository_autolink_reference.go
```diff
@@ -4,65 +4,49 @@ import (
...
 		},
 
-		SchemaVersion: 1,
+		CustomizeDiff: diffRepository,
+
+		SchemaVersion: 2,
```
stevehipwell, line 31: Is there an upgrader missing?

github/resource_github_repository_autolink_reference.go
```diff
@@ -140,44 +136,113 @@ func resourceGithubRepositoryAutolinkReferenceRead(d *schema.ResourceData, meta
...
+		return diag.FromErr(err)
 	}
 
 	return nil
 }
 
-func resourceGithubRepositoryAutolinkReferenceDelete(d *schema.ResourceData, meta any) error {
-	client := meta.(*Owner).v3client
+func resourceGithubRepositoryAutolinkReferenceUpdate(ctx context.Context, d *schema.ResourceData, m any) diag.Diagnostics {
+	tflog.Warn(ctx, "Update function of autolink reference. This should not be called. But it's necessary when 'repository' doesn't have `ForceNew`", map[string]any{
```
stevehipwell: Could we make this a single line?

github/resource_github_repository_autolink_reference.go
```diff
@@ -140,44 +136,113 @@ func resourceGithubRepositoryAutolinkReferenceRead(d *schema.ResourceData, meta
...
+	}
+
+	return nil
+}
+
+func resourceGithubRepositoryAutolinkReferenceImport(ctx context.Context, d *schema.ResourceData, m any) ([]*schema.ResourceData, error) {
+	parts := strings.Split(d.Id(), "/")
+	if len(parts) != 2 {
+		return nil, fmt.Errorf("invalid ID specified: supplied ID must be written as <repository>/<autolink_reference_id>")
+	}
```
stevehipwell: I think the ID should be migrated as part of the schema migration so it can use the ID functions.

stevehipwell (commented):
Just a quick review.

github/resource_github_repository_autolink_reference.go
deiga: Nope, the resource was originally created with SchemaVersion: 1

github/resource_github_repository_autolink_reference.go
```diff
@@ -114,23 +98,35 @@ func resourceGithubRepositoryAutolinkReferenceCreate(d *schema.ResourceData, met
...
+	repo, _, err := client.Repositories.Get(ctx, owner, repoName)
+	if err != nil {
+		return diag.FromErr(err)
+	}
+
+	if err := d.Set("repository_id", int(repo.GetID())); err != nil {
+		return diag.FromErr(err)
+	}
+
+	return resourceGithubRepositoryAutolinkReferenceRead(ctx, d, m)
```
stevehipwell: We should be able to remove this call and just set the etag.

github/resource_github_repository_autolink_reference.go
```diff
@@ -114,23 +98,35 @@ func resourceGithubRepositoryAutolinkReferenceCreate(d *schema.ResourceData, met
...
 }
 
-func resourceGithubRepositoryAutolinkReferenceRead(d *schema.ResourceData, meta any) error {
-	client := meta.(*Owner).v3client
+func resourceGithubRepositoryAutolinkReferenceRead(ctx context.Context, d *schema.ResourceData, m any) diag.Diagnostics {
+	ctx = tflog.SetField(ctx, "id", d.Id())
```
stevehipwell: Let's keep with the current pattern where we only set attributes when we need them. And as a rule of thumb `id` shouldn't be set unless there is an unexpected error, we want the field named in a friendly way (e.g. `autolink_id`).

github/resource_github_repository_autolink_reference.go
```diff
@@ -114,23 +98,35 @@ func resourceGithubRepositoryAutolinkReferenceCreate(d *schema.ResourceData, met
...
+	client := meta.v3client
+	owner := meta.name
 
-	owner := meta.(*Owner).name
 	repoName := d.Get("repository").(string)
 	autolinkRefID, err := strconv.ParseInt(d.Id(), 10, 64)
 	if err != nil {
-		return unconvertibleIdErr(d.Id(), err)
+		return diag.FromErr(unconvertibleIdErr(d.Id(), err))
 	}
```
stevehipwell: Why not add a new `idStringToInt64OK` function with the signature `func(string) (bool, diag.Diagnostics)` (similar to the `checkOrganizationOK` I'm adding in #3537)?

github/resource_github_repository_autolink_reference.go
```diff
@@ -114,23 +98,35 @@ func resourceGithubRepositoryAutolinkReferenceCreate(d *schema.ResourceData, met
...
-	ctx := context.WithValue(context.Background(), ctxId, d.Id())
+
 	if !d.IsNewResource() {
 		ctx = context.WithValue(ctx, ctxEtag, d.Get("etag").(string))
 	}
```
stevehipwell: This can be removed as create shouldn't be calling through.

github/resource_github_repository_autolink_reference.go
```diff
@@ -140,44 +136,113 @@ func resourceGithubRepositoryAutolinkReferenceRead(d *schema.ResourceData, meta
 		var ghErr *github.ErrorResponse
 		if errors.As(err, &ghErr) {
 			if ghErr.Response.StatusCode == http.StatusNotFound {
-				log.Printf("[INFO] Removing autolink reference for repository %s/%s from state because it no longer exists in GitHub",
-					owner, repoName)
+				tflog.Info(ctx, "Autolink reference not found, removing from state.", map[string]any{
```
stevehipwell: Please make this a single line.

github/resource_github_repository_autolink_reference.go
```diff
@@ -140,44 +136,113 @@ func resourceGithubRepositoryAutolinkReferenceRead(d *schema.ResourceData, meta
 		var ghErr *github.ErrorResponse
 		if errors.As(err, &ghErr) {
 			if ghErr.Response.StatusCode == http.StatusNotFound {
```
stevehipwell: Please could we use the documented `errors.AsType` here?

github/resource_github_repository_autolink_reference.go
```diff
@@ -140,44 +136,113 @@ func resourceGithubRepositoryAutolinkReferenceRead(d *schema.ResourceData, meta
...
+		return diag.FromErr(err)
+	}
+
+	repo, _, err := client.Repositories.Get(ctx, owner, repoName)
+	if err != nil {
+		return diag.FromErr(err)
+	}
+	if err = d.Set("repository_id", int(repo.GetID())); err != nil {
+		return diag.FromErr(err)
 	}
```
stevehipwell: This isn't needed here.

github/resource_github_repository_autolink_reference.go
```diff
@@ -140,44 +136,113 @@ func resourceGithubRepositoryAutolinkReferenceRead(d *schema.ResourceData, meta
...
 }
 
-func resourceGithubRepositoryAutolinkReferenceDelete(d *schema.ResourceData, meta any) error {
-	client := meta.(*Owner).v3client
+func resourceGithubRepositoryAutolinkReferenceUpdate(ctx context.Context, d *schema.ResourceData, m any) diag.Diagnostics {
+	tflog.Warn(ctx, "Update function of autolink reference. This should not be called. But it's necessary when 'repository' doesn't have `ForceNew`", map[string]any{
+		"repository":    d.Get("repository"),
+		"repository_id": d.Get("repository_id"),
+		"id":            d.Id(),
+	})
```
stevehipwell: Please make this a single line.

github/resource_github_repository_autolink_reference.go
```diff
@@ -140,44 +136,113 @@ func resourceGithubRepositoryAutolinkReferenceRead(d *schema.ResourceData, meta
...
+	return nil
+}
+
+func resourceGithubRepositoryAutolinkReferenceDelete(ctx context.Context, d *schema.ResourceData, m any) diag.Diagnostics {
+	ctx = tflog.SetField(ctx, "id", d.Id())
```
stevehipwell: Same as above.

github/resource_github_repository_autolink_reference.go
```diff
@@ -140,44 +136,113 @@ func resourceGithubRepositoryAutolinkReferenceRead(d *schema.ResourceData, meta
...
-	return err
+	if err != nil {
+		return diag.FromErr(err)
+	}
+
+	return nil
+}
+
+func resourceGithubRepositoryAutolinkReferenceImport(ctx context.Context, d *schema.ResourceData, m any) ([]*schema.ResourceData, error) {
+	parts := strings.Split(d.Id(), "/")
```
stevehipwell: This should use the `:` separator. AFAIK we've changed some of these without a major version.

github/resource_github_repository_autolink_reference.go
```diff
@@ -140,44 +136,113 @@ func resourceGithubRepositoryAutolinkReferenceRead(d *schema.ResourceData, meta
...
+
+	_, err := strconv.Atoi(id)
+	if err != nil {
+		autolink, err := getAutolinkByKeyPrefix(ctx, client, owner, repository, id)
+		if err != nil {
+			return nil, err
+		}
+
+		id = strconv.FormatInt(*autolink.ID, 10)
+	}
```
stevehipwell:
```suggestion
	if _, err := strconv.Atoi(id); err != nil {
		autolink, err := getAutolinkByKeyPrefix(ctx, client, owner, repository, id)
		if err != nil {
			return nil, err
		}

		id = strconv.FormatInt(*autolink.ID, 10)
	}
```

github/resource_github_repository_autolink_reference.go
```diff
@@ -140,44 +136,113 @@ func resourceGithubRepositoryAutolinkReferenceRead(d *schema.ResourceData, meta
...
+
+	if err = d.Set("repository", repository); err != nil {
+		return nil, err
+	}
+	if err = d.Set("repository_id", int(repo.GetID())); err != nil {
+		return nil, err
+	}
+
+	return []*schema.ResourceData{d}, nil
```
stevehipwell, line 250: You need to set all of the fields as they're force new.

github/resource_github_repository_autolink_reference_migration.go
```diff
@@ -0,0 +1,70 @@
...
+	if v, ok := rawState["repository"]; ok {
+		if s, ok := v.(string); ok && s != "" {
+			repoName = s
+		}
+	}
+	if repoName == "" {
+		return nil, fmt.Errorf("state upgrade v1: repository is not a string or not set")
+	}
+
+	repo, _, err := client.Repositories.Get(ctx, owner, repoName)
```
stevehipwell: You can use the `migrateRepositoryWithID` function to avoid repeating the code.

github/resource_github_repository_autolink_reference_migration_test.go
```diff
@@ -0,0 +1,80 @@
...
+	"encoding/json"
+	"fmt"
+	"net/http"
+	"testing"
+
+	"github.com/google/go-cmp/cmp"
+	"github.com/google/go-github/v89/github"
+)
+
+func buildMockResponsesForRepositoryAutolinkReferenceMigrationV1toV2(mockOwner, mockRepo string, wantRepoID int) []*mockResponse {
```
stevehipwell: You shouldn't need this, please copy the pattern used for `Test_resourceGithubCustomPropertyStateUpgradeV0`.

github/resource_github_repository_autolink_reference_test.go
stevehipwell: Please switch to using the test helpers to create the repo instead of using TF.

github/resource_github_repository_autolink_reference_test.go
```diff
@@ -54,56 +59,34 @@ func TestAccGithubRepositoryAutolinkReference(t *testing.T) {
...
-		)
-
 		resource.Test(t, resource.TestCase{
 			PreCheck:          func() { skipUnauthenticated(t) },
 			ProviderFactories: providerFactories,
 			Steps: []resource.TestStep{
 				{
 					Config: config,
-					Check:  check,
+					ConfigStateChecks: []statecheck.StateCheck{
```
stevehipwell, line 62: I'm pretty sure you don't need the checks given that none are computed (and we should ignore etag).

stevehipwell (changes requested):

github/resource_github_repository_autolink_reference.go
deiga: I noticed that we weren't setting `etag` at all 😂

github/resource_github_repository_autolink_reference.go
deiga: Since this is "just" the import ID, it doesn't need a migration, right?

rilical (approved):

rilical (approved):

templates/resources/repository_autolink_reference.md.tmpl
```diff
@@ -1,49 +0,0 @@
----
```
stevehipwell, line 1: I think we need to keep the template per resource/data source pattern as in all but the most trivial cases there will be additional context required that can't be added to the schema (e.g. required permissions). So far all migrated docs have kept the template.

examples/resources/github_repository_autolink_reference/import.sh
```diff
@@ -0,0 +1,7 @@
+# Import by ID
```
stevehipwell, line 1: Please also add `import-by-string-id.tf`.

github/resource_github_repository_autolink_reference.go
```diff
@@ -4,65 +4,50 @@ import (
...
+		StateUpgraders: []schema.StateUpgrader{
+			{
+				Type:    resourceGithubRepositoryAutolinkReferenceV1().CoreConfigSchema().ImpliedType(),
+				Upgrade: resourceGithubRepositoryAutolinkReferenceStateUpgradeV1,
+				Version: 1,
 			},
 		},
 
-		SchemaVersion: 1,
+		Description: "Creates and manages an autolink reference for a single repository.",
```
stevehipwell:
```suggestion
		Description: "Resource to manage an autolink reference for a repository.",
```
For consistency with refactored documentation.

stevehipwell (changes requested):
This is looking good, just a few minor docs changes. Probably also want to add them to the contributing guide in another PR for consistency.

templates/resources/repository_autolink_reference.md.tmpl
deiga: I wouldn't honestly document the permissions needed for resources/data sources. Instead I'd add links to the GH API docs to read the permissions from. That should fit into the `Description` field nicely and reduce the burden of trying to keep up-to-date with upstream changes with permissions.

And with that IMO, we can use the generic templates for most IMO

stevehipwell: The description is used for the page title, so it should be a simple sentence/paragraph with the template containing the bulk of the documentation. There is no real benefit to bulking out the Go code with a significant amount of resource level documentation, but there are a number of downsides.

stevehipwell (changes requested):
I'm still of the opinion that we have a template per resource/data-source and keep the high level description concise with additional content added to the template.

PR #3272, opened 2026-03-13
ci: honest greeting + split PR/issue response jobs

eyalgal:
Replace the misleading "reviews twice a week, every Monday and Friday" greeting with an honest note, and split the single `respond` job into event-scoped jobs. The Copilot-powered issue triage that was previously in this PR is deferred to a follow-up, per discussion in this thread.

----

### Before the change?

- A single `respond` job posts one greeting on both issues and PRs.
- The greeting promises "reviews twice a week, every Monday and Friday" — which hasn't matched reality for a long time.

### After the change?

**Honest greeting.** The comment body no longer promises a cadence. It states that the repo is maintained by GitHub and community members on a best-effort basis, and keeps the existing `Status: Up for grabs` community pointer.

**Split jobs.** `respond` becomes two jobs:
- `respond-to-pr` — triggered on `pull_request_target`, requests only `pull-requests: write`.
- `respond-to-issue` — triggered on `issues`, requests only `issues: write`.

Each job has a tailored opener ("thank you for this contribution" vs. "thank you for opening this issue") and the same honest body.

No new dependencies, no new external calls, no template parsing. The Copilot-powered triage (template completeness checks, duplicate search, Models API analysis) will come in a follow-up PR with the concerns from this thread addressed (extracted script, model id as env var, repo-variable kill switch, quiet-on-complete-issues behavior).

### Pull request checklist

- [ ] Yes
- [x] No

----


.github/workflows/immediate-response.yaml
```diff
@@ -11,23 +11,366 @@ on:
...
+            seen.add(issue.number);
+
+            for (const term of searchTerms) {
+              try {
+                const results = await github.rest.search.issuesAndPullRequests({
+                  q: `repo:integrations/terraform-provider-github is:issue state:open ${term}`,
+                  per_page: 5,
+                  sort: 'reactions',
+                  order: 'desc'
+                });
```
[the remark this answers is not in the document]
eyalgal: Wrong. sort supports `reactions` ([source](https://docs.github.com/en/rest/search/search?apiVersion=2026-03-10#search-issues-and-pull-requests))

.github/workflows/immediate-response.yaml
```diff
@@ -11,23 +11,366 @@ on:
...
+              // Truncate body to avoid token limits & reduce injection surface
+              const sanitizedBody = body.substring(0, 3000);
+              const duplicateContext = duplicateCandidates.length > 0
+                ? duplicateCandidates.map(d => `#${d.number}: ${d.title}`).join('\n')
+                : 'None found';
+
+              const response = await fetch('https://models.github.ai/inference/chat/completions', {
+                method: 'POST',
+                headers: {
+                  'Authorization': `Bearer ${process.env.GITHUB_TOKEN}`,
```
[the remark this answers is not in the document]
eyalgal: I don't think that's true. @robert-crandall ?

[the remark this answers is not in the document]
deiga: I don't think it's available as env variable? It's in secrets.GITHUB_TOKEN, which doesn't automatically translate to env

.github/workflows/immediate-response.yaml
```diff
@@ -11,23 +11,367 @@ on:
...
+        uses: actions/github-script@60a0d83039c74a4aee543508d2ffcb1c3799cdea # v7.0.1
+        with:
+          script: |
+            const issue = context.payload.issue;
+            const body = issue.body || '';
+            const title = issue.title || '';
+
+            // ── 1. Determine issue type from title prefix ──
+            let issueType = 'unknown';
+            if (title.startsWith('[BUG]')) issueType = 'bug';
```
austenstone: This handles `[BUG]`, `[FEAT]`, and `[DOCS]`, but the repo also has a `[MAINT]` issue template. Right now maintenance issues fall into `unknown`, so they won't get the same validation or prompt context. Can we include `[MAINT]` here too?

.github/workflows/immediate-response.yaml
```diff
@@ -11,23 +11,367 @@ on:
...
+  # ──────────────────────────────────────────────
+  # PR Response — static greeting + Copilot review notice
+  # ──────────────────────────────────────────────
+  respond-to-pr:
+    name: Respond to PR
+    if: >
+      github.event_name == 'pull_request_target' &&
+      github.actor != 'dependabot[bot]' &&
+      github.actor != 'renovate[bot]' &&
+      github.actor != 'githubactions[bot]' &&
```
austenstone: Minor: I think this bot name should be `github-actions[bot]`, not `githubactions[bot]`. Same thing a bit lower in the workflow too.

austenstone (commented):
Left two comments: one on missing `[MAINT]` handling in the issue-type parsing, and one small fix on the bot name (`github-actions[bot]`).

.github/workflows/immediate-response.yaml
eyalgal: Good catch! Fixed in 2da6b86 — changed to `github-actions[bot]` in both the PR and issue jobs.

.github/workflows/immediate-response.yaml
eyalgal: Good call — added `[MAINT]` → `maintenance` to the type detection and also included `MAINT` in the `cleanTitle` regex. Fixed in 2da6b86.

.github/workflows/immediate-response.yaml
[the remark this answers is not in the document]
eyalgal: You're right — `actions/github-script` injects the token into the `github` Octokit client but doesn't set it as an env var. Added `env: GITHUB_TOKEN: ${{ github.token }}` to the step. Fixed in 2da6b86.

deiga (dismissed):

robert-crandall (changes requested):
I want to hold off on this PR. I view this as a cool and clever use case for what Copilot can do, but I do not want to experiment on this repository with unproven Copilot uses. This repository is already under-staffed, and potential improvements like this come with a real risk of increasing workload instead of reducing it. For now, let's implement Copilot only when it has proven its value. For example, I view CCR as a proven value.

When coming back to this PR, some non-exhaustive thoughts:
- I don't think this organization has Copilot enabled. That needs to be resolved before implementing this or CCR.
- The logic to review issues should move to a new file, and actions should call that file. That will make it easier to test and to review.
- The model chosen, `gpt-40-mini`, should be a variable

eyalgal:
@robert-crandall I hear the concern about experimenting on an under-staffed repo, but I'd push back a bit on the framing here.

The current workflow is actively unhelpful: it promises "reviews twice a week, every Monday and Friday" which hasn't been true in a long time, and gives zero triage value. A maintainer's first touch on every issue is spent asking for basic info that the template already requested. With 40-50 new issues a month, that adds up fast.

This isn't an aggressive use of Copilot. It's template validation, duplicate search, and targeted follow-up questions. I simulated it against real issues from this repo, and in every case it asked for exactly what maintainers end up asking manually, sometimes hours later. I'm happy to share those simulations.

I'll monitor it closely after merge and we can roll back immediately if it creates noise instead of reducing it. But waiting for "proven value" on a workflow that replaces something already broken feels like the wrong bar.

Re: the specific suggestions: extracting the logic into a separate file and making the model a variable are good improvements, but I'd treat those as follow-ups rather than blockers. The workflow works as-is, and for this use case `gpt-4o-mini` is the right fit. On Copilot being enabled: I requested a Copilot review on this very PR and it ran successfully, so it's enabled in some capacity. Happy to verify the exact scope, but it shouldn't block merging.

nickfloyd (dismissed):

.github/workflows/immediate-response.yaml
```diff
@@ -11,23 +11,371 @@ on:
...
+  # ──────────────────────────────────────────────
+  respond-to-pr:
+    name: Respond to PR
+    if: >
+      github.event_name == 'pull_request_target' &&
+      github.actor != 'dependabot[bot]' &&
+      github.actor != 'renovate[bot]' &&
+      github.actor != 'github-actions[bot]' &&
+      github.actor != 'octokitbot' &&
+      github.repository == 'integrations/terraform-provider-github'
```
stevehipwell, line 25: Is this a possible combination, isn't `github.repository` going to point at the source repository?

stevehipwell (changes requested):
At a high level I agree that the current workflow doesn't really fit but I'm not sure about the value of adding more signal (we already have too much signal). Wouldn't a better MVP solution be to fix the current behaviour? How about an initial PR to remove this workflow and add a separate workflow for issues and PRs just adding a correct comment in the vein of the original. I'm confident that this would be non-contentious.

After there are separate MVP workflows you could re-add these suggestions as 2 seperate PRs which would be easier to review.

.github/workflows/immediate-response.yaml
eyalgal: @stevehipwell Thanks for the review!

Good catch — you're right that it's redundant here. pull_request_target always runs in the base repo context, and issues only fire on this repo's own tracker. Removed it from both jobs in [7945d60](https://github.com/integrations/terraform-provider-github/pull/3272/changes/7945d6073362bb648c99ea1218ef1f7e0832007b).

I considered splitting into multiple PRs, but the workflow is a single file and the two jobs are independent — splitting them into 3 sequential PRs (remove old → add greeting → add triage) would triple the review cycles without making any individual PR easier to review. I'd rather land this and iterate.

The current workflow actively misleads reporters by promising reviews "twice a week, every Monday and Friday," which hasn't been accurate for a while. This PR replaces that with honest expectations and adds triage that asks for the same missing info maintainers end up requesting manually. If it turns out to add noise instead of reducing it, I'll revert quickly — but leaving the broken workflow in place has a real daily cost.

Would you be willing to convert to an approval given the above? Happy to discuss further.

deiga:
@eyalgal Could you split this into multiple PRs with the first being least changes to make the "immediate-response" workflow useful again?

eyalgal:
@deiga — splitting it up. Revised plan after thinking it through:

**This PR (now rescoped) — Fix the greeting + split PR/issue jobs.** One small PR because the two changes share a rationale ("make the workflow useful + stop promising a cadence we don't have"), and once the messages diverge the job split is the natural shape. Concretely:
- Rewrites the comment body so it no longer claims "reviews twice a week, every Monday and Friday." Replaces it with an honest note that the repo is maintained by GitHub and community members on a best-effort basis, plus the existing `Status: Up for grabs` nudge.
- Splits `respond` into `respond-to-pr` (trigger: `pull_request_target`) and `respond-to-issue` (trigger: `issues`), with scoped permissions per job.
- No new dependencies, no Copilot, no template parsing.

I've just force-pushed this branch to that scope (net diff: one file, +53/-14). Please take another look.

**Follow-up PR — Copilot-powered issue triage (the real thing).** I'd originally planned a deterministic-only version as an intermediate step, but I think that's the wrong move: a rules-only triage comment (missing-field checklist + duplicate list) would look like the final product while lacking the two things that make it feel smart — asking the one specific follow-up that matters, and *staying quiet* when the issue is already well-described. Shipping that as an intermediate state risks proving the "noise instead of signal" concern right. So the follow-up will be the full thing, with the concerns from this thread baked in:
- Script extracted to `.github/scripts/triage-issue.js` (per @robert-crandall's earlier feedback).
- Model id + endpoint as workflow-level `env:` vars so swapping models is a one-line change.
- **Kill switch:** `if: vars.ENABLE_COPILOT_TRIAGE != 'false'` on the job — on by default, flip a repo variable to disable instantly without a revert.
- Deterministic checks (template completeness, duplicate search) run first and gate the LLM output; if the issue is complete, no duplicates found, and the model returns `LGTM`, the job posts nothing beyond a minimal greeting.
- PR description will include a rollout note: monitor the first ~20 issues post-merge; any false-positive complaint → flip the repo var off.

Does this sequencing work?

deiga (dismissed):

robert-crandall (dismissed):

stevehipwell (changes requested):
Could we remove the up for grabs part as I'd rather focus the community on reporting issues, discussing changes and reviewing PRs. We already have more code than we can currently handle, so let's get to a state where we can handle it before inviting more contributions. I feel that with some more community engagement on the issues we could get better visibility into what end-users need; I'm a fan of the OpenTofu top issues pinned issue.

eyalgal:
@stevehipwell done. Dropped the `Status: Up for grabs` line from both greetings and replaced it with a nudge toward what you described: joining discussions, sharing details, and reviewing PRs. The issue greeting also asks for details, repro steps, or a 👍 to improve signal. Happy to do the OpenTofu-style pinned "top issues" as a follow-up. Pushed in 4f3fe91b, mind taking another look?

stevehipwell (dismissed):
LGTM

stevehipwell (approved):
LGTM

deiga (approved):

PR #3274, opened 2026-03-15
feat: add hosted runner private networking support

### LINKED ISSUE #2302: [FEAT]: github_enterprise_actions_runner_group - Support network configurations and static IP
### Describe the need

For the resources `github_enterprise_actions_runner_group` and `github_actions_runner_group` I would like support for configuring private networking via Terraform. This would help ease setup and maintenance of a great and cost efficient feature.

```hcl

data "github_enterprise" "example" {
  slug = "example-co"
}

resource "azapi_resource" "subnet_integration" {
  type                      = "GitHub.Network/networkSettings@2024-04-02"
  name                      = "test-subnet-integration"
  location                  = module.vnet.vnet.location
  parent_id                 = module.vnet.rg.id
  schema_validation_enabled = false
  body = jsonencode({
    properties = {
      subnetId   = module.vnet.subnets["test-subnet"].id
      businessId = data.github_enterprise.example.database_id
    }
  })

  lifecycle {
    ignore_changes = [tags]
  }
}

resource "github_enterprise_actions_runner_group" "example" {
  name                       = "my-awesome-runner-group"
  enterprise_slug            = data.github_enterprise.enterprise.slug
  selected_organization_ids  = [github_enterprise_organization.enterprise_organization.database_id]
  
  # NEW
  network_configuration_id   = azapi_resource.subnet_integration.tags["GithubId"] # optional
  disable_network_configuration = false # optional. Only has an effect if network_configuration_id is set
}

resource "azapi_resource" "subnet_integration2" {
  type                      = "GitHub.Network/networkSettings@2024-04-02"
  name                      = "test-subnet-integration"
  location                  = module.vnet.vnet.location
  parent_id                 = module.vnet.rg.id
  schema_validation_enabled = false
  body = jsonencode({
    properties = {
      subnetId   = module.vnet.subnets["test-subnet2"].id
      businessId = data.github_enterprise.example.database_id
    }
  })

  lifecycle {
    ignore_changes = [tags]
  }
}


resource "github_actions_runner_group" "example" {
  name                    = "xxx"
  selected_repository_ids = ["yyy"]

  # NEW
  network_configuration_id   = azapi_resource.subnet_integration2.tags["GithubId"] # optional
  disable_network_configuration = false # optional. Only has an effect if network_configuration_id is set
}

```

Alternatively an optional block can be used to configue this

```hcl
resource "github_actions_runner_group" "example" {
...
{
  network_configuration_id   = azapi_resource.subnet_integration2.tags["GithubId"] # optional
  disable_network_configuration = false # optional. Only has an effect if network_configuration_id is set
}
...
}
```

Enterprise:
- https://docs.github.com/en/enterprise-cloud@latest/admin/configuring-settings/configuring-private-networking-for-hosted-compute-products/configuring-private-networking-for-github-hosted-runners-in-your-enterprise#1-add-a-new-network-configuration-for-your-enterprise

Github team:
- https://docs.github.com/en/organizations/managing-organization-settings/configuring-private-networking-for-github-hosted-runners-in-your-organization#about-azure-private-networking-for-github-hosted-runners

### SDK Version

_No response_

### API Version

_No response_

### Relevant log output

_No response_

### Code of Conduct

austenstone:
Resolves #2302

----

### Before the change?

Hosted-runner network configurations and their runner-group assignments cannot be managed through the provider.

### After the change?

Manage network configurations and runner-group assignments at organization and enterprise scope using native go-github methods. Both runner-group create endpoints accept `network_configuration_id`; no separate networking PATCH is needed after creation.

`network_configuration_id` is Optional+Computed. Omitting it or setting `null` retains the existing assignment. Setting `""` clears it by replacing the runner group.

**Draft: replacement on clearing is not merge-ready.** [google/go-github#4541](https://github.com/google/go-github/pull/4541) must merge, ship, and be adopted here for non-destructive removal. The REST API already supports explicit null. Upstream main is merged, including its released go-github v92.0.0 dependency, but v92 does not contain the detachment fix.

The requested split between the new resources and runner-group integration has not been done. That scope decision remains unresolved.

Implementation and review fixes were substantially AI-assisted with Copilot.

### Pull request checklist

- [x] Schema migrations have been created if needed ([example](https://github.com/F-Secure-web/terraform-provider-github/blob/main/github/migrate_github_actions_organization_secret.go)). No migration is needed for these additions.

September 18 local checks passed on Go 1.26.0: full unit suite, both mocked Terraform runner-group lifecycles, baseline and strict new-code lint, build, example formatting, and import-script syntax. The targeted networking race run passed 150 tests/subtests with zero skips. Docs regenerated unchanged and validated with explicit `--provider-name github` for the isolated worktree.

Live acceptance and Azure connectivity have **not** been rerun for this revision. The September 10 live results predate these review fixes. Fixture requirements are in [CONTRIBUTING.md](CONTRIBUTING.md#private-networking-acceptance-tests).

[CI](https://github.com/integrations/terraform-provider-github/actions/runs/35390042939) and [CodeQL](https://github.com/integrations/terraform-provider-github/actions/runs/35390042919) require maintainer approval. The actual [acceptance test job](https://github.com/integrations/terraform-provider-github/actions/runs/35390040031/job/105746032762) was skipped; its wrapper result is not an acceptance-test pass.

- [ ] Yes
- [x] No

Existing configurations without the new attributes retain their behavior.

----


deiga:
Your validation commands would not actually run the tests. You should use make testacc and make test

tag-assistant (commented):
## Code Review

Nice work on a complex feature spanning org + enterprise scopes. Overall this is solid, well-structured, and well-tested. A few observations:

### ✅ Highlights

- **DRY refactor** — Extracting `setGithubActionsRunnerGroupState()` and `setGithubActionsEnterpriseRunnerGroupState()` eliminates repeated `d.Set()` blocks across Create/Read. Welcome cleanup.
- **Defensive nil checks** — Adding `ghErr.Response != nil` before checking `StatusCode` fixes potential nil pointer panics.
- **`log.Printf` → `tflog`** — Structured logging with `tflog.SetField` is the right call for Terraform plugins.
- **Create-then-PATCH pattern** — Well-documented workaround for the API limitation. Comments explain clearly.
- **Test coverage** — Full lifecycle: create without networking → add → import → remove. Both org and enterprise scopes.
- **Docs** — The Azure `GitHubId` vs ARM resource ID callout is especially helpful.

### ⚠️ Suggestions

**1. Duplicate structs could be shared**
`organizationRunnerGroup` and `enterpriseRunnerGroup` are identical. Same for `get*RunnerGroupNetworking` and `update*RunnerGroupNetworking` — they differ only in URL path. A shared type + helper taking a base path could cut ~60 lines.

**2. `StatusNotModified` fix is a bug fix mixed into a feature PR**
`getOrganizationRunnerGroup` previously swallowed all `ErrorResponse` errors. The narrowing to `StatusNotModified` only is a correctness fix worth calling out separately in the PR description.

**3. Missing newline at EOF**
`enterprise_network_configuration.html.markdown` is missing a trailing newline.

**4. `context.Background()` in Read for networking call**
```go
runnerGroupNetworking, _, err := getOrganizationRunnerGroupNetworking(client,
    context.WithValue(context.Background(), ctxId, d.Id()), ...)
```
A `ctx` with the etag is already in scope — any reason to create a fresh `context.Background()` here?

**5. Create returns `nil` instead of calling Read**
Intentional (state set directly), but means server-side defaults not in the Create response won't appear until next refresh. Worth noting the pattern change.

**6. No explicit `CheckDestroy` in tests**
Terraform's framework usually handles cleanup, but an explicit `CheckDestroy` function would be stronger for the new network configuration resources.

### 📊 Summary

| Aspect | Rating |
|--------|--------|
| Code quality | ⭐⭐⭐⭐ |
| Test coverage | ⭐⭐⭐⭐ |
| Documentation | ⭐⭐⭐⭐⭐ |
| DRY / maintainability | ⭐⭐⭐ (room for dedup) |

Ship it 🚢

austenstone:
`normalizeEtag` is staying because removing it breaks the behavior introduced in this PR, not because we’re broadening scope.

The GitHub REST API docs explicitly describe ETag-based conditional requests and show weak ETags in examples: https://docs.github.com/en/rest/using-the-rest-api/best-practices-for-using-the-rest-api#use-conditional-requests-if-appropriate

Without normalization, the targeted enterprise runner group import acceptance test fails because the API can return the same validator with and without the `W/` weak ETag prefix. That creates Terraform state drift on `etag` alone even when the resource is otherwise equivalent.

So this helper is just normalizing wire-format variance to keep create/read/import stable for the runner group resources touched here.

tag-assistant (approved):
## Re-review ✅

All previous feedback addressed, plus some nice extras:

**Addressed:**
- ✅ Shared `runnerGroupNetworking` struct + helpers in `resource_github_actions_runner_group_networking.go` — clean dedup
- ✅ `getRunnerGroupNetworking` now takes a path string, works for both org and enterprise scopes
- ✅ Trailing newline fixed on enterprise docs
- ✅ `StatusNotModified` handling added to `getRunnerGroupNetworking` — consistent with the rest of the codebase

**Bonus improvements (not requested):**
- 🧪 173-line unit test file for all three shared helpers (`get`, `update`, `setState`) — including 304 handling and null payload for removal. Solid.
- 🔧 `normalizeEtag()` in util.go to handle weak vs strong ETag drift between create and read paths — nice catch, prevents state churn.
- 🧪 Unit tests for `normalizeEtag` covering empty, strong, weak, and whitespace cases.

LGTM. Ship it 🚢

github/resource_github_actions_runner_group.go
deiga, line 1: Since you're modifying this file, please migrate the CRUD functions to be Context-aware

github/resource_github_actions_runner_group.go
```diff
@@ -92,10 +92,69 @@ func resourceGithubActionsRunnerGroup() *schema.Resource {
 				Optional:    true,
 				Description: "List of workflows the runner group should be allowed to run. This setting will be ignored unless restricted_to_workflows is set to 'true'.",
 			},
+			"network_configuration_id": {
+				Type:         schema.TypeString,
+				Optional:     true,
+				ValidateFunc: validation.StringLenBetween(1, 255),
```
deiga: Issue: use ValidateDiagFunc

github/resource_github_actions_runner_group.go
```diff
@@ -92,10 +92,69 @@ func resourceGithubActionsRunnerGroup() *schema.Resource {
...
 	}
 }
 
+func getOrganizationRunnerGroup(client *github.Client, ctx context.Context, org string, groupID int64) (*github.RunnerGroup, *github.Response, error) {
+	runnerGroup, resp, err := client.Actions.GetOrganizationRunnerGroup(ctx, org, groupID)
+	if err != nil {
+		var ghErr *github.ErrorResponse
+		if errors.As(err, &ghErr) && ghErr.Response != nil && ghErr.Response.StatusCode == http.StatusNotModified {
+			// ignore error StatusNotModified
+			return runnerGroup, resp, nil
```
deiga: Issue: runnerGroup is nil here, plead return that explicitly

github/resource_github_actions_runner_group.go
```diff
@@ -148,57 +207,26 @@ func resourceGithubActionsRunnerGroupCreate(d *schema.ResourceData, meta any) er
...
-		return err
+	if networkConfigurationID, ok := d.GetOk("network_configuration_id"); ok {
+		networkConfigurationIDValue := networkConfigurationID.(string)
+		// The create endpoint does not accept network_configuration_id, so private networking
+		// must be attached with a follow-up PATCH after the runner group has been created.
+		if _, err = updateRunnerGroupNetworking(client, ctx, fmt.Sprintf("orgs/%s/actions/runner-groups/%d", orgName, runnerGroup.GetID()), &networkConfigurationIDValue); err != nil {
+			return err
+		}
+
+		return resourceGithubActionsRunnerGroupRead(d, meta)
```
deiga: issue: Never call CRUD functions directly.

github/resource_github_actions_runner_group_helpers.go
```diff
@@ -0,0 +1,66 @@
...
+
+	return &runnerGroup, resp, nil
+}
+
+func updateRunnerGroupNetworking(client *github.Client, ctx context.Context, path string, networkConfigurationID *string) (*github.Response, error) {
+	payload := map[string]any{
+		"network_configuration_id": networkConfigurationID,
+	}
+
+	req, err := client.NewRequest("PATCH", path, payload)
```
deiga: issue: use only go-github provided functions

github/resource_github_actions_runner_group.go
```diff
@@ -233,42 +261,13 @@ func resourceGithubActionsRunnerGroupRead(d *schema.ResourceData, meta any) erro
 		return err
 	}
 
-	// if runner group is nil (typically not modified) we can return early
-	if runnerGroup == nil {
-		return nil
+	runnerGroupEtag := normalizeEtag(d.Get("etag").(string))
+	if resp != nil {
+		runnerGroupEtag = normalizeEtag(resp.Header.Get("ETag"))
 	}
```
deiga: What's the purpose of this structure? In what cases would resp be nil?

github/resource_github_actions_runner_group.go
```diff
@@ -339,6 +352,18 @@ func resourceGithubActionsRunnerGroupUpdate(d *schema.ResourceData, meta any) er
...
 
+	if d.HasChange("network_configuration_id") {
+		var networkConfigurationIDValue *string
+		if networkConfigurationID, ok := d.GetOk("network_configuration_id"); ok {
+			value := networkConfigurationID.(string)
+			networkConfigurationIDValue = &value
+		}
+
+		if _, err := updateRunnerGroupNetworking(client, ctx, fmt.Sprintf("orgs/%s/actions/runner-groups/%d", orgName, runnerGroupID), networkConfigurationIDValue); err != nil {
```
deiga: Why is this separate patching necessary at this stage? Can't it use the normal API operations?

github/resource_github_actions_runner_group_helpers.go
```diff
@@ -0,0 +1,66 @@
...
+	"github.com/google/go-github/v83/github"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+)
+
+type runnerGroupNetworking struct {
+	NetworkConfigurationID *string `json:"network_configuration_id,omitempty"`
+}
+
+func getRunnerGroupNetworking(client *github.Client, ctx context.Context, path string) (*runnerGroupNetworking, *github.Response, error) {
+	req, err := client.NewRequest("GET", path, nil)
```
deiga: issue: use go-github provided functions

github/resource_github_actions_runner_group_networking.go
```diff
@@ -0,0 +1,66 @@
...
+func setRunnerGroupNetworkingState(d *schema.ResourceData, runnerGroup *runnerGroupNetworking) error {
+	if runnerGroup != nil && runnerGroup.NetworkConfigurationID != nil && *runnerGroup.NetworkConfigurationID != "" {
+		if err := d.Set("network_configuration_id", *runnerGroup.NetworkConfigurationID); err != nil {
+			return err
+		}
+	} else {
+		if err := d.Set("network_configuration_id", nil); err != nil {
+			return err
+		}
+	}
```
deiga: What's the purpose of this structure? d.Set is able to handle pointers automatically AFAIK

github/resource_github_actions_runner_group_helpers_test.go
deiga, line 1: issue: Acceptance tests are missing

github/resource_github_actions_runner_group_test.go
```diff
@@ -4,13 +4,26 @@ import (
...
+func testCheckRunnerGroupNetworkConfigurationMatches(resourceName, networkConfigurationResourceName string) statecheck.StateCheck {
+	return statecheck.CompareValuePairs(
+		resourceName,
+		tfjsonpath.New("network_configuration_id"),
+		networkConfigurationResourceName,
+		tfjsonpath.New("id"),
+		compare.ValuesSame(),
+	)
+}
+
```
deiga: Inline this, the deduplication reduces readability

github/resource_github_actions_runner_group_test.go
```diff
@@ -97,6 +110,109 @@ func TestAccGithubActionsRunnerGroup(t *testing.T) {
...
+					ConfigStateChecks: []statecheck.StateCheck{
+						statecheck.ExpectKnownValue(resourceName, tfjsonpath.New("network_configuration_id"), knownvalue.NotNull()),
+						testCheckRunnerGroupNetworkConfigurationMatches(resourceName, networkConfigurationResourceName),
+					},
+				},
+				{
+					ResourceName:      resourceName,
+					ImportState:       true,
+					ImportStateVerify: true,
+				},
```
deiga: Use separate import test

github/resource_github_actions_runner_group_test.go
```diff
@@ -97,6 +110,109 @@ func TestAccGithubActionsRunnerGroup(t *testing.T) {
...
+				{
+					Config: configWithoutNetworkConfiguration,
```
deiga: What's the purpose of this?

github/resource_github_enterprise_actions_runner_group.go
```diff
@@ -74,6 +74,12 @@ func resourceGithubActionsEnterpriseRunnerGroup() *schema.Resource {
 				Optional:    true,
 				Description: "List of workflows the runner group should be allowed to run. This setting will be ignored unless restricted_to_workflows is set to 'true'.",
 			},
+			"network_configuration_id": {
+				Type:         schema.TypeString,
+				Optional:     true,
+				ValidateFunc: validation.StringLenBetween(1, 255),
```
deiga: ValidateDiagFunc

github/resource_github_enterprise_actions_runner_group.go
```diff
@@ -125,7 +169,7 @@ func resourceGithubActionsEnterpriseRunnerGroupCreate(d *schema.ResourceData, me
 
 	ctx := context.Background()
 
-	enterpriseRunnerGroup, resp, err := client.Enterprise.CreateEnterpriseRunnerGroup(ctx,
+	runnerGroup, resp, err := client.Enterprise.CreateEnterpriseRunnerGroup(ctx,
```
deiga: What do we gain from making the name less descriptive?

github/resource_github_enterprise_actions_runner_group.go
```diff
@@ -139,49 +183,34 @@ func resourceGithubActionsEnterpriseRunnerGroupCreate(d *schema.ResourceData, me
...
+
+	if networkConfigurationID, ok := d.GetOk("network_configuration_id"); ok {
+		networkConfigurationIDValue := networkConfigurationID.(string)
+		// The create endpoint does not accept network_configuration_id, so private networking
+		// must be attached with a follow-up PATCH after the runner group has been created.
+		if _, err = updateRunnerGroupNetworking(client, ctx, fmt.Sprintf("enterprises/%s/actions/runner-groups/%d", enterpriseSlug, runnerGroup.GetID()), &networkConfigurationIDValue); err != nil {
+			return err
+		}
+
+		return resourceGithubActionsEnterpriseRunnerGroupRead(d, meta)
```
deiga:
```suggestion
		return nil
```

github/resource_github_enterprise_actions_runner_group.go
deiga, line 1: While modifying this file, migrate to use Context-aware CRUD functions

github/resource_github_enterprise_actions_runner_group.go
```diff
@@ -198,6 +227,8 @@ func resourceGithubActionsEnterpriseRunnerGroupRead(d *schema.ResourceData, meta
 		return err
 	}
 	ctx := context.WithValue(context.Background(), ctxId, d.Id())
+	ctx = tflog.SetField(ctx, "enterprise_slug", enterpriseSlug)
```
deiga: Remove these

github/resource_github_enterprise_actions_runner_group.go
```diff
@@ -342,8 +372,11 @@ func resourceGithubActionsEnterpriseRunnerGroupDelete(d *schema.ResourceData, me
...
+	ctx = tflog.SetField(ctx, "id", d.Id())
+	ctx = tflog.SetField(ctx, "name", d.Get("name"))
```
deiga: Remove all except id

github/resource_github_enterprise_actions_runner_group_test.go
```diff
@@ -2,13 +2,28 @@ package github
...
+func testCheckEnterpriseRunnerGroupNetworkConfigurationMatches(resourceName, networkConfigurationResourceName string) statecheck.StateCheck {
+	return statecheck.CompareValuePairs(
+		resourceName,
+		tfjsonpath.New("network_configuration_id"),
+		networkConfigurationResourceName,
+		tfjsonpath.New("id"),
+		compare.ValuesSame(),
+	)
+}
+
```
deiga: Inline this

deiga (changes requested):
Partial review

github/resource_github_actions_runner_group.go
austenstone: Done — this file now uses context-aware CRUD handlers (`CreateContext`/`ReadContext`/`UpdateContext`/`DeleteContext`).

github/resource_github_actions_runner_group_helpers_test.go
austenstone: The helper file is unit coverage only; the acceptance coverage is in the runner group resource tests. I also split the org networking import coverage into its own acceptance scenario.

github/resource_github_enterprise_actions_runner_group.go
austenstone: Fair point — I switched the create response variable back to `enterpriseRunnerGroup`.

github/resource_github_actions_runner_group_test.go
austenstone: Done — the networking import coverage is now its own acceptance test.

github/resource_github_actions_runner_group_helpers.go
austenstone: I checked `go-github v83` locally before changing this — there isn’t a first-class runner-group networking method for this endpoint yet, so this still has to stay as a thin `NewRequest`/`Do` shim for now.

github/resource_github_enterprise_actions_runner_group.go
austenstone: Updated — delete logging now keeps only `id`.

github/resource_github_actions_runner_group_test.go
austenstone: That was the combined import step from the earlier test shape — removed as part of splitting the separate import scenario.

github/resource_github_enterprise_actions_runner_group.go
austenstone: Done — migrated this resource to context-aware CRUD handlers.

github/resource_github_enterprise_actions_runner_group.go
austenstone: Removed — the extra read-time log fields are gone.

github/resource_github_actions_runner_group_helpers.go
austenstone: Same constraint here: I checked the current `go-github` surface and there isn’t a provided function for the runner-group networking GET/PATCH endpoints yet.

github/resource_github_enterprise_actions_runner_group.go
austenstone: Updated to `ValidateDiagFunc` here as well.

github/resource_github_enterprise_actions_runner_group_test.go
austenstone: Updated — inlined this in the enterprise acceptance test too.

github/resource_github_actions_runner_group_test.go
austenstone: Updated — inlined this in the org runner group acceptance test.

github/resource_github_enterprise_actions_runner_group.go
austenstone: Applied — create now returns after setting state directly instead of bouncing through `Read`.

github/resource_github_actions_runner_group.go
austenstone: Updated — create/update no longer call the resource CRUD methods directly. They now set state from the API response instead.

github/resource_github_actions_runner_group.go
austenstone: Because the normal runner-group create/update APIs don’t accept `network_configuration_id`. The association is only possible via the separate networking endpoint, so the follow-up PATCH is still required.

github/resource_github_actions_runner_group_networking.go
austenstone: Good call — I simplified this and now let `d.Set()` handle the pointer directly.

github/resource_github_actions_runner_group.go
austenstone: Updated to `ValidateDiagFunc` for `network_configuration_id`.

github/resource_github_actions_runner_group.go
austenstone: Fixed — the `304 Not Modified` path now returns `nil` explicitly for the runner group.

github/resource_github_actions_runner_group.go
austenstone: The fallback keeps the current normalized ETag if the response doesn’t provide a fresh header. The nil guard is just defensive in case we ever get a transport/client path without a populated response.

github/resource_github_actions_runner_group.go
```diff
@@ -131,7 +190,7 @@ func resourceGithubActionsRunnerGroupCreate(d *schema.ResourceData, meta any) er
 		}
 	}
 
-	ctx := context.Background()
+	ctx = context.WithValue(ctx, ctxId, d.Id())
```
deiga:
```suggestion
```

github/resource_github_actions_runner_group.go
```diff
@@ -145,76 +204,49 @@ func resourceGithubActionsRunnerGroupCreate(d *schema.ResourceData, meta any) er
...
-	}
-	if err = d.Set("selected_repository_ids", selectedRepositoryIDs); err != nil { // Note: runnerGroup has no method to get selected repository IDs
-		return err
-	}
-	if err = d.Set("restricted_to_workflows", runnerGroup.GetRestrictedToWorkflows()); err != nil {
-		return err
-	}
-	if err = d.Set("selected_workflows", runnerGroup.SelectedWorkflows); err != nil {
-		return err
+	ctx = context.WithValue(ctx, ctxId, d.Id())
```
deiga:
```suggestion
```

github/resource_github_actions_runner_group.go
```diff
@@ -145,76 +204,49 @@ func resourceGithubActionsRunnerGroupCreate(d *schema.ResourceData, meta any) er
...
+		return diag.FromErr(err)
 	}
 
-	return resourceGithubActionsRunnerGroupRead(d, meta)
-}
+	if networkConfigurationID, ok := d.GetOk("network_configuration_id"); ok {
+		networkConfigurationIDValue := networkConfigurationID.(string)
+		// The create endpoint does not accept network_configuration_id, so private networking
+		// must be attached with a follow-up PATCH after the runner group has been created.
+		if _, err = updateRunnerGroupNetworking(client, ctx, fmt.Sprintf("orgs/%s/actions/runner-groups/%d", orgName, runnerGroup.GetID()), &networkConfigurationIDValue); err != nil {
```
deiga: Isn't this `UpdateOrganizationRunnerGroup` what you're looking for?

github/resource_github_actions_runner_group.go
```diff
@@ -145,76 +204,49 @@ func resourceGithubActionsRunnerGroupCreate(d *schema.ResourceData, meta any) er
...
 	client := meta.(*Owner).v3client
 	orgName := meta.(*Owner).name
 
 	runnerGroupID, err := strconv.ParseInt(d.Id(), 10, 64)
 	if err != nil {
-		return err
+		return diag.FromErr(err)
 	}
-	ctx := context.WithValue(context.Background(), ctxId, d.Id())
+	ctx = context.WithValue(ctx, ctxId, d.Id())
```
deiga, line 238:
```suggestion

```

github/resource_github_actions_runner_group.go
deiga: Please elaborate the edge-cases you foresee? AFAIK resp should only be nil if an error happened in the API call and we fail to return the error there

github/resource_github_actions_runner_group.go
```diff
@@ -294,17 +297,30 @@ func resourceGithubActionsRunnerGroupRead(d *schema.ResourceData, meta any) erro
 		options.Page = resp.NextPage
 	}
 
-	if err = d.Set("selected_repository_ids", selectedRepositoryIDs); err != nil {
-		return err
+	if runnerGroup != nil {
+		if err = setGithubActionsRunnerGroupState(d, runnerGroup, runnerGroupEtag, selectedRepositoryIDs); err != nil {
+			return diag.FromErr(err)
+		}
+	} else {
```
deiga: Please explain this branching logic? Why aren't we setting the state values to nil if api returned nil?

github/resource_github_actions_runner_group.go
deiga: What about UpdateOrganizationRunnerGroup in go-github?

github/resource_github_actions_runner_group_helpers.go
deiga: If the functions don't exist in go-github, then you should open a PR in there to add them. We don't accept direct API endpoint interactions

github/resource_github_actions_runner_group_helpers_test.go
deiga: This file is called the test file of the runner_group_networking resource. As such it needs to contain acceptance tests for that resource

deiga (changes requested):
Partial review

github/resource_github_actions_runner_group_helpers.go
austenstone: Opened https://github.com/google/go-github/pull/4099 — adds `NetworkConfigurationID` and `HostedRunnersURL` to `EnterpriseRunnerGroup`, `CreateEnterpriseRunnerGroupRequest`, and `UpdateEnterpriseRunnerGroupRequest`.

Once that merges, I'll update this PR to use the native go-github functions and drop the raw `NewRequest`/`Do` calls.

github/resource_github_actions_runner_group_helpers.go
austenstone: With https://github.com/google/go-github/pull/4099, `UpdateOrganizationRunnerGroup` now accepts `NetworkConfigurationID` in the request struct, and `GetOrganizationRunnerGroup` returns it on the response. So yes — once that merges, this raw call can be replaced with the native go-github `UpdateOrganizationRunnerGroup`/`UpdateEnterpriseRunnerGroup` functions and the separate `getRunnerGroupNetworking` GET becomes unnecessary too.

github/resource_github_actions_runner_group.go
austenstone: You're right — `UpdateOrganizationRunnerGroup` (and `UpdateEnterpriseRunnerGroup`) is the right call here. The go-github structs were just missing `NetworkConfigurationID` on the enterprise side. I've opened https://github.com/google/go-github/pull/4099 to add it. Once merged, I'll refactor this to use the native update functions and drop the raw PATCH.

github/resource_github_actions_runner_group.go
austenstone: Yes — once https://github.com/google/go-github/pull/4099 merges, `UpdateOrganizationRunnerGroup` will carry `NetworkConfigurationID` natively and this raw call goes away.

github/resource_github_actions_runner_group.go
austenstone: Fair point — you're right. If the API call errors, we return early above. If it succeeds (including the 304 path), `resp` is always non-nil. The guard is unnecessary. I'll remove it.

github/resource_github_actions_runner_group_helpers_test.go
austenstone: To clarify — `runner_group_networking` isn't a standalone Terraform resource. It's a shared internal helper file (`getRunnerGroupNetworking`, `updateRunnerGroupNetworking`, `setRunnerGroupNetworkingState`) used by both `github_actions_runner_group` and `github_enterprise_actions_runner_group`.

The acceptance coverage for networking is in the runner group resource test files:
- `resource_github_actions_runner_group_test.go` — org networking lifecycle
- `resource_github_enterprise_actions_runner_group_test.go` — enterprise networking lifecycle

This test file covers the shared helpers with unit tests (mocked HTTP). Happy to rename it to something like `runner_group_networking_helpers_test.go` if the current name is misleading.

austenstone: Renamed to `_helpers.go` / `_helpers_test.go` to make it clear this isn't a resource file.

github/resource_github_actions_runner_group.go
austenstone: Removed.

github/resource_github_actions_runner_group.go
austenstone: Removed.

github/resource_github_actions_runner_group.go
austenstone: Removed.

github/resource_github_actions_runner_group.go
austenstone: The `nil` case happens on a `304 Not Modified` — the API returned no body, so `runnerGroup` is nil but `resp` is valid. In that case we can't call `setGithubActionsRunnerGroupState` (it'd nil-pointer on the getter methods), but we still need to update `selected_repository_ids` and `etag` since those came from separate API calls that did succeed. The rest of state stays untouched from the previous read, which is correct for a not-modified response.

austenstone:
All review feedback addressed except the raw `NewRequest`/`Do` calls — those are blocked on [google/go-github#4099](https://github.com/google/go-github/pull/4099) which adds `NetworkConfigurationID` to the runner group structs. Once that merges and we bump the dependency, the helpers file goes away and networking flows through the native `UpdateOrganizationRunnerGroup`/`UpdateEnterpriseRunnerGroup` functions.

github/resource_github_actions_runner_group.go
deiga: That doesn't make sense. The function already exists

github/resource_github_actions_runner_group.go
deiga: If the GET returned nil, then we need to exit early, instead of doing conditional setting

github/resource_github_actions_runner_group.go
deiga: The Org functions are not missing that field.

github/resource_github_actions_runner_group_helpers_test.go
deiga: If it's a file for utilities, then it needs to follow the existing file naming convention

github/resource_github_actions_runner_group.go
austenstone: You're right — my earlier reply was wrong. I confused the org and enterprise types. `UpdateOrganizationRunnerGroup` already had `NetworkConfigurationID` in go-github; the missing field was only on the enterprise side.

Fixed — org Create and Update now pass `NetworkConfigurationID` natively via `CreateRunnerGroupRequest` / `UpdateRunnerGroupRequest`. The separate PATCH is gone. Read also pulls `NetworkConfigurationID` directly from the `RunnerGroup` response instead of the separate networking GET.

Enterprise still uses raw calls since go-github doesn't have `NetworkConfigurationID` on the enterprise types yet (added in https://github.com/google/go-github/pull/4099, merged but not released). Will switch enterprise to native functions once that ships in a go-github release.

github/resource_github_actions_runner_group.go
austenstone: Done — both org and enterprise Read now exit early with `return nil` when the GET returns nil (304 Not Modified). No more conditional state-setting branches.

github/resource_github_actions_runner_group_helpers.go
austenstone: Org side is now fully native — no more raw calls for org runner groups. The helpers file is still needed for enterprise since go-github hasn't released a version with `NetworkConfigurationID` on `EnterpriseRunnerGroup` / `UpdateEnterpriseRunnerGroupRequest` yet (https://github.com/google/go-github/pull/4099 merged but isn't in a release). Once that ships, a follow-up PR will switch enterprise to native functions and remove the helpers entirely.

github/resource_github_actions_runner_group.go
austenstone: You're right — the org functions already had `NetworkConfigurationID`. The separate PATCH was only ever needed for the enterprise side. Org Create/Update now use the native field directly, no more follow-up PATCH.

deiga (changes requested):
Please mark this as Draft until the go-github version is released and an upgrade PR is merged. We will not accept PRs with direct API interactions

austenstone:
Hey @deiga — quick status update: google/go-github#4099 (adds `NetworkConfigurationID` to enterprise runner group types) merged on Mar 18, but go-github hasn't cut a new release since v84.0.0 (Feb 27). ~30 commits are sitting on master unreleased.

I've also opened google/go-github#4101 (custom image endpoints) which merged and is waiting on the same release.

Once v85 drops, I'll bump the dep, remove all raw `NewRequest`/`Do` calls, and convert enterprise networking to native SDK functions. Will move this back to Ready for Review at that point.

deiga:
@austenstone Thanks! Please create the v85 bump in a separate PR and not bundled with this :)

huczas:
@austenstone Did you create separate PR? I'm currently on this and trying to do it using api calls, but your PR would defenitly make a simpler and better TF code.

austenstone:
Status update: just rebased on `main`. Upstream already moved go-github to v88 since @deiga's review, so the v85 bump is already landed and there's nothing left to split out. Adapted to the v88 API drift (`NewRequest`/`Do`/`NewClient` signatures), migrated the new docs to `docs/resources/*.md`, `go build` and `go vet` are green.

@huczas — no separate PR, this is it. Don't duplicate in API calls.

Re-requesting review.

github/resource_github_actions_runner_group_helpers.go
austenstone: Done in [`bef2b1d5`](https://github.com/integrations/terraform-provider-github/pull/3274/commits/bef2b1d509e69128b1d020fb3efd4774ee511536). `resource_github_actions_runner_group_helpers.go` is deleted. go-github v88 carries `NetworkConfigurationID` on `EnterpriseRunnerGroup` / `CreateEnterpriseRunnerGroupRequest` / `UpdateEnterpriseRunnerGroupRequest` (added in google/go-github#4099), so enterprise create/read/update now go through the native `Enterprise.*RunnerGroup` functions. No more raw `NewRequest`/`Do` calls anywhere in this PR.

github/resource_github_actions_runner_group_helpers.go
austenstone: google/go-github#4099 merged and shipped (v85+, branch is on v88), so the helper is gone in [`bef2b1d5`](https://github.com/integrations/terraform-provider-github/pull/3274/commits/bef2b1d509e69128b1d020fb3efd4774ee511536) and enterprise uses the native functions. The create endpoint does accept `network_configuration_id` directly, so the follow-up PATCH was dropped too.

github/resource_github_actions_runner_group.go
austenstone: Right, org was already switched to native `UpdateOrganizationRunnerGroup`/`CreateRunnerGroupRequest` earlier. As of [`bef2b1d5`](https://github.com/integrations/terraform-provider-github/pull/3274/commits/bef2b1d509e69128b1d020fb3efd4774ee511536) enterprise matches: `CreateEnterpriseRunnerGroupRequest` and `UpdateEnterpriseRunnerGroupRequest` carry `NetworkConfigurationID` natively and the raw GET/PATCH helper is deleted.

austenstone:
@deiga ready for another look. The go-github blocker is resolved:

- google/go-github#4099 merged and shipped; this branch is on go-github v88, which has `NetworkConfigurationID` on `EnterpriseRunnerGroup`, `CreateEnterpriseRunnerGroupRequest`, and `UpdateEnterpriseRunnerGroupRequest`.
- [`bef2b1d5`](https://github.com/integrations/terraform-provider-github/pull/3274/commits/bef2b1d509e69128b1d020fb3efd4774ee511536) switches enterprise runner group create/read/update to the native `Enterprise.*RunnerGroup` go-github functions and deletes `resource_github_actions_runner_group_helpers.go` (+ its test). No raw `NewRequest`/`Do` calls remain in the PR.
- The enterprise create endpoint accepts `network_configuration_id` directly, so the follow-up PATCH is gone. Implementation now mirrors the already-native org runner group resource.

Build, vet, gofmt, and unit tests pass.

github/resource_github_actions_runner_group_helpers.go
austenstone: Done. This helper file is deleted; the org CRUD now uses go-github native `client.Actions.*RunnerGroup` functions with no raw API calls.

github/resource_github_actions_runner_group_helpers.go
austenstone: Done. File removed and replaced with go-github native `client.Actions`/`client.Enterprise` runner-group functions, no direct endpoint interaction remains.

github/resource_github_actions_runner_group.go
austenstone: Yes. Switched to `client.Actions.UpdateOrganizationRunnerGroup` using go-github v88 `NetworkConfigurationID`.

github/resource_github_actions_runner_group.go
```diff
@@ -92,15 +93,73 @@ func resourceGithubActionsRunnerGroup() *schema.Resource {
...
+				ValidateDiagFunc: validation.ToDiagFunc(validation.StringLenBetween(1, 255)),
+				Description:      "The identifier of the hosted compute network configuration to associate with this runner group for GitHub-hosted private networking.",
+			},
 		},
 	}
 }
 
-func resourceGithubActionsRunnerGroupCreate(d *schema.ResourceData, meta any) error {
-	err := checkOrganization(meta)
+func getOrganizationRunnerGroup(client *github.Client, ctx context.Context, org string, groupID int64) (*github.RunnerGroup, *github.Response, error) {
```
stevehipwell: This logic shouldn't be extracted as it make the code much harder to follow. In this case there is also the fact that the error handling may need to differ per call site.

github/resource_github_actions_runner_group.go
```diff
@@ -92,15 +93,73 @@ func resourceGithubActionsRunnerGroup() *schema.Resource {
...
+		var ghErr *github.ErrorResponse
+		if errors.As(err, &ghErr) && ghErr.Response != nil && ghErr.Response.StatusCode == http.StatusNotModified {
+			// ignore error StatusNotModified
+			return nil, resp, nil
+		}
+	}
+	return runnerGroup, resp, err
+}
+
+func setGithubActionsRunnerGroupState(d *schema.ResourceData, runnerGroup *github.RunnerGroup, etag string, selectedRepositoryIDs []int64) error {
```
stevehipwell: This logic shouldn't be extracted as it make the code much harder to follow.

github/resource_github_actions_runner_group.go
```diff
@@ -131,7 +190,11 @@ func resourceGithubActionsRunnerGroupCreate(d *schema.ResourceData, meta any) er
 		}
 	}
 
-	ctx := context.Background()
+	var networkConfigurationIDPtr *string
+	if networkConfigurationID, ok := d.GetOk("network_configuration_id"); ok {
+		value := networkConfigurationID.(string)
+		networkConfigurationIDPtr = &value
+	}
```
stevehipwell: Create the request object up front and set the field in the if block rather than create a separate variable for it.

github/resource_github_actions_runner_group.go
```diff
@@ -142,79 +205,36 @@ func resourceGithubActionsRunnerGroupCreate(d *schema.ResourceData, meta any) er
...
 	client := meta.(*Owner).v3client
 	orgName := meta.(*Owner).name
 
 	runnerGroupID, err := strconv.ParseInt(d.Id(), 10, 64)
 	if err != nil {
-		return err
+		return diag.FromErr(err)
 	}
-	ctx := context.WithValue(context.Background(), ctxId, d.Id())
+	ctx = context.WithValue(ctx, ctxId, d.Id())
```
stevehipwell:
```suggestion
```
This is uneccessary.

github/resource_github_actions_runner_group.go
```diff
@@ -223,55 +243,20 @@ func resourceGithubActionsRunnerGroupRead(d *schema.ResourceData, meta any) erro
 	if err != nil {
 		var ghErr *github.ErrorResponse
 		if errors.As(err, &ghErr) {
```
stevehipwell, line 241: This should be refactored to use `errors.AsType`.

github/resource_github_actions_runner_group.go
```diff
@@ -223,55 +243,20 @@ func resourceGithubActionsRunnerGroupRead(d *schema.ResourceData, meta any) erro
...
-			if ghErr.Response.StatusCode == http.StatusNotFound {
+			if ghErr.Response != nil && ghErr.Response.StatusCode == http.StatusNotFound {
```
stevehipwell: Why does this need changing?

github/resource_github_actions_runner_group_unit_test.go
```diff
@@ -0,0 +1,116 @@
+package github
```
stevehipwell: All tests should live in `<name>_test.go` not `<name>_unit_test.go`.

stevehipwell (changes requested):
There seems to be a lot of unnecessary and non-idiomatic changes to existing resources. It also looks like you've made changes to _docs/resources_ rather than _templates/resources_. Please can you split this PR up so the net new resources are in a separate PR to the changes for the existing resources. Please could you also make sure you've run the automation locally to catch issues such as the docs one.

austenstone:
Thanks for the patience on this one, and sorry for the long gap. I've rebuilt the PR against current `main` and worked through all the review feedback. Summary of what changed:

### Scope reduction

The biggest problem with the previous state was that it dragged both runner group resources through an unrelated context/diag migration. Both `resource_github_actions_runner_group.go` and `resource_github_enterprise_actions_runner_group.go` are now reverted to `main` with only the `network_configuration_id` additions layered on top. That drops the helper extraction, the `ctx = context.WithValue(ctx, ctxId, …)` churn, the `errors.As` → `errors.AsType` rewrite, and the `ghErr.Response != nil` guard changes — all of which you flagged and none of which this feature needs.

The `_unit_test.go` file is gone too. `normalizeEtag` went with it, since nothing called it after the revert.

### Review threads

| Thread | Resolution |
|---|---|
| `getOrganizationRunnerGroup` / `setGithubActionsRunnerGroupState` extraction | Reverted, untouched |
| `networkConfigurationIDPtr` dance | Request object is now built up front and the optional field is set inside the `if`, as you suggested |
| `ctx = context.WithValue(ctx, ctxId, …)` | Reverted. Also dropped from the new resources — `ctxId` is written in ~40 places and read in exactly zero, only `ctxEtag` is consumed by the transport |
| `errors.As` / `ghErr.Response != nil` | Reverted in the runner groups; the new resources use `errors.AsType` with no nil guard, matching the current idiom |
| `_unit_test.go` naming | File deleted |

### New resources

`github_organization_network_configuration` and `github_enterprise_network_configuration` are rewritten against go-github v89's native `Organizations.*NetworkConfiguration*` and `Enterprise.*EnterpriseNetworkConfiguration*` methods. No raw REST anywhere, which resolves @deiga's blocker.

They follow the pattern the organization role data sources use: `m any` asserted once into a typed `meta`, `checkOrganizationOK` for the org-scope guard, `CreateContext`/`ReadContext`/… returning `diag.Diagnostics`. `enterprise_slug` is `ForceNew`. Every attribute has a `Description` and bounded inputs have `ValidateDiagFunc` (name length/charset mirroring the API constraint, `compute_service` restricted to `none`/`actions` since go-github's own client-side validation rejects `codespaces`).

422s from a cross-scope network settings ID get annotated, since mixing up an org-scoped and an enterprise-scoped `GitHub.Network/networkSettings` GitHub ID is the easiest mistake to make here and the bare API error doesn't say so.

### `network_configuration_id` removal forces replacement

Worth calling out explicitly. `PATCH /orgs/{org}/actions/runner-groups/{id}` declares `network_configuration_id` as `type: string, nullable: true`, so an explicit `null` unsets it. But go-github's `UpdateRunnerGroupRequest.NetworkConfigurationID` is `*string` with `json:",omitempty"`, so a nil pointer omits the key entirely and there's no way to send explicit `null`. Rather than leave a perpetual diff, removing the attribute now forces replacement via `customdiff.ForceNewIfChange`. Changing to a different non-empty ID is still an in-place update. Happy to swap this for an upstream go-github fix if you'd rather wait for one.

### Docs

I hadn't realised `docs/` is generated — the earlier revision hand-edited those pages, which `make checkdocs` would have blown away. There are now `templates/resources/*.md.tmpl` and `examples/resources/*/example_1.tf` for both new resources, plus `network_configuration_id` documented on both runner groups and a note on the hosted runner page clarifying that private networking is a runner-group concern (`public_ip_enabled` is unrelated). The `examples/hosted_runner` config also used `enable_static_ip`, which is the API payload key, not the schema attribute — fixed to `public_ip_enabled`.

### Verification

`make lintcheck`, `make lintcheck-new`, `make test`, `make checkdocs`, `make validatedocs`, and `rumdl check ./docs` all pass locally. `go.sum` had picked up entries `go mod tidy` doesn't want; that's cleaned up too. CI runs are sitting in `action_required` — a maintainer approval would be appreciated.

Acceptance tests are gated behind `GITHUB_TEST_NETWORK_SETTINGS_ID` / `GITHUB_TEST_ENTERPRISE_NETWORK_SETTINGS_ID` since they need a real Azure vnet registered against the org/enterprise.

@stevehipwell @deiga — would appreciate another look when you have a moment.

github/resource_github_actions_runner_group.go
```diff
@@ -383,3 +402,14 @@ func resourceGithubActionsRunnerGroupDelete(d *schema.ResourceData, m any) error
 	_, err = client.Actions.DeleteOrganizationRunnerGroup(ctx, orgName, runnerGroupID)
 	return err
 }
+
+// networkConfigurationRemoved reports whether a hosted compute network configuration was
+// removed from a runner group. The REST API only accepts an explicit null to clear the
+// assignment, which the client library's request type cannot express, so the runner group is
+// replaced instead of leaving the practitioner with a perpetual diff.
```
deiga, line 409: question: Please elaborate on this. 
The API docs don't contain any mention of this and it sounds like an `go-github` ~bug that the field is `omitempty`. Could you open a PR there to modify the struct?

github/resource_github_actions_runner_group.go
```diff
@@ -170,6 +179,9 @@ func resourceGithubActionsRunnerGroupCreate(d *schema.ResourceData, m any) error
 	if err = d.Set("name", runnerGroup.GetName()); err != nil {
 		return err
 	}
+	if err = d.Set("network_configuration_id", runnerGroup.GetNetworkConfigurationID()); err != nil {
+		return err
+	}
```
deiga, line 184: question: Since we're setting the value here, should the field be marked as `Computed: true`?

github/resource_github_actions_runner_group.go
```diff
@@ -383,3 +402,14 @@ func resourceGithubActionsRunnerGroupDelete(d *schema.ResourceData, m any) error
...
+// networkConfigurationRemoved reports whether a hosted compute network configuration was
+// removed from a runner group. The REST API only accepts an explicit null to clear the
+// assignment, which the client library's request type cannot express, so the runner group is
+// replaced instead of leaving the practitioner with a perpetual diff.
+func networkConfigurationRemoved(_ context.Context, oldValue, newValue, _ any) bool {
+	oldID, _ := oldValue.(string)
+	newID, _ := newValue.(string)
+
+	return oldID != "" && newID == ""
+}
```
deiga, line 415: issue: Since this function is shared, it needs to be moved to a util file (new or existing)

github/resource_github_enterprise_actions_runner_group_test.go
```diff
@@ -203,4 +208,142 @@ func TestAccGithubActionsEnterpriseRunnerGroup(t *testing.T) {
...
+		groupName := fmt.Sprintf("tf-acc-test-%s", randomID)
+		networkConfigurationName := fmt.Sprintf("%senterprise-network-config-%s", testResourcePrefix, randomID)
+
+		configWithoutNetworking := fmt.Sprintf(`
+			resource "github_enterprise_network_configuration" "test" {
+			  enterprise_slug      = %q
+			  name                 = %q
+			  compute_service      = "actions"
+			  network_settings_ids = [%q]
+			}
```
deiga, line 226: Please create a `mustCreateTestEnterpriseNetworkConfiguration` function and use that for test setup

github/resource_github_enterprise_actions_runner_group_test.go
```diff
@@ -203,4 +208,142 @@ func TestAccGithubActionsEnterpriseRunnerGroup(t *testing.T) {
...
+
+		resource.Test(t, resource.TestCase{
+			PreCheck:          func() { skipUnlessEnterprise(t) },
+			ProviderFactories: providerFactories,
+			Steps: []resource.TestStep{
+				{
+					Config: configWithoutNetworking,
+					ConfigStateChecks: []statecheck.StateCheck{
+						statecheck.ExpectKnownValue(resourceName, tfjsonpath.New("name"), knownvalue.StringExact(groupName)),
+						statecheck.ExpectKnownValue(resourceName, tfjsonpath.New("visibility"), knownvalue.StringExact("all")),
```
deiga, line 259: Please only check Computed values in tests, configured values are verified automatically

deiga (changes requested):
Partial review

deiga:
Please also note that some of the resource you modify here are being changed in [this PR](https://github.com/integrations/terraform-provider-github/pull/3521). It's recommended to use that as a base

templates/resources/enterprise_network_configuration.md.tmpl
```diff
@@ -0,0 +1,37 @@
+---
```
stevehipwell: Please use the default template as a base and make sure that you're using the automation instead of manually writing the docs.

templates/resources/actions_hosted_runner.md.tmpl
```diff
@@ -34,6 +34,8 @@ The following arguments are supported:
 - `public_ip_enabled` - (Optional) Whether to enable static public IP for the runner. Note there are account limits. To list limits, use the GitHub API: `GET /orgs/{org}/actions/hosted-runners/limits`. Defaults to false.
 - `image_version` - (Optional) The version of the runner image to deploy. This is only relevant for runners using custom images.
 
+~> **Note:** Private networking is configured on the runner group rather than on the hosted runner. Assign a `github_organization_network_configuration` to a runner group through `github_actions_runner_group.network_configuration_id`, then place the hosted runner in that group.
```
stevehipwell:
```suggestion
~> Private networking is configured on the runner group rather than on the hosted runner. Assign a `github_organization_network_configuration` to a runner group through `github_actions_runner_group.network_configuration_id`, then place the hosted runner in that group.
```

github/resource_github_organization_network_configuration.go
```diff
@@ -0,0 +1,248 @@
...
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/validation"
+)
+
+// networkConfigurationNamePattern mirrors the name constraint the REST API enforces on hosted
+// compute network configurations so that invalid names are rejected at plan time.
+var networkConfigurationNamePattern = regexp.MustCompile(`^[a-zA-Z0-9._-]+$`)
+
+func resourceGithubOrganizationNetworkConfiguration() *schema.Resource {
+	return &schema.Resource{
```
stevehipwell, line 16: Please follow the schema organisation as described in the contributing documentation.

github/resource_github_organization_network_configuration.go
```diff
@@ -0,0 +1,248 @@
...
+		Description:   "Manages a hosted compute network configuration for a GitHub organization.",
```
stevehipwell:
```suggestion
		Description:   "Resource to manage a hosted compute network configuration for a GitHub organization.",
```

stevehipwell (changes requested):

GrantBirki:
bump @austenstone we would love to see this land! ❤️

github/resource_github_actions_runner_group.go
austenstone: It didn't need changing for networking. Removed that nil-guard change.

github/resource_github_actions_runner_group.go
austenstone: Using errors.AsType now, matching main.

github/resource_github_actions_runner_group_unit_test.go
austenstone: Removed the separate _unit_test.go file. Runner-group tests live in the existing _test.go files.

github/resource_github_actions_runner_group.go
austenstone: Dropped the new shared error wrapper. Error handling stays at the existing resource call sites.

github/resource_github_actions_runner_group.go
austenstone: Removed the runner-group state helper extraction. The state assignments stay inline.

github/resource_github_actions_runner_group.go
austenstone: Dropped the context-callback migration from this PR. The existing callback and context setup are unchanged.

github/resource_github_actions_runner_group.go
austenstone: The request is now built before the conditional, which sets NetworkConfigurationID directly on it.

github/resource_github_actions_runner_group.go
austenstone: Kept this Optional without Computed intentionally. Reading the configured value back is also needed for refresh/import; it doesn't make the field server-controlled. Optional+Computed would retain the previous assignment when the attribute is omitted, rather than treating omission as a request to remove it. The separate SDK null-serialization thread is still open.

github/resource_github_enterprise_actions_runner_group_test.go
austenstone: Added mustCreateTestEnterpriseNetworkConfiguration and the organization equivalent. Runner-group tests now use native API fixtures with cleanup rather than managing the prerequisite configuration in their Terraform configs.

github/resource_github_organization_network_configuration.go
austenstone: Matched the Resource Structure layout in ARCHITECTURE.md: CRUD callbacks and importer first, then Description and Schema. Shared network-configuration helpers now live in a utility file.

github/resource_github_actions_runner_group.go
```diff
@@ -135,18 +142,20 @@ func resourceGithubActionsRunnerGroupCreate(d *schema.ResourceData, m any) error
...
+		Visibility:               &visibility,
+		RestrictedToWorkflows:    &restrictedToWorkflows,
+		SelectedRepositoryIDs:    selectedRepositoryIDs,
+		SelectedWorkflows:        selectedWorkflows,
+		AllowsPublicRepositories: &allowsPublicRepositories,
+	}
+	if networkConfigurationID, ok := d.GetOk("network_configuration_id"); ok {
+		networkConfigurationIDValue, _ := networkConfigurationID.(string)
+		createOptions.NetworkConfigurationID = new(networkConfigurationIDValue)
+	}
```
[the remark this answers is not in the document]
austenstone: Added matching organization acceptance flows for create-time assignment, adding networking, import, and removal, plus offline POST/PATCH coverage. Live acceptance still needs a configured network-settings fixture. In-place removal remains blocked by the separate SDK thread.

templates/resources/actions_hosted_runner.md.tmpl
austenstone: Applied the suggested wording in the template and regenerated the page.

github/resource_github_actions_runner_group.go
```diff
@@ -135,18 +142,20 @@ func resourceGithubActionsRunnerGroupCreate(d *schema.ResourceData, m any) error
...
+
+	runnerGroup, resp, err := client.Actions.CreateOrganizationRunnerGroup(ctx, orgName, createOptions)
```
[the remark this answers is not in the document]
austenstone: The PR description was wrong. The organization POST contract includes network_configuration_id, so the native create request is sufficient. Corrected the description and added POST payload coverage.

[Official request schema](https://github.com/github/rest-api-description/blob/0a144e030abbe358000f0d32874e197b59a8f7ed/descriptions/api.github.com/api.github.com.json#L15784-L15791).

github/resource_github_enterprise_actions_runner_group_test.go
austenstone: Removed the redundant configured-value assertions from the networking acceptance steps. They now compare the computed group ID across updates; import verification stays in place.

github/resource_github_enterprise_actions_runner_group.go
```diff
@@ -125,18 +132,20 @@ func resourceGithubActionsEnterpriseRunnerGroupCreate(d *schema.ResourceData, me
...
+		AllowsPublicRepositories: &allowsPublicRepositories,
+		RestrictedToWorkflows:    &restrictedToWorkflows,
+		SelectedWorkflows:        selectedWorkflows,
+	}
+	if networkConfigurationID, ok := d.GetOk("network_configuration_id"); ok {
+		networkConfigurationIDValue, _ := networkConfigurationID.(string)
+		createOptions.NetworkConfigurationID = new(networkConfigurationIDValue)
+	}
+
+	enterpriseRunnerGroup, resp, err := client.Enterprise.CreateEnterpriseRunnerGroup(ctx, enterpriseSlug, createOptions)
```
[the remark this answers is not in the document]
austenstone: Corrected the PR description. Enterprise POST also accepts network_configuration_id; there is no need for a networking-only PATCH after create. Added native request coverage.

[Official enterprise request schema](https://github.com/github/rest-api-description/blob/0a144e030abbe358000f0d32874e197b59a8f7ed/descriptions/ghec/ghec.json#L7999-L8006).

github/resource_github_actions_runner_group.go
```diff
@@ -260,6 +272,9 @@ func resourceGithubActionsRunnerGroupRead(d *schema.ResourceData, m any) error {
 	if err = d.Set("name", runnerGroup.GetName()); err != nil {
 		return err
 	}
+	if err = d.Set("network_configuration_id", runnerGroup.GetNetworkConfigurationID()); err != nil {
+		return err
+	}
```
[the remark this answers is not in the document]
austenstone: Fixed both organization and enterprise helpers to suppress only 304. Regression cases cover 304 retaining state, 404 removing it, and 403/500 propagating errors rather than silently keeping stale state.

templates/resources/enterprise_network_configuration.md.tmpl
austenstone: Removed both custom network-configuration templates. The docs now use the default template, schema descriptions, and discovered usage/import examples, with no handwritten attribute lists.

github/resource_github_actions_runner_group.go
austenstone: Moved the shared diff function into util_diff.go. The SDK limitation remains tracked separately.

github/resource_github_organization_network_configuration.go
austenstone: Used the suggested description, with the equivalent wording for the enterprise resource.

austenstone:
Can we keep the configuration resources and runner-group attachment together here? They cover one end-to-end workflow; the unrelated runner-group refactors are gone.

The non-SDK inline feedback is addressed. [CI](https://github.com/integrations/terraform-provider-github/actions/runs/34536217871) and [CodeQL](https://github.com/integrations/terraform-provider-github/actions/runs/34536217909) need maintainer approval. Native detachment remains blocked on [google/go-github#4541](https://github.com/google/go-github/pull/4541) and its release; replacement on removal is not the intended merge-ready behavior.

github/resource_github_actions_runner_group.go
deiga: https://developer.hashicorp.com/terraform/plugin/sdkv2/resources/data-consistency-errors#finding-data-consistency-errors

> Resources should never set or change an attribute value without the schema Computed flag.

examples/resources/github_enterprise_network_configuration/import.sh
```diff
@@ -0,0 +1,4 @@
+set -euo pipefail
```
deiga, line 1:
```suggestion
```

examples/resources/github_organization_network_configuration/import.sh
```diff
@@ -0,0 +1,4 @@
+set -euo pipefail
+
```
deiga, line 2:
```suggestion
```

github/resource_github_actions_runner_group.go
deiga: This is still undone

github/resource_github_actions_runner_group.go
```diff
@@ -260,6 +273,9 @@ func resourceGithubActionsRunnerGroupRead(d *schema.ResourceData, m any) error {
...
```
deiga, line 278: issue: this needs to be Computed

github/resource_github_actions_runner_group_test.go
```diff
@@ -2,16 +2,157 @@ package github
...
+				{
+					Config: configWithoutNetworking,
+					ConfigStateChecks: []statecheck.StateCheck{
+						sameID.AddStateValue(resourceName, tfjsonpath.New("id")),
+					},
+				},
+				{
+					Config: configWithNetworking,
+					ConfigStateChecks: []statecheck.StateCheck{
+						sameID.AddStateValue(resourceName, tfjsonpath.New("id")),
```
deiga, line 108: question: what's the purpose of this check?

deiga (changes requested):

github/resource_github_actions_runner_group.go
austenstone: You're right. The migration was dropped during later scope cleanup, so the old "done" reply was stale. Restored context-aware CRUD for both runner-group resources in a7924177, with cancellation checks for all four operations.

examples/resources/github_organization_network_configuration/import.sh
austenstone: Removed in 37d3e902, including the generated docs.

github/resource_github_actions_runner_group_test.go
austenstone: It verifies that assigning networking updates the existing group rather than replacing it. The comparer is now named `groupIDUnchanged`; import coverage is a separate scenario in a7924177.

examples/resources/github_enterprise_network_configuration/import.sh
austenstone: Removed in 37d3e902, including the generated docs.

github/resource_github_actions_runner_group.go
austenstone: Fixed in a7924177. The field is Optional+Computed in both scopes.

github/resource_github_actions_runner_group.go
austenstone: You're right. Both resources now use Optional+Computed in a7924177. Omission or null preserves the assignment. Explicit empty-string clearing still replaces the group until the SDK detachment fix is available; that remains a merge blocker.

github/acc_helpers_network_configuration_test.go
```diff
@@ -5,21 +5,32 @@ import (
...
+	"github.com/hashicorp/terraform-plugin-testing/compare"
 	"github.com/hashicorp/terraform-plugin-testing/helper/acctest"
+	"github.com/hashicorp/terraform-plugin-testing/helper/resource"
+	"github.com/hashicorp/terraform-plugin-testing/knownvalue"
+	"github.com/hashicorp/terraform-plugin-testing/statecheck"
+	"github.com/hashicorp/terraform-plugin-testing/tfjsonpath"
 )
 
-func testRunnerGroupNetworking(t *testing.T, runnerGroup *schema.Resource, scope string, attributes map[string]any, create, update func(*schema.ResourceData, any) error) {
+func testRunnerGroupNetworking(t *testing.T, runnerGroup *schema.Resource, scope string, attributes map[string]any) {
```
deiga: If this is not from auto-formatting, please keep all args on the same line

deiga: issue: please use a descriptive name for the test or use a sub-test

github/resource_github_actions_runner_group.go
austenstone: Opened and implemented [google/go-github#4541](https://github.com/google/go-github/pull/4541). A nil `*string` omits the field with either `omitempty` or `omitzero`; a pointer to `""` sends an empty string, not JSON `null`. Both native update requests now have a Go-only `RemoveNetworkConfiguration` flag and value-receiver marshaler, following `UpdateTeamRequest.RemoveParentTeam`.

That addresses the request to investigate the encoding and open the SDK PR. The provider still needs the fix merged, released, and adopted before clearing can be non-destructive. Keeping that merge blocker in this PR's description rather than leaving this question unanswered.

austenstone:
All 49 review threads are now resolved, and the branch is conflict-free against main. Local unit/race tests, lint, build, and docs checks pass on the current head.

@deiga @stevehipwell could you approve the current [CI](https://github.com/integrations/terraform-provider-github/actions/runs/35390042939) and [CodeQL](https://github.com/integrations/terraform-provider-github/actions/runs/35390042919) runs, and confirm whether the configuration resources and runner-group integration can stay together here? The actual acceptance job was skipped; the successful wrapper is not an acceptance pass.

The remaining code dependency is [google/go-github#4541](https://github.com/google/go-github/pull/4541). Its explicit-null fix and latest requested wording are pushed, and CLA passes. Once it merges and ships, this PR needs to adopt it and replace group recreation with native in-place detachment. Keeping this draft until that behavior is implemented and verified.

austenstone:
Hey I addressed the review feedback. Ready for CI

PR #3277, opened 2026-03-15
fix: add plan-time validation to key_prefix in autolink reference resource (#3176)

### LINKED ISSUE #3176: [BUG]: autolink allows numbers during plan, but fails during apply
### Expected Behavior

Plan should have failed with bad key_prefix, but passed

### Actual Behavior

Plan echoed successful diff
```diff
+ resource "github_repository_autolink_reference" "all" {
+       etag                = (known after apply)
+       id                  = (known after apply)
+       is_alphanumeric     = false
+       key_prefix          = "PTFY25"
+       repository          = "apiv2"
+       target_url_template = "https://***.atlassian.net/browse/PTFY25-<num>"
```

But during apply

```
Error: POST https://api.github.com/repos/****/apiv2/autolinks: 422 Validation Failed [{Resource:KeyLink Field:key_prefix Code:custom Message:key_prefix must not end with a number} {Resource:KeyLink Field:key_prefix Code:custom Message:key_prefix must only contain letters, numbers, or .-_+=:/#}]
│ 
│   with github_repository_autolink_reference.all["apiv2_PTFY25"],
│   on repository.tf line 189, in resource "github_repository_autolink_reference" "all":
│  189: resource "github_repository_autolink_reference" "all" {
```

### Terraform Version

terraform ">= 1.9.8, < 1.10"
integration "v6.6.0"

### Affected Resource(s)

- github_repository_autolink_reference

```hcl

```

### Steps to Reproduce

_No response_

### Debug Output

```shell

```

### Panic Output

```shell

```

### Code of Conduct

LeC-D:
## Summary

Fixes #3176

## Problem

`terraform plan` accepts any value for `key_prefix` in `github_repository_autolink_reference`, but `terraform apply` can fail with a 422 from the GitHub API if the value is invalid:

```
key_prefix must not end with a number
key_prefix must only contain letters, numbers, or .-_+=:/#
```

This creates a confusing plan/apply inconsistency where the plan shows a clean diff but apply fails.

## Fix

Added `ValidateDiagFunc` to the `key_prefix` schema field using a regexp that mirrors GitHub's API constraints:

- Allowed characters: letters, digits, and `.-_+=:/#`
- Must **not** end with a digit

```go
ValidateDiagFunc: validation.ToDiagFunc(validation.StringMatch(
    regexp.MustCompile(`^[a-zA-Z0-9.=+:/#_-]*[a-zA-Z.=+:/#_-]$`),
    "must only contain letters, numbers, or .-_+=:/# and must not end with a number",
)),
```

Both `regexp` and `validation` were already imported in this file (used by `target_url_template`). No new dependencies needed.

## Testing

Existing acceptance tests use `key_prefix` values like `TEST1-`, `TEST2-`, etc. which all pass the new validation. Invalid values like `PTFY25` (ends with digit) will now be caught at plan time.

### DIFF
--- a/github/resource_github_repository_autolink_reference.go
+++ b/github/resource_github_repository_autolink_reference.go
@@ -69,6 +69,10 @@ func resourceGithubRepositoryAutolinkReference() *schema.Resource {
 				Required:    true,
 				ForceNew:    true,
 				Description: "This prefix appended by a number will generate a link any time it is found in an issue, pull request, or commit",
+				ValidateDiagFunc: validation.ToDiagFunc(validation.StringMatch(
+					regexp.MustCompile(`^[a-zA-Z0-9.=+:/#_-]*[a-zA-Z.=+:/#_-]$`),
+					"must only contain letters, numbers, or .-_+=:/# and must not end with a number",
+				)),
 			},
 			"target_url_template": {
 				Type:             schema.TypeString,
--- a/github/resource_github_repository_autolink_reference_test.go
+++ b/github/resource_github_repository_autolink_reference_test.go
@@ -296,4 +296,24 @@ func TestAccGithubRepositoryAutolinkReference(t *testing.T) {
 			},
 		})
 	})
+
+	t.Run("rejects key_prefix ending with a digit at plan time", func(t *testing.T) {
+		config := `
+			resource "github_repository_autolink_reference" "autolink_invalid" {
+				repository          = "some-repo"
+				key_prefix          = "PTFY25"
+				target_url_template = "https://example.com/<num>"
+			}
+		`
+
+		resource.UnitTest(t, resource.TestCase{
+			ProviderFactories: providerFactories,
+			Steps: []resource.TestStep{
+				{
+					Config:      config,
+					ExpectError: regexp.MustCompile(`must not end with a number`),
+				},
+			},
+		})
+	})
 }


austenstone (commented):
This validation change looks correct and matches the API behavior described in #3176. One thing I’d suggest before merging: can we add a regression test for an invalid `key_prefix` (for example one that ends with a digit, like `PTFY25`) and assert that it fails at plan time? Right now the tests only cover valid prefixes, so we’re not actually locking in the behavior this PR is fixing.

github/resource_github_repository_autolink_reference.go
austenstone, line 1: This validation looks right. I think the missing piece is a regression test that proves invalid prefixes now fail during `terraform plan`, since that’s the behavior this PR is fixing.

Because this PR only changes this file, I can’t attach a native GitHub suggestion directly to `github/resource_github_repository_autolink_reference_test.go`, but something along these lines would lock it in:

```go
t.Run("rejects invalid key_prefix at plan time", func(t *testing.T) {
	repoName := fmt.Sprintf("%srepo-autolink-%s", testResourcePrefix, randomID)
	config := fmt.Sprintf(`
			name        = "%s"
			description = "Test autolink validation"
		}

		resource "github_repository_autolink_reference" "invalid" {
			repository = github_repository.test.name

			key_prefix          = "PTFY25"
			target_url_template = "https://example.com/PTFY-<num>"
		}
	`, repoName)

		PreCheck:          func() { skipUnauthenticated(t) },
		Steps: []resource.TestStep{
			{
				Config:      config,
				ExpectError: regexp.MustCompile(`must only contain letters, numbers, or .* must not end with a number`),
			},
		},
	})
})
```

That would make sure we don’t regress back to the current plan/apply mismatch later.

austenstone (commented):
Left a concrete regression-test suggestion so the fix is covered at plan time as well as in the schema.

LeC-D:
Good call! Added a `resource.UnitTest` sub-test in `resource_github_repository_autolink_reference_test.go` that passes `key_prefix = "PTFY25"` (ends with a digit) and asserts a plan-time error matching `must not end with a number`. This locks in the validation behavior without requiring a real GitHub account.

LeC-D:
Rebased on upstream `main` to bring the branch up to date.

github/resource_github_repository_autolink_reference.go
```diff
@@ -69,6 +69,10 @@ func resourceGithubRepositoryAutolinkReference() *schema.Resource {
 				Required:    true,
 				ForceNew:    true,
 				Description: "This prefix appended by a number will generate a link any time it is found in an issue, pull request, or commit",
+				ValidateDiagFunc: validation.ToDiagFunc(validation.StringMatch(
+					regexp.MustCompile(`^[a-zA-Z0-9.=+:/#_-]*[a-zA-Z.=+:/#_-]$`),
+					"must only contain letters, numbers, or .-_+=:/# and must not end with a number",
+				)),
```
deiga: suggestion: use 2 validations, one for suffix and one for allowed chars.
And the must not end with could be simplified to `\D$` I think

deiga (changes requested):

LeC-D:
Applied @deiga's review feedback: split the `key_prefix` validation into two separate `StringMatch` validators inside `validation.All()` — one for allowed characters (`^[a-zA-Z0-9.=+:/#_-]+$`) and one for the digit-suffix check (`\D$`). Error messages are now more targeted. Thanks for the suggestion!

LeC-D:
Rebased on upstream `main` (2026-03-24) to bring the branch back up to date.

github/resource_github_repository_autolink_reference.go
LeC-D: Done — split into two `StringMatch` validators inside `validation.All()`: one for allowed characters (`^[a-zA-Z0-9.=+:/#_-]+$`) and one for the digit-suffix check using `\D$` as you suggested. Error messages are now separate and targeted.

github/resource_github_repository_autolink_reference.go
LeC-D: Good catch! Added a `resource.UnitTest` sub-test that passes `key_prefix = "PTFY25"` (ends with a digit) and asserts a plan-time error matching `must not end with a number`. The test uses `providerFactories` and doesn't require a real GitHub account.

LeC-D:
Rebased on upstream `main` (2026-03-26) to bring the branch back up to date.

LeC-D:
@deiga The CHANGES_REQUESTED feedback has been addressed in the latest commit (b1015f6) — the `key_prefix` validation is now split into two separate `StringMatch` validators inside `validation.All()` as suggested. Would appreciate a re-review when you get a chance!

LeC-D:
Rebased on upstream `main` (2026-03-28) to bring the branch back up to date.

deiga (dismissed):

LeC-D:
Rebased on upstream `main` (2026-03-28) to bring the branch back up to date.

deiga:
@LeC-D Sorry that this has taken long to get to. Would you be available to rebase this (possibly multiple times) during the next week? We're trying to get 6.13.0 out and will try to merge as many ~ready PRs as is reasonable

github/resource_github_repository_autolink_reference.go
```diff
@@ -69,6 +69,16 @@ func resourceGithubRepositoryAutolinkReference() *schema.Resource {
...
+				ValidateDiagFunc: validation.ToDiagFunc(validation.All(
+					validation.StringMatch(
+						regexp.MustCompile(`^[a-zA-Z0-9.=+:/#_-]+$`),
+						"must only contain letters, numbers, or the characters .-_+=:/#",
+					),
+					validation.StringMatch(
+						regexp.MustCompile(`\D$`),
+						"must not end with a number",
+					),
+				)),
```
stevehipwell:
```suggestion
				ValidateDiagFunc: validation.ToDiagFunc(validation.StringMatch(
						regexp.MustCompile(`^[a-zA-Z0-9.=+:/#_-]+\D$`),
						"must only contain letters, numbers, or the characters .-_+=:/# and not end with a number",
					)),
```

There is no reason to use two validators for this. Also could you confirm the logic for the start of the prefix can actually include all or the allowed characters?

stevehipwell (changes requested):
@LeC-D please could you rebase this PR?

stevehipwell:
@LeC-D are you in a position to continue working on this?

github/resource_github_repository_autolink_reference.go
deiga: 😂 I think I specifically told them to split into two validators to make the regex easier to maintain and to make the validation error clearer for the user 😬

LeC-D:
Rebased on upstream `main` (2026-06-26) to bring the branch back up to date. Ready when you are!

github/resource_github_repository_autolink_reference.go
LeC-D: Thank you for the suggestion! The split into two separate `StringMatch` validators was intentionally done based on @deiga's earlier request (2026-03-18) to make the error messages clearer for the user — one message for invalid characters and another for the digit-suffix rule.

Since there's now a difference of opinion between reviewers on this, I'm happy to follow whichever approach the maintainers prefer. @deiga @stevehipwell — could you align on which direction you'd like to go? I can switch to the single combined regex or keep the split — just let me know.

stevehipwell: Please use a single regex validator, a single validation error expressing the logic as a whole will be clearer than having two different messages. It also makes the code significantly easier to read.

LeC-D: Done — switched to a single `StringMatch` validator with one combined error message:

```go
ValidateDiagFunc: validation.ToDiagFunc(validation.StringMatch(
    regexp.MustCompile(`^[a-zA-Z0-9.=+:/#_-]*[a-zA-Z.=+:/#_-]$`),
    "must only contain letters, numbers, or .-_+=:/# and must not end with a number",
)),
```

One nuance: I used `[a-zA-Z.=+:/#_-]$` for the final character rather than `\D$`. A bare `\D` matches *any* non-digit, so `...+\D$` would let a character outside the allowed set slip into the last position (e.g. `abc@` would pass). Constraining the final char to an allowed non-digit keeps the character allow-list intact end-to-end while enforcing the no-trailing-digit rule in a single validator.

To your question: yes — the leading `[a-zA-Z0-9.=+:/#_-]*` permits every allowed character (digits included) at the start and middle of the prefix; only the final character is restricted to a non-digit.

Also rebased onto the latest `main`. The plan-time regression test (`PTFY25`) still passes.

stevehipwell (approved):
LGTM

deiga (approved):

robert-crandall (approved):
Thanks for tackling this - moving the failure from apply-time 422 to plan-time is a really nice UX win. The regex matches GitHub's documented constraints exactly, and I like that you covered it with a fast unit test using `ExpectError`. Looks good to me!

PR #3222, opened 2026-02-24
fix: Correct logic for github_actions_organization_workflow_permissions

stevehipwell:
Resolves #ISSUE_NUMBER

----

### Before the change?

- The `github_actions_organization_workflow_permissions` resource can't set `can_approve_pull_request_reviews` to `false`

### After the change?

- The `github_actions_organization_workflow_permissions` resource works correctly

### Pull request checklist

- [x] Schema migrations have been created if needed ([example](https://github.com/F-Secure-web/terraform-provider-github/blob/main/github/migrate_github_actions_organization_secret.go))
- [x] Tests for the changes have been added (for bug fixes / features)
- [x] Docs have been reviewed and added / updated if needed (for bug fixes / features)

### Does this introduce a breaking change?

Please see our docs on [breaking changes](https://github.com/octokit/.github/blob/master/community/breaking_changes.md) to help!

- [ ] Yes
- [x] No

----

PR #3281, opened 2026-03-16
fix: polish github_actions_organization_workflow_permissions for merge readiness

austenstone:
## Summary

This PR carries forward the `github_actions_organization_workflow_permissions` fix from #3222 and applies the additional cleanup needed to make it fit the existing provider patterns more cleanly.

## What changed

- preserves the boolean fix for `can_approve_pull_request_reviews = false`
- restores the conflict-specific enterprise restriction error handling
- uses `id` as the canonical lookup key in read/update/delete
- switches import back to passthrough
- does read-after-write on create/update so state is authoritative

## Why

PR #3222 fixes the core bug, but it also introduced a few shape changes that made the resource less consistent with adjacent resources in this provider and dropped the clearer enterprise-conflict error path.

This branch keeps the important behavior fix while tightening the implementation so it is easier to merge as-is.

## Validation

- `go test ./github -run TestAccGithubActionsOrganizationWorkflowPermissions -count=1`

## Related

- Carries forward / follows up on #3222

### DIFF
--- a/github/resource_github_actions_organization_workflow_permissions.go
+++ b/github/resource_github_actions_organization_workflow_permissions.go
@@ -2,6 +2,11 @@ package github
 
 import (
 	"context"
+	"encoding/json"
+	"errors"
+	"fmt"
+	"io"
+	"net/http"
 
 	"github.com/google/go-github/v83/github"
 	"github.com/hashicorp/terraform-plugin-log/tflog"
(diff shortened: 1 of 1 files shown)


austenstone:
Fair call — that’s on me. I was batch-reviewing and didn’t dig into this one enough before opening the follow-up. I’m closing this out and will keep any input on #3222 scoped to the bool fix.

PR #3298, opened 2026-03-23
docs: mark private attribute as deprecated in github_repository

### LINKED ISSUE #3297: [DOCS]: Update `github_repository` docs to reflect implementation
> In version 6.8.3, I get the following message:
> 
> │ Warning: Deprecated attribute
> │
> │   on repository_settings.tf line 18, in resource "github_repository" "repository_settings":
> │   18:       private,
> │
> │ The attribute "private" is deprecated. Refer to the provider documentation
> │ for details.
> 
> However, this deprecation is not documented by the provider documentation:
> 
> https://registry.terraform.io/providers/integrations/github/6.8.3/docs/resources/repository
>  

 _Originally posted by @drmcclelland in [#2569](https://github.com/integrations/terraform-provider-github/issues/2569#issuecomment-3564530042)_

mvanhorn:
## Summary

Migrates `github_repository` to the provider's generated-docs pipeline, and marks the `private` attribute as deprecated in the process.

## Why this matters

The `private` attribute triggers a Terraform deprecation warning at plan time ("use visibility instead"), but the published docs page never mentioned it. The Go schema has carried `Deprecated: "use visibility instead"` on that field for a while with nothing reflecting it in the docs.

Fixing that by hand-editing `website/docs/r/repository.html.markdown` would have kept the resource on the legacy hand-maintained page, where the schema and the docs drift apart exactly like this. So the fix is the migration: the page is now generated from the schema, and the deprecation shows up because the schema says so.

## Changes

- `templates/resources/repository.md.tmpl` — template for the generated page
- `docs/resources/repository.md` — regenerated output
- `examples/resources/github_repository/*.tf`, `import.sh` — examples extracted so the template can include them
- `github/resource_github_repository.go` — schema descriptions, including the `private` deprecation. Description-only: no schema shape or state behaviour changes.

## Testing

Regenerated the docs locally and confirmed the output matches what is committed.

Fixes #3297

This contribution was developed with AI assistance (Claude Code).


deiga (dismissed):

mvanhorn:
@deiga - friendly ping. Approved a while back, no merge activity - let me know if anything else is needed.

deiga:
@mvanhorn your branch has conflicts. Please make sure that you're branch is directly mergeable before requesting work from maintainers.

mvanhorn:
Apologies for the noise. Rebased on main, resolved the conflict, pushed as 1beca719. Should be directly mergeable now.

deiga (dismissed):

mvanhorn:
Thanks @deiga — looks like main has moved since you checked. The branch is showing mergeable now (last commit 1beca71 from 2026-03-23). Happy to rebase again if anything has shifted on your end before review.

deiga:
@mvanhorn Please stop pinging people. We will get to your PR when there is capacity for it. 
We do see activity notifications without pings.

We really appreciate that you are contributing and value that. At the same time it's important to remember that we are hobby maintainers. It is our free time we use here.

mvanhorn:
Understood, sorry for the noise. I'll leave this one alone - no more pings. Appreciate the maintenance work, this is my last comment on it.

stevehipwell (changes requested):
Thanks for the PR @mvanhorn. We've actually update how the provider docs are generated, so there are currently two (different) ways to rebase and make this change.

The first way, which would be our preference, is to update the schema descriptions to match the documentation, update the [examples](https://github.com/integrations/terraform-provider-github/tree/main/examples/resources/repository), and update the [template](https://github.com/integrations/terraform-provider-github/blob/main/templates/resources/repository.md.tmpl) to match the [generic resource template](https://github.com/integrations/terraform-provider-github/blob/main/templates/resources.md.tmpl) pattern.

The second easier way is to just update the [examples](https://github.com/integrations/terraform-provider-github/tree/main/examples/resources/repository) and [template](https://github.com/integrations/terraform-provider-github/blob/main/templates/resources/repository.md.tmpl) with the desired content.

mvanhorn:
Went with the preferred way. The branch is rebuilt on current main: argument descriptions from the old hand-written template are now schema Description fields (private documents the deprecation and points at visibility), the resource-level Description carries the GitHub App authentication note, templates/resources/repository.md.tmpl now matches the generic resource template pattern, and I added examples/resources/repository/import.sh so the generated Import section renders. docs/resources/repository.md is regenerated via tfplugindocs. go build and gofmt are clean; tfplugindocs validate still reports the pre-existing github_users gap on main, unrelated to this change.

mvanhorn:
5e72f53 restores both sections. The template now points `tffile` and `codefile` at the actual file names in the repo (`example_1.tf`, `example_2.tf`, `import.sh`) in an else branch, since `.HasExamples` and `.HasImport*` look for names this provider does not use.

On the regression you flagged: `docs/resources/repository.md` here is generated output, not hand-edited. I ran `tfplugindocs generate` twice and got byte-identical results, and no other doc page drifts, so `make checkdocs` stays green.

Still worth a maintainer call: you also asked why a docs-only PR touches many schema `Description` strings in `resource_github_repository.go`. Those are on the branch from the original scope. Happy to split them into their own PR if you would rather keep this one purely template plus generated docs.

mvanhorn:
Took the preferred path.

`private` is marked `Deprecated: "use visibility instead"` in the schema. Examples now live under `examples/resources/github_repository/` (`resource.tf`, `resource_forked.tf`, `import.sh`) so tfplugindocs picks them up. `templates/resources/repository.md.tmpl` dropped the hardcoded fallbacks and uses the same `.HasExamples` / `.HasImport` pattern as the generic template. Regenerated docs show `private` as `(Boolean, Deprecated)`.

deiga (dismissed):
LGTM!

deiga:
> ## Pull request overview
> 
> Copilot reviewed 4 out of 6 changed files in this pull request and generated no new comments.
> 
> 
> 
> 
> <details>
> <summary>Suppressed comments (5)</summary>
> 
> **github/resource_github_repository.go:69**
> * `private` is declared with `ConflictsWith: []string{"visibility"}`, so `visibility` cannot override it—Terraform rejects configurations that set both. Document this as the replacement for `private`, not as an override.
> ```
> 				Description:      "Can be 'public' or 'private'. If your organization is associated with an enterprise account using GitHub Enterprise Cloud or GitHub Enterprise Server 2.20+, visibility can also be 'internal'. The visibility parameter overrides the private parameter.",
> ```
> **github/resource_github_repository.go:304**
> * The new schema-generated row only displays `(Deprecated)` and does not render the `Deprecated` field's message, so the documented replacement with `github_branch_default` has been lost. Include that migration guidance in `Description` so it remains visible in generated docs.
> ```
> 				Description: "The name of the default branch of the repository. NOTE: This can only be set after a repository has already been created, and after a correct reference has been created for the target branch inside the repository. This means a user will have to omit this parameter from the initial repository creation and create the target branch inside of the repository prior to setting this attribute.",
> ```
> **github/resource_github_repository.go:332**
> * After switching to `SchemaMarkdown`, the rendered `pages` row no longer says to use `github_repository_pages`; tfplugindocs only emits the deprecated marker, not the `Deprecated` message. Keep the replacement/removal guidance in `Description` so users can migrate.
> ```
> 				Description: "The repository's GitHub Pages configuration.",
> ```
> **docs/resources/repository.md:67**
> * This generated row now tells users how to enable downloads but drops the prior warning that the attribute is no longer used and will be removed. Move the deprecation details/link from the schema's `Deprecated` field into the `has_downloads` `Description`, then regenerate this file.
> ```
> - `has_downloads` (Boolean, Deprecated) Set to 'true' to enable the (deprecated) downloads features on the repository.
> ```
> **templates/resources/repository.md.tmpl:5**
> * The PR description presents this as two targeted documentation edits, but this replaces the full resource template and also migrates examples, import docs, and many schema descriptions. That materially broader scope is not documented and has introduced unrelated documentation changes; either split the tfplugindocs migration from the deprecation fix or update the PR description to explain and validate the expanded scope.
> ```
> subcategory: ""
> description: |-
> {{ .Description | plainmarkdown | trimspace | prefixlines "  " }}
> ```
> </details>
> 
> 


@mvanhorn Please check these comments out. There might have been some information lost in the docs

github/resource_github_repository.go
```diff
@@ -22,6 +22,7 @@ func resourceGithubRepository() *schema.Resource {
 		ReadContext:   resourceGithubRepositoryRead,
 		UpdateContext: resourceGithubRepositoryUpdate,
 		DeleteContext: resourceGithubRepositoryDelete,
+		Description:   "This resource allows you to create and manage repositories within your GitHub organization or personal account.\n\n~> **Note** When used with GitHub App authentication, even GET requests must have the `contents:write` permission. Without it, the following arguments will be ignored, leading to unexpected behavior and confusing diffs: `allow_merge_commit`, `allow_squash_merge`, `allow_rebase_merge`, `merge_commit_title`, `merge_commit_message`, `squash_merge_commit_title` and `squash_merge_commit_message`.",
```
stevehipwell: Please use the pattern `"Resource to ..."` for the description and keep it short. Additional content can be added directly to the template. Please also follow the resource structure guidance in _ARCHITECTURE.md_ for where this should be placed.

stevehipwell (commented):
Please add _import-by-string-id.tf_ to the examples and use the format _resource\_N.tf_ (e.g. _resource\_1.tf_ & _resource\_2.tf_) for the example resources adding in a comment to explain the purpose.

docs/resources/repository.md
```diff
@@ -1,18 +1,21 @@
...
-  Creates and manages repositories within GitHub organizations or personal accounts
+  Resource to create and manage a GitHub repository.
 ---
 
 # github_repository (Resource)
 
-This resource allows you to create and manage repositories within your GitHub organization or personal account.
+Resource to create and manage a GitHub repository.
 
 ~> **Note** When used with GitHub App authentication, even GET requests must have the `contents:write` permission. Without it, the following arguments will be ignored, leading to unexpected behavior and confusing diffs: `allow_merge_commit`, `allow_squash_merge`, `allow_rebase_merge`, `merge_commit_title`, `merge_commit_message`, `squash_merge_commit_title` and `squash_merge_commit_message`.
```
stevehipwell:
```suggestion
~> When used with GitHub App authentication, even GET requests must have the `contents:write` permission. Without it, the following arguments will be ignored, leading to unexpected behavior and confusing diffs: `allow_merge_commit`, `allow_squash_merge`, `allow_rebase_merge`, `merge_commit_title`, `merge_commit_message`, `squash_merge_commit_title` and `squash_merge_commit_message`.
```

stevehipwell (dismissed):
LGTM

CC @deiga

mvanhorn:
@stevehipwell applied your suggestion at docs/resources/repository.md, and the same change in `templates/resources/repository.md.tmpl` so it survives regeneration. The examples are already on the `resource_N.tf` pattern with `import-by-string-id.tf` alongside, and the generated page has its Example Usage and Import sections.

deiga (dismissed):
LGTM

deiga:
@mvanhorn Please rebase so that we can get this merged ☺️

deiga:
@mvanhorn please fix the docs failures!

robert-crandall (dismissed):
Nice work! I regenerated the docs locally and confirmed they match, and every Go change is description-only, so there's no schema or state impact. One small ask: could you update the PR description? It still describes a tweak to `website/docs/r/repository.html.markdown`, but this actually migrates the resource to generated docs, which is a much better change than the body gives it credit for. LGTM otherwise.

mvanhorn:
@deiga docs failures are fixed and pushed.

The cause was in the schema description rather than the markdown: `html_url` quoted its example URL with single quotes, so the regenerated page rendered a bare URL and rumdl's MD034 flagged it. On main that URL sits in a code span, so I restored that form in `github/resource_github_repository.go` and regenerated. `Check` was only failing because it aggregates Documentation, so both should go green.

Verified locally with rumdl 0.2.60 (same version the workflow pins): no issues across 168 files. Branch is already current with main, so no rebase was needed.

deiga (approved):

robert-crandall (approved):

mvanhorn:
Description updated, @robert-crandall — thanks for the nudge, and for regenerating the docs to check them.

You were right that the body undersold it. It still described a hand edit to `website/docs/r/repository.html.markdown`, which is not a file this PR touches any more; the change is the migration to the generated pipeline, with the deprecation falling out of the schema rather than being written into a page by hand. The rewritten description says that.

PR #3300, opened 2026-03-24
fix: add nil check for ref_name element in expandConditions

### LINKED ISSUE #3299: [BUG]: Panic in expandConditions when conditions.ref_name is nil/empty
### Expected Behavior

When applying a github_repository_ruleset resource with an empty array conditions.ref_name, the provider should either produce a validation error or treat it as empty (not panic).

### Actual Behavior

Terraform plan/apply fails with a panic:

```
```

Stack trace:
by using the github.com/integrations/terraform-provider-github/v6/github.expandConditions function (util_rules.go:194) when conditions.ref_name is empty array or contains nil, the provider panics rather than failing gracefully.

### Terraform Version

Terraform v1.9.5
provider registry.terraform.io/integrations/github v6.11.1

### Affected Resource(s)

- github_repository_ruleset

```hcl
  repository = "some-repo"
  name = "test"
  target = "branch"
  enforcement = "disabled"

  conditions {
    ref_name {
      # No include/exclude specified or empty arrays
    }
  }
  rules {
    creation = true
  }
}
```

### Steps to Reproduce

1. Write the above github_repository_ruleset resource with an empty array conditions.ref_name block
2. Run terraform apply

### Debug Output

```shell

```

### Panic Output

```shell

github.com/integrations/terraform-provider-github/v6/github.expandConditions({0xc00067d2e0, ...}, 0x0)
	github.com/integrations/terraform-provider-github/v6/github/util_rules.go:194 +0xb6b
...
Error: The terraform-provider-github_v6.11.1 plugin crashed!
```

### Code of Conduct

majiayu000:
## Summary

Prevent repository ruleset condition expansion from panicking on nil internal
entries, while validating empty `ref_name` blocks according to the target
contract.

## Changes

- ignore nil first entries while expanding `ref_name`, `repository_name`, and
  `repository_property`
- require a non-empty `ref_name` block for branch and tag targets
- reject explicit empty or populated `ref_name` blocks for push targets
- continue allowing push rulesets that omit `ref_name`
- add focused helper and real Terraform schema and `CustomizeDiff` planning
  regressions

The earlier file split remains reverted so the implementation stays in the
existing rules files.

## Configuration matrix covered

- branch with empty `ref_name` block: rejected without panic
- push with empty `ref_name` block: rejected
- repository push with `conditions` omitted: accepted
- organization push with `repository_name` and `ref_name` omitted: accepted

## Validation

- focused helper and HCL PlanOnly tests
- focused race tests
- `make test`
- `go vet ./...`
- `CGO_ENABLED=0 go build ./...`
- `gofmt` and `git diff --check`

The PlanOnly regressions use the real provider schema and `CustomizeDiff`
without GitHub credentials. Live acceptance execution still requires repository
credentials.

Fixes #3299.


github/util_rules_test.go
```diff
@@ -1066,6 +1066,25 @@ func TestExpandRepositoryPropertyConditions_NilPropertyValues(t *testing.T) {
...
+
+	result := expandConditions(input, true)
+
+	if result == nil {
+		t.Fatal("Expected result to not be nil")
+	}
+	if result.RefName != nil {
+		t.Errorf("Expected RefName to be nil, got %+v", result.RefName)
+	}
+}
```
deiga: suggestion: Thanks for the test, could we use a test array here?
```suggestion
func TestExpandConditions(t *testing.T) {
	t.Parallel()

	for _, d := range []struct {
		testName string
		input    []any
		want     *github.RepositoryRulesetConditions
	}{
		{
			testName: "returns_nil_for_empty_input",
			input:    []any{},
			want:     nil,
		},
		{
			testName: "returns_nil_for_empty_input_slice",
			input:    []any{nil},
			want:     nil,
		},
		{
			testName: "returns_empty_conditions_for_empty_input_slice",
			input:    []any{map[string]any{}},
			want:     &github.RepositoryRulesetConditions{},
		},
		{
			testName: "returns_empty_conditions_for_empty_ref_name",
			input:    []any{map[string]any{"ref_name": []any{}}},
			want:     &github.RepositoryRulesetConditions{},
		},
		{
			testName: "returns_empty_conditions_for_empty_ref_name_arrays",
			input:    []any{map[string]any{"ref_name": []any{map[string]any{"include": []any{}, "exclude": []any{}}}}},
			want:     &github.RepositoryRulesetConditions{RefName: &github.RepositoryRulesetRefConditionParameters{Include: []string{}, Exclude: []string{}}},
		},
		{
			testName: "returns_empty_conditions_for_nil_ref_name_arrays",
			input:    []any{map[string]any{"ref_name": []any{nil}}},
			want:     &github.RepositoryRulesetConditions{},
		},
	} {
		t.Run(d.testName, func(t *testing.T) {
			t.Parallel()

			got := expandConditions(d.input, false)

			if diff := cmp.Diff(got, d.want); diff != "" {
				t.Fatalf("got %+v, want %+v", got, d.want)
			}
		})
	}
}
```

deiga (changes requested):
Amazing, thank you! 🎉

github/util_rules_test.go
majiayu000: Done! Converted to a table-driven test with all 6 cases as suggested. All subtests pass. Thanks for the review!

deiga (dismissed):

majiayu000:
Resolved the merge conflict with current upstream main. I also adjusted the added TestExpandConditions assertions to avoid introducing a new import conflict, then verified both the PR branch and the local upstream-main merge result with: go test ./github.

stevehipwell (commented):
Thanks for the PR @majiayu000, could you please rebase it.

majiayu000:
Rebased this branch onto current `main` and addressed the Copilot review threads.

Changes in the latest push:
- Added the same nil-element guard for `repository_name` that `ref_name` already had.
- Added a `repository_name: []any{nil}` regression case in `TestExpandConditions`.
- Renamed the two misleading `TestExpandConditions` cases called out in review.
- Split the oversized rules helpers/tests into focused files; the functional change is the `repository_name` guard plus the test case above.

Verified locally:
- `go test ./github -run TestExpandConditions -count=1`
- `go test ./github -count=1`
- `go test ./... -count=1`
- `go build ./...`

majiayu000:
Addressed the latest Copilot thread in `aa89c835`.

Follow-up changes:
- Added the same nil-first guard for `repository_property`.
- Added a `repository_property: []any{nil}` regression case in `TestExpandConditions`.
- Replaced the new-file unchecked type assertions with checked helper functions so strict new-code lint no longer reports `forcetypeassert` issues.

Verified locally:
- `go test ./github -run TestExpandConditions -count=1`
- `go test ./github -count=1`
- `go test ./... -count=1`
- `go build ./...`
- `make build -o lintcheck`
- `make test`
- `/tmp/codex-bin/golangci-lint run ./... --config .golangci.yml --enable forcetypeassert --new-from-merge-base main` -> 0 issues

The latest upstream workflow runs for CI/CodeQL/acceptance are currently `action_required` with no jobs created, so they appear to need repository-side approval before they can run.

devantler:
Just ran into this issue as well at work. The error that it creates when failing is not very actionable, so a getting this merged would be super helpful. I am not sure everyone will figure out that there is a workaround, as it is not clearly stated in the logs, as it just throws a stack trace.

So here is a gentle nudge to hopefully get this prioritized and merged 🙏🏻

stevehipwell:
@majiayu000 could you explain why you needed to split up the files? It makes it harder to review when the code location and ordering has been modified in this way.

majiayu000:
The file split was not required for this fix. I reverted it and rebuilt the PR on current main, keeping the nil guards and focused regression tests in the existing rules files to minimize review churn. The new head is a 4-file, single-commit diff; current-head CI is running.

majiayu000:
The current head is now based directly on current main and the fork-safe pull_request_target checks are green. The CI and CodeQL runs are currently `action_required` with no jobs, so they need maintainer approval before the required current-head checks can execute. Could a maintainer approve those workflow runs and review the reduced 4-file diff?

github/util_ruleset_validation_test.go
```diff
@@ -32,6 +32,11 @@ func Test_validateConditionsFieldForPushTarget(t *testing.T) {
 			conditions:  map[string]any{"ref_name": []any{}},
 			expectError: false,
 		},
+		{
+			name:        "valid push target with nil ref_name element",
+			conditions:  map[string]any{"ref_name": []any{nil}},
+			expectError: false,
+		},
```
deiga, line 40: issue: This seems to assert the wrong thing, `ref_name` should not be set for `push` target, so an empty block is not valid config

deiga (changes requested):
Please also update ruleset acceptance tests to verify which combinations were failing and which shouldn't be allowed

github/util_ruleset_validation_test.go
majiayu000: Addressed in 6bbddf32. Push targets now reject an explicit empty `ref_name {}` block, while an omitted `ref_name` remains valid. I also added real Terraform schema/CustomizeDiff PlanOnly coverage for branch empty-block rejection, push empty-block rejection, repository push omission, and organization push with `repository_name` plus omitted `ref_name`. Focused, race, `make test`, vet, build, formatting, and diff checks pass.

majiayu000:
The requested validation and planning matrix is now on current head `6bbddf32`, with the review thread answered. The fresh CI and CodeQL runs are `action_required`; could a maintainer approve those current-head workflows and re-review when convenient?

github/resource_github_organization_ruleset_test.go
```diff
@@ -924,6 +924,42 @@ resource "github_organization_ruleset" "test" {
...
+			}
+		`
+
+		resource.UnitTest(t, resource.TestCase{
+			ProviderFactories: providerFactories,
+			Steps: []resource.TestStep{
+				{
+					Config:             config,
+					PlanOnly:           true,
+					ExpectNonEmptyPlan: true,
```
deiga, line 957: question: why does it expect this? Please use configplanchecks to have granular explanations

github/resource_github_repository_ruleset_test.go
```diff
@@ -779,6 +779,100 @@ resource "github_repository_ruleset" "test" {
...
+			}
+		`
+
+		resource.UnitTest(t, resource.TestCase{
+			ProviderFactories: providerFactories,
+			Steps: []resource.TestStep{
+				{
+					Config:             config,
+					PlanOnly:           true,
+					ExpectNonEmptyPlan: true,
```
deiga, line 870: https://github.com/integrations/terraform-provider-github/pull/3300#discussion_r3775272372

deiga (changes requested):

PR #3301, opened 2026-03-24
fix: allow empty required_workflow blocks in required_workflows rule

### LINKED ISSUE #3217: [BUG]: required_workflows rule cannot represent empty workflows list (UI / API inconsistency)
### Expected Behavior

The provider should allow `required_workflows` with zero `required_workflow` sub-blocks, matching the behavior of the GitHub UI and API:

```hcl
resource "github_organization_ruleset" "example" {
  name        = "strict-master"
  target      = "branch"
  enforcement = "active"

  conditions {
    ref_name {
      include = ["~DEFAULT_BRANCH"]
      exclude = []
    }
    repository_name {
      include = ["~ALL"]
      exclude = []
    }
  }

  rules {
    required_workflows {
      do_not_enforce_on_create = false
      # No required_workflow blocks — should be valid
    }
  }
}
```

### Actual Behavior

```
Error: Insufficient required_workflow blocks

  on main.tf line XX, in resource "github_organization_ruleset" "example":

  At least 1 "required_workflow" blocks are required.
```

### Terraform Version

Terraform v1.5.7
on darwin_arm64

### Affected Resource(s)

- `github_repository_ruleset`

```hcl

```

### Steps to Reproduce

1. In the GitHub UI, create an organization ruleset
2. Enable "Require workflows to pass" without adding any specific workflows
3. Attempt to import or recreate this ruleset in Terraform

### Debug Output

```shell

```

### Panic Output

```shell

```

### Code of Conduct

sc0rp10:
Resolves #3217

----

### Before the change?

- The `required_workflow` sub-block inside `required_workflows` was defined with `Required: true` and `MinItems: 1`, forcing users to specify at least one workflow.
- This made it impossible to represent or import rulesets created in the GitHub UI with "Require workflows to pass" enabled but no specific workflows configured.

### After the change?

- Changed `required_workflow` from `Required: true` to `Optional: true` and removed the `MinItems: 1` constraint.
- An empty `required_workflows {}` block (with no `required_workflow` sub-blocks) is now valid, matching the behavior of the GitHub UI and API.

### Pull request checklist

- [ ] Yes
- [x] No

----

### DIFF
--- a/github/resource_github_organization_ruleset_test.go
+++ b/github/resource_github_organization_ruleset_test.go
@@ -184,6 +184,53 @@ resource "github_organization_ruleset" "test" {
 		})
 	})
 
+	t.Run("create_branch_ruleset_with_empty_required_workflows", func(t *testing.T) {
+		randomID := acctest.RandStringFromCharSet(5, acctest.CharSetAlphaNum)
+		rulesetName := fmt.Sprintf("%s-empty-wf-ruleset-%s", testResourcePrefix, randomID)
+
+		config := fmt.Sprintf(`
+resource "github_organization_ruleset" "test" {
+	name        = "%s"
+	target      = "branch"
+	enforcement = "active"
+
+	conditions {
+		repository_name {
+			include = ["~ALL"]
+			exclude = []
+		}
+
+		ref_name {
+			include = ["~DEFAULT_BRANCH"]
+			exclude = []
+		}
+	}
+
+	rules {
+		required_workflows {
+			do_not_enforce_on_create = false
+		}
+	}
+}
+`, rulesetName)
+
+		resource.Test(t, resource.TestCase{
+			PreCheck:          func() { skipUnlessHasPaidOrgs(t) },
+			ProviderFactories: providerFactories,
+			Steps: []resource.TestStep{
+				{
+					Config: config,
+					Check: resource.ComposeTestCheckFunc(
+						resource.TestCheckResourceAttr("github_organization_ruleset.test", "name", rulesetName),
+						resource.TestCheckResourceAttr("github_organization_ruleset.test", "enforcement", "active"),
+						resource.TestCheckResourceAttr("github_organization_ruleset.test", "rules.0.required_workflows.0.do_not_enforce_on_create", "false"),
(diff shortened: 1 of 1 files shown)


deiga:
@sc0rp10 Thanks for the PR!

Question: Does the "Require workflows to pass" toggle without specified workflows do anything? AFAIK it only affects specified workflows

sc0rp10:
@deiga Thanks for the review!

You're right that an empty `required_workflows` rule doesn't actively block anything — with no workflows specified, there's nothing to enforce. However, the GitHub UI and API both allow this state, which creates a practical problem for the provider:

1. **Import failures**: A user enables "Require workflows to pass" in the UI without adding specific workflows (e.g., intending to add them later). When they try to `terraform import` that ruleset, it fails because the provider rejects the empty workflow list.
2. **Plan/apply drift**: If a ruleset is managed in Terraform and someone removes all workflows via the UI, `terraform plan` can't represent the current state, leading to confusing errors.

The provider should faithfully represent any state the API can return. Since the API accepts and returns `required_workflows` with an empty `workflows` array, the provider should too.

sc0rp10:
To add more context: the primary use case here is enforcing that PRs have successful checks before merge. The "Require workflows to pass" toggle in rulesets serves as a gate — even without specifying particular workflows, it enforces that PR checks must succeed. Without being able to represent this state in Terraform, users are forced to disable the rule entirely, losing that enforcement.

deiga:
Even if the UI allows you to specify an empty list of required workflows, it doesn't do anything. The UI states "Require all changes made to a targeted branch to pass the **specified** workflows before they can be merged." and the API has similar wording https://docs.github.com/en/rest/orgs/rules?apiVersion=2026-03-10#create-an-organization-repository-ruleset

There is no use-case for the provider to allow this.

> The provider should faithfully represent any state the API can return.

This is not true, the provider should be an abstraction layer on top of the API

>  the primary use case here is enforcing that PRs have successful checks before merge. The "Require workflows to pass" toggle in rulesets serves as a gate

I think you're mixing something up. `required_workflows` is a way to ensure all Org PRs have a shared workflow run. When you talk about "enforcing that PRs have successful checks before merge", then you might mean `required_status_checks`. Which also needs a specified list of status checks.


AFAIK Import doesn't fail if only the boolean has been set, it would just cause drift. Can you show me output of a failing import?

PR #3305, opened 2026-03-25
Fix: prevent unexpected status check contexts deletion for archived repositories

### LINKED ISSUE #3306: [BUG]: unexpected 'contexts' deletion in github_branch_protection when archiving repository
### Expected Behavior

 Archiving a repository should not trigger changes to the required_status_checks.0.contexts list in the
  Terraform state. The provider should recognize that the status checks still exist in the modern checks array
  (which the GitHub API continues to return) and maintain the legacy contexts state to avoid unnecessary plan
  churn.

### Actual Behavior

 After archiving a repository, terraform plan shows an unexpected change to delete the contexts list:

```bash
   1   ~ resource "github_branch_protection_v3" "example" {
   2       ~ required_status_checks {
   3           - contexts = [
   4               - "ci/test",
   5               - "ci/build",
   6             ] -> null
   7             # ... (other fields like 'checks' remain unchanged)
   8         }
   9     }
```

This caused the error below when archiving a repo:

```bash
│ Error: PUT https://api.github.com/repos/ORG/REPO_NAME/branches/master/protection: 403 Repository was archived so is read-only. []
--
│
```

  This happens because the legacy contexts field is being returned as an empty list by the GitHub API for
  archived repositories, while the modern checks array remains populated.

  Root Cause Analysis
   1. Dual-Population: When a branch protection rule is created using the modern checks array, GitHub's
      backend auto-populates the legacy contexts field for backward compatibility.
   2. Computed State: Because contexts is marked as Computed: true in the provider, Terraform saves these
      auto-populated values into the .tfstate during the initial Read.
   3. API Change on Archive: When a repository is archived, the GitHub REST (v3) API returns an empty list for
      the legacy contexts field, but correctly returns the data in the checks array.
   4. Drift Detection: The provider's Read function receives the empty contexts from the API and, since it
      differs from the saved state, Terraform interprets this as a deletion.

  Proposed Fix
  The provider's Read logic should be updated to be more resilient to this API inconsistency:
   * Fallback Logic: In github/resource_github_branch_protection_v3_utils.go, the
     flattenAndSetRequiredStatusChecks function should fallback to populating the contexts state from the
     names in the checks array if the legacy contexts field is returned as empty.
   * Schema Update: In github/resource_github_branch_protection.go, ensure the contexts field is marked as
     Computed: true to prevent drift detection for auto-populated legacy values.

### Terraform Version

- terraform 1.4.7

### Affected Resource(s)

- github_branch_protection (GraphQL/v4)
- github_branch_protection_v3 (REST/v3)

```hcl

```

### Steps to Reproduce

- Create a repo with terraform
- Update archive = true

Introduced in 

https://github.com/integrations/terraform-provider-github/releases/tag/v5.16.0

[https://github.com/integrations/terraform-provider-github/pull/1415/changes#diff-a483da9852b3cd77bdd18a439d5d0[…]8f2c5970b4a1e4e03d8747526e29R56-R63](https://github.com/integrations/terraform-provider-github/pull/1415/changes#diff-a483da9852b3cd77bdd18a439d5d0422c6598f2c5970b4a1e4e03d8747526e29R56-R63) (edited) 
[10:13 AM]https://github.com/integrations/terraform-provider-github/pull/1415/changes#diff-63892fd5ebce048bce10d1f8c7defe1356f154a22acf8cc2f82ece6c3c6914b7L44

### Debug Output

```shell

```

### Panic Output

```shell

```

### Code of Conduct

hanyouqing:
Resolves #3306

  This PR addresses an issue where archiving a repository would cause Terraform to show a plan diff to delete
  the contexts field in github_branch_protection and github_branch_protection_v3 resources, even when status
  checks are still active via the newer checks API.

----

### Before the change?

   * REST (v3): When a repository is archived, the GitHub REST API stops returning data in the legacy contexts
     field but continues to return it in the checks array. Since contexts is Computed, the provider detects
     this as a deletion in the state.
   * GraphQL (v4): The contexts field was not marked as Computed, leading to drift detection when it was
     populated by the API but absent from the user's HCL configuration.

### After the change?

   1. REST (v3) Fallback: Modified the Read logic in github/resource_github_branch_protection_v3_utils.go to
      automatically populate the contexts state from the checks array if the legacy contexts field is empty.
      This keeps the legacy field in sync with the actual status checks.
   2. GraphQL (v4) Schema Update: Marked the contexts field as Computed: true in
      github/resource_github_branch_protection.go. This prevents Terraform from seeing the field as drift that
      needs to be "corrected" (deleted) when not explicitly specified in the configuration.

### Pull request checklist

- [ ] Yes
- [x] No

----

### DIFF
--- a/github/resource_github_branch_protection_test.go
+++ b/github/resource_github_branch_protection_test.go
@@ -9,7 +9,10 @@ import (
 
 	"github.com/hashicorp/terraform-plugin-testing/helper/acctest"
 	"github.com/hashicorp/terraform-plugin-testing/helper/resource"
+	"github.com/hashicorp/terraform-plugin-testing/knownvalue"
+	"github.com/hashicorp/terraform-plugin-testing/statecheck"
 	"github.com/hashicorp/terraform-plugin-testing/terraform"
+	"github.com/hashicorp/terraform-plugin-testing/tfjsonpath"
 )
 
 func TestAccGithubBranchProtectionV4(t *testing.T) {
@@ -167,30 +170,22 @@ func TestAccGithubBranchProtectionV4(t *testing.T) {
 	t.Run("removes from state when repository is archived", func(t *testing.T) {
 		randomID := acctest.RandStringFromCharSet(5, acctest.CharSetAlphaNum)
 		testRepoName := fmt.Sprintf("%sbranch-protection-%s", testResourcePrefix, randomID)
-		config := fmt.Sprintf(`
+
+		configTemplate := `
 			resource "github_repository" "test" {
 			  name      = "%s"
 			  auto_init = true
+			  archived  = %t
 			}
 
 			resource "github_branch_protection" "test" {
 			  repository_id = github_repository.test.node_id
 			  pattern       = "main"
 			}
-		`, testRepoName)
+		`
 
-		configArchived := fmt.Sprintf(`
-			resource "github_repository" "test" {
-			  name      = "%s"
-			  auto_init = true
-			  archived  = true
-			}
-
-			resource "github_branch_protection" "test" {
-			  repository_id = github_repository.test.node_id
-			  pattern       = "main"
-			}
-		`, testRepoName)
+		config := fmt.Sprintf(configTemplate, testRepoName, false)
+		configArchived := fmt.Sprintf(configTemplate, testRepoName, true)
 
 		resource.Test(t, resource.TestCase{
 			PreCheck:          func() { skipUnauthenticated(t) },
@@ -198,15 +193,15 @@ func TestAccGithubBranchProtectionV4(t *testing.T) {
 			Steps: []resource.TestStep{
 				{
 					Config: config,
-					Check: resource.ComposeAggregateTestCheckFunc(
-						resource.TestCheckResourceAttr("github_branch_protection.test", "pattern", "main"),
-					),
+					ConfigStateChecks: []statecheck.StateCheck{
+						statecheck.ExpectKnownValue("github_branch_protection.test", tfjsonpath.New("pattern"), knownvalue.StringExact("main")),
+					},
 				},
 				{
 					Config: configArchived,
-					Check: resource.ComposeAggregateTestCheckFunc(
-						resource.TestCheckResourceAttr("github_repository.test", "archived", "true"),
-					),
+					ConfigStateChecks: []statecheck.StateCheck{
+						statecheck.ExpectKnownValue("github_repository.test", tfjsonpath.New("archived"), knownvalue.Bool(true)),
+					},
 				},
 				{
 					Config:             configArchived,


github/resource_github_branch_protection.go
```diff
@@ -366,6 +374,25 @@ func resourceGithubBranchProtectionRead(d *schema.ResourceData, meta any) error
...
+	client := meta.(*Owner).v4client
+	err := client.Query(ctx, &query, variables)
+	if err == nil {
+		protection := query.Node.Node
+		if protection.Repository.IsArchived {
+			log.Printf("[INFO] Skipping update of branch protection (%s) because the repository (%s) is archived", d.Id(), protection.Repository.Name)
+			return nil
+		}
+	}
+
```
deiga: This shouldn't be necessary as Read will always be called before Update

github/resource_github_branch_protection.go
```diff
@@ -444,6 +471,25 @@ func resourceGithubBranchProtectionUpdate(d *schema.ResourceData, meta any) erro
...
+	client := meta.(*Owner).v4client
+	err := client.Query(ctx, &query, variables)
+	if err == nil {
+		protection := query.Node.Node
+		if protection.Repository.IsArchived {
+			log.Printf("[INFO] Skipping deletion of branch protection (%s) because the repository (%s) is archived", d.Id(), protection.Repository.Name)
+			return nil
+		}
+	}
+
```
deiga: This shouldn't be necessary as Read will always be called before Delete

github/resource_github_branch_protection_v3.go
```diff
@@ -349,6 +369,16 @@ func resourceGithubBranchProtectionV3Update(d *schema.ResourceData, meta any) er
...
+	orgName := meta.(*Owner).name
+	ctx := context.WithValue(context.Background(), ctxId, d.Id())
+
+	repo, _, err := client.Repositories.Get(ctx, orgName, repoName)
+	if err == nil {
+		if repo.GetArchived() {
+			log.Printf("[INFO] Skipping update of branch protection %s/%s (%s) because the repository is archived", orgName, repoName, branch)
+			return nil
+		}
+	}
```
deiga: This shouldn't be necessary as Read will always be called before Update

github/resource_github_branch_protection_v3.go
```diff
@@ -407,6 +437,14 @@ func resourceGithubBranchProtectionV3Delete(d *schema.ResourceData, meta any) er
...
 	ctx := context.WithValue(context.Background(), ctxId, d.Id())
 
+	repo, _, err := client.Repositories.Get(ctx, orgName, repoName)
+	if err == nil {
+		if repo.GetArchived() {
+			log.Printf("[INFO] Skipping deletion of branch protection %s/%s (%s) because the repository is archived", orgName, repoName, branch)
+			return nil
+		}
+	}
+
```
deiga: This shouldn't be necessary as Read will always be called before Delete

github/resource_github_branch_protection_v3_utils.go
deiga, line 1: Please update tests to cover regression of this

hanyouqing:
Hi @deiga, 

PR updated, feel free to take a look. Thanks for your review and suggestion.

github/resource_github_branch_protection_test.go
```diff
@@ -164,6 +164,62 @@ func TestAccGithubBranchProtectionV4(t *testing.T) {
...
+			  auto_init = true
+			}
+
+			resource "github_branch_protection" "test" {
+			  repository_id = github_repository.test.node_id
+			  pattern       = "main"
+			}
+		`, testRepoName)
+
+		configArchived := fmt.Sprintf(`
```
deiga: Please useaa single template string with variables

github/resource_github_branch_protection_test.go
```diff
@@ -164,6 +164,62 @@ func TestAccGithubBranchProtectionV4(t *testing.T) {
...
+			}
+		`, testRepoName)
+
+		resource.Test(t, resource.TestCase{
+			PreCheck:          func() { skipUnauthenticated(t) },
+			ProviderFactories: providerFactories,
+			Steps: []resource.TestStep{
+				{
+					Config: config,
+					Check: resource.ComposeAggregateTestCheckFunc(
```
deiga: Please use ConfigStateChecks

github/resource_github_branch_protection_v3_utils.go
deiga: I still don't see any new unit tests

deiga (changes requested):

hanyouqing:
Hi @deiga, any more suggestions for this PR?

Issue #3133, opened 2026-01-27
[MAINT]: Refactor `github_repository` resource


stevehipwell:
### Describe the need

The `github_repository` resource needs to be refactored to follow best practice including not calling update & read after create or read after update. This will allow us to refactor the logic as there will be separation between the lifecycles.

### Relevant log output

```shell

```

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

deiga:
Due to the nature of the Repo create API endpoint all repo settings can't be set during create, which is why it calls update to manage the "rest" of the settings. 

With that in mind, what do we think about separating repo and repo_settings into 2 resources? Similar to ORG and org_settings

stevehipwell:
@deiga which settings are you thinking of? Off the top of my head vulnerability alerts and pages need new resources. I'd also be tempted to move topics into their own resource. Have I missed any other settings?

deiga:
> [@deiga](https://github.com/deiga) which settings are you thinking of? Off the top of my head vulnerability alerts and pages need new resources. I'd also be tempted to move topics into their own resource. Have I missed any other settings?

Naive approach: The diff between `github.createRepoRequest` and `github.Repository`. That would ensure that we wouldn't have to call the Create and Edit endpoint in the Create function.
But that might be too naive 😬 
And the overlap could be hard to fathom

stevehipwell:
@deiga it's not that simple, the `github.Repository` type has all possible fields for a repository. We'd probably need to look at the docs to get a full understanding. But basically the "create" function **SHOULD NOT** call the "update" function even if we need to hit a second endpoint on create.

deiga:
> [@deiga](https://github.com/deiga) it's not that simple, the `github.Repository` type has all possible fields for a repository. We'd probably need to look at the docs to get a full understanding. But basically the "create" function **SHOULD NOT** call the "update" function even if we need to hit a second endpoint on create.

Definitely agree on the latter point.
And I think moving everything out isn't a great option either.

Some potential ones in addition to the mentioned ones:
- `security_and_analysis`: Since these can be limited by organization/enterprise maybe it could be a nicer UX to manage these as a separate resource
- `topics`: Since they anyway need to be managed by a separate endpoint

deiga:
I just realised that `topics` already have an existing resource, so we can just deprecate the field

deiga:
I linked #2090 as I think the refactor could include changes to how `template` changes are being enacted. 
[...]

Issue #3070, opened 2026-01-10
[MAINT] Tracking issue for refactoring logging to use `tflog` instead of `log`


deiga:
## Summary

This issue tracks the migration from Go's standard `log` package to HashiCorp's structured logging package [`tflog`](https://pkg.go.dev/github.com/hashicorp/terraform-plugin-log/tflog) across the provider codebase.

Resolves #2629

## Why?

The `tflog` package provides several benefits over standard `log`:

1. **Structured logging** - Separate log messages from filterable field data for programmatic parsing
2. **Better integration** - Works properly with Terraform's logging system (`TF_LOG`, `TF_LOG_PROVIDER`)
3. **Filtering capabilities** - Users can filter logs by fields, making debugging easier
4. **Consistent patterns** - Aligns with HashiCorp's official Terraform provider development practices
5. **Sensitive data masking** - Built-in support for masking sensitive values in logs

## Related Issues

- **#2996** - [MAINT] Migrate all resources and data sources to Context-aware CRUD functions - **Should be paired with this work** (see note below)
- **#2679** - [BUG]: Insufficient logging - Requests more logging output; better structured logging framework would help address this
- **#1226** - Provider hangs, without giving any logging or feedback - Rate limit logging would be improved with tflog in `transport.go`
- **#2925** - [MAINT]: Update transport implementation - Transport changes may want to coordinate with logging refactoring

> **Important**: The `tflog` package requires a `context.Context` parameter. Files that still use legacy CRUD functions (e.g., `Create`, `Read` instead of `CreateContext`, `ReadContext`) should be migrated to Context-aware functions as part of the same PR. See #2996 for details on the Context migration.

### Log Levels (least to most verbose)
- **Error** - Unexpected conditions before halting execution
- **Warn** - Unexpected conditions that don't stop execution (deprecations, external changes)
- **Info** - Documents logic conditions or events (state changes, decisions)
- **Debug** - Operational milestones and behaviors
- **Trace** - Intra-function steps and raw data details

### Structured Fields
Use structured log fields instead of embedding values in message strings:

```go
// Good: Structured fields for filtering
tflog.Debug(ctx, "Deleting secret", map[string]any{
    "repository":  repoName,
    "secret_name": secretName,
})

// Bad: Embedded values
log.Printf("[DEBUG] Deleting secret: %s/%s", repoName, secretName)
```

### Persistent Context
Use `tflog.SetField()` to attach fields to all subsequent logs in a function:

```go
[...]

PR #3310, opened 2026-03-28
[MAINT] Modernize `github_repository` resource

### LINKED ISSUE #2264: [BUG]: No warning/error on conflicting resource properties for `github_repository`
### Expected Behavior

The Terraform provider currently exhibits a problematic behavior where it silently ignores certain configuration settings when conflicting configurations are present. 

Specifically, if a user defines `squash_merge_commit_message` or `squash_merge_commit_title` while disabling squash merges through `allow_squash_merge = false`, these settings are rendered meaningless. Similarly, specifying `merge_commit_message` or `merge_commit_title` without enabling `allow_merge_commit` has no effect.

Although this behavior is documented, relying solely on documentation is insufficient. Users may easily overlook these details, leading to unexpected outcomes and configuration confusion.

To enhance user experience and prevent silent misconfigurations, the provider should implement stronger validation mechanisms.  Instead of silently ignoring conflicting settings, it should either:

* **Issue a warning** during plan or apply stages, explicitly informing users about the conflict and the settings being ignored.
* **Throw an error**, preventing the deployment of a configuration containing these contradictions, ensuring users address the issue before proceeding. 

This proactive approach would provide clearer feedback, reduce user frustration, and promote a more robust and reliable configuration process.

### Actual Behavior

see above

### Terraform Version

v1.8.3

### Affected Resource(s)

- `github_repository`

_No response_

### Steps to Reproduce

_No response_

### Debug Output

_No response_

### Panic Output

_No response_

### Code of Conduct

### LINKED ISSUE #2604: [BUG]: There appears to be a race condition in the github api when creating new repos
### Expected Behavior

When creating a new repo by making use of a github action runner from github hosted runners repos can be created and configured as required.

### Actual Behavior

When running the same code on a computer that is not close to the github api all functionality works as expected. Run the same code with the same credentials within an actions runner and configurations appear to fail.

I think this is because the repo is not yet available even though the API has responded to say that it is. If you are just a few ms slower in the requests, it will work.

I suspect a simple "sleep 2" after the repo create api call returns would be enough to fix the problem. Or check after the create has happened to confirm that the repo is actually visible now.

### Terraform Version

Terraform: 1.10.3
registry.terraform.io/integrations/github = 6.6.0"

### Affected Resource(s)

- github_repository
- github_repository_environment

```hcl
variable "name" {
  description = "Repo name"
  type = string
}

variable "description" {
  description = "Repo description"
  type = string
}

variable "visibility" {
  description = "Repo visibility"
  type = string
  default = "internal"
}

variable "environment_protections" {
  description = "Environment protection rules. Map key is the environment name."
  type = map(object({
    prevent_self_review = optional(bool, true)
    can_admins_bypass = optional(bool, true)
    reviewers = object({
      teams = optional(list(string), [])
      users = optional(list(string), [])
    })
  }))
  default = {}
}

resource github_repository "current" {
  name        = var.name
  description = var.description
  visibility  = var.visibility
  auto_init   = true
  archive_on_destroy = true
  
  # This one appears to cause a 422 error
  vulnerability_alerts = true
}

# This also causes a 422
resource github_repository_environment "current" {
  for_each = var.environment_protections
  environment = each.key
  repository  = github_repository.current.name

  prevent_self_review = each.value.prevent_self_review
  can_admins_bypass   = each.value.can_admins_bypass
  reviewers {
    teams = each.value.reviewers.teams
    users = each.value.reviewers.users
  }
}
```

### Steps to Reproduce

You need to run this code from within a github actions runner so you get really low latency on the API.
You'll need to create your variables values as required in the code.

### Debug Output

```shell
# This is the response from the console logs.

│ Error: PUT https://api.github.com/repos/XXX/devops-junk-3/vulnerability-alerts: 422 Failed to change dependabot alerts status. []
│ 
│   with module.zeam_generic_repos["devops-junk-3"].github_repository.current,
│   on ../modules/github/repo/main.tf line 47, in resource "github_repository" "current":
│   47: resource github_repository "current" {
```

### Panic Output

```shell
na
```

### Code of Conduct

### LINKED ISSUE #2700: [BUG]:
### Expected Behavior

Repo created and visibility changed to `internal`

### Actual Behavior

The terraform apply action fails with a `422 Failed to update visibility. A previous repository operation is still in progress.`

The issue is sporadic.  Sometimes I can retry right away and it works on the second attempt.  Other times it will fail 10 times in a row.  I believe it's caused by a race condition with the GitHub API -- making the assumption that it's because the repository create operation isn't quite finished before the visibility update is attempted.

```
terraform apply -auto-approve \
  -var "org=myorg" \
  -var "repo=myrepo \
  -var "repo_visibility=internal" \
  -var "template=mytemplate" \
  -var "team=myteam"
............
Plan: 30 to add, 0 to change, 0 to destroy.
github_repository.repo: Creating...

│ Error: PATCH https://api.github.com/repos/myorg/myrepo: 422 Failed to update visibility. A previous repository operation is still in progress. []
│ 
│   with github_repository.repo,
│   on repo.tf line 4, in resource "github_repository" "repo":
│    4: resource "github_repository" "repo" {
│ 
╵
Error: Process completed with exit code 1.
```

### Terraform Version

1.1.7

This is happening on our self-hosted github agents and without testing thoroughly, I don't want to break our 100s of other terraform workflows.

### Affected Resource(s)

- github_repository

```hcl
locals {
  template = toset([var.template])
}
resource "github_repository" "repo" {
  name        = var.repo
  description = var.repo_desc
  visibility  = var.repo_visibility
  auto_init   = true

  dynamic "template" {
    for_each = local.template
    content {
      owner                = var.org
      repository           = template.value
      include_all_branches = true
    }
  }
}

resource "github_branch_default" "default" {
  repository = github_repository.repo.name
  branch     = var.default_branch
}
```

### Steps to Reproduce

`terraform apply` -- see output above for full command.  I can provide all the `.tf` files if necessary.  Given that it happens on the very first step, I opted to not provide the other 6 `.tf` files

This is difficult to reproduce because it doesn't happen every time a repo creation happens.  

### Debug Output

```shell

```

### Panic Output

```shell

```

### Code of Conduct

### LINKED ISSUE #3302: [BUG]: 422 error: Race condition updating visibility to internal after template repo creation (no retry/backoff on PATCH)
### Expected Behavior

When creating a repository from a template with visibility = "internal", the operation should complete successfully without spurious 422 errors, and the desired visibility should be set reliably by Terraform. The provider should handle any normal GitHub-side settling delays internally (via retry/backoff) so users do not need to tune write_delay_ms or split operations across multiple applies.

### Actual Behavior

Creating a repo from a template with visibility = "internal" frequently fails with:

Error: PATCH https://api.github.com/repos/[org]/[repo]: 422 Failed to update visibility. A previous repository operation is still in progress. []

This is because the provider uses the template generate endpoint (which only supports 'private', not 'internal'), and then immediately issues a PATCH to set 'internal' visibility without any retry/backoff when GitHub is not ready. Users need to add large write_delay_ms to get reliable applies, but is not feasible for this one issue.

### Terraform Version

Terraform v1.9.5
+ provider registry.terraform.io/integrations/github v6.11.1 (and v6.11.0)

Reproducible on current releases.

### Affected Resource(s)

- github_repository (template create + visibility = "internal")

```hcl
resource "github_repository" "team_repos" {
  name        = "my-repo"
  template {
    owner      = "my-org"
    repository = "template-repo"
  }
  visibility     = "internal"
  auto_init      = true
  # additional fields as needed
}
```

### Steps to Reproduce

1. Create a template repo (must be "internal" or "private")
2. Use github_repository resource with that template and visibility = "internal"
3. terraform apply

Result:
Randomly (often), the operation fails with 422 error above if provider attempts to PATCH visibility too quickly after repo creation.

### Debug Output

```shell
Terraform apply output:

Error: PATCH https://api.github.com/repos/my-org/my-repo: 422 Failed to update visibility. A previous repository operation is still in progress. []

This repeats unless write_delay_ms is set to large values (2-10 seconds or more).

The provider code path is resourceGithubRepositoryCreate() followed by resourceGithubRepositoryUpdate() which issues the PATCH for visibility = "internal" with no retry.

Proposed Fix:
- In resourceGithubRepositoryUpdate(), add retry/backoff for PATCH /repos/:org/:repo when 422 'previous repository operation' is returned for visibility changes after template creation.
- See also PR #3123 (fixed template create, but not the PATCH retry).
```

### Panic Output

```shell
N/A
```

### Code of Conduct

### LINKED ISSUE #3554: [BUG]: github_repository reports drift on merge_commit_message and merge_commit_title if allow_merge_commit is false
### Expected Behavior

When `allow_merge_commit = false` in `github_repository` I would expect the values of `merge_commit_title` and `merge_commit_message` to be ignored as they can't be changed if merge commits are disabled. 

### Actual Behavior

I had the following settings applied

```
allow_merge_commits = true
merge_commit_message = "PR_BODY"
merge_commit_title = "PR_TITLE"
```

then when I flipped `allow_merge_commmits = false` 

```
allow_merge_commits = true
merge_commit_message = "PR_BODY"
merge_commit_title = "PR_TITLE"
```
and applied that. 

Now if I plan again I get the following drift:

```
!       merge_commit_message                    = "PR_TITLE" -> "PR_BODY"
!       merge_commit_title                      = "MERGE_MESSAGE" -> "PR_TITLE"
```

### Terraform Version

Terraform v1.15.8
on linux_arm64
+ provider registry.terraform.io/hashicorp/google v7.38.0

### GitHub Installation Type

- [x] I don't know

### Affected Resource(s)

- github_repository

```hcl

```

### Steps to Reproduce

I can't copy my actual terraform configuration here,  but I think this is the gist of it 

terraform config
```
resource "github_repository" "this" {
  name                        = "test"
  description                 = ""
  delete_branch_on_merge      = true
  auto_init                   = true
  allow_update_branch         = true
  allow_merge_commit          = true
  allow_rebase_merge          = true
  allow_squash_merge          = true
  squash_merge_commit_title   = "PR_TITLE"
  squash_merge_commit_message = "PR_BODY"
  merge_commit_title          = "PR_TITLE"
  merge_commit_message        = "PR_BODY"

}

```

terraform plan & apply that. 

Then change `allow_merge_commit = false`

terraform plan & apply that (it will work fine). 

Then `terraform plan` again, it will show drift in `merge_commit_title` and `merge_commit_message`:

```
 resource "github_repository" "this" {
        id                                      = "xxxxx"
!       merge_commit_message                    = "PR_TITLE" -> "PR_BODY"
!       merge_commit_title                      = "MERGE_MESSAGE" -> "PR_TITLE"
        name                                    = "xxxx"
        # (40 unchanged attributes hidden)

        # (1 unchanged block hidden)
    }
```

So it believes that current value of `merge_commit_message = PR_TITLE` which I guess it the GitHub defaults ( GitHub repo > Settings > General > Pull Requests > Allow merge commits > Default message will set `merge_commit_title: "MERGE_MESSAGE", merge_commit_message: "PR_TITLE"`) 

### Debug Output

```shell

```

### Code of Conduct

### LINKED ISSUE #3562: [BUG]:  Provider stalls for hours on unauthenticated refresh instead of failing fast
### Expected Behavior

When the provider has no token and is talking to `api.github.com`, an operation that can't
succeed anonymously should fail fast with a clear "authentication required" error.

### Actual Behavior

With an empty token the provider silently enters anonymous mode and proceeds. On a plan/refresh
over an org with many repositories it can stall for hours before failing.

I believe this is due to the anonymous primary rate limit (60 requests/hour/IP) combined with
`RateLimitTransport`'s sleep-until-reset behavior https://github.com/integrations/terraform-provider-github/blob/add2157906efe6ce657d6fda681635bf84336021/github/transport.go#L105-L111: 

Reported by a user of the downstream Pulumi provider in https://github.com/pulumi/pulumi-github/issues/1242.

### Terraform Version

Provider v6.14.0

### GitHub Installation Type

- [ ] I don't know

### Affected Resource(s)

- `github_repository`
- `github_repository_vulnerability_alerts`
- Any anonymous read that trips the 60/hour anonymous rate limit

```hcl
provider "github" {}  # no token in the environment

  name = "example"
}
```

### Steps to Reproduce

1. Manage several `github_repository` resources.
2. Run `terraform plan` with no `GITHUB_TOKEN`/`token` set.
3. Observe the run sleep for ~1 hour at a time and take hours to fail, rather than erroring
   immediately on the unauthenticated response.

### Debug Output

```shell

```

### Code of Conduct

### LINKED ISSUE #3592: github_repository Read doesn't refresh id/name after an out-of-band rename (unlike Create/Update)
## Terraform Version

1.5.7 (and later, unaffected by TF core version)

## Provider Version

v6.11.1 (confirmed still present at v6.13.0, the latest release as of this report)

## Affected Resource(s)

- github_repository

## Terraform Configuration Files

N/A — reproduces on any `github_repository` resource that is renamed out-of-band (i.e. `name` changes on an already-applied resource) between two `terraform apply`/`terraform plan -refresh-only` runs.

## Expected Behavior

After a rename, `terraform refresh`/`plan` should observe the *actual current* repository name and either report no diff (if the config's `name` already matches) or a correct diff.

## Actual Behavior

`resourceGithubRepositoryRead` (`github/resource_github_repository.go`) does:

```go
repoName := d.Id()
...
repo, resp, err := client.Repositories.Get(ctx, owner, repoName)
...
_ = d.Set("name", repoName)
```

`client.Repositories.Get` succeeds even when `repoName` is the *old* name, because the GitHub API transparently redirects a renamed repository's old-name lookup to the current one — `repo` in the response correctly reflects the new name (`repo.GetName()`), and every other field is set from that response (`repo.GetDescription()`, `repo.GetFullName()`, etc.). But `name` is set from the pre-read `repoName` local variable instead of `repo.GetName()`, so it never reflects the actual current name. `Read` also never calls `d.SetId(...)`, so the resource's ID likewise never advances past the original name.

By contrast, both `resourceGithubRepositoryCreate` and `resourceGithubRepositoryUpdate` correctly call `d.SetId(repo.GetName())` after their respective API calls (Update even has a comment noting exactly this hazard: "It's possible that `repo.GetName()` is different from `repoName` if the repository is renamed"). That same handling was never applied to Read.

Practical impact for consumers that track external identity by ID/name (e.g. Crossplane's upjet-based providers, which re-derive their external-name annotation from the Terraform ID on every reconcile): once a repository is renamed, `Read` keeps reporting the pre-rename name indefinitely, and any control loop that treats that as authoritative will keep trying to "correct" the live repository back to its old name.

## Steps to Reproduce

1. `terraform apply` a `github_repository` resource, e.g. `name = "foo"`.
2. Rename the repository out-of-band (GitHub UI, or `gh repo rename bar -R owner/foo`), so the live repo is now `owner/bar` but the Terraform ID in state is still `foo`.
3. `terraform plan -refresh-only` (or any `Read`/`RefreshWithoutUpgrade` invocation using the stale ID).
4. Observe: the read succeeds (via GitHub's redirect), but the refreshed `name` attribute (and the resource's ID) still shows `foo`, not `bar`.

## Suggested fix

In `resourceGithubRepositoryRead`, immediately after the `client.Repositories.Get` call:

```go
d.SetId(repo.GetName())
...
_ = d.Set("name", repo.GetName())
```

mirroring the existing `Create`/`Update` handling.

deiga:
Resolves #2604
Resolves #2264 
Resolves #3302
Resolves #3562
Resolves #3554
Resolves #3592
Addresses many parts of #3133 
Addresses #3070 

----

### Before the change?

- Tests used legacy testing checks
- Tests weren't being run in parallel
- Resource was still using `log`
- There was an deprecated gate against `archived` in the `Read` func
- `Create` would call `Update` and `Update` would call `Read` which is an anti-pattern

### After the change?

- Ensure that each CRUD function is fully self-contained
- Tests use modern `ConfigStateChecks`
- Tests can run in parallel, which reduces time to run tests immensely
- Logging has been refactored to use `tflog`
- After verification the "if not archived" gate in `Read` could be removed
- Updates the `resource` and `data-source` docs to use template generation
- Enable some `resource` tests that needed Org settings changes
- Adds `web_commit_signoff_required` computed attribute to `data-source`
- Adds timeouts to `resource` operations
- Adds deprecation warning to `topics` attribute in `resource`
- Adds validation that correct settings for `allow_squash_merge` and `allow_commit_merge` are set

### Pull request checklist

- [ ] Yes
- [x] No

----


deiga:
@doughlass very rude to ping us with reminders. 
We maintainers have comms channels where we discuss priorities and such. 

You're free to add comments to motivate some issue or PR, but pinging is wholly unnecessary

rilical (commented):
There are two state-behavior changes here that I think need either coverage or clearer deprecation handling. The PR removes the request and Read wiring for `has_downloads`, but the schema and generated docs still present it as a usable argument, so configurations can now set a value that is silently ignored. It also removes the archived-repository guard around several merge and forking fields even though the old code noted that GitHub may omit those values after archival. Could we either preserve the existing behavior or make the new behavior explicit, then add focused tests for both cases?

PR #3316, opened 2026-04-01
feat: use provider owner as default for `github_organization` data source

mattias-fjellstrom:
## Summary

When no `name` is provided to the `github_organization` data source, it now falls back to the provider's configured `owner` (or the authenticated user if no owner is set). This makes `name` optional.

## Motivation

Users who already configure an `owner` in the provider block shouldn't need to repeat it in every `github_organization` data source. This is consistent with how other data sources in this provider handle the owner fallback.

## Changes

- `github/data_source_github_organization.go`: changed `name` from `Required` to `Optional`+`Computed`, and fall back to `meta.(*Owner).name` when not supplied
- `github/data_source_github_organization_test.go`: added acceptance test verifying the fallback behaviour
- `website/docs/d/organization.html.markdown`: updated docs to reflect `name` is now optional, with a second example showing omitted `name`

## Example

Before (name was required):
```hcl
  name = "my-org"
}
```

After (name is  falls back to provider `owner`):optional 
```hcl
provider "github" {
  owner = "my-org"
}

data "github_organization" "example" {}
```

## Testing

All acceptance tests pass:
-  `uses_provider_owner_when_name_is_not_set` (new)
-  `queries_for_an_organization_without_error`
-  `queries_for_an_organization_with_archived_repos`
-  `queries_for_an_organization_summary_only_without_error`

### DIFF
--- a/github/data_source_github_organization_test.go
+++ b/github/data_source_github_organization_test.go
@@ -6,9 +6,32 @@ import (
 
 	"github.com/hashicorp/terraform-plugin-testing/helper/acctest"
 	"github.com/hashicorp/terraform-plugin-testing/helper/resource"
+	"github.com/hashicorp/terraform-plugin-testing/knownvalue"
+	"github.com/hashicorp/terraform-plugin-testing/statecheck"
+	"github.com/hashicorp/terraform-plugin-testing/tfjsonpath"
 )
 
 func TestAccGithubOrganizationDataSource(t *testing.T) {
+	t.Run("uses provider owner when name is not set", func(t *testing.T) {
+		config := `data "github_organization" "test" {}`
+
+		resource.Test(t, resource.TestCase{
+			PreCheck:          func() { skipUnlessHasOrgs(t) },
+			ProviderFactories: providerFactories,
+			Steps: []resource.TestStep{
+				{
+					Config: config,
+					ConfigStateChecks: []statecheck.StateCheck{
+						statecheck.ExpectKnownValue("data.github_organization.test", tfjsonpath.New("login"), knownvalue.StringExact(testAccConf.owner)),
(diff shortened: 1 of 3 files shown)


github/data_source_github_organization_test.go
```diff
@@ -9,6 +9,30 @@ import (
 )
 
 func TestAccGithubOrganizationDataSource(t *testing.T) {
+	t.Run("uses provider owner when name is not set", func(t *testing.T) {
+		config := `
+			data "github_organization" "test" {}
+		`
+
+		check := resource.ComposeTestCheckFunc(
```
deiga: Please don't use a variable for checks

github/data_source_github_organization_test.go
```diff
@@ -9,6 +9,30 @@ import (
...
+			resource.TestCheckResourceAttrSet("data.github_organization.test", "plan"),
+		)
+
+		resource.Test(t, resource.TestCase{
+			PreCheck:          func() { skipUnlessHasOrgs(t) },
+			ProviderFactories: providerFactories,
+			Steps: []resource.TestStep{
+				{
+					Config: config,
+					Check:  check,
```
deiga: Please use ConfigStateChecks instead

deiga (changes requested):

github/data_source_github_organization_test.go
mattias-fjellstrom: Noted! I've updated this to not use variables. I notice this very test file seems to use both approaches (variables and not variables) for the checks.

github/data_source_github_organization_test.go
mattias-fjellstrom: Done.

PR #3325, opened 2026-04-09
refactor: migrate resource_github_actions_environment_variable to tflog

texasich:
Replace Go standard `log` package with HashiCorp structured `tflog` for consistent logging across the provider.

Changes:
- Replace `log` import with `fmt` and add `tflog` import
- - Convert `log.Printf` to `tflog.Info` with structured fields
Resolves #3070 (partial — this PR covers one file)

---

### Before the change?

`resource_github_actions_environment_variable.go` uses Go's standard `log.Printf` for logging, which produces unstructured output inconsistent with the rest of the provider.

### After the change?

Logging uses HashiCorp's `tflog.Info` with structured key-value fields (`variable_id`), consistent with the pattern established in other migrated files like `resource_github_membership.go`.

### Pull request checklist

- [x] Schema migrations have been created if needed
- [ ] - [x] Acceptance tests have been run (no behavioral changes — logging only)
- [ ] - [x] Documentation has been updated if needed (N/A — internal refactor)

PR #3328, opened 2026-04-10
refactor: migrate data_source_github_branch logging to tflog

texasich:
Migrates `data_source_github_branch.go` from the stdlib `log` package to `tflog`, part of the work tracked in #3070.

The one debug log statement (404 branch-not-found path) gets structured key-value fields attached alongside the formatted message, matching the pattern from the already-migrated files (`resource_github_membership.go`, `resource_github_actions_environment_secret.go`).

**Changes:**
- Swap `"log"` import for `"fmt"` + `"github.com/hashicorp/terraform-plugin-log/tflog"`
- Convert `log.Printf("[DEBUG] ...")` ? `tflog.Debug(ctx, fmt.Sprintf(...), map[string]any{...})`

Tested: the function already uses `ReadContext`, so `ctx` is in scope. Import path verified against existing migrated files.

### DIFF
--- a/github/data_source_github_branch.go
+++ b/github/data_source_github_branch.go
@@ -3,10 +3,11 @@ package github
 import (
 	"context"
 	"errors"
-	"log"
+	"fmt"
 	"net/http"
 
 	"github.com/google/go-github/v84/github"
+	"github.com/hashicorp/terraform-plugin-log/tflog"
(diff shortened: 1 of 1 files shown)


texasich:
Same deal as #3329 and #3330 -- closing to respect the 1-open-PR limit for new contributors. #3325 is the one in review; will queue this back up after that one lands.

PR #3342, opened 2026-04-17
refactor: migrate resource_github_issue_label to context-aware CRUD a…

slymanmrcan:
Addresses parts of  #3070
Addresses parts of  #2996

----

### Before the change?

- `resource_github_issue_label` uses legacy `Create/Read/Update/Delete` 
  schema functions and `log.Printf` for logging.

### After the change?

- Migrated to `CreateContext/ReadContext/UpdateContext/DeleteContext`
- Function signatures updated to `(ctx context.Context, d *schema.ResourceData, meta any) diag.Diagnostics`
- Replaced `log.Printf` with `tflog.Info` for structured logging
- Removed manual `context.Background()` calls
- Wrapped error returns with `diag.FromErr()`

### Pull request checklist

- [x] Schema migrations have been created if needed — N/A, no schema change
- [x] Tests for the changes have been added — N/A, no behavior change
- [x] Docs have been reviewed and added / updated if needed — N/A

- [x] No

### DIFF
--- a/github/resource_github_issue_label.go
+++ b/github/resource_github_issue_label.go
@@ -3,19 +3,20 @@ package github
 import (
 	"context"
 	"errors"
-	"log"
 	"net/http"
 
 	"github.com/google/go-github/v84/github"
+	"github.com/hashicorp/terraform-plugin-log/tflog"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
 )
 
 func resourceGithubIssueLabel() *schema.Resource {
 	return &schema.Resource{
-		Create: resourceGithubIssueLabelCreateOrUpdate,
-		Read:   resourceGithubIssueLabelRead,
-		Update: resourceGithubIssueLabelCreateOrUpdate,
-		Delete: resourceGithubIssueLabelDelete,
+		CreateContext: resourceGithubIssueLabelCreateOrUpdate,
+		ReadContext:   resourceGithubIssueLabelRead,
+		UpdateContext: resourceGithubIssueLabelCreateOrUpdate,
+		DeleteContext: resourceGithubIssueLabelDelete,
 		Importer: &schema.ResourceImporter{
 			StateContext: schema.ImportStatePassthroughContext,
 		},
@@ -70,7 +71,7 @@ func resourceGithubIssueLabel() *schema.Resource {
 // otherwise it will create. This is also advantageous in that we get to use the
 // same function for two schema funcs.
 
-func resourceGithubIssueLabelCreateOrUpdate(d *schema.ResourceData, meta any) error {
+func resourceGithubIssueLabelCreateOrUpdate(ctx context.Context, d *schema.ResourceData, meta any) diag.Diagnostics {
 	client := meta.(*Owner).v3client
 	orgName := meta.(*Owner).name
 	repoName := d.Get("repository").(string)
@@ -81,7 +82,6 @@ func resourceGithubIssueLabelCreateOrUpdate(d *schema.ResourceData, meta any) er
 		Name:  new(name),
 		Color: new(color),
 	}
-	ctx := context.Background()
 	if !d.IsNewResource() {
 		ctx = context.WithValue(ctx, ctxId, d.Id())
 	}
@@ -95,14 +95,14 @@ func resourceGithubIssueLabelCreateOrUpdate(d *schema.ResourceData, meta any) er
 		var err error
 		_, originalName, err = parseTwoPartID(d.Id(), "repository", "name")
 		if err != nil {
-			return err
+			return diag.FromErr(err)
 		}
 	}
 
 	existing, resp, err := client.Issues.GetLabel(ctx,
 		orgName, repoName, originalName)
 	if err != nil && resp.StatusCode != http.StatusNotFound {
-		return err
+		return diag.FromErr(err)
 	}
 
 	if existing != nil {
@@ -117,14 +117,14 @@ func resourceGithubIssueLabelCreateOrUpdate(d *schema.ResourceData, meta any) er
 			var err error
 			_, originalName, err = parseTwoPartID(d.Id(), "repository", "name")
 			if err != nil {
-				return err
+				return diag.FromErr(err)
 			}
 		}
 
 		_, _, err := client.Issues.EditLabel(ctx,
 			orgName, repoName, originalName, label)
 		if err != nil {
-			return err
+			return diag.FromErr(err)
 		}
 	} else {
 		if v, ok := d.GetOk("description"); ok {
@@ -134,24 +134,23 @@ func resourceGithubIssueLabelCreateOrUpdate(d *schema.ResourceData, meta any) er
 		_, _, err := client.Issues.CreateLabel(ctx,
 			orgName, repoName, label)
 		if err != nil {
-			return err
+			return diag.FromErr(err)
 		}
 	}
 
 	d.SetId(buildTwoPartID(repoName, name))
 
-	return resourceGithubIssueLabelRead(d, meta)
+	return resourceGithubIssueLabelRead(ctx, d, meta)
 }
 
-func resourceGithubIssueLabelRead(d *schema.ResourceData, meta any) error {
+func resourceGithubIssueLabelRead(ctx context.Context, d *schema.ResourceData, meta any) diag.Diagnostics {
 	client := meta.(*Owner).v3client
 	repoName, name, err := parseTwoPartID(d.Id(), "repository", "name")
 	if err != nil {
-		return err
+		return diag.FromErr(err)
 	}
 
 	orgName := meta.(*Owner).name
-	ctx := context.WithValue(context.Background(), ctxId, d.Id())
 	if !d.IsNewResource() {
 		ctx = context.WithValue(ctx, ctxEtag, d.Get("etag").(string))
 	}
@@ -165,45 +164,48 @@ func resourceGithubIssueLabelRead(d *schema.ResourceData, meta any) error {
 				return nil
 			}
 			if ghErr.Response.StatusCode == http.StatusNotFound {
-				log.Printf("[INFO] Removing label %s (%s/%s) from state because it no longer exists in GitHub",
-					name, orgName, repoName)
+				tflog.Info(ctx, "Removing label from state because it no longer exists in GitHub", map[string]any{
+					"name":      name,
+					"org_name":  orgName,
+					"repo_name": repoName,
+				})
 				d.SetId("")
 				return nil
 			}
 		}
-		return err
+		return diag.FromErr(err)
 	}
 
 	if err = d.Set("etag", resp.Header.Get("ETag")); err != nil {
-		return err
+		return diag.FromErr(err)
 	}
 	if err = d.Set("repository", repoName); err != nil {
-		return err
+		return diag.FromErr(err)
 	}
 	if err = d.Set("name", name); err != nil {
-		return err
+		return diag.FromErr(err)
 	}
 	if err = d.Set("color", githubLabel.GetColor()); err != nil {
-		return err
+		return diag.FromErr(err)
 	}
 	if err = d.Set("description", githubLabel.GetDescription()); err != nil {
-		return err
+		return diag.FromErr(err)
 	}
 	if err = d.Set("url", githubLabel.GetURL()); err != nil {
-		return err
+		return diag.FromErr(err)
 	}
 
 	return nil
 }
 
-func resourceGithubIssueLabelDelete(d *schema.ResourceData, meta any) error {
+func resourceGithubIssueLabelDelete(ctx context.Context, d *schema.ResourceData, meta any) diag.Diagnostics {
 	client := meta.(*Owner).v3client
 
 	orgName := meta.(*Owner).name
 	repoName := d.Get("repository").(string)
 	name := d.Get("name").(string)
-	ctx := context.WithValue(context.Background(), ctxId, d.Id())
+	ctx = context.WithValue(ctx, ctxId, d.Id())
 
 	_, err := client.Issues.DeleteLabel(ctx, orgName, repoName, name)
-	return handleArchivedRepoDelete(err, "issue label", name, orgName, repoName)
+	return diag.FromErr(handleArchivedRepoDelete(err, "issue label", name, orgName, repoName))
 }
--- a/github/resource_github_issue_label.go
+++ b/github/resource_github_issue_label.go
@@ -3,6 +3,7 @@ package github
 import (
 	"context"
 	"errors"
+	"fmt"
 	"net/http"
 
 	"github.com/google/go-github/v88/github"
@@ -61,7 +62,6 @@ func resourceGithubIssueLabel() *schema.Resource {
 	}
 }
 
-// resourceGithubIssueLabelCreate creates an issue label.
 func resourceGithubIssueLabelCreate(ctx context.Context, d *schema.ResourceData, m any) diag.Diagnostics {
 	meta, _ := m.(*Owner)
 	client := meta.v3client
@@ -91,7 +91,8 @@ func resourceGithubIssueLabelCreate(ctx context.Context, d *schema.ResourceData,
 		}
 		label.Description = &description
 	}
-	githubLabel, resp, err := client.Issues.GetLabel(ctx, orgName, repoName, name)
+	var githubLabel *github.Label
+	_, resp, err := client.Issues.GetLabel(ctx, orgName, repoName, name)
 	if err != nil {
 		if resp == nil || resp.StatusCode != http.StatusNotFound {
 			return diag.FromErr(err)
@@ -247,7 +248,7 @@ func resourceGithubIssueLabelDelete(ctx context.Context, d *schema.ResourceData,
 func resourceGithubIssueLabelImport(_ context.Context, d *schema.ResourceData, _ any) ([]*schema.ResourceData, error) {
 	repoName, name, err := parseID2(d.Id())
 	if err != nil {
-		return nil, err
+		return nil, fmt.Errorf("invalid import ID %q; expected format %q: %w", d.Id(), "<repository name>:<label name>", err)
 	}
 
 	if err := d.Set("repository", repoName); err != nil {
--- a/github/resource_github_issue_label.go
+++ b/github/resource_github_issue_label.go
@@ -3,21 +3,22 @@ package github
 import (
 	"context"
 	"errors"
-	"log"
 	"net/http"
 
 	"github.com/google/go-github/v88/github"
+	"github.com/hashicorp/terraform-plugin-log/tflog"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
 )
 
 func resourceGithubIssueLabel() *schema.Resource {
 	return &schema.Resource{
-		Create: resourceGithubIssueLabelCreateOrUpdate,
-		Read:   resourceGithubIssueLabelRead,
-		Update: resourceGithubIssueLabelCreateOrUpdate,
-		Delete: resourceGithubIssueLabelDelete,
+		CreateContext: resourceGithubIssueLabelCreate,
+		ReadContext:   resourceGithubIssueLabelRead,
+		UpdateContext: resourceGithubIssueLabelUpdate,
+		DeleteContext: resourceGithubIssueLabelDelete,
 		Importer: &schema.ResourceImporter{
-			StateContext: schema.ImportStatePassthroughContext,
+			StateContext: resourceGithubIssueLabelImport,
 		},
 
 		Schema: map[string]*schema.Schema{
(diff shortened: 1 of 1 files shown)


deiga:
Please remove the resolves comment, since it adresses the issue only partly. And "resolves" will close the issue

deiga (approved):
LGTM!

Thanks for the contribution!

slymanmrcan:
I just removed it from the description

deiga:
Would you be available to rebase this (possibly multiple times) during the next week? We're going to release 6.13.0 and will try to land some of these open PRs

slymanmrcan:
Yes, I can rebase it during the next week as needed.

I’ll rebase now and address the Copilot review comments as well.

deiga (dismissed):
@stevehipwell even though there are other parts here that could be refactored, I would suggest we do the minimum here to get this merged.

stevehipwell (changes requested):
For me the bare minimum here would be to separate create & update and remove the call to read from them.

slymanmrcan:
Thanks, I separated the create and update handlers and removed the explicit read call from both paths while keeping the existing idempotent create behavior for default GitHub labels.

github/resource_github_issue_label.go
```diff
@@ -81,77 +80,68 @@ func resourceGithubIssueLabelCreateOrUpdate(d *schema.ResourceData, meta any) er
...
-	// parsed ID. If not, it's the value given to the resource.
-	var originalName string
-	if d.Id() == "" {
-		originalName = name
-	} else {
-		var err error
-		_, originalName, err = parseID2(d.Id())
-		if err != nil {
-			return err
+	existing, resp, err := client.Issues.GetLabel(ctx, orgName, repoName, name)
```
stevehipwell: This is the create function, it should error if there is already an existing resource.

github/resource_github_issue_label.go
```diff
@@ -81,77 +80,68 @@ func resourceGithubIssueLabelCreateOrUpdate(d *schema.ResourceData, meta any) er
...
-		if err != nil {
-			return err
-		}
+		_, _, err = client.Issues.CreateLabel(ctx, orgName, repoName, label)
+	}
+	if err != nil {
+		return diag.FromErr(err)
 	}
-
 	d.SetId(buildTwoPartID(repoName, name))
```
stevehipwell:
```suggestion
	id, err := buildID(repoName, name)
	if err != nil {
		return diag.FromErr(err)
	}
	d.SetId(id)
```

github/resource_github_issue_label.go
```diff
@@ -81,77 +80,68 @@ func resourceGithubIssueLabelCreateOrUpdate(d *schema.ResourceData, meta any) er
...
+	return nil
 
-	return resourceGithubIssueLabelRead(d, meta)
 }
+func resourceGithubIssueLabelUpdate(ctx context.Context, d *schema.ResourceData, meta any) diag.Diagnostics {
+	client := meta.(*Owner).v3client
+	orgName := meta.(*Owner).name
```
stevehipwell:
```suggestion
}

func resourceGithubIssueLabelUpdate(ctx context.Context, d *schema.ResourceData, m any) diag.Diagnostics {
	meta := m.(*Owner)
	client := meta.v3client
	orgName := meta.name
```

Also the functions should be in CRUD order.

github/resource_github_issue_label.go
```diff
@@ -3,19 +3,20 @@ package github
...
-		Create: resourceGithubIssueLabelCreateOrUpdate,
-		Read:   resourceGithubIssueLabelRead,
-		Update: resourceGithubIssueLabelCreateOrUpdate,
-		Delete: resourceGithubIssueLabelDelete,
+		CreateContext: resourceGithubIssueLabelCreate,
+		ReadContext:   resourceGithubIssueLabelRead,
+		UpdateContext: resourceGithubIssueLabelUpdate,
+		DeleteContext: resourceGithubIssueLabelDelete,
 		Importer: &schema.ResourceImporter{
 			StateContext: schema.ImportStatePassthroughContext,
```
stevehipwell: The updated pattern is to have an import function that correctly sets the whole state.

github/resource_github_issue_label.go
```diff
@@ -81,77 +80,68 @@ func resourceGithubIssueLabelCreateOrUpdate(d *schema.ResourceData, meta any) er
...
-			return err
-		}
+		_, _, err = client.Issues.CreateLabel(ctx, orgName, repoName, label)
+	}
+	if err != nil {
+		return diag.FromErr(err)
 	}
-
 	d.SetId(buildTwoPartID(repoName, name))
+	return nil
```
stevehipwell: It looks like you're missing the logic to update the state.

github/resource_github_issue_label.go
```diff
@@ -81,77 +80,68 @@ func resourceGithubIssueLabelCreateOrUpdate(d *schema.ResourceData, meta any) er
...
 	client := meta.(*Owner).v3client
 	repoName, name, err := parseID2(d.Id())
 	if err != nil {
-		return err
+		return diag.FromErr(err)
 	}
 
 	orgName := meta.(*Owner).name
-	ctx := context.WithValue(context.Background(), ctxId, d.Id())
+	ctx = context.WithValue(ctx, ctxId, d.Id())
```
stevehipwell:
```suggestion
```
This line is unnecessary.

stevehipwell (changes requested):

github/resource_github_issue_label.go
```diff
@@ -81,77 +74,32 @@ func resourceGithubIssueLabelCreateOrUpdate(d *schema.ResourceData, meta any) er
...
-
-	d.SetId(buildTwoPartID(repoName, name))
-
-	return resourceGithubIssueLabelRead(d, meta)
+	id, err := buildID(repoName, name)
+	if err != nil {
+		return diag.FromErr(err)
+	}
+	d.SetId(id)
+	return nil
```
stevehipwell, line 109: You need to set url and etag before returning.

github/resource_github_issue_label.go
```diff
@@ -81,77 +74,32 @@ func resourceGithubIssueLabelCreateOrUpdate(d *schema.ResourceData, meta any) er
...
-func resourceGithubIssueLabelRead(d *schema.ResourceData, meta any) error {
-	client := meta.(*Owner).v3client
+func resourceGithubIssueLabelRead(ctx context.Context, d *schema.ResourceData, m any) diag.Diagnostics {
+	meta := m.(*Owner)
+	client := meta.v3client
 	repoName, name, err := parseID2(d.Id())
 	if err != nil {
-		return err
+		return diag.FromErr(err)
 	}
```
stevehipwell: Please use the data and not the ID.

github/resource_github_issue_label.go
```diff
@@ -165,45 +113,103 @@ func resourceGithubIssueLabelRead(d *schema.ResourceData, meta any) error {
 				return nil
 			}
 			if ghErr.Response.StatusCode == http.StatusNotFound {
-				log.Printf("[INFO] Removing label %s (%s/%s) from state because it no longer exists in GitHub",
-					name, orgName, repoName)
+				tflog.Info(ctx, "Removing label from state because it no longer exists in GitHub", map[string]any{
```
stevehipwell: Please make these single line.

github/resource_github_issue_label.go
```diff
@@ -165,45 +113,103 @@ func resourceGithubIssueLabelRead(d *schema.ResourceData, meta any) error {
...
+	_, originalName, err := parseID2(d.Id())
+	if err != nil {
+		return diag.FromErr(err)
+	}
 
-	orgName := meta.(*Owner).name
+	label := &github.Label{
+		Name:        new(name),
+		Color:       new(color),
+		Description: new(d.Get("description").(string)),
```
stevehipwell: This doesn't match the pattern in create.

github/resource_github_issue_label.go
```diff
@@ -165,45 +113,103 @@ func resourceGithubIssueLabelRead(d *schema.ResourceData, meta any) error {
...
+	meta := m.(*Owner)
+	client := meta.v3client
+	orgName := meta.name
+	repoName := d.Get("repository").(string)
+	name := d.Get("name").(string)
+	color := d.Get("color").(string)
 
-func resourceGithubIssueLabelDelete(d *schema.ResourceData, meta any) error {
-	client := meta.(*Owner).v3client
+	_, originalName, err := parseID2(d.Id())
```
stevehipwell: Same as for read, please use the data not the ID.

github/resource_github_issue_label.go
```diff
@@ -165,45 +113,103 @@ func resourceGithubIssueLabelRead(d *schema.ResourceData, meta any) error {
...
+	d.SetId(id)
+	if err := d.Set("name", name); err != nil {
+		return diag.FromErr(err)
+	}
+	if err := d.Set("color", color); err != nil {
+		return diag.FromErr(err)
+	}
+	if err := d.Set("description", d.Get("description").(string)); err != nil {
+		return diag.FromErr(err)
+	}
```
stevehipwell: These aren't computed, so why are they being updated? But url and etag are and they're not being updated.

github/resource_github_issue_label.go
```diff
@@ -165,45 +113,103 @@ func resourceGithubIssueLabelRead(d *schema.ResourceData, meta any) error {
...
+	return nil
+}
+func resourceGithubIssueLabelDelete(ctx context.Context, d *schema.ResourceData, m any) diag.Diagnostics {
```
stevehipwell, line 217:
```suggestion
}

func resourceGithubIssueLabelDelete(ctx context.Context, d *schema.ResourceData, m any) diag.Diagnostics {
```

github/resource_github_issue_label.go
```diff
@@ -165,45 +113,103 @@ func resourceGithubIssueLabelRead(d *schema.ResourceData, meta any) error {
...
+	meta := m.(*Owner)
+	client := meta.v3client
+
+	orgName := meta.name
 	repoName := d.Get("repository").(string)
 	name := d.Get("name").(string)
-	ctx := context.WithValue(context.Background(), ctxId, d.Id())
+	ctx = context.WithValue(ctx, ctxId, d.Id())
```
stevehipwell:
```suggestion
```

stevehipwell (changes requested):

stevehipwell (changes requested):
@slymanmrcan this is looking good. There are still a couple of outstanding comments to address and it looks like you haven't run the strict linting for new code?

slymanmrcan:
Thanks, my mistake. I missed the strict new-code lint locally. 
I’ve run it now and addressed the findings in the latest push.

github/resource_github_issue_label.go
```diff
@@ -60,150 +61,220 @@ func resourceGithubIssueLabel() *schema.Resource {
...
-	orgName := meta.(*Owner).name
-	repoName := d.Get("repository").(string)
-	name := d.Get("name").(string)
-	color := d.Get("color").(string)
+// resourceGithubIssueLabelCreate creates an issue label.
+func resourceGithubIssueLabelCreate(ctx context.Context, d *schema.ResourceData, m any) diag.Diagnostics {
+	meta, diags := getIssueLabelOwner(m)
+	if diags.HasError() {
+		return diags
+	}
```
stevehipwell:
```suggestion
	meta, _ := m.(*Owner)
```

github/resource_github_issue_label.go
```diff
@@ -60,150 +61,220 @@ func resourceGithubIssueLabel() *schema.Resource {
...
+	client := meta.v3client
+	orgName := meta.name
+	repoName, diags := getIssueLabelStringAttribute(d, "repository")
+	if diags.HasError() {
+		return diags
+	}
```
stevehipwell:
```suggestion
	repoName, ok := d.Get("repository").(string); !ok {
		return diag.Errorf(`expected "repository" to be string`)
	}
```
This is much less readable than the equivalent inline implementation.

github/resource_github_issue_label.go
```diff
@@ -60,150 +61,220 @@ func resourceGithubIssueLabel() *schema.Resource {
...
+	name, diags := getIssueLabelStringAttribute(d, "name")
+	if diags.HasError() {
+		return diags
+	}
```
stevehipwell: Same as above.

github/resource_github_issue_label.go
```diff
@@ -60,150 +61,220 @@ func resourceGithubIssueLabel() *schema.Resource {
...
+	color, diags := getIssueLabelStringAttribute(d, "color")
+	if diags.HasError() {
+		return diags
+	}
```
stevehipwell: Same as above.

github/resource_github_issue_label.go
```diff
@@ -60,150 +61,220 @@ func resourceGithubIssueLabel() *schema.Resource {
...
+	return meta, nil
+}
+
+func getIssueLabelStringAttribute(d *schema.ResourceData, key string) (string, diag.Diagnostics) {
+	value, ok := d.Get(key).(string)
+	if !ok {
+		return "", diag.Errorf("expected %q to be string", key)
+	}
+	return value, nil
 }
```
stevehipwell:
```suggestion
```

stevehipwell (changes requested):
Thanks for making the changes, I've added some feedback so you can follow the idiomatic patterns.

stevehipwell:
Thanks @slymanmrcan, this looks good. Could you please rebase and let us know if you've run the acceptance tests for the resource locally?

stevehipwell (dismissed):
LGTM

deiga (dismissed):

github/resource_github_issue_label.go
```diff
@@ -60,150 +61,187 @@ func resourceGithubIssueLabel() *schema.Resource {
...
+		return diag.FromErr(err)
 	}
 
 	if err = d.Set("etag", resp.Header.Get("ETag")); err != nil {
-		return err
+		return diag.FromErr(err)
 	}
-	if err = d.Set("repository", repoName); err != nil {
-		return err
+	if err = d.Set("url", githubLabel.GetURL()); err != nil {
```
robert-crandall, line 154: One thing worth a look: the old `Read` also set `color` and `description` from the API response, and this version only sets `etag` and `url`. I think that regresses two things - drift detection (an external color/description change won't be noticed on refresh) and import (the new importer sets only `repository`/`name`, so after `terraform import` color/description come back empty and the next plan shows a diff). Could we add `d.Set("color", githubLabel.GetColor())` and `d.Set("description", githubLabel.GetDescription())` back here?

github/resource_github_issue_label.go
```diff
@@ -60,150 +61,187 @@ func resourceGithubIssueLabel() *schema.Resource {
...
+			return diag.Errorf(`expected "description" to be string`)
 		}
+		label.Description = &description
 	}
 
-	existing, resp, err := client.Issues.GetLabel(ctx,
-		orgName, repoName, originalName)
-	if err != nil && resp.StatusCode != http.StatusNotFound {
-		return err
+	githubLabel, resp, err := client.Issues.CreateLabel(ctx, orgName, repoName, label)
```
robert-crandall, line 92: Quick question on intent: the old combined function did a `GetLabel` first and called `EditLabel` if the label already existed (the removed comment explained it - new repos ship default labels, and a name clash returns 422). This version calls `CreateLabel` directly, so a pre-existing label would now 422 instead of being adopted. That may well be the better Terraform behavior (import instead of adopt) - just flagging since the checklist says "no behavior change." Totally fine if intentional.

robert-crandall (commented):
Thanks for taking this on - moving to context-aware CRUD and `tflog` is a nice cleanup, and I appreciate the careful function split. Both questions I left feel like behavior changes worth a second look. Happy to help think through it!

slymanmrcan:
Thanks for the review! I’ve pushed the fixes now: color/description are set from Read again, and Create preserves the previous GetLabel-then-EditLabel behavior for existing labels.

github/resource_github_issue_label.go
```diff
@@ -60,150 +61,202 @@ func resourceGithubIssueLabel() *schema.Resource {
...
-// otherwise it will create. This is also advantageous in that we get to use the
-// same function for two schema funcs.
-
-func resourceGithubIssueLabelCreateOrUpdate(d *schema.ResourceData, meta any) error {
-	client := meta.(*Owner).v3client
-	orgName := meta.(*Owner).name
-	repoName := d.Get("repository").(string)
-	name := d.Get("name").(string)
-	color := d.Get("color").(string)
+// resourceGithubIssueLabelCreate creates an issue label.
```
deiga: Let's not keep comments around if they don't explain something non-obvious
```suggestion
```

github/resource_github_issue_label.go
```diff
@@ -60,150 +61,202 @@ func resourceGithubIssueLabel() *schema.Resource {
...
+	}
 	_, err := client.Issues.DeleteLabel(ctx, orgName, repoName, name)
-	return handleArchivedRepoDelete(err, "issue label", name, orgName, repoName)
+	return diag.FromErr(handleArchivedRepoDelete(err, "issue label", name, orgName, repoName))
+}
+
+func resourceGithubIssueLabelImport(_ context.Context, d *schema.ResourceData, _ any) ([]*schema.ResourceData, error) {
+	repoName, name, err := parseID2(d.Id())
+	if err != nil {
+		return nil, err
```
deiga: To make the DX nicer, lets add a helpful error message here :)
```suggestion
		return nil, fmt.Errorf("Import failed due to ID format mismatch in %s. Please ensure import ID follows the pattern `<repository name>:<label name>.", d.Id())
```

deiga (changes requested):

github/resource_github_issue_label.go
stevehipwell: I think this should be Terraform based logic, so fail if the label already exists. We have `github_issue_labels` to act as an authoritative source.

deiga:
@slymanmrcan could you please rebase the branch?

github/resource_github_issue_label.go
```diff
@@ -60,150 +62,198 @@ func resourceGithubIssueLabel() *schema.Resource {
...
-			}
-		}
-
-		_, _, err := client.Issues.EditLabel(ctx,
-			orgName, repoName, originalName, label)
-		if err != nil {
-			return err
+	label.Description = &description
+	var githubLabel *github.Label
+	_, resp, err := client.Issues.GetLabel(ctx, orgName, repoName, name)
```
stevehipwell: The TF create function shouldn't be getting a remote resource as it should error if it already exists.

stevehipwell (changes requested):
As per my previous comment, this resource should fail on create if the label already exists. You can either import or use `github_isue_labels` to handle the existing label scenario.

slymanmrcan:
thanks for review
@stevehipwell @deiga 
- removed GetLabel/EditLabel adopt logic from create, now fails if label already exists
- fixed description to use d.Get instead of d.GetOk to allow clearing it

github/resource_github_issue_label.go
```diff
@@ -60,150 +62,189 @@ func resourceGithubIssueLabel() *schema.Resource {
...
 	if !d.IsNewResource() {
 		ctx = context.WithValue(ctx, ctxEtag, d.Get("etag").(string))
 	}
 
-	githubLabel, resp, err := client.Issues.GetLabel(ctx,
-		orgName, repoName, name)
+	githubLabel, resp, err := client.Issues.GetLabel(ctx, orgName, repoName, name)
 	if err != nil {
 		var ghErr *github.ErrorResponse
 		if errors.As(err, &ghErr) {
```
stevehipwell: Could you please swap this out to `errors.AsType` (it's documented in the contribution guide if you want an example).

github/resource_github_issue_label.go
```diff
@@ -60,150 +62,189 @@ func resourceGithubIssueLabel() *schema.Resource {
...
+	name, ok := d.Get("name").(string)
+	if !ok {
+		return diag.Errorf(`expected "name" to be string`)
+	}
 	_, err := client.Issues.DeleteLabel(ctx, orgName, repoName, name)
-	return handleArchivedRepoDelete(err, "issue label", name, orgName, repoName)
+	return diag.FromErr(handleArchivedRepoDelete(err, "issue label", name, orgName, repoName))
+}
+
+func resourceGithubIssueLabelImport(_ context.Context, d *schema.ResourceData, _ any) ([]*schema.ResourceData, error) {
```
stevehipwell: I think you need to get the label here and update all of the fields as `color` is required so we shouldn't be writing to state without it set.

github/resource_github_issue_label.go
```diff
@@ -60,150 +62,189 @@ func resourceGithubIssueLabel() *schema.Resource {
...
+	if !ok {
+		return diag.Errorf(`expected "repository" to be string`)
+	}
+	name, ok := d.Get("name").(string)
+	if !ok {
+		return diag.Errorf(`expected "name" to be string`)
+	}
 	_, err := client.Issues.DeleteLabel(ctx, orgName, repoName, name)
-	return handleArchivedRepoDelete(err, "issue label", name, orgName, repoName)
+	return diag.FromErr(handleArchivedRepoDelete(err, "issue label", name, orgName, repoName))
```
stevehipwell: Please remove this helper function and explicitly check for not found (currently missing) and forbidden in the code.

stevehipwell (changes requested):

stevehipwell:
@slymanmrcan could you please fix the lint failures?

slymanmrcan:
Thanks, fixing it now.

stevehipwell:
@slymanmrcan you appear to have been caught out by a regression.

@deiga any thoughts?

deiga (approved):

stevehipwell (approved):
LGTM

robert-crandall (approved):
Thanks for this! The migration to context-aware CRUD with `diag` and `tflog` is clean and consistent, and I really appreciate how quickly you turned around the feedback. The `color`/`description` read-back restores drift detection, and setting the description unconditionally so it can be cleared is exactly right. Nice, well-scoped change. 🙏

Issue #2892, opened 2025-11-13
[MAINT]: Remove read call from create/update


stevehipwell:
### Describe the need

This provider make heavy use of calling the read function instead of duplicating logic in the create/update functions; while this may look clean and DRY it has a significant impact. Firstly it means that changes generally use twice as many API calls to make as they should (or more in cases where a single resource is mapped to multiple API calls); which is impactful given the low API rate limits imposed by GitHub. Secondly it means that systems which don't have strict read after write guarantees may be returning incorrect values to be stored in the state.

### Relevant log output

```shell

```

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

PR #3349, opened 2026-04-19
feat: Add github_repository_pull_request_creation_policy resource

### LINKED ISSUE #3348: [FEAT]: support pull_request_creation_policy on github_repository
## Describe the need

The provider currently has no support for the repository-level `pull_request_creation_policy` setting, which controls who can create pull requests in a repository.

I maintain a separate `github-infra` repo that manages my GitHub repositories through this provider. Because this setting is missing, that repo currently has to patch it out-of-band with a `curl` call against the GitHub API during Terraform apply:

- desired Terraform surface: `github_repository.pull_request_creation_policy`
- values currently in use: `collaborators_only` and `all`
- current workaround: a local `terraform_data` + `local-exec` `curl -X PATCH /repos/{owner}/{repo}` call in the infra module

This means the repository resource cannot fully express the repo settings being managed, and the workaround has to exist outside normal provider state handling.

## SDK Version

terraform-provider-github (current main branch)

## API Version

GitHub REST API 2022-11-28 in the current workaround; GitHub GraphQL also exposes `pullRequestCreationPolicy` and update support.

## Relevant log output

Current workaround in `github-infra`:

```shell
curl -s -f \
  -X PATCH \
  -H "Authorization: Bearer $GITHUB_TOKEN" \
  -H "Accept: application/vnd.github+json" \
  -H "X-GitHub-Api-Version: 2022-11-28" \
  "https://api.github.com/repos/${var.owner}/${var.name}" \
  -d '{"pull_request_creation_policy":"${var.pull_request_creation_policy}"}'
```

## Code of Conduct

RoFz:
Resolves #3348

----

### Before the change?

- `github_repository` could not manage `pull_request_creation_policy`.
- Users who needed to enforce it had to rely on out-of-band API calls or local workarounds.

### After the change?

- Add a new `github_repository_pull_request_creation_policy` resource that manages the pull request creation policy for a repository via the GitHub GraphQL API.
- Support the `all` and `collaborators_only` values.
- The resource reads and updates the setting through the GitHub GraphQL API and supports import by repository name.

### Why a separate resource?

`github_repository` uses only the REST API. Because `pullRequestCreationPolicy` is only exposed via GraphQL, this change follows the existing pattern used by other GraphQL-backed resources in this provider (e.g. `github_team_settings`, `github_branch_protection`) and introduces a dedicated resource rather than mixing transport layers in `github_repository`.

### Evidence

- `go test ./github -run 'TestPullRequestCreationPolicyMapping|TestRepositoryPullRequestCreationPolicyGraphQL' -count=1`
- `make test`
- `GH_TEST_AUTH_MODE=individual GITHUB_OWNER=RoFz GITHUB_USERNAME=RoFz make testacc T='TestAccGithubRepositoryPullRequestCreationPolicy'`
- `make lintcheck`
- Manual test with a local provider override on disposable repo `tf-pr-creation-policy-manual-20260419-120606`: `terraform apply` succeeded, GitHub API readback returned `collaborators_only`, and a follow-up `terraform plan -detailed-exitcode` returned no changes.

### Pull request checklist

- [ ] Schema migrations have been created if needed

- [ ] Yes
- [x] No

----


deiga (changes requested):
The github_repository resource relies solely on the REST API currently. If you need to use GQL API, then please do that in a separate resource.

RoFz (commented):
Thanks for the review. Reworked to a dedicated `github_repository_pull_request_creation_policy` resource that uses only the GraphQL API, keeping `github_repository` REST-only.

github/resource_github_repository_pull_request_creation_policy.go
```diff
@@ -0,0 +1,120 @@
...
+		CreateContext: resourceGithubRepositoryPullRequestCreationPolicyCreate,
+		ReadContext:   resourceGithubRepositoryPullRequestCreationPolicyRead,
+		UpdateContext: resourceGithubRepositoryPullRequestCreationPolicyUpdate,
+		DeleteContext: resourceGithubRepositoryPullRequestCreationPolicyDelete,
+		Importer: &schema.ResourceImporter{
+			StateContext: resourceGithubRepositoryPullRequestCreationPolicyImport,
+		},
+
+		Schema: map[string]*schema.Schema{
+			"repository": {
```
deiga, line 25: Please add also repository_id and the diffRepo CustomizeDiff configuration to support renaming repositories

github/resource_github_repository_pull_request_creation_policy.go
```diff
@@ -0,0 +1,120 @@
...
+	if err != nil {
+		return diag.Errorf("error resolving repository node ID for %s: %s", repoName, err)
+	}
+
+	if err := updateRepositoryPullRequestCreationPolicy(ctx, nodeID, policy, meta); err != nil {
+		return diag.Errorf("error setting pull request creation policy for %s: %s", repoName, err)
+	}
+
+	d.SetId(repoName)
+	return resourceGithubRepositoryPullRequestCreationPolicyRead(ctx, d, meta)
```
deiga: Never call any of the CRUD functions directly. If you need to set Computed fields, set them in the same function

github/resource_github_repository_pull_request_creation_policy.go
```diff
@@ -0,0 +1,120 @@
...
+}
+
+func resourceGithubRepositoryPullRequestCreationPolicyRead(ctx context.Context, d *schema.ResourceData, meta any) diag.Diagnostics {
+	owner := meta.(*Owner).name
+	repoName := d.Id()
+
+	policy, err := getRepositoryPullRequestCreationPolicy(ctx, owner, repoName, meta)
+	if err != nil {
+		return diag.Errorf("error reading pull request creation policy for %s/%s: %s", owner, repoName, err)
+	}
```
deiga, line 82: Please differentiate between an actual error and for example the "repo doesn't exist anymore" case where this resource can be removed from state directly

github/resource_github_repository_pull_request_creation_policy.go
```diff
@@ -0,0 +1,120 @@
...
+
+	return nil
+}
+
+func resourceGithubRepositoryPullRequestCreationPolicyUpdate(ctx context.Context, d *schema.ResourceData, meta any) diag.Diagnostics {
+	repoName := d.Id()
+	policy := d.Get("policy").(string)
+
+	nodeID, err := getRepositoryID(repoName, meta)
```
deiga: Instead of wasting API calls on fetching the repo Node ID in each CRUD function, could we add a computed field to the schema to store it?

github/resource_github_repository_pull_request_creation_policy.go
```diff
@@ -0,0 +1,120 @@
...
+	repoName := d.Id()
+
+	nodeID, err := getRepositoryID(repoName, meta)
+	if err != nil {
+		return diag.Errorf("error resolving repository node ID for %s: %s", repoName, err)
+	}
+
+	if err := updateRepositoryPullRequestCreationPolicy(ctx, nodeID, "all", meta); err != nil {
+		return diag.Errorf("error resetting pull request creation policy for %s: %s", repoName, err)
+	}
```
deiga, line 135: Please check if it's an actual error before returning an error. Is the repo gone? Not an actual error, delete resource from state and be happy.

github/resource_github_repository_pull_request_creation_policy.go
```diff
@@ -0,0 +1,120 @@
...
+	repoName := d.Id()
+
+	if err := d.Set("repository", repoName); err != nil {
+		return nil, err
+	}
+
+	diags := resourceGithubRepositoryPullRequestCreationPolicyRead(ctx, d, meta)
+	if diags.HasError() {
+		return nil, fmt.Errorf("%s", diags[0].Summary)
+	}
```
deiga: There is no need to call Read inside Import, the lifecycle works so that the next call will be Read anyway

github/resource_github_repository_pull_request_creation_policy_test.go
```diff
@@ -0,0 +1,100 @@
...
+	"github.com/hashicorp/terraform-plugin-testing/helper/acctest"
+	"github.com/hashicorp/terraform-plugin-testing/helper/resource"
+)
+
+func TestAccGithubRepositoryPullRequestCreationPolicy(t *testing.T) {
+	t.Run("sets policy without error", func(t *testing.T) {
+		randomID := acctest.RandStringFromCharSet(5, acctest.CharSetAlphaNum)
+		repoName := fmt.Sprintf("%srepo-pr-policy-%s", testResourcePrefix, randomID)
+		initial := `policy = "collaborators_only"`
+		updated := `policy = "all"`
```
deiga: It's clearer if you use the variables just for the policy values and not the whole property and value pair

github/resource_github_repository_pull_request_creation_policy_test.go
```diff
@@ -0,0 +1,100 @@
...
+				auto_init  = true
+			}
+
+			resource "github_repository_pull_request_creation_policy" "test" {
+				repository = github_repository.test.name
+				%%s
+			}
+		`, repoName)
+
+		checks := map[string]resource.TestCheckFunc{
```
deiga: Don't use variables for this, inline it

github/resource_github_repository_pull_request_creation_policy_test.go
```diff
@@ -0,0 +1,100 @@
...
+			),
+		}
+
+		resource.Test(t, resource.TestCase{
+			PreCheck:          func() { skipUnauthenticated(t) },
+			ProviderFactories: providerFactories,
+			Steps: []resource.TestStep{
+				{
+					Config: fmt.Sprintf(config, initial),
+					Check:  checks["before"],
```
deiga: Please use ConfigStateChecks instead

RoFz:
Merged the latest `main` to resolve the conflicts that had this PR marked as conflicting. Two conflicts came up, both from changes that landed on `main` after this branch was opened:

**1. `github/provider.go`**

`main` reformatted this file and added several new resources/data sources. The only change from this branch is the single registration line for `github_repository_pull_request_creation_policy`, which I re-applied on top of the updated file.

**2. Documentation layout migration**

`main` migrated the provider docs from the legacy `website/docs/` tree to the `tfplugindocs`-generated `docs/` layout, so the hand-written `website/docs/r/repository_pull_request_creation_policy.html.markdown` from this branch conflicted against a directory that no longer exists. To fit the new workflow I:

- removed the old hand-written page,
- added `templates/resources/repository_pull_request_creation_policy.md.tmpl` (narrative, example reference, and import section),
- added `examples/resources/repository_pull_request_creation_policy/example_1.tf`,
- regenerated `docs/resources/repository_pull_request_creation_policy.md` with `make generatedocs`.

The argument reference rows come from the resource's schema `Description` fields, so they stay the source of truth. I dropped a sentence about the GraphQL transport that was in the original page, since that is design rationale for reviewers rather than something a Registry reader needs.

**Verification**

- `go build ./...` and `go vet ./github/` clean
- `make test` (full unit suite) passing, including the provider schema validation
- `golangci-lint run ./github/`: 0 issues
- `make validatedocs` clean, and a `make generatedocs` re-run produces no diff, so the docs CI check should pass

RoFz:
Follow-up to the conflict resolution above. While re-reviewing the resource I also pushed a few robustness and docs improvements (commit b2f15e8e). These were not requested in the earlier review, so if you'd rather keep this PR limited to the new resource and the conflict fix, I'm happy to pull them into a separate follow-up PR.

What changed:

- **Drift handling:** `Read` now removes the resource from state when the repository no longer exists, so a repo deleted out of band recreates on the next apply instead of hard-failing the plan. This matches the existing pattern in `github_branch_protection`.
- **Empty-policy read:** an empty `pullRequestCreationPolicy` from the API now returns an explicit error instead of silently setting `policy = ""`, which would otherwise produce a perpetual diff against the `Required` schema.
- **Docs:** noted that destroying the resource resets the policy to `all` (it does not delete anything on GitHub).

On the not-found detection: the v4 client (`shurcooL/graphql`) discards the structured `type: NOT_FOUND` extension and only surfaces the error message, so this matches on the message text, the same approach the other GraphQL-backed resources use. I extracted it into a documented `isRepositoryNotFoundError` helper and added a test that feeds GitHub's actual not-found payload through the real client, so the assumption is pinned rather than implicit.

Verification: `make test`, `go vet`, `golangci-lint run ./github/` (0 issues), and `make generatedocs` with no drift.

deiga (changes requested):
Partial review

github/resource_github_repository_pull_request_creation_policy.go
RoFz: Done in b2f15e8e: `Read` now detects the repository-not-found GraphQL error and removes the resource from state instead of failing. The detection lives in an `isRepositoryNotFoundError` helper with a unit test that pins GitHub's actual not-found payload, and the log uses structured `tflog`.

github/resource_github_repository_pull_request_creation_policy.go
RoFz: Done in 890cab59: `Delete` treats a missing repository as success. The guard sits on the by-name repository resolution, since that is where a missing repo surfaces with the error string the helper verifies. I did not guard the subsequent node-ID mutation: GitHub returns a different error shape there ("Could not resolve to a node with the global id"), it is only reachable in a delete-between-calls race, and a retry converges through the by-name guard anyway.

github/resource_github_repository_pull_request_creation_policy.go
RoFz: Done in 890cab59: replaced the custom importer with `schema.ImportStatePassthroughContext`. `Read` derives `repository` from the ID, so the custom importer added nothing, and `ImportStateVerify` passes against the live API.

github/resource_github_repository_pull_request_creation_policy.go
RoFz: Done in 33377d33: Create and Update now return nil directly. One observation for context: the codebase currently has around 80 Create/Update implementations that end by calling Read (including `github_branch_protection`, which this resource was modeled on), while ARCHITECTURE.md and the schema-and-state instructions document the no-read-after-write convention from #2892. This resource now follows the documented convention, and the legacy call sites may be worth a tracking issue if a migration is planned. Computed fields are set in Create and Update directly (see the `repository_id` change for rename support).

github/resource_github_repository_pull_request_creation_policy.go
RoFz: Done in 97d8104b: removed `ForceNew` from `repository`, added a computed `repository_id`, and set `CustomizeDiff: diffRepository`, per the "Repository as a Required Argument" convention. An acceptance subtest renames the repository and asserts the resource is updated in place rather than replaced, with the policy preserved.

github/resource_github_repository_pull_request_creation_policy.go
RoFz: Addressed in 97d8104b. Create and Update now resolve the repository through a single by-name GraphQL query that returns both the node ID and the numeric ID, replacing the previous helper that probed the input as a node ID twice before falling back to the name. One note for transparency: the stored `repository_id` is the numeric ID required by `diffRepository`, not the node ID the mutation needs, so each write still resolves the node ID, but now in one query instead of three calls.

github/resource_github_repository_pull_request_creation_policy_test.go
RoFz: Done in f9a3ed3c. Variables now hold just the policy values; the `policy` attribute lives in the config template.

github/resource_github_repository_pull_request_creation_policy_test.go
RoFz: Done in f9a3ed3c. The checks map is gone; each step carries its assertions inline.

github/resource_github_repository_pull_request_creation_policy_test.go
RoFz: Done in f9a3ed3c. Both subtests now use `ConfigStateChecks` with `knownvalue` matchers, and the suite (including create/update, import, and the rename scenario) passes against the live API.

github/resource_github_repository_pull_request_creation_policy.go
```diff
@@ -0,0 +1,126 @@
...
+	if err != nil {
+		return diag.Errorf("error resolving repository node ID for %s: %s", repoName, err)
+	}
+
+	if err := updateRepositoryPullRequestCreationPolicy(ctx, nodeID, "all", meta); err != nil {
+		return diag.Errorf("error resetting pull request creation policy for %s: %s", repoName, err)
+	}
+
+	return nil
+}
```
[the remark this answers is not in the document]
RoFz: Resolved in 890cab59. `Delete` returns nil when the repository no longer exists, guarding on the by-name resolution that matches the error path described here.

github/resource_github_repository_pull_request_creation_policy.go
```diff
@@ -0,0 +1,126 @@
...
+				Description: "The name of the GitHub repository.",
+			},
+			"policy": {
+				Type:             schema.TypeString,
+				Required:         true,
+				Description:      "Controls who can create pull requests for the repository. Can be `all` or `collaborators_only`.",
+				ValidateDiagFunc: validation.ToDiagFunc(validation.StringInSlice([]string{"all", "collaborators_only"}, false)),
+			},
+		},
+	}
```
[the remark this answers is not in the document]
RoFz: Resolved in 97d8104b. Added the computed `repository_id`, `CustomizeDiff: diffRepository`, and removed `ForceNew`, with an acceptance test covering the rename-in-place behavior.

usmonster:
Is this superseded by #3479?

ap-1:
> Is this superseded by #3479?

yes, that one uses the SDK

PR #3247, opened 2026-03-02
fix: correct typo in retry_delay_ms configuration

laughedelic:
I think this is just a typo or a copy-paste bug.

----

### Before the change?

- `read_delay_ms` is read into `retryDelay` instead of the corresponding `rerty_delay_ms`. 

### After the change?

- `retryDelay` is set to the `rerty_delay_ms` argument

### Pull request checklist

- [x] Schema migrations have been created if needed ([example](https://github.com/F-Secure-web/terraform-provider-github/blob/main/github/migrate_github_actions_organization_secret.go))
- [x] Tests for the changes have been added (for bug fixes / features)
- [x] Docs have been reviewed and added / updated if needed (for bug fixes / features)

### Does this introduce a breaking change?

- [x] No

----

PR #3352, opened 2026-04-19
fix: retry_delay_ms reads from wrong config key (read_delay_ms)

slymanmrcan:
problem 
`retry_delay_ms` is being read from the wrong config key.
Currently it reads from `read_delay_ms` instead of `retry_delay_ms`.

This means the `retry_delay_ms` configuration is silently ignored
and always takes the value of `read_delay_ms`.
## Fix
Changed `d.Get("read_delay_ms")` to `d.Get("retry_delay_ms")`
for the `retryDelay` variable.


deiga:
Duplicate of #3247

slymanmrcan:
Sorry for the duplicate, closing this one

PR #3371, opened 2026-04-24
fix: guard nil invitation from AddCollaborator for org members

bofus10:
Resolves #3370

----

### Before the change?

- `github_repository_collaborators` panics with a nil pointer dereference when creating or updating a resource that lists an organization member who is not yet a direct collaborator on the target repository.
- GitHub returns `204 No Content` from `PUT /repos/{owner}/{repo}/collaborators/{username}` on that path (the user is promoted to direct collaborator immediately, no invitation is issued). `client.Repositories.AddCollaborator` therefore returns `inv == nil, err == nil`.
- Since #3233 the code at `github/resource_github_repository_collaborators.go:694` unconditionally dereferences `inv.ID`, producing `SIGSEGV` inside `updateUserCollaboratorsAndInvites`.

Stack trace (trimmed):

```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0xde9469]

goroutine 77 [running]:
github.com/integrations/terraform-provider-github/v6/github.updateUserCollaboratorsAndInvites(...)
	github/resource_github_repository_collaborators.go:694 +0xe09
github.com/integrations/terraform-provider-github/v6/github.resourceGithubRepositoryCollaboratorsCreate(...)
	github/resource_github_repository_collaborators.go:199 +0x3f1
```

### After the change?

- The invitation-tracking branch is guarded on `inv != nil`. On the 204 path we skip recording an invitation ID because none was issued; the user is already a direct collaborator and will be picked up on the next refresh via `listUserCollaborators`.
- Same code path that was already safe for permission updates at line 648 (which already discards the `AddCollaborator` return value) is now safe for the invite-new-user path too.

Documentation reference for the 204 return:
https://docs.github.com/rest/collaborators/collaborators#add-a-repository-collaborator

### Pull request checklist

- [ ] Schema migrations have been created if needed ([example](https://github.com/F-Secure-web/terraform-provider-github/blob/main/github/migrate_github_actions_organization_secret.go))
- [ ] Tests for the changes have been added (for bug fixes / features)
- [x] Docs have been reviewed and added / updated if needed (for bug fixes / features)

Notes on the checklist:

- No schema changes — purely a runtime nil guard.
[...]

PR #3373, opened 2026-04-25
[BUG] Fix drift with external collaborators invite

deiga:
----

### Before the change?

- Adding an external collaborator would cause constant drift for as long as the invite would not have been accepted

### After the change?

- Adding an external collaborator no longer causes drift in the `user` state

### Pull request checklist

- [ ] Yes
- [x] No

----

### DIFF
--- a/github/resource_github_repository_collaborators_test.go
+++ b/github/resource_github_repository_collaborators_test.go
@@ -57,7 +57,7 @@ resource "github_repository_collaborators" "test" {
 }
 `, repoName, teamName0, teamName1, collaboratorUser)
 
-		resource.ParallelTest(t, resource.TestCase{
+		resource.Test(t, resource.TestCase{
 			PreCheck:          func() { skipUnlessHasOrgs(t) },
 			ProviderFactories: providerFactories,
 			Steps: []resource.TestStep{
@@ -118,7 +118,7 @@ resource "github_repository_collaborators" "test" {
 }
 `, repoName, testAccConf.testExternalUser)
 
-		resource.ParallelTest(t, resource.TestCase{
+		resource.Test(t, resource.TestCase{
 			PreCheck:          func() { skipUnauthenticated(t) },
 			ProviderFactories: providerFactories,
 			Steps: []resource.TestStep{
@@ -196,7 +196,7 @@ resource "github_repository_collaborators" "test" {
 }
 `, configPre)
 
-		resource.ParallelTest(t, resource.TestCase{
+		resource.Test(t, resource.TestCase{
 			PreCheck:          func() { skipUnlessHasOrgs(t) },
 			ProviderFactories: providerFactories,
 			Steps: []resource.TestStep{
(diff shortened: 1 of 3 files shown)


robert-crandall (dismissed):

deiga:
@stevehipwell Could you check this PR so that we can get 6.12.1 release cut?

github/resource_github_repository_collaborators_test.go
```diff
@@ -57,7 +57,7 @@ resource "github_repository_collaborators" "test" {
 }
 `, repoName, teamName0, teamName1, collaboratorUser)
 
-		resource.Test(t, resource.TestCase{
+		resource.ParallelTest(t, resource.TestCase{
```
stevehipwell: Is there a specific reason for this change? I'd be concerned that this would hit the secondary rate limit as it currently stands. My suggestion for test improvements would be to remove the setup resources from TF and ue the SDK directly and then make the tests parallel. We should also look to remove the dependency on any global resources.

github/util_user.go
```diff
@@ -21,10 +21,6 @@ func (u userCollaborator) flatten() any {
 		"permission": u.permission,
 	}
 
-	if u.invitationID != nil {
-		m["invitation_id"] = *u.invitationID
-	}
```
stevehipwell, line 26: Was this just not being used? My original plan for the changes was to add invitation ID to the user and then do the merge as you did above, with the separate invitations deprecated.

stevehipwell (changes requested):
Thanks for the fix @deiga (and getting #3371 merged), I'm not sure why this code looks like it does but it looks like I messed up the end of the work on this. I've just got a couple of questions, but the code change looks good.

github/resource_github_repository_collaborators_test.go
deiga: The Main reason is: tests go faster. 
But we should maybe default to parallel=2 or 3 so that secondary rate limits don't become a problem.

github/util_user.go
deiga: This caused a mapping mismatch since invitation ID isn't an allowed field in one of the transformations (can't remember where)

And since it wasn't utilised I decided to remove it. 
Instead it possibly could be removed in the transformation step if we want to keep it

deiga:
@stevehipwell thanks. Responded to the comments. Let me know if I should make changes to those

github/resource_github_repository_collaborators_test.go
stevehipwell: Personally I'd hold off on making tests parallel, especially in a patch PR. I had loads of issues with parallel tests both due to the rate limits and the current client design.

github/util_user.go
stevehipwell: Yeah, I saw it in the code but not the schema and was wondering why it was there. I still can't figure out why this code is halfway between the two patterns I tested.

stevehipwell (approved):
LGTM

robert-crandall (approved):

PR #3376, opened 2026-04-27
feat: add github_actions_runner_group_repository_access resource

### LINKED ISSUE #3375: [FEAT]: Allow configuring runner group repository access from multiple resources independently
### Describe the need

Currently the only way to control which repositories have access to a specific runner group is via the central runner group resource `github_actions_runner_group`/`github_enterprise_actions_runner_group`, which manages the list of all repositories that have access to the runner group. 
In many cases, especially surrounding building Internal Development Platforms which create managed repositories as needed, it is not desirable to centralize the list of repositories in a central state and rather allow independent, potentially concurrent deployment processes to add/remove a specific repository.

The relevant API Endpoints to add/remove single repos [have existed since at least the first API Version `2022-11-28`](https://raw.githubusercontent.com/github/rest-api-description/refs/heads/main/descriptions-next/ghes-3.20/ghes-3.20.2022-11-28.yaml#:~:text=operationId%3A%20actions%2Fadd%2Drepo%2Daccess%2Dto%2Dself%2Dhosted%2Drunner%2Dgroup%2Din%2Dorg) and should therefore be available in all versions of the octokit sdk.

Example of how the new resource could look like
```hcl
resource "github_actions_runner_group_repository_access" "example" {
  runner_group_id = 12345
  repository_id = 3256
}
```

Note that apparently [the enterprise runner group resource returns its id](https://registry.terraform.io/providers/integrations/github/latest/docs/resources/enterprise_actions_runner_group#id-8) but [the org runner group resource doesn't](https://registry.terraform.io/providers/integrations/github/latest/docs/resources/actions_runner_group#attributes-reference) although this is not a blocker, but a separate DX problem, since the id can still be extracted from the urls returned in the attributes
Edit: This seems to be an oversight in the manually created docs, the id attribute is accessible

### SDK Version

_No response_

### API Version

_No response_

### Relevant log output

_n/a_

### Code of Conduct

BronzeDeer:
Resolves #3375

Currently the only way to control which repositories have access to a specific runner group is via the central runner group resource `github_actions_runner_group`, which manages the list of all repositories that have access to the runner group. In many cases, especially surrounding building Internal Development Platforms which create managed repositories as needed it is not desirable to centralize the list of repositories in a central state and rather allow independent, potentially concurrent deployment processes to add/remove a specific repository.

This commit adds a new resource which allows to configure this access on a single repo basis without needing to track the state of all authorized repositories for a runner group

----

### Before the change?

- Unable to manage runner group repo access from multiple resources/states 

### After the change?

- new resource added to fulfill that usecase

### Pull request checklist

- [ ] Yes
- [x] No

----

### DIFF
--- a/docs/resources/actions_runner_group_repository_access.html.markdown
+++ b/docs/resources/actions_runner_group_repository_access.html.markdown
@@ -1,0 +1,51 @@
+---
+layout: "github"
+page_title: "GitHub: github_actions_runner_group_repository_access"
+description: |-
+  Manages a repository's access to an Actions Runner Group within a GitHub organization
+---
+
+# github_actions_runner_group_repository_access
+
+This resource allows you to manage repository access to GitHub Actions runner groups within your GitHub (enterprise) organizations independently for each repository.
+You must have runner group admin access to an organization to use this resource.
+
+~> **Note:** The action runners group's `visibility` must be `selected` and if also managing the runner group via terraform: `selected_repository_ids` must **not** be set.
+
+## Example Usage
+
+```hcl
+resource "github_repository" "example" {
+  name = "my-repository"
+}
+
+resource "github_actions_runner_group" "example" {
+  name                    = github_repository.example.name
+  visibility              = "selected"
+}
+
+resource "github_actions_runner_group_repository_access" "example" {
+  runner_group_id = github_actions_runner_group.id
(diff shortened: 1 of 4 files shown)


BronzeDeer:
Rebased on origin/HEAD and fixed merge conflict

PR #3383, opened 2026-05-01
fix: respect context cancellation in RetryTransport

yoshi-taka:
- Replace `time.Sleep` with context-aware `sleep()` in `RetryTransport.RoundTrip`
- Skip the delay after the final retry attempt (`retry < maxRetries`)

`RateLimitTransport` already uses the existing `sleep(ctx, dur)` helper (transport.go:122), which respects context cancellation. `RetryTransport` was using bare `time.Sleep`, causing it to ignore `ctx.Done()` during retry delays.

This aligns `RetryTransport` with the existing pattern in the codebase.

- Add `TestRetryTransport_cancelled`: verifies context cancellation interrupts retry delay
- Add `TestRetryTransport_no_sleep_after_last_retry`: verifies no delay after final attempt

Resolves #ISSUE_NUMBER

----

### Before the change?

- `RetryTransport` used `time.Sleep` between retry attempts.
- During retry delays, request context cancellation was not respected.
- `RetryTransport` also slept after the final retryable attempt, even when no further retry would be made.

### After the change?

- `RetryTransport` now uses the existing context-aware `sleep(ctx, dur)` helper.
- Retry delays can now be interrupted when the request context is cancelled.
- `RetryTransport` no longer sleeps after the final retryable attempt.

### Pull request checklist

- [x] Schema migrations have been created if needed (not needed)

- [ ] Yes
- [x] No


deiga:
Please rebase this branch ☺️

deiga (dismissed):
LGTM

stevehipwell (changes requested):
@yoshi-taka please could you look at the Copilot review comments.

yoshi-taka:
Addressed the Copilot review comments in 280d6c6d:
- Stop retrying immediately when the request context is cancelled.
- Use test-local HTTP clients instead of mutating http.DefaultClient.
- Fix the final-attempt test so it only verifies that no sleep occurs after the last attempt.

The RetryTransport and related transport tests pass. Could you please take another look?

stevehipwell (changes requested):
@yoshi-taka please address the Copilot review comments. If you want a reference for how this should work you can check out the new client implementation of the logic which uses the `hashicorp/retriable-http` package.

PR #3393, opened 2026-05-04
fix: support changed username

### LINKED ISSUE #3392: [BUG]: Support suspended EMU users
### Expected Behavior

Able to remove a suspended GitHub EMU user from a team or repository.

### Actual Behavior

The user cannot be found because when a user is suspend its username goes from `{normalized_idp_username}_{enterprise_shortcode}` to an obfuscated hash: `{username_hash}_{enterprise_shortcode}`

See https://docs.github.com/en/enterprise-cloud@latest/admin/managing-iam/provisioning-user-accounts-with-scim/deprovisioning-and-reinstating-users#effects-of-soft-deprovisioning

All username based operations will fail on suspended users or silently fail.

### Terraform Version

v6.12.1

It was happening for a long time any way

### Affected Resource(s)

- resource_github_team_members
- but also all other resources referencing a user by username

```hcl

```

### Steps to Reproduce

Get a EMU account
Create a team and add the EMU user as a member
Suspend the user and wait for the username change
Try to remove the user

### Debug Output

```shell

```

### Panic Output

```shell

```

### Code of Conduct

acouvreur:
In GitHub EMU, suspended users may have their username changed to a hashed version.

Resolves #3392

----

### Before the change?

- Suspended users cannot be deleted

### After the change?

- Suspended users will have their username retrieved using the user ID before deletion

### Pull request checklist

- [ ] Yes
- [x] No

----

AI Disclaimer: developed with Claude Sonnet 4.6

### DIFF
--- a/github/resource_github_membership.go
+++ b/github/resource_github_membership.go
@@ -5,6 +5,7 @@ import (
 	"errors"
 	"fmt"
 	"net/http"
+	"strconv"
 
 	"github.com/google/go-github/v85/github"
 	"github.com/hashicorp/terraform-plugin-log/tflog"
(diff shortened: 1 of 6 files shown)


acouvreur:
Hey @stevehipwell @deiga 

Let me know if this implementation makes sense to you. I would really like to get this merged to properly clean up users that got their licenses revoked.

Right now they would be either removed from the state because of a 404 not found, or error (depending on refresh=true or false).

With this fix, we compute the underlying user ID and we are able to retrieve it's new username if it changed and properly remove the user.

stevehipwell (changes requested):
@acouvreur I don't think the fundamentals of current approach are correct, if you want to add in `user_id` you should follow the pattern we used for `repository_id` which requires a state migration per resource. It also looks like you missed the collaborators resource too.

PR #3394, opened 2026-05-04
fix: import or update autolinks on creation

### LINKED ISSUE #2565: [FEAT]: Automatically import autolinks when trying to create one for a repository
### Describe the need

When using the Terraform provider with organization scoped rights (GitHub App especially), repository transfers are complicated to manage.

Every 422 could be mitigated by trying to import the resource automatically first.

Autolinks are unique by the prefix key, so they could be imported quite easily. 

### SDK Version

_No response_

### API Version

_No response_

### Relevant log output

```shell

```

### Code of Conduct

acouvreur:
Resolves #2565

----

### Before the change?

Creating a `github_repository_autolink_reference` resource whose `key_prefix` already existed in the repository failed immediately with a `422 Unprocessable Entity` error from the GitHub API. There was no way to adopt an existing autolink into Terraform state through a normal apply; the only workaround was a manual terraform import.

### After the change?

The Create function now looks up existing autolinks by key_prefix before calling the GitHub API:

- Same settings (target_url_template and is_alphanumeric match): the existing autolink is silently imported into state, making the operation idempotent.
- Different settings: the existing autolink is deleted first, then a new one is created with the desired configuration, providing an in-place update on first apply.
- No existing autolink: behaviour is unchanged; the autolink is created normally.

### Pull request checklist

- [ ] Yes
- [x] No

----

AI Disclaimer: This was developed with the help of Claude Sonnet 4.6

### DIFF
--- a/github/resource_github_repository_autolink_reference.go
+++ b/github/resource_github_repository_autolink_reference.go
@@ -102,6 +102,26 @@ func resourceGithubRepositoryAutolinkReferenceCreate(d *schema.ResourceData, met
 	isAlphanumeric := d.Get("is_alphanumeric").(bool)
 	ctx := context.Background()
 
+	// Check if an autolink with this key prefix already exists.
+	// The GitHub API returns 422 when attempting to create a duplicate, so we
+	// handle it ourselves: import the existing one if its settings match, or
+	// delete and recreate it if they differ.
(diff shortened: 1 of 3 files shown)


deiga:
This is not behaviour we want to add, since using terraforms `import` blocks should address the same concern.

Issue #3180, opened 2026-02-11
[MAINT]: Tracking issue to fully support repository renaming for resources with required repo name fields


deiga:
### Describe the need

                                                                                                                                
                                                                                                                                                          
  This issue tracks adding full support for repository renaming across all resources that have a required repository (or similar) field. When a GitHub     
  repository is renamed, resources referencing the old name break on the next terraform plan/apply because the stored repository name in state no longer   
  matches the GitHub API.                                                                                                                                  
                                                                                                                                                           
##   Why This Is Needed                                                                                                                                       

  When a repository is renamed (either via github_repository name change or externally through the GitHub UI/API), any dependent resource that stores the
  repository name as part of its ID or state will:

  1. Fail with a 404 on the next Read, causing Terraform to believe the resource was deleted
  2. Force unnecessary recreation of resources that could otherwise survive a rename
  3. Break terraform plan with confusing diffs showing the old vs new repository name

  The provider already has a pattern for handling this — the diffRepository function in util_diff.go — but it is only applied to 6 out of 33 affected
  resources.

###   Existing Pattern

  The diffRepository function (util_diff.go) handles renames by:

  1. Storing a repository_id (numeric) alongside the repository name
  2. On CustomizeDiff, checking if the repository name changed
  3. Looking up the new name via the API and comparing numeric IDs
  4. Only forcing a new resource if the actual repository changed (different ID), not just the name

####  Resources Already Using diffRepository

  - github_actions_secret
  - github_actions_variable
  - github_actions_environment_secret
  - github_actions_environment_variable
  - github_dependabot_secret
  - github_repository_file

###  Before (current — breaks on rename)
```
  func resourceGithubExample() *schema.Resource {
      return &schema.Resource{
          // ...
          Schema: map[string]*schema.Schema{
              "repository": {
                  Type:     schema.TypeString,
                  Required: true,
                  ForceNew: true, // Forces recreation on ANY name change
              },
          },
      }
  }

  func resourceGithubExampleRead(ctx context.Context, d *schema.ResourceData, meta any) diag.Diagnostics {
      repoName := d.Get("repository").(string)
      // Uses old name from state → 404 if renamed
  }
```
###   After (target — survives rename)
```
  func resourceGithubExample() *schema.Resource {
      return &schema.Resource{
          // ...
          CustomizeDiffFunc: customdiff.Sequence(
              diffRepository,
          ),
          Schema: map[string]*schema.Schema{
              "repository": {
[...]

PR #3396, opened 2026-05-05
fix: support repository rename in resource_github_actions_repository_…

acouvreur:
…oidc_subject_claim_customization_template

Related to https://github.com/integrations/terraform-provider-github/issues/3180

----

### Before the change?

- The OIDC claim customization would recreate/reset on rename 

### After the change?

- The OIDC claim customization will be unchanged if the new repository name is actually a rename

### Pull request checklist

- [ ] Yes
- [ ] No

----


github/resource_github_actions_repository_oidc_subject_claim_customization_template.go
```diff
@@ -53,77 +76,98 @@ func resourceGithubActionsRepositoryOIDCSubjectClaimCustomizationTemplateCreateO
...
 	if err != nil {
-		return deleteResourceOn404AndSwallow304OtherwiseReturnError(err, d, "actions repository oidc subject claim customization template (%s, %s)", owner, repository)
+		var ghErr *github.ErrorResponse
+		if errors.As(err, &ghErr) && ghErr.Response.StatusCode == http.StatusNotFound {
+			tflog.Info(ctx, "Repository not found, removing from state.", map[string]any{"repository": repoName})
+			d.SetId("")
+			return nil
+		}
+		return diag.FromErr(err)
+	}
```
acouvreur, line 132: I could use the same wrapper deleteResourceOn404AndSwallow304OtherwiseReturn I guess

acouvreur:
@deiga @stevehipwell let me know if I can do anything to facilitate the review of this change

github/resource_github_actions_repository_oidc_subject_claim_customization_template.go
```diff
@@ -53,77 +76,98 @@ func resourceGithubActionsRepositoryOIDCSubjectClaimCustomizationTemplateCreateO
...
+		return diag.FromErr(err)
 	}
 
 	d.SetId(repository)
-	return resourceGithubActionsRepositoryOIDCSubjectClaimCustomizationTemplateRead(d, meta)
+	if err := d.Set("repository_id", int(repo.GetID())); err != nil {
+		return diag.FromErr(err)
+	}
+
+	return resourceGithubActionsRepositoryOIDCSubjectClaimCustomizationTemplateRead(ctx, d, meta)
```
stevehipwell, line 112: For modernising resources, removing the call to read is a minimum.

stevehipwell (changes requested):
Thanks for the PR @acouvreur. Could you add an empty test for the migration so we know what we need to enables when we have a pattern. Also you may want to take a look at #3476 which is the latest migration work I've done and should be the best pattern to follow.

github/resource_github_actions_repository_oidc_subject_claim_customization_template.go
acouvreur: Done

acouvreur:
@stevehipwell I've updated the PR to the latest recommendations and architecture guidelines, can you take a look?

PR #3380, opened 2026-04-29
chore: Use tf plugin docs

stevehipwell:
Resolves #2924
Resolves #2961 
Resolves #2897
Resolves #3241
Closes #2987
Closes #3199

----

### Before the change?

- Docs were manually written using the legacy website pattern

### After the change?

- Docs auto-generated using `tfplugindocs`
  - Not really
  - They're using (mostly) static templates
  - We need to migrate each doc to either use either the generic template or to use dynamic sections in their templates

### Pull request checklist

- [x] Schema migrations have been created if needed ([example](https://github.com/F-Secure-web/terraform-provider-github/blob/main/github/migrate_github_actions_organization_secret.go))
- [x] Tests for the changes have been added (for bug fixes / features)
- [x] Docs have been reviewed and added / updated if needed (for bug fixes / features)

### Does this introduce a breaking change?

Please see our docs on [breaking changes](https://github.com/octokit/.github/blob/master/community/breaking_changes.md) to help!

- [ ] Yes
- [x] No

----

PR #3414, opened 2026-05-09
doc(data github_app): Add id on data source GitHub app documentation

### LINKED ISSUE #2819: [FEAT]: Add more output to `github_app` datasource
### Describe the need

I would like to add `id` to the output of the `github_app`. This is for example used in the `github_repository_ruleset` resource when defining `bypass_actors`.

### SDK Version

_No response_

### API Version

_No response_

### Relevant log output

```shell

```

### Code of Conduct

ViacheslavKudinov:
Resolves #2819 

----

### Before the change?

- The attribute `id` is not present in documentation but is actually [there](https://github.com/integrations/terraform-provider-github/blob/main/github/data_source_github_app.go#L46) 
- "Static" doc

### After the change?

- The attribute `id` is on `github_app` data source documentation
- Doc generated based on schema and example by tfdoc

### Pull request checklist

- [ ] Yes
- [x] No

----


stevehipwell (changes requested):
@ViacheslavKudinov you could automate this document either by merging _templates/data-sources.md.tmpl_ with this specific template, or deleting this template and using the default. In both cases you just need to make sure the schema comments are correct so as to not lose any information.

templates/data-sources/app.md.tmpl
```diff
@@ -1,29 +1,16 @@
...
 ---
 
 # {{.Name}} ({{.Type}})
 
-Use this data source to retrieve information about an app.
+{{ .Description | trimspace }}
 
 ## Example Usage
 
 {{ tffile "examples/data-sources/app/example_1.tf" }}
```
stevehipwell:
```suggestion
{{ if .HasExamples -}}
## Example Usage

{{- range .ExampleFiles }}

{{ tffile . }}
{{- end }}
{{- end }}
```

This should work if the example is in the correct location.

ViacheslavKudinov: Updated template

ViacheslavKudinov:
@stevehipwell updated PR.
Please, take a look if anything else needs to be changed.
Thank you in advance.

PS what is expected to be in "subcategory" ?

stevehipwell:
> PS what is expected to be in "subcategory" ?

It's expected to be empty.

stevehipwell (dismissed):
LGTM

stevehipwell:
@robert-crandall this is one of the kinds of PRs that I don't think there is much value requiring a staff approval. What do you think?

templates/data-sources/app.md.tmpl
deiga, line 1: suggestion: This file is now ~identical to `templates/data-sources.md.tmpl`, which AFAIK means you can delete this as it will automatically use the generic template

deiga (changes requested):
Could you verify if my hunch is correct?

stevehipwell:
@deiga it could be removed, it's in my comment from earlier. I'm kind of on the fence about which is better. Technically having a specific template is quicker and makes it easier to improve the documentation. But having a template does potentially require maintenance in the future.

deiga:
> @deiga it could be removed, it's in my comment from earlier. I'm kind of on the fence about which is better. Technically having a specific template is quicker and makes it easier to improve the documentation. But having a template does potentially require maintenance in the future.

Copying the template when needed sounds super quick and reduces overhead of keeping up with generic template changes

templates/data-sources/app.md.tmpl
ViacheslavKudinov: Thanks for the comment. 
Probably it became identical while I was tuning initial one i started with.
I will double check and remove if its the same as default one.

ViacheslavKudinov: Now i recall why i added custom template.

If there is no "example" it generates some extra empty lines.

I will update default template  for data sources to render by default without extra empty lines if no examples.

With default  template was like on screenshot  (when no examples)
image
but when i added example then it looks as from default one.

ViacheslavKudinov: Updated PR

deiga (dismissed):
LGTM

templates/data-sources.md.tmpl
```diff
@@ -9,11 +9,10 @@ description: |-
 
 {{ .Description | trimspace }}
 
-{{ if .HasExamples -}}
-## Example Usage
-
-{{- range .ExampleFiles }}
+{{- if .HasExamples }}
```
stevehipwell: Why this change?

stevehipwell:
@ViacheslavKudinov could you please rebase this PR, I'd like to get it merged in.

ViacheslavKudinov:
> @ViacheslavKudinov could you please rebase this PR, I'd like to get it merged in.

done

stevehipwell (approved):
LGTM

stevehipwell:
@deiga let's get this merged as is and then we can figure out if the template can be dropped.

robert-crandall (approved):
Nice cleanup - thanks for surfacing id in the docs and converting the template to the tfplugindocs auto-generated pattern from #3380. Build and provider validation pass locally; LGTM. 🙏

PR #3415, opened 2026-05-10
[BUG]: fix github repository file id escaping

### LINKED ISSUE #3335: [BUG]: github_repository_file rejects valid file paths containing ':' in v6.11.1
### Expected Behavior
`github_repository_file` should continue to support valid GitHub repository file paths that include a literal colon (`:`) in the filename, such as `repro/example:one.yaml`.

After creating a file at exactly `repro/example:one.yaml` with provider `6.11.0`, upgrading to `6.11.1` should still allow refresh and plan to succeed.

### Actual Behavior
After upgrading `integrations/github` from `6.11.0` to `6.11.1`, `terraform plan` / `tofu plan` fails during refresh for existing `github_repository_file` resources whose `file` path contains `:`.

The provider returns:

```shell
Error: failed to build ID: unescaped separator in non-final part "repro/example:one.yaml"
```

This appears to come from `github_repository_file` building its resource ID with `buildID(repo, file, branch)`, while `buildID` rejects `:` in all non-final parts.

### Terraform Version
```shell
OpenTofu v1.11.5
on linux_amd64
```

### Affected Resource(s)
- `github_repository_file`

```hcl
terraform {
  required_providers {
    github = {
      version = "6.11.1"
    }
  }
}

variable "github_token" {
  type      = string
  sensitive = true
}

variable "owner" {
  type = string
}

variable "repository" {
  type = string
}

provider "github" {
  token = var.github_token
  owner = var.owner
}

resource "github_repository_file" "marker" {
  repository          = var.repository
  branch              = "main"
  file                = "repro/README.md"
  commit_message      = "create repro directory"
  overwrite_on_create = true
  content             = "Files in this directory are managed by Terraform."
}

resource "github_repository_file" "colon_filename" {
  repository          = var.repository
  branch              = "main"
  file                = "repro/example:one.yaml"
  commit_message      = "manage colon filename repro"
  overwrite_on_create = true
  content = yamlencode({
    apiVersion = "v1"
    kind       = "ConfigMap"
    metadata = {
      name = "example-one"
    }
  })
}
```

### Steps to Reproduce
1. Create an empty test repository with a `main` branch.
2. Apply the configuration above with `integrations/github` `6.11.0` so the resource manages the literal path `repro/example:one.yaml` and stores it in state.
3. Change only the provider version to `6.11.1`.
4. Run `terraform plan` or `tofu plan` again.
5. Observe refresh fail before planning completes.

### Debug Output
```shell
Error: failed to build ID: unescaped separator in non-final part "repro/example:one.yaml"
```

In my original production case, the same failure occurred on existing file paths like:

```shell
trivial-apps/infra-misc/kyverno/irsa/tf/nes-backstage:nes-backstage.yaml
trivial-apps/infra-misc/kyverno/irsa/tf/ack:ack-rds-controller.yaml
trivial-apps/infra-misc/kyverno/irsa/tf/kube-system:s3-csi-driver-sa.yaml
```

### Panic Output
N/A

### Code of Conduct

deiga:
Resolves #3335 

----

### Before the change?

- Terraform would throw an error if `filePath` in `github_repository_resource` contained a colon
  - This was introduced by me in https://github.com/integrations/terraform-provider-github/pull/3175 when I changed the ID generation code

### After the change?

- Terraform successfully operates on resources when `filePath` in `github_repository_resource` contains a colon

### Pull request checklist

- [ ] Yes
- [x] No

----


stevehipwell (commented):
Have we tested this doesn't break existing IDs? It looks OK, but worth checking. Also should we add this to the next minor release milestone as we can always cherry pick it into a patch if required?

deiga:
@stevehipwell Thanks for asking, I ran v6.11.0 locally and then tried to run this code: same failure, since it was already the v0 migration that was failing!

deiga:
I think we should release this as a patch, considering that the changes in v6.11.1 are blocking upgrading for anyone with colons in `repository_file` resources

stevehipwell:
@deiga do you want to rebase this and then we can get it merged in. Once it's been merged we can decide if we want to cherry pick it onto a patch branch.

examples/dev.tfrc
```diff
@@ -1,7 +1,9 @@
...
   dev_overrides {
-    "integrations/github" = "~/go/bin/"
+    "integrations/github" = "$HOME/go/bin/"
   }
 
-  direct {}
+  direct {
+    exclude = ["registry.terraform.io/hashicorp/github"]
+  }
 }
```
stevehipwell: Could we fix the EOF?

stevehipwell: Actually should this file even be in this PR?

github/resource_github_repository_file_migration.go
```diff
@@ -95,28 +95,35 @@ func resourceGithubRepositoryFileStateUpgradeV0(ctx context.Context, rawState ma
...
 
 	repo, _, err := client.Repositories.Get(ctx, owner, repoName)
 	if err != nil {
-		return nil, fmt.Errorf("failed to retrieve repository '%s': %w", repoName, err)
+		return nil, fmt.Errorf("state upgrade v0: failed to retrieve repository '%s': %w", repoName, err)
 	}
 
 	rawState["repository_id"] = int(repo.GetID())
 
+	branch, ok := rawState["branch"].(string)
```
stevehipwell: Is there a reason this has been changed from a single line?

stevehipwell (changes requested):

github/resource_github_repository_file_migration.go
deiga: `forcetypeassert` linter bit it :)

examples/dev.tfrc
deiga: No, it's not directly related. It's a Scout Rule thing, since I used the example to verify the changes and noticed that the config file didn't work as we had it

github/resource_github_repository_file_migration.go
stevehipwell: I'm not sure I follow why, the code is identical as there isn't a new check for the map lookup (which could be nested single lines anyway).

deiga: In [this commit](https://github.com/integrations/terraform-provider-github/pull/3415/changes/24563567b8c0cc2438fe8e709feb3a8fb014340a#diff-30cb7eb7814ead2d68bd2b1982a6586c82b7442e8fed947a1f04ff65e6f0a39dL113-R113) I needed to touch that line, which caused it to be checked by `make lintcheck-new`

And since we didn't use `forcetypeassert` when that migration was originally created: it had a lot to complain about

github/resource_github_repository_file_migration.go
```diff
@@ -95,28 +95,35 @@ func resourceGithubRepositoryFileStateUpgradeV0(ctx context.Context, rawState ma
...
+	if !ok || branch == "" {
+		branch = repo.GetDefaultBranch()
+		rawState["branch"] = branch
 	}
 
-	newResourceID, err := buildID(rawState["repository"].(string), rawState["file"].(string), rawState["branch"].(string))
+	filePath, ok := rawState["file"].(string)
+	if !ok {
+		return nil, fmt.Errorf("state upgrade v0: file path is not a string")
+	}
```
stevehipwell:
```suggestion
	branchSet := false
	if v, ok := rawState["branch"]; ok {
		if _, ok := v.(string); ok || s != "" {
			branchSet = true
		}
	}	
	if !branchSet {
		rawState["branch"] = repo.GetDefaultBranch()
	}

	pathSet := false
	if v, ok := rawState["file"]; ok {
		if _, ok := v.(string); ok {
			pathSet = true
		}
	}
	if !pathSet {
		return nil, fmt.Errorf("state upgrade v0: file path is not a string")
	}
```
This would be my prefered pattern here, otherwise we're still masking a conversion.

stevehipwell (changes requested):

github/resource_github_repository_file_migration.go
deiga: Done

deiga:
@stevehipwell Implemented your requested changes and I commented out the unit test to propose this format again, since I needed them to determine that my code changes didn't break anything

examples/dev.tfrc
```diff
@@ -1,7 +1,9 @@
 provider_installation {
```
stevehipwell, line 1: I don't think this file should be in the repo at all. The usage pattern could be documented in the contributing guide.

stevehipwell (dismissed):
LGTM

Let's get the _dev.tfrc_ file removed in a separate PR.

deiga:
@robert-crandall Could you please review this? :)

deiga:
@stevehipwell copilot caught a bug, please re-review :)

stevehipwell (approved):
LGTM

robert-crandall (approved):

PR #1978, opened 2023-10-25
feat: read timeout for github_ip_ranges data source

kimxogus:
Resolves #1961 

----

### Before the change?

*  Adds 5 minutes default timeout for `github_ip_ranges` data source.

### After the change?

* Terraform will stop reading github ip ranges after 5 minutes timeout

### Pull request checklist
- [ ] Tests for the changes have been added (for bug fixes / features)
- [x] Docs have been reviewed and added / updated if needed (for bug fixes / features)

### Does this introduce a breaking change?

Please see our docs on [breaking changes](https://github.com/octokit/.github/blob/master/community/breaking_changes.md) to help!

- [ ] Yes
- [x] No

----

Issue #1961, opened 2023-10-16
[FEAT]: Configurable timeout for github_ip_ranges data source


kimxogus:
### Describe the need

Sometimes github_ip_ranges takes too much time. Today I waited 40 minutes but it was still reading (see the logs below).
I wanted to set timeout for github_ip_ranges but I found that it's not configurable.

Can you make this configurable or set default timeout about 1 or 5 minutes?

### API Version

5.34.0

### Relevant log output

```shell
data.github_ip_ranges.github[0]: Still reading... [20m0s elapsed]
data.github_ip_ranges.github[0]: Still reading... [20m10s elapsed]
data.github_ip_ranges.github[0]: Still reading... [20m20s elapsed]
data.github_ip_ranges.github[0]: Still reading... [36m41s elapsed]
data.github_ip_ranges.github[0]: Still reading... [36m51s elapsed]
data.github_ip_ranges.github[0]: Still reading... [37m1s elapsed]
data.github_ip_ranges.github[0]: Still reading... [37m11s elapsed]
data.github_ip_ranges.github[0]: Still reading... [37m21s elapsed]
data.github_ip_ranges.github[0]: Still reading... [37m31s elapsed]
data.github_ip_ranges.github[0]: Still reading... [37m41s elapsed]
data.github_ip_ranges.github[0]: Still reading... [37m51s elapsed]
data.github_ip_ranges.github[0]: Still reading... [38m1s elapsed]
data.github_ip_ranges.github[0]: Still reading... [38m11s elapsed]
data.github_ip_ranges.github[0]: Still reading... [38m21s elapsed]
data.github_ip_ranges.github[0]: Still reading... [38m31s elapsed]
data.github_ip_ranges.github[0]: Still reading... [38m41s elapsed]
data.github_ip_ranges.github[0]: Still reading... [38m51s elapsed]
data.github_ip_ranges.github[0]: Still reading... [39m1s elapsed]
data.github_ip_ranges.github[0]: Still reading... [39m11s elapsed]
data.github_ip_ranges.github[0]: Still reading... [39m21s elapsed]
data.github_ip_ranges.github[0]: Still reading... [39m31s elapsed]
data.github_ip_ranges.github[0]: Still reading... [39m41s elapsed]
data.github_ip_ranges.github[0]: Still reading... [39m51s elapsed]
data.github_ip_ranges.github[0]: Still reading... [40m1s elapsed]
[...]

PR #3421, opened 2026-05-12
feat: read timeout for github_ip_ranges data source

ffreitas-te:
### Before the change?

* The `github_ip_ranges` data source had no `timeouts` block, and the metadata request used `owner.StopContext` instead of the Terraform read context, so a read timeout would not reliably bound the HTTP call.

### After the change?

* Default **5 minute** read timeout on `github_ip_ranges` (overridable via `timeouts { read = "..." }` in configuration).
* The metadata API call uses the **`ReadContext` `ctx`**, so Terraform's read deadline is honored.

### Pull request checklist

**Schema migrations:** N/A — this updates only the `github_ip_ranges` data source; no managed resource identity or Terraform state schema changes, so no migration code is required.

- [ ] Yes
- [x] No

----

Related to #1961  
Follow-up to #1978 (previous change went stale; this re-applies the same intent on the current Terraform Plugin SDK v2 codebase.)

/cc @kimxogus @kfcampbell @KatKorotkikh


ffreitas-te:
Bumping for visibility: this addresses the long-running `github_ip_ranges` read behavior described in #1961 (timeouts + using the read context for the metadata call).

@KatKorotkikh — same class of pain as in #1961; feel free to weigh in if this matches what you hit in practice.

github/data_source_github_ip_ranges_test.go
```diff
@@ -2,13 +2,29 @@ package github
...
+		t.Fatal("expected Timeouts to be configured on github_ip_ranges data source")
+	}
+	if r.Timeouts.Read == nil {
+		t.Fatal("expected default read timeout to be configured")
+	}
+	if got := *r.Timeouts.Read; got != 5*time.Minute {
+		t.Fatalf("expected default read timeout 5m, got %v", got)
+	}
+}
+
```
deiga: issue: This test is useless. It only checks that timeout has been configured, it's very brittle. Tests should verify behaviour, and not be tightly coupled to code and it's structure

deiga (changes requested):

github/data_source_github_ip_ranges_test.go
ffreitas-te: Thanks @deiga — fair point, the old test asserted the schema struct rather than any behaviour. Replaced it in c7164125 with tests that exercise the read path:

- a mock metadata endpoint that withholds its response, asserting the read aborts at the caller's deadline instead of waiting for the server. This one fails against the previous `owner.StopContext` behaviour, which is the actual bug being fixed.
- a normal `/meta` response, asserting the IP ranges are still populated.
- an acceptance sub-test whose config sets `timeouts { read = "2m" }`, so the configurable timeout is covered through Terraform rather than by introspecting the schema.

Happy to adjust further if you would rather see this covered purely at the acceptance-test level.

deiga (dismissed):
LGTM

ffreitas-te:
@deiga your approval was cleared when I pushed two follow-up commits, so flagging you here.

The CI run on `dd9601a2` failed on two checks, both caused by my own changes (the `Check` job is just the aggregate gate over `ci`, `docs` and `lint-new`, so it had nothing of its own):

- **Strict linting of new code** — `forcetypeassert` fired twice in `github/data_source_github_ip_ranges_test.go` (lines 78 and 81). The comparisons put a type assertion on the right hand side of a multi-value assignment, which that linter rejects. Fixed in 9f9afe93 by comparing with `reflect.DeepEqual` against `[]any{...}`, which drops the assertion entirely and matches how the rest of the test suite compares values.
- **Documentation** — `rumdl` flagged the `*` list marker in the new Timeouts section; `.rumdl.toml` pins `MD004` to `style = "dash"`. Fixed in 2ca66790 by switching the marker to `-` in `templates/data-sources/ip_ranges.md.tmpl` and regenerating `docs/` with `make generatedocs`, so the template stays the source of truth.

Verified locally with the repo's own tooling before pushing: `golangci-lint` v2.12.2 built against the Go version in `go.mod` reports 0 issues for both the baseline config and `--new-from-merge-base main --config .golangci.new.yml`, `golangci-lint fmt --diff` is empty, `make build -o lintcheck` and `make test` pass, and re-running `make generatedocs` leaves no diff so `make checkdocs` stays clean. No `*` list markers remain anywhere under `docs/`.

The new CI and CodeQL runs are queued as `action_required` since this is a fork PR — could you approve the workflow run and take another look? The branch also shows as behind `main` again; happy to bring it up to date if you'd prefer that before merge.

stevehipwell (changes requested):
I don't think this is how we want to introduce per resource/data-source timeouts as they require us to think about the big picture and design something that fit's the current provider and the provider we're working towards.

CC @deiga

stevehipwell:
> The github_ip_ranges data source had no timeouts block, and the metadata request used owner.StopContext instead of the Terraform read context, so a read timeout would not reliably bound the HTTP call.

@ffreitas-te I don't think the before state is correct, the current implementation inherits the default read timeout (AFAIK 20m).

ffreitas-te:
@stevehipwell you're right, and thanks for pushing back — I went and read the SDK rather than argue from memory. `Resource.read()` wraps the call in `context.WithTimeout(ctx, d.Timeout(TimeoutRead))`, and `ResourceData.Timeout()` falls back to a hardcoded 20-minute system default when no `Timeouts` block is set (terraform-plugin-sdk v2.40.1, `helper/schema/resource.go:864` and `helper/schema/resource_data.go:552-560`). Data sources reach that same path via `ReadDataApply` → `read`, so `github_ip_ranges` has always had a 20-minute read deadline and my description was wrong to imply otherwise.

Checking that turned up something more important: the defect this PR was opened for is already fixed on `main`. The read used to issue the metadata request with `owner.StopContext`, ignoring whatever deadline the SDK had set; #3448 replaced that with the read context as part of the new client, so `main` already calls `Meta.Get(ctx)`. My successive merges of `main` absorbed it, and this diff no longer contains that fix. What remains is a `Timeouts{Read: 5m}` block plus docs and tests — and that 5-minute default would have silently lowered the effective default from 20 minutes for everyone who never sets it, which I shouldn't have bundled into something framed as a bug fix. `internal/ghclient` also already caps each request at `clientTimeout = 5 * time.Minute`, so the call was never unbounded in the way I described.

One thing for the record, aimed at your design point rather than at this PR: per-resource timeouts already exist here. `github_actions_hosted_runner` has shipped `Timeouts{Delete: 10 * time.Minute}` since #2893. I don't think that justifies a second one-off — if anything it argues your side, since it arrived bundled with a new resource for a long-running delete rather than as a deliberate pattern. But it does mean users can set `timeouts` on exactly one resource and nowhere else, and whatever design you land on has that instance to reconcile. Seems worth capturing in whatever decision gets recorded.

So I'm closing this. I'd rather contribute to the design than to this diff: is the shape you have in mind a provider-level timeout setting, or a shared helper applied consistently to resources and data sources as they're touched? Happy to open a design issue with the above as background, or to implement whatever you and @deiga settle on.

Last thing — the two unit sub-tests here don't depend on the `timeouts` block. They call the read function directly with their own context and assert it aborts at the caller's deadline instead of waiting on a stalled `/meta`, which is precisely the behavior #3448 introduced. I applied them to a clean `main` checkout and they pass unchanged (76 lines, no production changes), which looks like a fit for #2996. Say the word and I'll send them as a test-only PR.

PR #3422, opened 2026-05-13
fix: org detection failure on ghec-dr endpoints

### LINKED ISSUE #2626: [BUG]: Create github_team resource causes error when created under EMU EU Data Residency Organization
### Expected Behavior

Hi Team.

We are working on using Terraform for platform automation. 

The code at:
https://github.com/integrations/terraform-provider-github/blob/6008909a212294e332cb23f3bbaf1089e0d9ed1e/github/util.go#L26

Produces:
Error: this resource can only be used in the context of an organization, "org_name" is a user

org_name = the org defined as owner in provider configuration. The org_name is not a user, but the actual org_name of the customer Enterprise Org to be used as owner.

Version 6.6.0.

Provider definition: 
https://github.com/orgorgorgorgorg1/platform/blob/main/tf/providers.tf

The code sample in https://github.com/orgorgorgorgorg1/platform/blob/main/tf/ works correctly in the public GH Org used for testing and developing. Then breaks on the error above when used in EMU Data Residency Environment.

### Actual Behavior

Org is successfully identified as being an organization. Hence the error not being produced, causing org level automation to break.

### Terraform Version

3.7.0

### Affected Resource(s)

- github_teams

```hcl
https://github.com/orgorgorgorgorg1/platform/tree/main/tf
```

### Steps to Reproduce

Create github_teams resource and configure provider to be set to an organization that resides in a GH Enteprise with Data residency & EMU.

### Debug Output

```shell
Error: this resource can only be used in the context of an organization, "org_name" is a user
```

### Panic Output

```shell

```

### Code of Conduct

ebk45:
Resolves #2626 

----

### Before the change?
 - When `base_url` is a GHEC-DR endpoint (`https://api.*.ghe.com/`), `Organizations.Get()` fails due to incorrect URL construction in the `go-github` library. The error is silently discarded, leaving `IsOrganization` as `false`. Any resource or data source that calls `checkOrganization()` (e.g. `github_external_groups`) then fails with the misleading error: `this resource can only be used in the context of an organization, "XYZ" is a user`.

### After the change?
- When `Organizations.Get()` fails, the provider logs a `[WARN]` and falls back to `Users.Get()` to check the owner's `type` field. Since `/users/{owner}` works correctly on GHEC-DR endpoints, the organization is detected and `IsOrganization` is set to `true`. Downstream resources work as expected.

### Pull request checklist
- [N/A] Schema migrations have been created if needed ([example](https://github.com/F-Secure-web/terraform-provider-github/blob/main/github/migrate_github_actions_organization_secret.go))
- [N/A] Docs have been reviewed and added / updated if needed (for bug fixes / features)

- [ ] Yes
- [x] No

----


portellir (approved):

willjr (approved):

stevehipwell (commented):
This PR shouldn't be required as the [Get Org](https://docs.github.com/en/rest/orgs/orgs?apiVersion=2022-11-28#get-an-organization) API should work correctly for GHEC-DR. Let's see if we can figure out why the API isn't responding correctly before looking for workarounds.

eyalgal:
Closing this after wrapping up the maintainer thread discussion.

Root cause turned out to be an installation ID mismatch, not URL construction or org detection on GHEC-DR. The App is installed per-org with a distinct `installation_id` each, and using the wrong org's ID makes `GET /orgs/{owner}` fail auth. The provider silently treats that as "not an org" and surfaces the misleading `"<owner> is a user"`. With the correct installation token, `GET /orgs/{owner}` returns `200` with `"type": "Organization"` and detection works as expected.

Since this PR's fallback masks the symptom rather than fixing a provider bug, I'm closing it. Thanks @ebk45 for the thorough investigation. The useful follow-up (a clearer error when the org lookup fails on auth) will be tracked separately.

PR #3430, opened 2026-05-14
[MAINT]: Introduce stricter linter rules for new code

deiga:
### Before the change?

- New code and existing code forced to use same linter rules

### After the change?

- Enables holding new code to a stricter standard
- Adds `make lintcheck-new` to run stricter linters on only code since HEAD of `main`
- Adds `make lintcheck-strict` to run stricter linters on whole codebase
	- Useful to see if some strict linters can be moved to default config 
- Refactor `ci.yaml` workflow, to run linters in parallel using `golangci-lint-action`
	- This enables inline linter complaints in PRs

### Pull request checklist

- [ ] Yes
- [x] No

----

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -43,6 +43,7 @@ jobs:
   lint-new:
     name: Strict linting of new code
     runs-on: ubuntu-latest
+    if: github.event_name == 'pull_request' || github.event_name == 'pull_request_target'
     permissions:
       contents: read
       pull-requests: read
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -37,7 +37,7 @@ jobs:
         uses: golangci/golangci-lint-action@1e7e51e771db61008b38414a730f564565cf7c20 # v9.2.0
         with:
           version: latest
-      - run: make build -o lintcheck # Stop running `make lintcheck` before build. Linting is handled by a separate job
+      - run: make build -o lintcheck # Stop running `make lintcheck` before build. Linting is handled separately
       - run: make test
 
   lint-new:
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -33,26 +33,12 @@ jobs:
         with:
           go-version-file: go.mod
           cache: true
-      - run: make build -o lintcheck # Stop running `make lintcheck` before build. Linting is handled by a separate job
-      - run: make test
-
-  lint:
-    name: Lint code
-    runs-on: ubuntu-latest
-    permissions:
-      contents: read
-    steps:
-      - name: Checkout
-        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
-      - name: Set-up Go
-        uses: actions/setup-go@4a3601121dd01d1626a1e23e37211e3254c1c06c # v6.4.0
-        with:
-          go-version-file: go.mod
-          cache: true
       - name: golangci-lint
         uses: golangci/golangci-lint-action@1e7e51e771db61008b38414a730f564565cf7c20 # v9.2.0
         with:
           version: latest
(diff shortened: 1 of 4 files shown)


.github/workflows/ci.yaml
```diff
@@ -34,9 +34,50 @@ jobs:
...
+      contents: read
+      pull-requests: read
+    steps:
+      - name: Checkout
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+      - name: Set-up Go
+        uses: actions/setup-go@7a3fe6cf4cb3a834922a1244abfce67bcef6a0c5 # v6.2.0
+        with:
+          go-version-file: go.mod
+          cache: true
```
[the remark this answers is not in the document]
deiga: `yq` is contained in. `ubuntu-latest`

GNUmakefile
```diff
@@ -46,6 +52,17 @@ lintcheck:
...
+# Checks all code against the strict rules. This is expected to fail until all issues have been fixed, but it allows us to track progress on fixing the issues and ensures that no new issues are introduced.
+lintcheck-strict: .golangci.new.yml
+	@branch=$$(git rev-parse --abbrev-ref HEAD); \
+	printf "==> Checking source code against strict linters on branch: \033[1m%s\033[0m...\n" "🌿 $$branch 🌿"
+	golangci-lint run ./... --config .golangci.new.yml
+
+lintcheck-new: .golangci.new.yml
+	@branch=$$(git rev-parse --abbrev-ref HEAD); \
+	printf "==> Checking changed source code against linters on branch: \033[1m%s\033[0m...\n" "🌿 $$branch 🌿"
+	golangci-lint run ./... --new-from-merge-base main --config .golangci.new.yml
```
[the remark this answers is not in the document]
deiga: `lintcheck-new` is not used in CI

.github/workflows/ci.yaml
```diff
@@ -33,10 +33,50 @@ jobs:
...
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+      - name: Set-up Go
+        uses: actions/setup-go@4a3601121dd01d1626a1e23e37211e3254c1c06c # v6.4.0
+        with:
+          go-version-file: go.mod
+          cache: true
+      - name: golangci-lint
+        uses: golangci/golangci-lint-action@1e7e51e771db61008b38414a730f564565cf7c20 # v9.2.0
+        with:
+          version: latest
```
stevehipwell: Could we keep this in `ci` and then have a `ci-new` job doing what `lint-new` currently does? I think that would be cleaner.

deiga: @stevehipwell Could you elaborate the reasoning here? I would prefer moving linting to a parallel job as that will make execution faster and we can surface both linting and testing errors from the same run.

If you're worried about the duplication between `lint` and `lint-new` I could merge them into a single job with a matrix strategy instead

stevehipwell: Both linting and testing generally have a dependency on the Go code actually building, we could fan these out into 3 jobs (`build`, `lint`, `test`) but IMHO it's cleaner and easier to keep them together in the `ci` job for now. In addition the new code linting is a short lived task for while we make al code meet the strict linting rules, so we shouldn't be shaping the automation around it. In summary having a primary `ci` job and adding an additional job on for new code follows the principle of least surprise.

stevehipwell (dismissed):
LGTM

deiga:
@stevehipwell addressed Copilot comments

@robert-crandall could you please review this PR? :)

stevehipwell (approved):
LGTM

stevehipwell:
@robert-crandall could you please review this PR as it's the next one we need to merge.

robert-crandall (approved):
Thanks for taking this on @deiga !

PR #3436, opened 2026-05-18
[FEAT] Support managing github_membership and lookups by stable user_id (#524)

### LINKED ISSUE #524: Feature Request > ability to use github unique ID for github_membership
Currently we can only provide a username to in the `github_membership` and `github_user` data sources. We're interested in using the unique identifier found at `https://api.github.com/users/{USERNAME}` in the `"id"` field. This is because username can be changed by the user, which causes our terraform scripts for github memberships to be out of date. The `id` cannot be changed and is globally unique, which makes it a much better candidate.

Instead of providing the username to the `/users` endpoint, you can get the same information by providing the `id` to the `/user` endpoint (e.g. `https://api.github.com/user/{id}`).

niklasbeinghaus:
## Summary

- Adds optional `user_id` (TypeInt) to `github_membership` resource and data source, and to `github_user` data source — mutually exclusive with `username` via `ExactlyOneOf`.
- Memberships managed by `user_id` are now resilient to GitHub username renames: Read resolves the live login via `GET /user/{id}` and silently updates the `username` attribute in state, producing no diff.
- Resource ID for new memberships is `org:user_id`; legacy `org:username` IDs are migrated lazily on the next Read. Imports accept both shapes.

Closes #524.

## Why

The GitHub `id` is globally unique and immutable; `username` is not. Today, when a user renames their account, modules using `github_membership` drift and trigger recreates. Issue #524 asks for the stable ID to be a valid lookup key. This PR adds that, while keeping the existing `username`-keyed workflow fully backward-compatible.

### `data "github_user"`
- New optional input `user_id` (mutually exclusive with `username`).
- Computed `user_id` attribute always populated.
- Resolves via `Users.GetByID` when `user_id` is set.

### `data "github_membership"`
- New optional input `user_id` (mutually exclusive with `username`).
- Resolves `user_id` → login first (GitHub's org membership endpoints only accept logins), then queries the membership.
- Computed `user_id` attribute always populated.

### `resource "github_membership"`
- New optional input `user_id` (TypeInt, ForceNew, exactly-one-of with `username`).
- Create resolves both halves of the (login, numeric_id) pair so state always carries both.
- ID format is `org:user_id` for new resources.
- Read auto-detects the ID shape: digits → numeric path (`GetByID`); non-digits → legacy username path, and rewrites the ID in place to `org:user_id` on first refresh.
- Delete uses the live `username` from state, which Read keeps fresh — so deletion still works correctly after a rename.

## Tests

Acceptance tests (require `TF_ACC=1` plus the env vars in `CONTRIBUTING.md`):

- `TestAccGithubUserDataSource` — added subtests for `user_id` happy path + error paths (not-found, mutual exclusion).
- `TestAccGithubMembershipDataSource` — same coverage for the data source.
- `TestAccGithubMembership` — added subtest for `user_id`-keyed create + mutual-exclusion error paths.
- `TestAccGithubMembershipRenameResilience` — end-to-end live-rename test. Creates a membership by `user_id`, renames the external user via `PATCH /user` with their own PAT (`GH_TEST_EXTERNAL_USER_TOKEN`), refreshes, asserts the `username` attribute updated with **no plan diff**, then restores the original login in cleanup. Skips automatically when `GH_TEST_EXTERNAL_USER_TOKEN` is unset.

Local validation:

```sh
go build ./...   # OK
go vet ./...     # OK
go test ./github/... -timeout=120s -parallel=4 -skip '^TestAcc' -count=1   # OK (5.0s)
```

## Docs

- `templates/data-sources/user.md.tmpl`, `templates/data-sources/membership.md.tmpl`, `templates/resources/membership.md.tmpl` updated with `user_id` arguments, import notes, and rename-resilience guidance.
- New example files: `examples/data-sources/user/example_2.tf`, `examples/data-sources/membership/example_2.tf`, `examples/resources/membership/example_2.tf`.
- `docs/*` regenerated to match templates (please re-run `make generatedocs` on a machine with `terraform` on PATH to confirm bit-for-bit match — my dev machine lacked it; I synced docs to templates manually).

## Backward compatibility

- Existing modules using `username` continue to work unchanged.
- Existing state with `org:username` IDs migrates lazily on next Read (single extra `Users.GetByID` round-trip, one-shot).
- `terraform import org:username` and `terraform import org:12345` both supported.

## Notes for reviewers

- `data "github_user"` exposes the stable ID as `user_id` rather than `id` (the issue's wording) because the Terraform SDKv2 reserves the top-level `id` attribute as the resource ID. Using `user_id` is also more consistent with the rest of this provider (e.g. `organization_role_users`).
- Read-side lazy migration was chosen over a `SchemaVersion` bump because StateUpgraders are supposed to be pure, and resolving username → user_id needs a network call. Read-side handles the migration self-healingly and incurs no upfront cost on `terraform plan`.

## Commits

- `[FEAT] Support looking up github_user data source by user_id`
- `[FEAT] Support looking up github_membership data source by user_id`
- `[FEAT] Support managing github_membership by stable user_id`

### DIFF
--- a/docs/data-sources/membership.md
+++ b/docs/data-sources/membership.md
@@ -16,15 +16,31 @@ data "github_membership" "membership_for_some_user" {
 }
 ```
 
+### Lookup by stable user ID
+
+```terraform
+# Look up a membership by the stable GitHub user ID.
+# The numeric ID does not change when the user renames their account.
+data "github_membership" "by_user_id" {
+  user_id = 1
+}
+```
+
 ## Argument Reference
 
-- `username` - (Required) The username to lookup in the organization.
+Exactly one of the following must be set:
 
-- `organization` - (Optional) The organization to check for the above username.
+- `username` - (Optional) The username (login) to lookup in the organization.
+- `user_id` - (Optional) The GitHub numeric user ID. Stable across username changes; prefer this for lookups that should survive renames.
+
+Other arguments:
+
+- `organization` - (Optional) The organization to check for the above user.
 
 ## Attributes Reference
 
-- `username` - The username.
+- `username` - The username (login). Always reflects the user's current login at refresh time.
+- `user_id` - The GitHub numeric user ID.
 - `role` - `admin` or `member` -- the role the user has within the organization.
 - `etag` - An etag representing the membership object.
 - `state` - `active` or `pending` -- the state of membership within the organization. `active` if the member has accepted the invite, or `pending` if the invite is still pending.
--- a/docs/data-sources/user.md
+++ b/docs/data-sources/user.md
@@ -26,13 +26,27 @@ output "current_github_login" {
 }
 ```
 
+### Lookup by stable user ID
+
+```terraform
+# Retrieve information about a GitHub user by their stable numeric ID.
+# Useful when the user may rename themselves: the lookup keeps working.
+data "github_user" "by_id" {
+  user_id = 1
+}
+```
+
 ## Argument Reference
 
-- `username` - (Required) The username. Use an empty string `""` to retrieve information about the currently authenticated user.
+Exactly one of the following must be set:
+
+- `username` - (Optional) The username (login). Use an empty string `""` to retrieve information about the currently authenticated user.
+- `user_id` - (Optional) The GitHub numeric user ID. Stable across username changes, so prefer this if you need lookups that survive a rename.
 
 ## Attributes Reference
 
 - `id` - the ID of the user.
+- `user_id` - the GitHub numeric user ID (same value as `id`, also settable as an input).
 - `node_id` - the Node ID of the user.
 - `login` - the user's login.
 - `avatar_url` - the user's avatar URL.
--- a/docs/resources/membership.md
+++ b/docs/resources/membership.md
@@ -20,18 +20,46 @@ resource "github_membership" "membership_for_some_user" {
 }
 ```
 
+### Identifying the user by stable numeric ID
+
+Using `user_id` instead of `username` makes the membership resilient to the user renaming their GitHub account. After a rename, the next `terraform refresh` updates the `username` attribute in state with no diff, and the resource continues to manage the same membership.
+
+```terraform
+# Manage organization membership by stable GitHub user ID.
+# Recommended over `username` for production: if the user renames their
+# account, the membership stays in sync without drift.
+resource "github_membership" "membership_by_user_id" {
+  user_id = 1
+  role    = "member"
+}
+```
+
 ## Argument Reference
 
 The following arguments are supported:
 
-- `username` - (Required) The user to add to the organization.
+Exactly one of:
+
+- `username` - (Optional) The user (login) to add to the organization. Note: usernames can change; if the user renames themselves, the resource will recreate unless `user_id` is used instead.
+- `user_id` - (Optional) The GitHub numeric user ID to add to the organization. Stable across username changes. Recommended for production use.
+
+Other arguments:
+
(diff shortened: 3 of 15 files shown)


niklasbeinghaus:
### Security implications worth calling out

Beyond fixing the state-drift annoyance from #524, this PR meaningfully reduces a real, exploitable class of attack against `github_membership`. Flagging here so reviewers can weigh it explicitly.

#### 1. Username-rename squatting against organizations

GitHub releases a username for reuse shortly after the original owner renames their account. Combined with a `username`-keyed `github_membership`, this enables a concrete attack:

1. An org member (call them `alice`) renames themselves to `alice-new`.
2. The old login `alice` becomes claimable by anyone on GitHub.
3. An attacker registers `alice`.
4. On the next `terraform apply`, the provider sees `alice` is no longer in the org and reconciles toward the declared config — **re-inviting `alice` into the organization**. The attacker accepts and is now a member (or admin, depending on the resource's `role`) of the org, fully via the org owner's own Terraform pipeline.

This isn't theoretical — it's the standard GitHub-username-squat pattern applied to IaC. The org owner has no signal anything is wrong; their drift detection passes because the config still says `alice = member`.

With this PR, when the membership is keyed by `user_id` (the immutable numeric ID), step 4 doesn't happen: the resource follows the renamed human, not the released string. The squatter inherits nothing.

This is an **opt-in** improvement — users who keep using `username` remain exposed. We may want to (in a follow-up) recommend `user_id` in the docs more strongly, or even mark `username` as soft-deprecated for the resource. Not in this PR; flagging for discussion.

#### 2. Admin-role continuity through rename

Same mechanism, sharper edge: an org admin renaming themselves currently causes the `username`-keyed resource to drift to "not a member". Depending on `downgrade_on_destroy` and apply ordering, this can produce confusing state — Terraform thinks the admin is gone while GitHub still has them as admin under the new login. `user_id`-keyed resources keep role enforcement tied to the human, so privilege drift stays observable.

#### 3. New attack surface this PR adds (honest accounting)

- **Lazy ID migration in `Read`**: one extra `Users.GetByID` request per legacy resource on first refresh. Same TLS, same configured token, same audit trail. No new trust boundary.
- **`TestAccGithubMembershipRenameResilience`**: the new live-rename test performs `PATCH /user` against the account whose PAT is supplied as `GH_TEST_EXTERNAL_USER_TOKEN`. Operationally important: this token must belong to a **dedicated throwaway test account**, never a real contributor's personal GitHub account, because the test mutates the authenticated user's login. The test skips automatically when the env var is unset (so CI without it is safe). Worth a line in `CONTRIBUTING.md` if maintainers want me to add it — happy to do so in this PR or a follow-up.

#### 4. What this does NOT do

- Doesn't fix existing `username`-keyed state retroactively until users explicitly migrate their HCL to `user_id`. The ID-format migration in `Read` only converts state representation; it doesn't change the identification semantics chosen in HCL.
- Doesn't apply to `github_team_membership` (out of scope for #524, but the same threat model applies there — possible follow-up).

Happy to add a `## Security` section to the resource docs spelling this out for end users if reviewers think it belongs in this PR rather than a follow-up.

mczaplinski:
> ### Security implications worth calling out
> 
> 1. An org member (call them `alice`) renames themselves to `alice-new`.
> 2. The old login `alice` becomes claimable by anyone on GitHub.
> 3. An attacker registers `alice`.
> 4. On the next `terraform apply`, the provider sees `alice` is no longer in the org and reconciles toward the declared config — **re-inviting `alice` into the organization**. The attacker accepts and is now a member (or admin, depending on the resource's `role`) of the org, fully via the org owner's own Terraform pipeline.
> 

That's why I'm not using this provider for user management as of now. Mapping members by mutable usernames is just too risky for me. I hope this gets merged!

PR #3444, opened 2026-05-25
chore: Update valid permissions for organization repository role

w1mvy:
### Before the change?

- `validRolePermissions` in `github/resource_github_organization_repository_role.go` was a snapshot of the GitHub repository fine-grained permissions API and had drifted from the upstream response. It rejected newer permissions (`read_copilot_agent_settings`, `write_copilot_agent_settings`, `write_repository_agent_secrets`, `write_repository_agent_variables`) at plan time via `CustomizeDiff`, even though the GitHub API would accept them.
- The doc-link comment above the slice referenced `apiVersion=2022-11-28`, which is no longer the current version on the linked GitHub docs page.

### After the change?

- The four permissions above are added to `validRolePermissions` (kept alphabetically sorted) so configurations using them now pass validation.
- The doc-link comment is bumped to `apiVersion=2026-03-10` to match the current GitHub docs page for `GET /orgs/{org}/repository-fine-grained-permissions`.
- The list was generated by calling `GET /orgs/{org}/repository-fine-grained-permissions` against a real org and diffing the returned `name`s against the existing slice; only additions were observed (no permissions removed upstream).

### Pull request checklist

Notes:
- `make generatedocs` produced no diff (the generated docs do not enumerate individual permissions).
- `make fmt`, `make lint`, and `make test` all pass locally.

- [ ] Yes
- [x] No

### DIFF
--- a/github/resource_github_organization_repository_role.go
+++ b/github/resource_github_organization_repository_role.go
@@ -200,7 +200,7 @@ func resourceGithubOrganizationRepositoryRoleDelete(ctx context.Context, d *sche
 	return nil
 }
 
-// Snapshot of the response to https://docs.github.com/en/enterprise-cloud@latest/rest/orgs/custom-roles?apiVersion=2022-11-28#list-repository-fine-grained-permissions-for-an-organization
+// Snapshot of the response to https://docs.github.com/en/enterprise-cloud@latest/rest/orgs/custom-roles?apiVersion=2026-03-10#list-repository-fine-grained-permissions-for-an-organization
 // The endpoint isn't covered in the SDK yet.
 var validRolePermissions = []string{
 	"add_assignee",
@@ -235,6 +235,7 @@ var validRolePermissions = []string{
 	"push_protected_branch",
 	"read_code_quality",
 	"read_code_scanning",
+	"read_copilot_agent_settings",
 	"reopen_discussion",
 	"reopen_issue",
 	"reopen_pull_request",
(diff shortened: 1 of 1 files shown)


github/resource_github_organization_repository_role.go
```diff
@@ -200,7 +200,7 @@ func resourceGithubOrganizationRepositoryRoleDelete(ctx context.Context, d *sche
 	return nil
 }
 
-// Snapshot of the response to https://docs.github.com/en/enterprise-cloud@latest/rest/orgs/custom-roles?apiVersion=2022-11-28#list-repository-fine-grained-permissions-for-an-organization
+// Snapshot of the response to https://docs.github.com/en/enterprise-cloud@latest/rest/orgs/custom-roles?apiVersion=2026-03-10#list-repository-fine-grained-permissions-for-an-organization
```
deiga: issue: Please don't change the API version, since the SDK hasn't upgraded API Version yet either.

github/resource_github_organization_repository_role.go
```diff
@@ -200,7 +200,7 @@ func resourceGithubOrganizationRepositoryRoleDelete(ctx context.Context, d *sche
...
 // The endpoint isn't covered in the SDK yet.
```
deiga: This line isn't accurate anymore: https://github.com/google/go-github/blob/master/github/orgs_custom_repository_roles.go#L168

deiga (changes requested):

stevehipwell (commented):
@deiga I think we were planning on revisiting this once I got the endpoint added to go-github. We could generate the data, but it would still need a release to keep it up to date. How about we look at adding a singleton pattern to make it available in the provider?

github/resource_github_organization_repository_role.go
w1mvy: Thanks for the review, fixed in this commit [2331a87](https://github.com/integrations/terraform-provider-github/pull/3444/commits/2331a87167da9ef53f98c9da012059f5c69b94c5)

PR #3448, opened 2026-05-26
feat: Add new gh client implementation

### LINKED ISSUE #849: RateLimitTransport incompatible with v4 api

### LINKED ISSUE #2709: [FEAT]: Bypass preemptive rate limit check  by go-github

### LINKED ISSUE #2925: [MAINT]: Update transport implementation

### LINKED ISSUE #3103: [MAINT]: Logic flaws with etag support

stevehipwell:
> [!WARNING]
> This is a work in progress. There is a known issue for tests on resources that rely on other resources to mutate them not passing due to internal GitHub logic. These would be fixed by adopting a test strategy where the fixtures are created outside of Terraform.

Resolves #2925
Resolves #3103
Resolves #2709
Resolves #849

----

### Before the change?

- Legacy client implementation

### After the change?

- New client implementation (behind `legacy_client = false` provider setting)
  - Safer provider configuration
  - Refreshable app token
  - Transport level response cache
    - Persistable using `cache_path` provider setting
  - Idiomatic retry handling
  - Idiomatic rate limit handling
  - Base support for future multi-owner pattern
  - Dropped support for `gh` CLI auth

### Pull request checklist

- [ ] Yes
- [x] No

----


github/config.go
```diff
@@ -16,17 +16,23 @@ import (
...
+	AppInstallationID *string
+	Token             string
+	ReadDelay         time.Duration
+	WriteDelay        time.Duration
+	RetryDelay        time.Duration
+	RetryableErrors   map[int]bool
+	MaxRetries        int
+	ParallelRequests  bool
+	Insecure          bool
+	CachePath         *string
```
deiga: suggestion: Could we sort these alphabetically? That would make eyeing through the list simpler IMO

github/provider.go
```diff
@@ -37,7 +41,6 @@ func NewProvider() func() *schema.Provider {
 				"organization": {
 					Type:        schema.TypeString,
 					Optional:    true,
-					DefaultFunc: schema.EnvDefaultFunc("GITHUB_ORGANIZATION", nil),
```
deiga, line 40: 🙏

github/provider.go
```diff
@@ -100,7 +103,7 @@ func NewProvider() func() *schema.Provider {
 					Type:        schema.TypeList,
 					Elem:        &schema.Schema{Type: schema.TypeInt},
 					Optional:    true,
-					Description: "List of HTTP status codes that should be retried; if not set this uses the provider defaults. This setting only applies when `max_retries` is greater than `0`.",
+					Description: "List of HTTP status codes that should be retried; if not set this uses the provider defaults. This setting only applies when `max_retries` is greater than `0`. This is ignored for the REST API when `legacy_client` is `false` since the new client implementation handles the retry logic.",
```
deiga, line 111: suggestion: Would it make sense to add validation that these aren't set with `legacy_client = false`?

github/provider.go
```diff
@@ -118,13 +121,26 @@ func NewProvider() func() *schema.Provider {
...
 					Default:     false,
-					Description: "Allow the provider to make parallel API calls; this is experimental and may cause concurrency and rate limiting issues.",
+					Description: "Allow the provider to make parallel API calls; this is experimental and may cause concurrency and rate limiting issues. This is ignored for the REST API when `legacy_client` is `false` since the new client implementation is designed to safely handle parallel requests.",
 				},
 				"insecure": {
 					Type:        schema.TypeBool,
 					Optional:    true,
 					Default:     false,
 					Description: "Allow insecure server connections when using SSL.",
+					Deprecated:  "This argument is deprecated as it's currently not used and will be removed in the next major version.",
```
deiga, line 138: 😂

github/provider.go
```diff
@@ -302,150 +318,189 @@ func NewProvider() func() *schema.Provider {
...
+
+		if config.AppID == nil {
+			if _, ok := d.GetOk("app_auth"); ok {
+				return nil, diag.Errorf("app_auth block is set but required fields are missing or contains empty values")
 			}
 
-			token = appToken
+			if v, ok := d.GetOk("token"); ok {
+				if s, ok := v.(string); ok && s != "" {
+					tflog.Info(ctx, "Using token from provider configuration.", map[string]any{"token": "****"})
```
deiga: question: Is the `map[string]any{"token": "****"}` useful here?

github/provider.go
```diff
@@ -302,150 +318,189 @@ func NewProvider() func() *schema.Provider {
...
-		log.Printf("[INFO] Setting write_delay_ms to %d", writeDelay)
 
-		readDelay := d.Get("read_delay_ms").(int)
-		if readDelay < 0 {
-			return nil, diag.Errorf("read_delay_ms must be greater than or equal to 0ms")
+		if v, ok := d.GetOk("read_delay_ms"); ok {
+			if i, ok := v.(int); ok {
+				if i < 0 {
+					return nil, diag.Errorf("read_delay_ms must be greater than or equal to 0")
+				}
```
deiga: suggestion: We could shorten this part if we'd use `ValidateDiagFunc` on the provider attribute (same for most of these validations that do independent validation)

github/provider.go
```diff
@@ -454,6 +509,93 @@ func configureProvider() func(context.Context, *schema.ResourceData) (any, diag.
 	}
 }
 
+// configureProviderMeta initializes the provider metadata, including setting up the GitHub API clients based on the provided configuration. It returns the initialized metadata or an error if the configuration is invalid or if there are issues initializing the clients.
+func configureProviderMeta(ctx context.Context, c *Config) (any, error) {
+	owner := &Owner{
+		name:        c.Owner,
+		StopContext: context.Background(),
```
deiga: question: With the context aware functions `StopContext` is unnecessary, right? Assuming that all resources where it's being used use Context-aware, then we could replace `StopContext` with `ctx`?

Not necessary to do in this PR

github/provider.go
```diff
@@ -468,34 +610,75 @@ func ghCLIHostFromAPIHost(host string) string {
...
 		ghCliPath = "gh"
 	}
 
 	host := ghCLIHostFromAPIHost(u.Host)
 
-	out, err := exec.Command(ghCliPath, "auth", "token", "--hostname", host).Output()
+	out, err := exec.CommandContext(ctx, ghCliPath, "auth", "token", "--hostname", host).Output()
 	if err != nil {
-		log.Printf("[DEBUG] Error getting token from GitHub CLI: %s", err.Error())
+		tflog.Debug(ctx, "Error getting token from GitHub CLI; ensure GitHub CLI is installed and authenticated if you intend to use it for authentication.", map[string]any{"error": err.Error()})
```
deiga, line 626: Should this use `Error` level as it's a fallback option for token?
```suggestion
		tflog.Error(ctx, "Error getting token from GitHub CLI; ensure GitHub CLI is installed and authenticated if you intend to use it for authentication.", map[string]any{"error": err.Error()})
```

internal/ghclient/graphql.go
```diff
@@ -0,0 +1,52 @@
...
+		return nil, fmt.Errorf("failed to create transport: %w", err)
+	}
+
+	client := &http.Client{Transport: tr, Timeout: clientTimeout}
+
+	if opts.GraphQLURL == nil {
+		return githubv4.NewClient(client), nil
+	}
+
+	return githubv4.NewEnterpriseClient(*opts.GraphQLURL, client), nil
```
deiga, line 51: question: Is `NewEnterpriseClient` correct here? Oh, it is, because is not `nil`, right?
I wonder if it would make sense to swap the condition and the "default" client to make the flow clearer?

deiga (commented):
This is looking quite nice!

github/provider.go
stevehipwell: There is no point to this as the property is deprecated and handed separately. I kept the owner one as it's wanted when we cut the next major and remove the hack.

github/provider.go
stevehipwell: I need to double check this, but couldn't see it being used.

github/provider.go
stevehipwell: Probably not.

github/provider.go
stevehipwell: I've tried to keep this SemVer compliant, so left it alone. It should never have been implemented like this anyway. I can remove all usage and just leave it orphaned.

In v7 I'd like to move the provider code into an internal `provider` package so there is no more SemVer on the actual code.

github/provider.go
stevehipwell: It's optional, we support anonymous. This needs to go in v7.

internal/ghclient/graphql.go
stevehipwell: We could do, but we always use the enterprise path. This can change when we remove legacy support.

github/provider.go
deiga: Yeah, I checked as well and came to the same conclusion

github/provider.go
stevehipwell: Just looking at the code, I don't think this is worth the effort given they have defaults.

stevehipwell:
@deiga I've addressed most of your comments.

github/config.go
```diff
@@ -92,30 +100,19 @@ func (c *Config) AnonymousHTTPClient() *http.Client {
...
-	v3client, err := github.NewClient(github.WithHTTPClient(client), github.WithURLs(new(c.BaseURL.JoinPath(path).String()), nil))
+	v3client, err := github.NewClient(github.WithHTTPClient(client), github.WithURLs(new(c.BaseURL.JoinPath(c.RESTAPIPath).String()), nil))
 	if err != nil {
 		return nil, err
 	}
 
 	return v3client, nil
 }
 
+// Deprecated: This is no longer required as [configureProviderMeta] is now used to configure the provider meta parameter with the necessary clients and owner information. Use [configureProviderMeta] instead.
```
deiga, line 115: suggestion: I think you removed the only call to this function in `c.Meta()` refactor. Can this be deleted?

github/provider.go
```diff
@@ -302,150 +324,178 @@ func NewProvider() func() *schema.Provider {
...
+		}
 
-			if v, ok := appAuthAttr["installation_id"].(string); ok && v != "" {
-				appInstallationID = v
-			} else {
-				return nil, diag.Errorf("app_auth.installation_id must be set and contain a non-empty value")
+		// TODO: In v7 remove organization and the associated backwards compatibility code, and require owner to be set either via the provider configuration or GITHUB_OWNER environment variable with the provider configuration taking precedence.
+		if v, ok := d.GetOk("organization"); ok {
+			if s, ok := v.(string); ok && s != "" {
+				tflog.Debug(ctx, "Using organization environment variable or attribute as owner.", map[string]any{"owner": s})
```
deiga, line 368: I think you removed the environment variable being read?
```suggestion
				tflog.Debug(ctx, "Using organization attribute as owner.", map[string]any{"owner": s})
```

github/provider.go
```diff
@@ -302,150 +324,178 @@ func NewProvider() func() *schema.Provider {
...
-			if isGHES {
-				apiPath = GHESRESTAPIPath
+		if config.Owner == "" {
+			if v, ok := d.GetOk("owner"); ok {
+				if s, ok := v.(string); ok && s != "" {
+					tflog.Debug(ctx, "Using owner attribute as owner.", map[string]any{"owner": s})
+					config.Owner = s
+				}
 			}
+		}
```
deiga, line 387: question: This prefers `GITHUB_OWNER` over the configured attribute, do we still want to keep it?
We have this note in the docs: 
> It is a bug that GITHUB_OWNER takes precedence over owner

internal/ghclient/anonymous_test.go
```diff
@@ -0,0 +1,105 @@
...
+		if err != nil {
+			t.Fatalf("failed to create anonymous source: %v", err)
+		}
+
+		defaultClient, err := source.RESTClient()
+		if err != nil {
+			t.Fatalf("failed to get default rest client: %v", err)
+		}
+
+		ownerClient, err := source.OwnerRESTClient(context.Background(), "octocat")
```
deiga:
```suggestion
		ownerClient, err := source.OwnerRESTClient(t.Context(), "octocat")
```

internal/ghclient/anonymous_test.go
```diff
@@ -0,0 +1,105 @@
...
+		if err != nil {
+			t.Fatalf("failed to create anonymous source: %v", err)
+		}
+
+		defaultClient, err := source.GraphQLClient()
+		if err != nil {
+			t.Fatalf("failed to get default graphql client: %v", err)
+		}
+
+		ownerClient, err := source.OwnerGraphQLClient(context.Background(), "octocat")
```
deiga:
```suggestion
		ownerClient, err := source.OwnerGraphQLClient(t.Context(), "octocat")
```

internal/ghclient/app_test.go
```diff
@@ -0,0 +1,217 @@
...
+						orgRequests.Add(1)
+						w.WriteHeader(http.StatusOK)
+						_, _ = w.Write([]byte(`{"id": 1001}`))
+						return
+					}
+
+					w.WriteHeader(http.StatusNotFound)
+				}))
+
+				firstClient, err := tt.callClient(context.Background(), source, "acme")
```
deiga:
```suggestion
				firstClient, err := tt.callClient(t.Context(), source, "acme")
```

internal/ghclient/app_test.go
```diff
@@ -0,0 +1,217 @@
...
+				if err != nil {
+					t.Fatalf("failed to get first owner client: %v", err)
+				}
+
+				secondClient, err := tt.callClient(context.Background(), source, "acme")
```
deiga:
```suggestion
				secondClient, err := tt.callClient(t.Context(), source, "acme")
```

internal/ghclient/app_test.go
```diff
@@ -0,0 +1,217 @@
...
+			t.Run(tt.name, func(t *testing.T) {
+				t.Parallel()
+
+				var orgRequests atomic.Int32
+				var userRequests atomic.Int32
+				source := mustTestAppSource(t, http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+					tt.handleRequest(w, r, &orgRequests, &userRequests)
+				}))
+
+				installationID, err := source.GetInstallationID(context.Background(), tt.owner)
```
deiga:
```suggestion
				installationID, err := source.GetInstallationID(t.Context(), tt.owner)
```

internal/ghclient/token_test.go
```diff
@@ -0,0 +1,131 @@
...
+		if err != nil {
+			t.Fatalf("failed to create token source: %v", err)
+		}
+
+		defaultClient, err := source.RESTClient()
+		if err != nil {
+			t.Fatalf("failed to get default rest client: %v", err)
+		}
+
+		ownerClient, err := source.OwnerRESTClient(context.Background(), "octocat")
```
deiga:
```suggestion
		ownerClient, err := source.OwnerRESTClient(t.Context(), "octocat")
```

internal/ghclient/token_test.go
```diff
@@ -0,0 +1,131 @@
...
+		if err != nil {
+			t.Fatalf("failed to create token source: %v", err)
+		}
+
+		defaultClient, err := source.GraphQLClient()
+		if err != nil {
+			t.Fatalf("failed to get default graphql client: %v", err)
+		}
+
+		ownerClient, err := source.OwnerGraphQLClient(context.Background(), "octocat")
```
deiga:
```suggestion
		ownerClient, err := source.OwnerGraphQLClient(t.Context(), "octocat")
```

deiga (changes requested):
Partial review

github/provider.go
```diff
@@ -37,7 +42,6 @@ func NewProvider() func() *schema.Provider {
 				"organization": {
 					Type:        schema.TypeString,
 					Optional:    true,
-					DefaultFunc: schema.EnvDefaultFunc("GITHUB_ORGANIZATION", nil),
 					Description: "GitHub organization to manage. This can also be set by the `GITHUB_ORGANIZATION` environment variable.",
 					Deprecated:  "This argument is deprecated and will be removed in a future major release; use `owner` instead.",
```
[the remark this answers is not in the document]
stevehipwell: This is incorrect as the env variable has always been independently evaluated so this is uneccessary.

github/provider.go
```diff
@@ -302,150 +324,178 @@ func NewProvider() func() *schema.Provider {
...
+					return nil, diag.FromErr(err)
+				}
+				config.Token = appToken
+			}
+
+			if config.Token == "" {
+				tflog.Debug(ctx, "No token found, using GitHub CLI to get token from base URL.", map[string]any{"base_url": config.BaseURL.String()})
+				config.Token = tokenFromGHCLI(ctx, config.BaseURL)
+			}
 		}
```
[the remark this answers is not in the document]
stevehipwell: This is the desired behaviour.

github/provider.go
```diff
@@ -454,6 +504,92 @@ func configureProvider() func(context.Context, *schema.ResourceData) (any, diag.
...
+			CachePath:    c.CachePath,
+			RetryMax:     c.MaxRetries,
+			RetryWaitMin: c.RetryDelay,
+			RetryWaitMax: c.RetryDelay,
+		}
+
+		var source ghclient.Source
+		if c.AppID != nil {
+			appSource, err := ghclient.NewAppSource(*c.AppID, c.AppPEM, options)
+			if err != nil {
```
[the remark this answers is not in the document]
stevehipwell: This is the intended behaviour and as an app can only be installed once per owner the behaviour has to be equivelent.

github/config.go
stevehipwell: Not while keeping to SemVer as it's exported.

github/provider.go
stevehipwell: I actually need to add the env variable back.

github/provider.go
stevehipwell: There is a note that this will be reverted on v7, it'd add additional complexity to just change this for the non-legacy client.

internal/ghclient/cache.go
```diff
@@ -0,0 +1,39 @@
...
+	var cacheDir string
+	if path != nil && *path != "" {
+		cacheDir = filepath.Join(*path, "terraform-provider-github")
+	} else {
+		tmpDir, err := os.MkdirTemp("", "terraform-provider-github-*")
+		if err != nil {
+			return nil, fmt.Errorf("failed to create temporary cache directory: %w", err)
+		}
+		cacheDir = tmpDir
+	}
```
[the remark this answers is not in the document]
stevehipwell: There is no provider shutdown hook to tap into so the provider falls back to the OS temp dir which should be automatically cleaned up.

github/config.go
deiga: How would that impact the provider? Could someone be using the provider as a library?

stevehipwell: They shouldn't be but I'm operating with Hyrum's law in mind. In v7 I propose we move all of the code into `internal/provider` so we don't have any of this.

stevehipwell:
@deiga please could you re-review?

deiga (approved):

stevehipwell:
@robert-crandall please could I get a review on this.

robert-crandall (approved):

kishaningithub:
Where does this disable go-github's annoying rate limit check before hitting the transport layer?

stevehipwell:
If you enable the new (experimental) client by setting `legacy_client` to `false` it [disables](https://github.com/integrations/terraform-provider-github/blob/e286cb04d9a8cb4dca56b041cbb230715850a92b/internal/ghclient/rest.go#L44) the native go-github rate limit check a well as the legacy provider rate limit implementation. It then uses an instance of [gofri/go-github-ratelimit](github.com/gofri/go-github-ratelimit) per client. It also adds [bored-engineer/github-conditional-http-transport](github.com/bored-engineer/github-conditional-http-transport) to transparently handle unchanged resources at the transport layer instead of adding complexity into each resource method.

PR #3456, opened 2026-05-28
feat: add resource github_repository_files

### LINKED ISSUE #2909: [MAINT]: Improve UX for github_repository_file when parallel_requests = true
### Describe the need

When the provider has set the option `parallel_requests = true`
`github_repository_file` tends to fail a lot with HTTP 409 errors when more than one file is updated in the same repository, see log output further down.

Ideally I would still like the requests to run in parallel for performance, just without the 409 errors occurring. Perhaps the resource could add an extra attribute to ignore intermediary changes to the commit hash? 

### SDK Version

_No response_

### API Version

_No response_

### Relevant log output

Sample config

```hcl
resource "github_repository_file" "tenant" {
  for_each = {
    for ns in local.tenant_namespaces :
    ns.name => ns
    if ns.flux.enabled
  }

  repository    = data.github_repository.cluster.name
  branch        = local.cluster_branch
  file          = "tenants/${local.cluster_id}/${each.key}.yaml"
...
}

```

```shell
Error: PUT https://api.github.com/repos/xxx/yyy/contents/tenants/test01/myfile.yaml: 409 is at 30e15e8485d2808d7f9cdf27cecbb145252e1427 but expected 46affa9e9f78ca99714ed0e12a9c44e94e1e9055 []
│ 
│   with github_repository_file.tenant["myfile"],
│   on fluxcd.tf line 204, in resource "github_repository_file" "tenant":
│  204: resource "github_repository_file" "tenant" {
```

### Code of Conduct

novucs:
Resolves #2909

FYI: Most of the code in this PR was AI generated. I have personally built, tested, and reviewed it, applying to a dummy project I have on Github.

### Before the change?

With `parallel_requests = true`, many `github_repository_file` resources writing the same branch race each other. Each one is a separate Contents API PUT and therefore a separate commit, so the second concurrent write sees a stale branch SHA and GitHub rejects it with 409.

### After the change?

Adds `github_repository_files`, which manages a set of files on a branch and writes them in one commit per apply via the Git Data API: get branch head, create a tree from the head tree with the changed entries, create the commit, update the ref. Files not listed in the resource are never touched. If the ref moves mid-commit the commit is rebuilt on the new head and retried. Adopting files whose content already matches produces no commit.

### Pull request checklist

- [x] Schema migrations have been created if needed

- [ ] Yes
- [x] No


mloiseleur:
This feature looks very interesting for our use case. 

@deiga @stevehipwell Is there anything we can do to help on this? 

For instance, would that help to have some real-world tests? Or a refresh of this code to include the new client?

deiga:
@mloiseleur what is the use case ypu have?

mloiseleur:
@deiga In fact, I have two that are solved with this:

1. Use the new client with parallelism **and** push CI files on CI-managed repositories. 
   - Currently, with the new client, I'm forced to set parallelism to 1 to avoid 409.
2. Apply all the CI files in one commit, like it would have been the case with a manual PR-based workflow updating CI.

stevehipwell (changes requested):
Thanks for the PR @novucs. Please could you review the contribution guide for how the code should be structured (or look for resources in `RESOURCES.md` with all ticks).

At first glance the following high level changes are required.

- Don't call read from create/update
- Use the latest testing pattern
- Use automated doc generation
- If you need to do non-resource work (e.g. building a commit from blob tree etc) put it into a util file

novucs:
@stevehipwell thanks for your review! I think this PR was collecting some dust. I've rebased and cleaned up the code a bunch. Create/Update no longer call Read, the tests use the new pattern I took from `github_repository_collaborators`, the docs are generated from the schema, and the commit building has moved to `util_git.go`.

github/resource_github_repository_files.go
```diff
@@ -0,0 +1,387 @@
...
+							Type:        schema.TypeString,
+							Computed:    true,
+							Description: "The blob SHA of the file's current content on the branch.",
+						},
+					},
+				},
+				Set: func(v any) int {
+					file, _ := v.(map[string]any)
+					path, _ := file["path"].(string)
+					return schema.HashString(path)
```
deiga: Note: this ignores all other fields changes after create, is that intentional?

github/resource_github_repository_files.go
```diff
@@ -0,0 +1,387 @@
...
+	if err != nil {
+		return diag.FromErr(err)
+	}
+
+	shas, err := getRepositoryBlobSHAs(ctx, client, owner, repoName, treeSHA)
+	if err != nil {
+		return diag.FromErr(err)
+	}
+
+	id, err := buildID(repoName, branch)
```
deiga: issue: this would prevent any other resource on this repo and branch, it needs something additional to distinguish

deiga (changes requested):
Partial review

github/resource_github_repository_files.go
novucs: Yes, this was intentional. This only sets the element identity. If content changes, we'll still see a diff. If we popped content into this hash, we'll end up seeing additions/deletions rather than changes of the same file. The configUpdated test covers this case.

github/resource_github_repository_files.go
novucs: This will not prevent multiple instances of this resource on the same repo and branch. Resources in terraform are tracked by their resource address, nothing in core or SDKv2 checks that two instances have distinct IDs. I've tested this against a locally built version of this code. We could add the set of paths to the ID, but that'll be introducing the config to the ID, which'll cause the ID to be unstable and will change on every file addition/removal on this resource.

novucs: Thanks again for your time in reviewing this by the way! Happy to adjust if you feel we could be handling this better another way. I suppose an alternative would be us introducing a user-supplied key

github/resource_github_repository_files.go
novucs: Really good check. It was intentional, but I was wrong. I'd even tested this, but screwed up. I'd thought changes here propagated as additions/deletions rather than in-place changes. Turns out, they didn't propagate at all 🤦. I've pushed a fix, content changes now produce diffs as you'd expect

github/resource_github_repository_files.go
deiga: AFAIK the assumption is that ID is unique, whether it's required to be so or not. 
I wonder if it would make sense to tie the ID to the commit hash? Since the resource seems like it's the commit of the files, not the files themselves 🤔 
Though the current logic updates commit hash on each read? That doesn't sound like a great experience, it would mean constant drift and a new commit on each apply? 🤔 

Could you add a acceptance test which does an out-of-band edit to the branch, without modifying the files? The test should assert that there are no changes to be done

github/util_git.go
```diff
@@ -0,0 +1,80 @@
...
+	}
+
+	return commit.GetSHA(), commit.GetTree().GetSHA(), nil
+}
+
+func getRepositoryBlobSHAs(ctx context.Context, client *github.Client, owner, repo, treeSHA string) (map[string]string, error) {
+	tree, _, err := client.Git.GetTree(ctx, owner, repo, treeSHA, true)
+	if err != nil {
+		return nil, err
+	}
```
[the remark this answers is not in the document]
novucs: https://github.com/integrations/terraform-provider-github/pull/3456/commits/a1f2a0a4a655745389d5e7d554832aa5b99a963f

github/resource_github_repository_files.go
```diff
@@ -0,0 +1,404 @@
...
+				Description: "The tree SHA of the branch head after the most recent commit created by this resource.",
+			},
+		},
+	}
+}
+
+func diffRepositoryFiles(_ context.Context, diff *schema.ResourceDiff, _ any) error {
+	_, err := expandRepositoryFiles(diff.Get("file"))
+	return err
+}
```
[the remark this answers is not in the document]
novucs: https://github.com/integrations/terraform-provider-github/pull/3456/commits/89505120e2c78b92f9dfbbb0a574a49709ebc485

github/resource_github_repository_files.go
```diff
@@ -0,0 +1,404 @@
...
+		return diag.FromErr(err)
+	}
+	d.SetId(id)
+
+	if err := d.Set("repository_id", int(repository.GetID())); err != nil {
+		return diag.FromErr(err)
+	}
+	if err := d.Set("commit_message", message); err != nil {
+		return diag.FromErr(err)
+	}
```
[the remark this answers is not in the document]
novucs: https://github.com/integrations/terraform-provider-github/pull/3456/commits/d866c1500c136c7cd44000459a34cbf6b5aee03f

github/resource_github_repository_files_test.go
```diff
@@ -0,0 +1,293 @@
...
+
+	"github.com/hashicorp/terraform-plugin-testing/compare"
+	"github.com/hashicorp/terraform-plugin-testing/helper/resource"
+	"github.com/hashicorp/terraform-plugin-testing/knownvalue"
+	"github.com/hashicorp/terraform-plugin-testing/plancheck"
+	"github.com/hashicorp/terraform-plugin-testing/statecheck"
+	"github.com/hashicorp/terraform-plugin-testing/tfjsonpath"
+)
+
+func TestAccGithubRepositoryFiles(t *testing.T) {
```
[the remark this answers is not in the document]
novucs: https://github.com/integrations/terraform-provider-github/pull/3456/commits/93b4f57294cd62f59d86969c7cc2e8381ab963c7

github/resource_github_repository_files.go
novucs: Yeah good idea, commit SHA it is. Still means the ID will move on changes being made, but I prefer that to manually speccing out a key on each resource and keeps IDs unique per-resource. I've pushed the changes with the OOB edit acceptance test

PR #3461, opened 2026-05-30
fix: skip github_team_members deletion when parent team is already gone

### LINKED ISSUE #3411: [MAINT]: Skip `github_team_members` deletion when the associated `github_team` is deleted
### Describe the need

When a `github_team` is deleted, all `github_team_members` are removed by the provider one by one.

We can completely skip that step because deleting the team will make these resources deleted either way.

### SDK Version

_No response_

### API Version

_No response_

### Relevant log output

```shell

```

### Code of Conduct

nitinjain999:
Resolves #3411

----

### Before the change?

When `github_team` and `github_team_members` are destroyed together, the team deletion removes all members server-side. The subsequent `github_team_members` destroy makes N individual `RemoveTeamMembershipByID` API calls that all return 404 and fail loudly.

Additionally, if a team is deleted outside of Terraform, the next `terraform plan` or `terraform refresh` would not remove `github_team_members` from state — it would either error or silently return an empty member list, causing a perpetual diff.

### After the change?

**Delete:** On a 404 from `RemoveTeamMembershipBySlug`/`RemoveTeamMembershipByID`, the function returns cleanly. The GitHub API returns 204 (not 404) when a membership doesn't exist, so a 404 exclusively means the team itself is gone mid-delete.

**Read:** Switched from GraphQL (`v4client`) to REST (`v3client`), making two paginated `ListTeamMembersByID/Slug` calls filtered by role (`member` and `maintainer`). A 404 response removes the resource from state cleanly.

**All functions (Create, Update, Read, Delete):** Use `newLegacyTeamIdentity` to dispatch to the slug-based or ID-based API methods without an upfront lookup, consistent with the pattern in `resource_github_repository_collaborators`.

### Pull request checklist

- [x] Schema migrations have been created if needed — not applicable, no schema changes
- [x] Docs have been reviewed and added / updated if needed — no doc changes needed, no schema or user-visible behaviour change

- [ ] Yes
- [x] No

----


nitinjain999:
Tested against a real GitHub org (`nitin-sv`) using `make testacc T=TestAccGithubTeamMembers`.

The new subtest `succeeds_when_parent_team_is_deleted_out-of-band_before_destroy` passes — team deleted via API before destroy, Terraform completes cleanly with no 404 error.

The existing `creates_a_team_&_members_configured_with_defaults` subtest fails in this environment because the test user is an org owner, and GitHub silently promotes owners to `maintainer` regardless of the requested role. This is a pre-existing issue unrelated to this change and documented in the provider notes.

nitinjain999:
Addressed both Copilot review comments:

**Delete logic (line 297):** On a 404 from `RemoveTeamMembershipByID`, the code now calls `GetTeamByID` to distinguish between "team is gone" and "just this membership was removed out-of-band". If the team is gone it short-circuits; if the team still exists it skips that one user and continues the loop.

**Test coverage (line 102):** Added a `RefreshState: true` step that deletes the team out-of-band via `PreConfig`, then asserts the refresh removes `github_team_members` from state cleanly — covering both the `getTeamSlug` 404 and `DatabaseId==0` guard paths in Read. Verified passing locally against `nitin-sv`.

github/resource_github_team_members.go
```diff
@@ -270,6 +289,20 @@ func resourceGithubTeamMembersDelete(ctx context.Context, d *schema.ResourceData
 
 		_, err = client.Teams.RemoveTeamMembershipByID(ctx, orgId, teamId, username)
 		if err != nil {
+			if ghErr, ok := errors.AsType[*github.ErrorResponse](err); ok && ghErr.Response.StatusCode == http.StatusNotFound {
+				// 404 can mean the membership was already removed out-of-band (drift)
```
deiga: > 404 can mean the membership was already removed out-of-band

issue: Is this true? The API doesn't state a 404 and a quick test locally returned with 204 if USERNAME didn't exist as team member

github/resource_github_team_members_test.go
```diff
@@ -75,6 +75,64 @@ resource "github_team_members" "test" {
...
+  }
+}
+`, teamName, testAccConf.testOrgUser)
+
+		deleteTeamOutOfBand := func() {
+			meta, err := getTestMeta()
+			if err != nil {
+				t.Fatal(err.Error())
+			}
+			ctx := context.Background()
```
deiga: issue: Please use `t.Context()` to ensure this gets cancelled correctly

deiga (changes requested):

nitinjain999:
Addressing @deiga's comment on the 404 behavior:

You're right. I verified directly against the GitHub API: `RemoveTeamMembershipByID` (DELETE `/organizations/{orgId}/team/{teamId}/memberships/{username}`) returns **204** when the membership doesn't exist for a real GitHub user — a **404** exclusively indicates the team itself is gone.

The secondary `GetTeamByID` check and the incorrect drift-handling `continue` branch have been removed. A 404 now immediately short-circuits the delete loop without an extra API call.

Also verified with a live acceptance test against a real org. ✅ Fixed

github/resource_github_team_members.go
```diff
@@ -257,6 +272,10 @@ func resourceGithubTeamMembersDelete(ctx context.Context, d *schema.ResourceData
 	teamIdString := d.Get("team_id").(string)
 	teamId, err := getTeamID(ctx, meta, teamIdString)
 	if err != nil {
+		if ghErr, ok := errors.AsType[*github.ErrorResponse](err); ok && ghErr.Response.StatusCode == http.StatusNotFound {
+			tflog.Info(ctx, "Team no longer exists, skipping member deletion", map[string]any{"team_id": teamIdString})
+			return nil
+		}
 		return diag.FromErr(err)
 	}
```
deiga: What's the point of this block? Couldn't we save on another API call by removing this block and just trying to delete members? 

It should silently remove all members from state if the team doesn't exist anymore anyway.

deiga (changes requested):

nitinjain999:
Hey @deiga, the `getTeamID` call can't be removed — it's not just a team-existence check, it's also a slug resolver. The `team_id` attribute accepts either a numeric ID or a team slug, and `RemoveTeamMembershipByID` requires an `int64`. If someone passed a slug as `team_id`, skipping this would break deletion entirely regardless of team state.

The 404 guard there is a side benefit: if the team's already gone before we start, we bail immediately instead of firing N membership deletes and getting N 404s back.

That said — are you thinking there's a way to avoid the extra API call when team_id is already numeric? In that case `getTeamID` does a local parse with no network call (`parseTeamID`), so it's free. The extra call only happens for slug-based IDs. Does that change your thinking?

stevehipwell:
I think a better fix for the read part would be to switch to using the REST client as I can't see any reason to use the GraphQL client. Our default posture is to use the REST client wherever possible.

stevehipwell:
@nitinjain999 RE ID vs slug you can use `newLegacyTeamIdentity` to handle both cases.

```go
t := newLegacyTeamIdentity(teamID)
if slug, ok := t.getSlugOK(); ok {
	// Use non-legacy slug method
} else {
	id := t.getID()
	// Use legacy ID method
}
```

nitinjain999:
Thanks @stevehipwell.

On the Read refactor to REST: understood, will switch to `v3client`. The current GraphQL query fetches members with roles in one call; the REST equivalent would be `ListTeamMembersByID/Slug` for the member list, then `GetTeamMembershipByID/Slug` per user for their role — so it becomes N+1 calls instead of one. Is that the expected tradeoff, or is there a way to get member roles in a single REST call I'm missing?

On `newLegacyTeamIdentity` for Delete: will apply that pattern, thanks for pointing to it.

stevehipwell:
@nitinjain999 you can use the `newLegacyTeamIdentity` pattern in all of the functions to save additional calls to lookup the ID. RE the REST API, you can make two calls to get the users filtered by role.

nitinjain999:
@stevehipwell Both suggestions applied

1. **`newLegacyTeamIdentity` across all functions** — Create, Update, Read, and Delete now all use `newLegacyTeamIdentity` to dispatch to the slug or ID-based API method directly, no upfront lookup needed.

2. **Read switched to REST** — replaced the GraphQL `v4client` query with two paginated `ListTeamMembersByID/Slug` calls filtered by role (`member` and `maintainer`). A 404 removes the resource from state cleanly. This also eliminates the `getTeamSlug` pre-call and the `DatabaseId == 0` null-team guard that were needed before.

Tested against nitin-sv, the out-of-band team deletion acceptance test passes.

github/resource_github_team_members.go
stevehipwell:
```suggestion
	return nil
```

This isn't needed.

stevehipwell:
```suggestion
	return nil
```

This isn't needed.

github/resource_github_team_members.go
```diff
@@ -198,48 +194,48 @@ func resourceGithubTeamMembersRead(ctx context.Context, d *schema.ResourceData,
...
-			})
+	for _, role := range []string{"member", "maintainer"} {
+		opts := &github.TeamListTeamMembersOptions{
+			Role:        role,
+			ListOptions: github.ListOptions{PerPage: maxPerPage},
 		}
-		if !q.Organization.Team.Members.PageInfo.HasNextPage {
-			break
+
+		for {
```
stevehipwell: You should be able to use the Iter methods on the go-github client which would mkae this pattern cleaner.

stevehipwell (changes requested):
This is looking good @nitinjain999. I've added some inline feedback, I also think it's worth replacing calls to `log` with `tflog`.

nitinjain999:
@stevehipwell All three addressed in the latest commit:

- **Create/Update**: both now return `nil` directly instead of calling Read inline.
- **Read pagination**: switched to `ListTeamMembersByIDIter`/`ListTeamMembersBySlugIter` — the manual next-page loop is gone.

Tested, passing.

stevehipwell (changes requested):
Thanks for the quick responses @nitinjain999, the code structure is looking good so we should be ready to approve/merge once you've fixed the lint issues.

stevehipwell:
@nitinjain999 it still looks like there are a few lint issues, the exact pattern for `errors.AsType` is in the contribution guide.

nitinjain999:
Switched to `errors.AsType[*github.ErrorResponse](err)` as documented in ARCHITECTURE.md — missed that the project uses its own generic helper rather than stdlib `errors.As`. Both Read and Delete updated. Lint passes locally with `.golangci.strict.yml`. ✅ Fixed

stevehipwell:
> Switched to errors.AsType[*github.ErrorResponse](err) as documented in ARCHITECTURE.md — missed that the project uses its own generic helper rather than stdlib errors.As. Both Read and Delete updated. Lint passes locally with .golangci.strict.yml. ✅ Fixed

@nitinjain999 [errors.AsType](https://pkg.go.dev/errors#AsType) is part of the Go stdlib.

nitinjain999:
Thanks for the correction — I mistakenly assumed it was a project helper. Good to know it's stdlib since Go 1.23.

nitinjain999:
@stevehipwell : Removed `getTeamSlug` and `lookupTeamSlug` from `util_team.go` — both became dead code once all functions were switched to `newLegacyTeamIdentity`. ✅ Fixed

stevehipwell (approved):
LGTM

nitinjain999:
> LGTM

i need to wait for @deiga

deiga (approved):

PR #3474, opened 2026-06-04
fix: Add etag diff suppression to remaining resources (supersedes #3311)

### LINKED ISSUE #3085: github_organization_ruleset gets in bad state and can't recover

### LINKED ISSUE #3103: [MAINT]: Logic flaws with etag support

### LINKED ISSUE #3291: [BUG]: address etag drift in resources

mtb-xt:
This is #3311 by @ei-grad, rebased onto current `main` (the original branch had fallen 64 commits behind). All commits retain original authorship; review feedback from that PR is already incorporated. Opening this to keep the fix moving — @ei-grad, happy to close in favor of yours if you'd rather rebase the original.

## Summary

PR #2840 added `DiffSuppressFunc` and `DiffSuppressOnRefresh` to etag fields in 7 resources to prevent spurious diffs on plan/apply. This PR applies the same fix to the remaining 22 resources that were missed:

- `github_branch_protection_v3`
- `github_emu_group_mapping`
- `github_enterprise_actions_runner_group`
- `github_issue`
- `github_membership`
- `github_organization_webhook`
- `github_project_card`
- `github_project_column`
- `github_release`
- `github_repository_autolink_reference`
- `github_repository_deploy_key`
- `github_repository_ruleset`
- `github_team`
- `github_team_membership`
- `github_team_repository`
- `github_team_sync_group_mapping`
- `github_user_gpg_key`
- `github_user_ssh_key`
- `organization_block`

`TestEtagSchemaConsistency` covers all 29 resources (7 existing + 22 new) and runs `InternalValidate` on each.

## Answering the open review questions from #3311

**Why `Optional: true` on the etag fields?** ([thread](https://github.com/integrations/terraform-provider-github/pull/3311#discussion_r3010989648)) — The SDK requires it: `InternalValidate` rejects `DiffSuppressFunc` on Computed-only fields with _"There is no config for computed-only field, nothing to compare."_ This matches how the 7 resources fixed in #2840 are declared.

**What's the benefit of running `InternalValidate` in the consistency test?** ([thread](https://github.com/integrations/terraform-provider-github/pull/3311#discussion_r3036689576)) — It guards exactly the constraint above: if someone later removes `Optional` from an etag field (a natural-looking cleanup — it was in fact requested in the first review round), the test fails immediately instead of the provider breaking at runtime. The two review threads on #3311 are each other's justification.

## Real-world impact

The etag churn makes no-op plans non-reproducible: a planfile saved at PR-review time never byte-matches the apply-time plan, which breaks plan-equality gates in CI (e.g. `atmos terraform plan-diff` / Cloud Posse's terraform-apply GitHub Action). Byte-level evidence from our production environment: a no-op plan differs from its immediate re-plan in 6 of 7352 lines — five occurrences of a refreshed `github_repository_ruleset` etag, plus the plan timestamp. `lifecycle.ignore_changes = [etag]` cannot work around this since the attribute is provider-managed. Downstream report with details: https://github.com/orgs/cloudposse/discussions/125

## Verification

- `go build ./...`, `go vet ./github/` clean
- Full unit suite passes; all 28 `TestEtagSchemaConsistency` subtests pass
- Audited resources added to `main` since the original PR — no new etag fields need coverage

Fixes #3103
Fixes #3291
Likely fixes #3085


deiga:
@mtb-xt You don't want to continue on this?

mtb-xt:
> @mtb-xt You don't want to continue on this?

I just saw there was some traction in https://github.com/integrations/terraform-provider-github/pull/3311 ? I can reopen it if needed.

deiga:
It stalled 2w ago, so honestly either one works. 
Not that I can promise you when we'll get to merging 😬

deiga (dismissed):
LGTM

mtb-xt:
> LGTM

Sorry, last push was failing the checks. I've fixed it, could you please reapprove the run and merge

deiga (changes requested):
Please add Description fields to all the modified etag fields 🙏

mtb-xt:
> Please add Description fields to all the modified etag fields 🙏

Done, please have a look when you have a moment!

deiga (dismissed):
LGTM

robert-crandall (commented):
Thanks for picking this up and rebasing it - the etag work looks good and the test coverage across all 29 resources is a nice touch. One thing to fix before merge: the `github_release` change re-adds `ForceNew: true` to `discussion_category_name`, which looks like a rebase artifact from #3526 (main removed it and handles the field in the update path now). Since that would force releases to be recreated, I want to make sure this was intentional. Everything else looks ready to go.

mtb-xt:
@robert-crandall good catch - that was exactly a rebase artifact. 

Fixed in a69c46f: removed the stray ForceNew: true from discussion_category_name so it matches main and stays handled in the update path (no forced recreation), and restored the description main uses. Diffed the whole file against main to confirm that was the only stray change - the only intended delta left is the etag diff-suppression. 

I also updated the branch, and all this means we need new approvals to run checks :) 

Thanks!

deiga (approved):

robert-crandall (approved):
Thank you for making that fix!

stevehipwell:
@deiga @robert-crandall I don't think this PR should have been merged. The `etag` attribute is fundamentally broken so any work on it needs better discussion. In this case the "fix" of making the field optional to add a suppress function makes zero sense and fundamentally changes how the provider operates.

mtb-xt:
> In this case the "fix" of making the field optional to add a suppress function makes zero sense

 It unblocks a real and very annoying issue for us in production.

mtb-xt:
> @deiga @robert-crandall I don't think this PR should have been merged. The `etag` attribute is fundamentally broken so any work on it needs better discussion. In this case the "fix" of making the field optional to add a suppress function makes zero sense and fundamentally changes how the provider operates.

Discussion is absolutely fine, and if "etag" needs a broader rethink, that’s worth having.

But this PR addressed a real and fairly disruptive issue in day-to-day use of the provider. The change was also based on an approach that had already been merged elsewhere in the provider, rather than introducing a completely new pattern.

There’s always value in discussing the ideal long-term design, but there’s also value in unblocking users who are being affected by a concrete problem today. I think those two things can happen independently.

stevehipwell:
@mtb-xt it might have a second order effect of suppressing a diff, but it breaks the TF data model as `etag` is very much computed only.

The fix for this problem is to use `CustomizeDiff` to clear the `etag` field so it never shows up in any diff output while remaining available in requests.

PR #3489, opened 2026-06-15
fix: refresh GitHub App installation token on legacy client

### LINKED ISSUE #977: app_auth credentials expire after an hour

yashrajdighe:
Resolves #977

----

### Before the change?

GitHub App installation tokens expire after an hour. With `app_auth`, we were minting the token once during provider setup and stuffing it into a static OAuth source that never refreshed. Long `terraform apply` runs would start failing with `401 Bad credentials` once that hour was up.

The new client in `internal/ghclient` already handles refresh via `go-githubauth`, but App auth still went through the legacy path unless you set `legacy_client = false`.

### After the change?

App auth now always uses the new client, even when `legacy_client` is left at the default. No changes to the legacy client itself.

- Dropped the `GenerateOAuthTokenFromApp` call from provider setup — tokens are minted on demand and refreshed automatically.
- Passed `WithEnterpriseURL` through in `internal/ghclient` so GHES/custom `base_url` token minting hits the right host.

PAT/token and anonymous auth on the legacy client are untouched.

One thing to be aware of: the new client looks up the installation from `owner` (API call), not `app_auth.installation_id`. That field is still required in the schema but isn't used for provider auth on this path. `owner` is already required for App auth, so this should be fine for the usual case.

Also, `read_delay_ms`, `write_delay_ms`, `parallel_requests`, and `retryable_errors` don't apply to REST on the new client — same as documented today.

### Pull request checklist

- [ ] Schema migrations have been created if needed — N/A
- [x] Tests for the changes have been added
- [x] Docs reviewed — no schema changes, nothing to regenerate

### Local test validation

- `make fmt`, `make lint`, `make test`, `make build` all pass
- `Test_appSource_installationTokenRefresh` — checks token reuse and refresh after expiry
- `Test_configureProviderMeta_appAuthUsesNewClient` — confirms `app_auth` hits the new client path even with `legacy_client = true`

- [ ] Yes
- [x] No

App-auth users get a different HTTP stack under the hood, but there's no schema change and the old behavior was broken past the 1-hour mark anyway.


yashrajdighe:
@nickfloyd @deiga  Would be great if this PR can be reviewed!

deiga:
Please don't ping us. We are fully aware of the open PRs. 
What we don't have is unlimited time, which means that we need to choose where to put our capacity. 

And in this case, you're pinging a person who is no longer working on the project.

yashrajdighe:
> Please don't ping us. We are fully aware of the open PRs. What we don't have is unlimited time, which means that we need to choose where to put our capacity.
> 
> And in this case, you're pinging a person who is no longer working on the project.

I apologize for the misdirected ping and completely respect your team's capacity constraints. I only flagged this because the PR is currently a blocker for us, and a review when your schedule permits would be greatly appreciated.

stevehipwell:
@yashrajdighe why can't you use the new client? It'd be much more useful to provide feedback on the new client than to attempt to backport functionality into the old one. FYI I'm not expecting us to make any changes to the legacy client as the plan is to remove it in the next major version.

yashrajdighe:
> @yashrajdighe why can't you use the new client? It'd be much more useful to provide feedback on the new client than to attempt to backport functionality into the old one. FYI I'm not expecting us to make any changes to the legacy client as the plan is to remove it in the next major version.

@stevehipwell It was an inspired fix from https://github.com/integrations/terraform-provider-github/pull/2695#issuecomment-3794272732.

But it makes sense to go with the new client rather than fixing with the legacy client.
Please review again.

stevehipwell:
@yashrajdighe I'm not sure I follow what the purpose of this PR is now?

PR #3491, opened 2026-06-16
feat: Add explicit concurrency control for new client

stevehipwell:
Resolves #ISSUE_NUMBER

----

### Before the change?

-

### After the change?

- The new client has explicit concurrency control to limit cross client request concurrency to the secondary rate limit
- The new client has improved default transport options
- The new client has a user agent with the provider version
- The `ghclient` package has improved unit tests

### Pull request checklist

- [ ] Yes
- [x] No

----

### DIFF
--- a/internal/ghclient/options.go
+++ b/internal/ghclient/options.go
@@ -10,6 +10,7 @@ import (
 type Options struct {
 	RESTAPIURL   string
 	GraphQLURL   string
+	UserAgent    string
 	CachePath    string
 	RetryMax     int
 	RetryWaitMin time.Duration
--- a/internal/ghclient/options.go
+++ b/internal/ghclient/options.go
@@ -1,15 +1,23 @@
 package ghclient
 
-import "time"
+import (
+	"time"
+
+	"golang.org/x/sync/semaphore"
+)
 
 // Options defines the configuration options for creating GitHub clients.
 type Options struct {
-	RESTAPIURL    *string
-	RESTUploadURL *string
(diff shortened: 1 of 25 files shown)


robert-crandall (approved):
Really like this - the per-source semaphore sized to the secondary rate limit is a clean way to bound concurrency, and moving the conditional cache to the outermost layer so cache hits don't burn a slot is a nice touch. I traced the body-close release path through the cache transport and it stays balanced, and the throttle tests cover it well. Builds and tests pass for me locally. Thanks for the thorough work here!

internal/ghclient/options.go
```diff
@@ -1,15 +1,24 @@
...
+	CachePath    string
+	RetryMax     int
+	RetryWaitMin time.Duration
+	RetryWaitMax time.Duration
+
+	// Set per usage.
+	sema            *semaphore.Weighted
+	maxIdleConns    int
+	idleConnTimeout time.Duration
+	cacheRef        string
```
deiga, line 23: suggestion: Could we sort fields inside the groups? It would be easier to parse the diff 🙏

deiga (approved):
Left a minor suggestion, no need to block merge for that

internal/ghclient/options.go
deiga: Would any go formatter have an option for sorting struct fields? 🤔

stevehipwell: I'll look at this in a follow-up PR.

PR #3492, opened 2026-06-17
feat: add ID in data_source_github_repository_teams

acouvreur:
### Before the change?

You cannot retrieve the ID from the `github_repository_teams` datasource, which is a pain when using associated resources that require a team ID.

### After the change?

- You can retrieve the ID from the `github_repository_teams` datasource, which is useful when using associated resources that require a team ID.

### Pull request checklist

- [x] Tests for the changes have been added (for bug fixes / features) Cannot test because ID is generated from created repository, unless ID is retrieved in the tests before assertion

- [ ] Yes
- [x] No

----


stevehipwell:
@acouvreur please could you open an issue before creating a PR to save everyone's time.

~~FYI the data source already has `id`, as is enforced by the Terraform Plugin SDk v2, which is a string type; the code you've provided here would likely error. If we want to add an integer team ID we'd need to call it `team_id`, but please can we have that discussion in an issue.~~

Sorry, wrong data source. I take back my objection but the main point that opening an issue to be triaged before a PR still stands.

acouvreur:
> @acouvreur please could you open an issue before creating a PR to save everyone's time.
> 
> ~FYI the data source already has `id`, as is enforced by the Terraform Plugin SDk v2, which is a string type; the code you've provided here would likely error. If we want to add an integer team ID we'd need to call it `team_id`, but please can we have that discussion in an issue.~
> 
> Sorry, wrong data source. I take back my objection but the main point that opening an issue to be triaged before a PR still stands.

Will do, the change was so minimal I thought it'd be okay. Duly noted for next contributions

stevehipwell (changes requested):
> Will do, the change was so minimal I thought it'd be okay. Duly noted for next contributions

It is a pretty minimal change, but we have very limited resources and unsolicited changes without the full context are problematic (see my mistake about the scope as a prime example).

Given that I've spent the time to get the context already, let's see if we can get this merged. Please can you update the tests to check that the new computed ID is being set.

acouvreur:
Closing. #3537 added the team `id` to the `github_repository_teams` data source.

PR #3498, opened 2026-06-18
fix: read immediate github_team_members only, not child-team members

### LINKED ISSUE #3497: github_team_members: Read includes child-team members
### Summary

PR #3461 (https://github.com/integrations/terraform-provider-github/pull/3461) switched the `github_team_members` Read function from a GraphQL query to the REST `ListTeamMembers` endpoint. Those two are not equivalent for organizations that use nested (parent/child) teams, and I think this introduces a perpetual diff for those orgs.

### Details

The old Read used GraphQL:

```graphql
members(membership: IMMEDIATE, first: 100, after: $after)
```

`IMMEDIATE` returns only the team's direct members.

The new Read uses the REST endpoint `GET /orgs/{org}/teams/{team_slug}/members` (via `ListTeamMembersBySlugIter` / `ListTeamMembersByIDIter`). That endpoint always includes the members of child teams. This is documented ("Team members will include the members of child teams", see the [List team members REST docs](https://docs.github.com/en/rest/teams/members?apiVersion=2022-11-28#list-team-members)) and is confirmed in the server implementation, which builds the member scope with a hard-coded `membership: "all"` (the valid values are `immediate`, `child_team`, and `all`). The REST endpoint exposes only a `role` filter, so there is no way to restrict it to immediate members.

### Impact

For an org with nested teams, `github_team_members.members` will now read back members inherited from child teams that the user never put in their configuration. Since `github_team_members` manages the member set authoritatively, I'd expect:

- a perpetual diff after apply, and
- potentially an update attempting to remove inherited members it does not actually manage.

Flat team structures (no child teams) are unaffected, which is probably why this passed review and acceptance tests.

### Suggested direction

Restore immediate-members-only semantics. Options I can think of:

- Filter the REST results down to immediate members (e.g. cross-reference each member's direct membership), or
- Move the Read back to the GraphQL `membership: IMMEDIATE` query, or
- Some hybrid that keeps the clean 404-handling from #3461 while preserving the immediate-only member list.

### Notes

The 404 handling added in #3461 (treating a 404 on member removal as "team already gone", and removing the resource from state on a 404 during Read) looks correct and worth keeping. This issue is specifically about the child-team member inclusion change in the Read path.

robert-crandall:
Resolves #3497

### Before the change?

#3499 refactored `github_team_members` and added a `team_with_child_team` regression test, but its Read path (`getTeamMembers`) still uses the REST list-team-members endpoint. That endpoint always includes members inherited from child teams and has no immediate-only option (the go-github `TeamListTeamMembersOptions` only exposes a `role` filter). So for organizations with nested teams, Read pulls in inherited members the user never configured, breaking the authoritative member set and surfacing drift. The `team_with_child_team` test currently fails on `main` for this reason.

### After the change?

`getTeamMembers` now uses the GraphQL API with `membership: IMMEDIATE`, which returns only the team's direct members. A null `team` node (team deleted out-of-band) is mapped to a 404 `github.ErrorResponse`, so the existing Read/Delete handling still removes the resource from state - preserving the `team_delete_handled` behaviour.

`getTeamMembers` and `updateTeamMembers` now take `*Owner` so the read path can reach the v4 (GraphQL) client while membership add/remove keep using the v3 (REST) client.

This is rebased onto `main` (post-#3499). Because #3499 rewrote the resource, this is now a single focused commit re-deriving the fix against the new `getTeamMembers` structure rather than the original pre-#3499 diff.

### How I verified it

I exercised the **real** `getTeamMembers` function (not a copy) against a live nested team (`integrations/hubbers`) by temporarily wiring it through `getTestMeta`:

- `getTeamMembers` (new GraphQL `IMMEDIATE` path) returned **14** members (immediate only).
- The old REST role-filtered approach returned **40** for the same team - the extra 26 are inherited child-team members. This is exactly the inflation that fails `team_with_child_team`'s `SetSizeExact` assertion on `main`.
- A non-existent team slug returned a 404 `github.ErrorResponse`, confirming the GraphQL null-node -> 404 mapping that keeps `team_delete_handled` working (Read removes from state).

Static checks: `go build`, `go vet`, `gofmt -l`, `go test -run TestAccGithubTeamMembers` (compiles + skips without creds), `golangci-lint run ./github/` (0 issues), and the repo's strict linter (`forcetypeassert`) against the changed code (0 issues).

### What I did NOT verify

- I did not run the full `TF_ACC` acceptance suite end-to-end (no org-write credentials here; `team_with_child_team` needs two org members + a child team). My verification drives the real `getTeamMembers` against a live nested team, which proves the immediate-vs-all behaviour and the 404 mapping, but it is not the literal Terraform acceptance run.

- [ ] Yes
- [x] No

### DIFF
--- a/github/resource_github_team_members.go
+++ b/github/resource_github_team_members.go
@@ -15,6 +15,7 @@ import (
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/customdiff"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/validation"
+	"github.com/shurcooL/githubv4"
 )
 
 func resourceGithubTeamMembers() *schema.Resource {
@@ -124,7 +125,7 @@ func resourceGithubTeamMembersCreate(ctx context.Context, d *schema.ResourceData
 		return diag.FromErr(fmt.Errorf("failed to parse team members: %w", err))
 	}
 
-	if err := updateTeamMembers(ctx, client, orgName, slug, teamMembers); err != nil {
+	if err := updateTeamMembers(ctx, meta, slug, teamMembers); err != nil {
 		return diag.FromErr(err)
 	}
 
@@ -167,7 +168,7 @@ func resourceGithubTeamMembersRead(ctx context.Context, d *schema.ResourceData, 
 
 	tflog.Debug(ctx, "Reading team members.", map[string]any{"team_slug": slug})
 
-	teamMembers, err := getTeamMembers(ctx, client, orgName, slug)
+	teamMembers, err := getTeamMembers(ctx, meta, slug)
 	if err != nil {
 		if ghErr, ok := errors.AsType[*github.ErrorResponse](err); ok && ghErr.Response.StatusCode == http.StatusNotFound {
 			tflog.Info(ctx, "Team not found during read, removing from state.", map[string]any{"team_slug": slug})
@@ -209,7 +210,7 @@ func resourceGithubTeamMembersUpdate(ctx context.Context, d *schema.ResourceData
 		return diag.FromErr(fmt.Errorf("failed to parse team members: %w", err))
 	}
 
-	if err := updateTeamMembers(ctx, client, orgName, slug, teamMembers); err != nil {
+	if err := updateTeamMembers(ctx, meta, slug, teamMembers); err != nil {
 		return diag.FromErr(err)
 	}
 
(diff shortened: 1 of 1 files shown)


stevehipwell:
@robert-crandall I think the first thing to do here is to add a test for the regression, I'll take a look at the tests and open a PR to modernize the existing ones and add a test for ignoring child team members.

github/resource_github_team_members.go
```diff
@@ -186,46 +185,75 @@ func resourceGithubTeamMembersRead(ctx context.Context, d *schema.ResourceData,
...
+			tflog.Info(ctx, "Team no longer exists, removing from state", map[string]any{"team_id": teamIdString})
+			d.SetId("")
+			return nil
+		}
+		return diag.FromErr(err)
+	}
+	teamSlug := team.GetSlug()
 
-	var teamMembersAndMaintainers []any
+	log.Printf("[DEBUG] Reading team members: %s", teamIdString)
```
deiga: Let's refactor all touched `log.*` lines to `tflog` if feasible

```suggestion
	tflog.Debug(ctx, "Reading team members",  map[string]any{"teamID": teamIdString})
```

deiga (changes requested):
Good catch! Do we have a regression test for this?

stevehipwell:
@robert-crandall you should be able to rebase this onto `main` and confirm that it fixes the currently failing regression test I added in #3499.

github/resource_github_team_members.go
robert-crandall: Done!

robert-crandall:
> @robert-crandall you should be able to rebase this onto `main` and confirm that it fixes the currently failing regression test I added in #3499.

@stevehipwell Done!

stevehipwell (approved):
LGTM

deiga (approved):

stevehipwell:
@robert-crandall please could you rebase so this can be merged.

PR #3500, opened 2026-06-21
Add cooldown config for dependabot

deiga:
Enable waiting for 7 days before offering updates, to reduce the risk of supply-chain attacks

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -2,6 +2,8 @@ version: 2
 
 updates:
   - package-ecosystem: github-actions
+    cooldown:
+      default-days: 7
     directories:
       - /
     groups:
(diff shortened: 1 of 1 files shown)


stevehipwell (changes requested):
I'm not sure I agree with the principal of cooldowns at a fundamental level, if everyone uses one then effectively nobody is using one (for all but the most trivial cases). I also think that both ecosystems here have completely different constraints so if we were enabling them it'd be two decisions rather than one.

In all ecosystems we should only be approving PRs if we fully understand the code implications or fully trust the code owners. In the latter case a small cooldown period may be justified to allow the owners to respond to issues but this is highly subjective per package not per ecosystem.

For GitHub Actions the likelihood of a supply chain hack is inherently higher, but the scope of the packages is significantly lower and should be trivial to review.

For Go the ecosystem was designed from the ground up to be resistant to supply chain hacks so the risk profile is different to other ecosystems and the code changes should be clear.

My personal opinion is that we just tighten up the process around approving Dependabot rather than playing theater by adding a cooldown.

PR #3510, opened 2026-06-30
feat: Refactor new client logic to use base url

stevehipwell:
Resolves #ISSUE_NUMBER

----

### Before the change?

- App auth using the new client would only work for github.com

### After the change?

- App auth works for all targets
- Improved test coverage for provider meta
- New client has been refactored

### Pull request checklist

- [ ] Yes
- [x] No

----

### DIFF
--- a/github/provider.go
+++ b/github/provider.go
@@ -558,14 +558,6 @@ func configureProviderMeta(ctx context.Context, version string, c *Config) (*Own
 			}
 			owner.name = user.GetLogin()
 		}
-
-		if owner.name == "" && c.Token != "" {
-			user, _, err := owner.v3client.Users.Get(ctx, "")
-			if err != nil {
-				return nil, fmt.Errorf("owner cannot be found by token: %w", err)
-			}
-			owner.name = user.GetLogin()
-		}
 	} else {
 		if !c.Anonymous() && owner.name == "" {
 			return nil, fmt.Errorf("owner must be set when authenticating using the new client implementation")
--- a/github/provider.go
+++ b/github/provider.go
@@ -2,15 +2,18 @@ package github
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"net/http"
 	"net/url"
 	"os"
 	"os/exec"
+	"path/filepath"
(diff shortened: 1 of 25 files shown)


robert-crandall (approved):
Thanks for the quick turnaround - the `getRESTURL`/`getGraphQLURL` helpers are a clean way to handle this, and passing the resolved `/api/v3/` URL into `WithBaseURL` fixes the GHES installation-token path. I reproduced it locally with `IsGHES: true` and the token request now correctly lands on `/api/v3/app/installations/{id}/access_tokens`. The new options tests and the extra `Response == nil` nil-guards are a nice touch too. Looks good to me!

stevehipwell:
@deiga this is just waiting for you.

github/provider.go
```diff
@@ -361,11 +363,10 @@ func configureProvider(version, commit string) func(context.Context, *schema.Res
 
 				tflog.Debug(ctx, "Using base URL from provider configuration.", map[string]any{"base_url": baseURL.String()})
 				config.BaseURL = baseURL
+				config.IsGHES = isGHES
 
 				if isGHES {
 					tflog.Debug(ctx, "Base URL indicates GitHub Enterprise Server (GHES) usage; enabling GHES mode.", map[string]any{"base_url": baseURL.String()})
-					config.RESTAPIPath = GHESRESTAPIPath
-					config.GraphQLAPIPath = GHESGraphQLAPIPath
 				}
```
deiga, line 370: suggestion(minor): The conditional feels redundant here just for logging. We could maybe get more information if we logged something like:

```suggestion:
				tflog.Debug(ctx, "Using base URL and isGHES from provider configuration.", map[string]any{"base_url": baseURL.String(), "isGHES": isGHES})
				config.BaseURL = baseURL
				config.IsGHES = isGHES
```

deiga (approved):
Really nice refactor!

github/provider.go
stevehipwell: I agree and I'll make this change in a future PR if that's OK so we can get this merged in?

PR #3518, opened 2026-07-02
Fix nil-pointer panic when deleting github_team_repository

### LINKED ISSUE #3517: [BUG]: nil-pointer panic deleting github_team_repository when the delete API call returns a nil response
### Expected Behavior

Deleting a `github_team_repository` when the underlying `Teams.RemoveTeamRepoByID` call fails at the transport layer (for example a cancelled context or an exceeded deadline) should surface the error as a Terraform diagnostic, so the operation fails cleanly and can be retried.

### Actual Behavior

The provider panics with a nil-pointer dereference. `resourceGithubTeamRepositoryDelete` reads `resp.StatusCode` without first checking whether `resp` is nil. When the API call returns a transport-level error, `google/go-github` returns a nil `*github.Response` (see `bareDo`: on a transport error `resp` is nil, so it returns `(nil, err)`), so the delete dereferences a nil pointer and the provider process crashes.

```
panic: runtime error: invalid memory address or nil pointer dereference [signal SIGSEGV]
github.com/integrations/terraform-provider-github/v6/github.resourceGithubTeamRepositoryDelete(...)
    github/resource_github_team_repository.go:226
```

This is the only unguarded `resp.StatusCode` dereference in the `github/` package; the other call sites already use the `resp != nil && resp.StatusCode` pattern.

### Terraform Version

Reproduced on `main` (v6, commit `4d34f10b`). The bug was surfaced by the acceptance test suite (`go test ./github/... -run TestAccGithubTeamRepository`) when a parallel delete hit `context deadline exceeded`. The affected code path is reached by any `terraform apply`/`destroy` that deletes a `github_team_repository` when the delete API call returns a transport-level error (nil HTTP response).

```
Terraform vX.Y.Z (provider built from main @ 4d34f10b)
```

### GitHub Installation Type

Observed against GitHub Enterprise Cloud, but the defect is installation-agnostic (a nil-pointer dereference on any transport error, independent of the GitHub edition).

### Affected Resource(s)

- github_team_repository

### Steps to Reproduce

Trigger a delete of a `github_team_repository` where `RemoveTeamRepoByID` returns a nil response (any transport error, such as a cancelled context or an exceeded deadline). This can be reproduced deterministically in a unit test that drives the delete with a cancelled context so the API call returns a nil response before reaching the server.

### Debug Output

```
    resource_github_actions_runner_group_test.go:90: Step 1/1 error: Error running apply: exit status 1
        Error: context deadline exceeded
          with github_repository.test,
          on terraform_plugin_test.tf line 12, in resource "github_repository" "test":
[signal SIGSEGV: segmentation violation code=0x2 addr=0x0]
github.com/integrations/terraform-provider-github/v6/github.resourceGithubTeamRepositoryDelete(...)
	github/resource_github_team_repository.go:226
github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema.(*Resource).delete(...)
```

### Code of Conduct

rilical:
Resolves #3517

----

### Before the change?

`resourceGithubTeamRepositoryDelete` read `resp.StatusCode` without checking whether `resp` was nil. When `Teams.RemoveTeamRepoByID` fails at the transport layer (for example a cancelled context or an exceeded deadline), `google/go-github` returns a nil `*github.Response` (in `bareDo`, a transport error yields `(nil, err)`), so the delete panicked with a nil-pointer dereference instead of returning the error:

```
    github/resource_github_team_repository.go:226
```

In a `terraform apply`/`destroy`, a transient network error while removing a team's repository access could crash the provider. This is the only unguarded `resp.StatusCode` dereference in the `github/` package; the other call sites already use `resp != nil && resp.StatusCode`.

### After the change?

- Guard the status check with `resp != nil`, matching the pattern already used elsewhere in the provider. When `resp` is nil, the underlying error now flows through `handleArchivedRepoDelete` and is returned as a diagnostic instead of panicking.
- Add a regression test (`Test_resourceGithubTeamRepositoryDelete_nilResponseDoesNotPanic`) that drives delete with a cancelled context so the API call returns a nil response, asserting delete returns an error diagnostic. It panics at line 226 without the fix and passes with it.

Verification:
- `go test ./github/ -run '^Test_resourceGithubTeamRepositoryDelete_nilResponseDoesNotPanic$'` panics before the change, passes after.
- Full unit suite green (`make test`, which skips `^TestAcc`).
- Happy path unchanged against a live org: `TestAccGithubTeamRepository` passes (create / read / import / update / destroy).

### Pull request checklist

- [ ] Schema migrations have been created if needed — N/A, no schema change
- [x] Docs have been reviewed and added / updated if needed — N/A, no user-facing schema or behavior change beyond not panicking

- [ ] Yes
- [x] No

----

> **AI-assisted disclosure:** this change was developed with AI assistance (GitHub Copilot). Per the repository's AI Use Policy I have reviewed and understand the change, verified the root cause in `google/go-github`'s `bareDo` (nil `*Response` on transport error), and tested it with a deterministic unit regression test plus a live acceptance run of `TestAccGithubTeamRepository`.

### DIFF
--- a/github/resource_github_team_repository_test.go
+++ b/github/resource_github_team_repository_test.go
@@ -35,7 +35,7 @@ func Test_resourceGithubTeamRepositoryDelete_nilResponseDoesNotPanic(t *testing.
 
 	// A cancelled context makes the underlying HTTP call return before it
 	// reaches the server, so go-github yields a nil response with an error.
-	ctx, cancel := context.WithCancel(context.Background())
+	ctx, cancel := context.WithCancel(t.Context())
 	cancel()
 
 	diags := resourceGithubTeamRepositoryDelete(ctx, d, meta)
--- a/github/resource_github_team_repository_test.go
+++ b/github/resource_github_team_repository_test.go
@@ -1,13 +1,48 @@
 package github
 
 import (
+	"context"
 	"fmt"
 	"strings"
 	"testing"
 
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
 	"github.com/hashicorp/terraform-plugin-testing/helper/acctest"
 	"github.com/hashicorp/terraform-plugin-testing/helper/resource"
 )
+
+// Test_resourceGithubTeamRepositoryDelete_nilResponseDoesNotPanic is a
+// regression test for a nil-pointer panic in the delete path. When the
+// RemoveTeamRepoByID call fails at the transport layer (for example a
+// cancelled context or an exceeded deadline), go-github returns a nil
+// *github.Response. Delete must surface that as an error diagnostic instead of
+// dereferencing resp.StatusCode and panicking.
+func Test_resourceGithubTeamRepositoryDelete_nilResponseDoesNotPanic(t *testing.T) {
+	t.Parallel()
+
+	ts := githubApiMock(nil)
+	defer ts.Close()
+
+	meta := &Owner{
+		name:           "test-org",
+		id:             12345,
+		v3client:       mustCreateTestGitHubClient(t, ts.URL),
+		IsOrganization: true,
+	}
+
+	d := schema.TestResourceDataRaw(t, resourceGithubTeamRepository().Schema, map[string]any{})
+	d.SetId(buildTwoPartID("123", "test-repo"))
+
+	// A cancelled context makes the underlying HTTP call return before it
+	// reaches the server, so go-github yields a nil response with an error.
+	ctx, cancel := context.WithCancel(context.Background())
+	cancel()
+
+	diags := resourceGithubTeamRepositoryDelete(ctx, d, meta)
+	if !diags.HasError() {
+		t.Fatal("expected an error diagnostic when the API response is nil, got none")
+	}
+}
 
 func TestAccGithubTeamRepository(t *testing.T) {
 	t.Parallel()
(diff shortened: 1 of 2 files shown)


robert-crandall (dismissed):
Thanks for this - really thorough writeup and root-cause analysis. I confirmed the fix and the regression test locally: it panics on the old code and passes with the guard, exactly as described. Nice deterministic repro using a cancelled context. 🙌

github/resource_github_team_repository_test.go
```diff
@@ -1,14 +1,49 @@
...
+		v3client:       mustCreateTestGitHubClient(t, ts.URL),
+		IsOrganization: true,
+	}
+
+	d := schema.TestResourceDataRaw(t, resourceGithubTeamRepository().Schema, map[string]any{})
+	d.SetId(buildTwoPartID("123", "test-repo"))
+
+	// A cancelled context makes the underlying HTTP call return before it
+	// reaches the server, so go-github yields a nil response with an error.
+	ctx, cancel := context.WithCancel(context.Background())
```
stevehipwell:
```suggestion
	ctx, cancel := context.WithCancel(t.Context())
```

stevehipwell (changes requested):
Thanks for the fix @rilical, just a minor change required. Did you happen to search for other cases where this pattern hasn't been implemented correctly?

stevehipwell (approved):
Thanks for the fix @rilical.

FYI for context; the error handling in this function isn't correctly implemented and there also appears to be unnecessary complexity. This whole block needs updating to check for an error and then use `errors.AsType` to check for a GitHub error with a response.

deiga (approved):

robert-crandall (approved):

PR #2356, opened 2024-08-20
Support repository property for GitHub organization ruleset

Moser-ss:
Resolves #2137
Introduces the support to use `repository_property` to target repositories in the ruleset

----

### Before the change?

* Cannot target ruleset using `repository_property`

### After the change?

* The ruleset can use the `repository_property` under the conditions block

### Pull request checklist
- [ ] Tests for the changes have been added (for bug fixes / features)
- [ ] Docs have been reviewed and added / updated if needed (for bug fixes / features)

### Does this introduce a breaking change?

Please see our docs on [breaking changes](https://github.com/octokit/.github/blob/master/community/breaking_changes.md) to help!

- [ ] Yes
- [ ] No

----

PR #3520, opened 2026-07-04
fix: prevent panic in expandConditions when ref_name is nil or empty

### LINKED ISSUE #3299: [BUG]: Panic in expandConditions when conditions.ref_name is nil/empty

nikolauspschuetz:
Resolves #3299

----

### Before the change?

Applying a `github_repository_ruleset` whose `conditions.ref_name` block is present but empty crashed the provider:

```
```

`expandConditions` asserted `ref_name[0].(map[string]any)` and `include`/`exclude` `.([]any)` without checking them, so a nil list element (or a `ref_name {}` block with no include/exclude keys) panicked instead of being handled.

### After the change?

The `ref_name` element and the `include`/`exclude` lookups now use comma-ok assertions. An empty/nil `ref_name` block is treated as no ref-name condition (no panic); populated and empty-array blocks behave exactly as before. This mirrors the existing nil-tolerant handling in `expandRepositoryPropertyConditions`.

### Pull request checklist

- [ ] Schema migrations have been created if needed ([example](https://github.com/F-Secure-web/terraform-provider-github/blob/main/github/migrate_github_actions_organization_secret.go)) — n/a, no schema change
- [x] Docs have been reviewed and added / updated if needed (for bug fixes / features) — n/a, no schema/behavior surface change

- [ ] Yes
- [x] No

----

<sub>AI assistance: implemented with Claude Code. I directed the fix, reviewed the diff, and verified it: the added regression tests panic on `main` at `util_rules.go:198` and pass after the change (`go test -run TestExpandConditions_ ./github/`), with `go vet`/`gofmt`/`go build` clean.</sub>

### DIFF
--- a/github/util_rules.go
+++ b/github/util_rules.go
@@ -194,26 +194,34 @@ func expandConditions(input []any, org bool) *github.RepositoryRulesetConditions
 	inputConditions := input[0].(map[string]any)
 
 	// ref_name is available for both repo and org rulesets
-	if v, ok := inputConditions["ref_name"].([]any); ok && v != nil && len(v) != 0 {
-		inputRefName := v[0].(map[string]any)
-		include := make([]string, 0)
-		exclude := make([]string, 0)
+	if v, ok := inputConditions["ref_name"].([]any); ok && len(v) != 0 {
+		// A present-but-empty ref_name block (e.g. `ref_name {}`) can yield a nil
+		// element or a map without include/exclude keys; guard both so we fail
+		// gracefully instead of panicking (#3299).
+		if inputRefName, ok := v[0].(map[string]any); ok {
+			include := make([]string, 0)
+			exclude := make([]string, 0)
 
-		for _, v := range inputRefName["include"].([]any) {
-			if v != nil {
-				include = append(include, v.(string))
+			if raw, ok := inputRefName["include"].([]any); ok {
+				for _, v := range raw {
+					if v != nil {
+						include = append(include, v.(string))
+					}
+				}
 			}
-		}
 
-		for _, v := range inputRefName["exclude"].([]any) {
-			if v != nil {
-				exclude = append(exclude, v.(string))
+			if raw, ok := inputRefName["exclude"].([]any); ok {
+				for _, v := range raw {
+					if v != nil {
+						exclude = append(exclude, v.(string))
+					}
+				}
 			}
-		}
 
-		rulesetConditions.RefName = &github.RepositoryRulesetRefConditionParameters{
-			Include: include,
-			Exclude: exclude,
+			rulesetConditions.RefName = &github.RepositoryRulesetRefConditionParameters{
+				Include: include,
+				Exclude: exclude,
+			}
 		}
 	}
 
--- a/github/util_rules_test.go
+++ b/github/util_rules_test.go
@@ -1365,3 +1365,113 @@ func TestFlattenRulesetRepositoryPropertyTargetParameters_NilPropertyValues(t *t
 		}
 	}
 }
+
+// Regression tests for #3299: expandConditions must not panic when a
+// conditions.ref_name block is present but nil/empty (e.g. `ref_name {}` with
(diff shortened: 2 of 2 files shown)


deiga:
Could you add an acceptance test case to check for regression as well?

nikolauspschuetz:
Thanks @deiga! I looked into an acceptance test and I don't think this one can be reproduced that way — reasoning below, plus why the unit tests match the repo's existing convention for this kind of fix.

The #3299 panic is in `expandConditions`, which runs at apply from `d.Get("conditions")`. It crashes when a `ref_name` element is nil (or missing `include`/`exclude`) — an internal representation. That input can't come from valid HCL: `ref_name`, `include`, and `exclude` are all `Required` (`resource_github_repository_ruleset.go:105-121`), so Terraform rejects any config that would create it, and a `dynamic "ref_name"` block only ever produces a fully-populated element or an empty list (already handled by the `len == 0` guard) — never a nil element. So an acceptance test can't deterministically construct the crashing input.

This is the same class of defensive nil-guard the provider already has for the sibling condition type: `expandRepositoryPropertyConditions` guards nil elements/values (`util_rules.go:207-297`), covered by unit tests `TestExpandRepositoryPropertyConditions_NilElements` / `_NilPropertyValues` (added in #2356). My change mirrors that for `ref_name`, with unit tests (`TestExpandConditions_*`) that reproduce the exact panic at `util_rules.go:198` on `main` and pass with the fix. The normal `ref_name` path is also already exercised end-to-end by the existing `TestAccGithubRepositoryRuleset`.

Happy to add more coverage if you'd like — just let me know what would be most useful. Thanks for reviewing!

deiga:
@nikolauspschuetz If this can't be reproduced via valid Terraform config, then what is this PR guarding against?
Where does the panic you pasted in the PR description come from?

nikolauspschuetz:
Good questions — and digging in, you're right to push.

The panic in the original report is at the *conditions* cast (`util_rules.go:194` in v6.11.1), which `main` already guards via the `input[0] == nil` check (added in #3050). My PR guards a deeper spot — the `ref_name` *element* cast — and I have to concede I can't construct a valid config that reaches it: with `include`/`exclude` `Required`, `ref_name` only yields a populated map or an empty list (already handled by the `len == 0` guard), never a nil element — not even via a `dynamic` block. So this is defensive-only against a state I can't demonstrate is reachable.

I also see #3300 already proposes the same `ref_name`-element guard. And the config from the original issue now fails validation ("Missing required argument") rather than panicking, which is the behavior #3299 asked for.

Given all that, I'll close this in favor of #3300 as the earlier PR (and since the reported case is already handled by validation). Thanks for the careful review — appreciate you catching it.

PR #3406, opened 2026-05-06
Fix `github_actions_hosted_runner` import  of the `image` block

terrabitz:
Resolves #3405 

----

### Before the change?

During imports, the `github_actions_hosted_runner` resource tried accessing incorrect JSON keys, resulting in the `image` block not being imported correctly. This in turn causes the resource to be replaced, since this is an immutable value:

```
Terraform will perform the following actions:

  # github_actions_hosted_runner.my_example_runner must be replaced
  # (imported from "1")
  # Warning: this will destroy the imported resource
-/+ resource "github_actions_hosted_runner" "my_example_runner" {
      ~ id                   = "1" -> (known after apply)
        image_gen            = false
      ~ last_active_on       = "2026-05-06T16:27:22.8271735Z" -> (known after apply)
      ~ machine_size_details = [
          - {
              - cpu_cores  = 2
              - id         = "2-core"
              - memory_gb  = 8
              - storage_gb = 75
            },
        ] -> (known after apply)
        maximum_runners      = 50
        name                 = "my-example-runner"
      ~ platform             = "linux-x64" -> (known after apply)
        public_ip_enabled    = false
      ~ public_ips           = [] -> (known after apply)
        runner_group_id      = 1
        size                 = "2-core"
      ~ status               = "Ready" -> (known after apply)

      + image { # forces replacement
          + id      = "2306"
          + size_gb = (known after apply)
          + source  = "github"
        }
    }

Plan: 1 to import, 1 to add, 0 to change, 1 to destroy.
```

### After the change?

The correct JSON keys are now used, and the resource imports correctly

```
Terraform will perform the following actions:

  # github_actions_hosted_runner.my_example_runner will be imported
    resource "github_actions_hosted_runner" "my_example_runner" {
        id                   = "1"
        image_gen            = false
        last_active_on       = "2026-05-06T16:27:22.8271735Z"
        machine_size_details = [
            {
                cpu_cores  = 2
                id         = "2-core"
                memory_gb  = 8
                storage_gb = 75
            },
        ]
        maximum_runners      = 50
        name                 = "my-example-runner"
        platform             = "linux-x64"
        public_ip_enabled    = false
        public_ips           = []
        runner_group_id      = 1
        size                 = "2-core"
        status               = "Ready"

        image {
            id      = "2306"
            size_gb = 75
            source  = "github"
        }
    }

Plan: 1 to import, 0 to add, 0 to change, 0 to destroy.
```

### Pull request checklist

[...]

PR #3521, opened 2026-07-05
[MAINT] Fix actions_hosted_runner tests

### LINKED ISSUE #3405: [BUG]: github_actions_hosted_runner resource must be replaced after import
### Expected Behavior

`github_actions_hosted_runner` resources should be importable without requiring any destructive replacements.

### Actual Behavior

When importing `github_actions_hosted_runner` resources, the imported state does not include the `image` block. Since the resource treats this as an immutable field, this causes Terraform to plan a destroy + recreation of the hosted runner.

### Terraform Version

Terraform v1.15.2
on darwin_arm64

### Affected Resource(s)

- github_actions_hosted_runner

```hcl
terraform {
  required_providers {
    github = {
      version = "~> 6"
    }
  }
}

provider "github" {
  owner = "my-example-org"
}

import {
  to = github_actions_hosted_runner.my_example_runner
  id = "1"
}

resource "github_actions_hosted_runner" "my_example_runner" {
  image_gen         = false
  image_version     = null
  maximum_runners   = 50
  name              = "my-example-runner"
  public_ip_enabled = false
  runner_group_id   = 1
  size              = "2-core"
  image {
    id     = "2306"
    source = "github"
  }
}
```

### Steps to Reproduce

1. In an enterprise organization, create a new GitHub-hosted runner in the **Default** runner group called `my-example-runner`. Use all the default values:
   - Platform: `Linux x64`
   - Size: `2-cores`
   - Image name: `Ubuntu Latest (24.04)`
3. Retrieve the runner ID from the URL (e.g. `https://github.com/organizations/my-example-org/settings/actions/github-hosted-runners/1` -> `1`)
4. Import the resource using an `import` block
5. Run the Terraform plan:
```bash
terraform plan
```

**Result**: Terraform wants to replace the runner

```

  # github_actions_hosted_runner.my_example_runner must be replaced
  # (imported from "1")
  # Warning: this will destroy the imported resource
-/+ resource "github_actions_hosted_runner" "my_example_runner" {
      ~ id                   = "1" -> (known after apply)
        image_gen            = false
      ~ last_active_on       = "2026-05-06T16:27:22.8271735Z" -> (known after apply)
      ~ machine_size_details = [
          - {
              - cpu_cores  = 2
              - id         = "2-core"
              - memory_gb  = 8
              - storage_gb = 75
            },
        ] -> (known after apply)
        maximum_runners      = 50
        name                 = "my-example-runner"
      ~ platform             = "linux-x64" -> (known after apply)
        public_ip_enabled    = false
      ~ public_ips           = [] -> (known after apply)
        runner_group_id      = 1
        size                 = "2-core"
      ~ status               = "Ready" -> (known after apply)

      + image { # forces replacement
          + id      = "2306"
          + size_gb = (known after apply)
          + source  = "github"
        }
    }

Plan: 1 to import, 1 to add, 0 to change, 1 to destroy.
```

### Debug Output

```shell

```

### Panic Output

```shell

```

### Code of Conduct

### LINKED ISSUE #3626: [BUG]: `github_actions_hosted_runner` does not wait for public IP provisioning
### Expected Behavior

When `public_ip_enabled = true`, creating or updating a
`github_actions_hosted_runner` should wait until:

1. The runner status is `Ready`.
2. The GitHub API returns the allocated `public_ips`.

After `terraform apply` completes, outputs referencing `public_ips` should be populated.

### Actual Behavior

The provider performs only one read immediately after creating or updating the runner. Because GitHub provisions hosted runners asynchronously, the resource can return while its status is still `Provisioning` and `public_ips` is empty.

The IPs appear in GitHub later, but Terraform state and dependent outputs remain empty until another refresh or apply.

### Terraform Version

1.14.9

### GitHub Installation Type

- [ ] I don't know

### Affected Resource(s)

github_actions_hosted_runner

```hcl
resource "github_actions_hosted_runner" "this" {
  name = "example"

  image {
    id     = "2306"
    source = "github"
  }

  public_ip_enabled = true
  maximum_runners   = 5
  size              = "2-core"
  runner_group_id   = github_actions_runner_group.this.id
}

output "github_actions_runner_public_ips" {
  value = github_actions_hosted_runner.this.public_ips
}
```

### Steps to Reproduce

terraform apply

### Debug Output

```shell
github_actions_runner_group.this: Creating...
github_actions_runner_group.this: Creation complete after 3s [id=5]
github_actions_hosted_runner.this: Creating...
github_actions_hosted_runner.this: Creation complete after 3s [id=8]

Apply complete! Resources: 2 added, 0 changed, 0 destroyed.

Outputs:

github_actions_runner_public_ips = tolist([])
```

### Code of Conduct

deiga:
Resolves #3405
Closes https://github.com/integrations/terraform-provider-github/pull/3406

----

### Before the change?

- `resource_github_actions_hosted_runner` tests have naming collisions
- `resource_github_actions_hosted_runner` uses legacy patterns

### After the change?

- `resource_github_actions_hosted_runner` tests use `ConfigStateChecks` and `ConfigPlanChecks`
- Ensure `resource_github_actions_hosted_runner` tests don't use conflicting names
- Refactors `resource_github_actions_hosted_runner` to use Context-aware CRUD functions
- Refactors `resource_github_actions_hosted_runner` to not call `Read` after `Update` and `Create`
- Refactors `resource_github_actions_hosted_runner` to use `go-github` API methods
- Adds sweeper for Organization Runner Groups

### Pull request checklist

- [ ] Yes
- [x] No

----


github/resource_github_actions_hosted_runner.go
```diff
@@ -183,366 +183,337 @@ func resourceGithubActionsHostedRunner() *schema.Resource {
 	}
 }
 
-func expandImage(imageList []any) map[string]any {
+func expandImage(imageList []any) github.HostedRunnerImage {
```
stevehipwell: It's probably worth moving the helper functions ot the bottom of the file to make the file layout clearer.

github/resource_github_actions_hosted_runner.go
```diff
@@ -183,366 +183,337 @@ func resourceGithubActionsHostedRunner() *schema.Resource {
...
+		ipResult["prefix"] = ip.GetPrefix()
+		ipResult["length"] = ip.GetLength()
 		result = append(result, ipResult)
 	}
 
 	return result
 }
 
-func resourceGithubActionsHostedRunnerCreate(d *schema.ResourceData, meta any) error {
+func resourceGithubActionsHostedRunnerCreate(ctx context.Context, d *schema.ResourceData, meta any) diag.Diagnostics {
```
stevehipwell:
```suggestion
func resourceGithubActionsHostedRunnerCreate(ctx context.Context, d *schema.ResourceData, m any) diag.Diagnostics {
```

github/resource_github_actions_hosted_runner.go
```diff
@@ -183,366 +183,337 @@ func resourceGithubActionsHostedRunner() *schema.Resource {
...
 
 	if v, ok := d.GetOk("maximum_runners"); ok {
-		payload["maximum_runners"] = v.(int)
+		maxRunners, _ := v.(int)
+		req.MaximumRunners = new(int64(maxRunners))
 	}
 
-	if v, ok := d.GetOk("public_ip_enabled"); ok {
-		payload["enable_static_ip"] = v.(bool)
+	if v, okExists := d.GetOkExists("public_ip_enabled"); okExists { //nolint:staticcheck // SA1019: d.GetOkExists is deprecated but necessary for bool fields
```
stevehipwell: This isn't required as the value is optional with a default so it's always set one way; `d.Get` would be fine here.

github/resource_github_actions_hosted_runner.go
```diff
@@ -183,366 +183,337 @@ func resourceGithubActionsHostedRunner() *schema.Resource {
...
+		req.EnableStaticIP = new(publicIPEnabled)
 	}
 
 	if v, ok := d.GetOk("image_version"); ok {
-		payload["image_version"] = v.(string)
+		runnerImageVersion, _ := v.(string)
+		req.Image.Version = new(runnerImageVersion)
 	}
 
 	if v, ok := d.GetOk("image_gen"); ok {
```
stevehipwell: This can be swapped for `d.Get` as `d.GotOk` isn't useful for bool values.

stevehipwell (changes requested):
Just a quick pre-review, the comments apply to all matching patterns not just the ones I commented on.

github/resource_github_actions_hosted_runner.go
```diff
@@ -183,366 +184,277 @@ func resourceGithubActionsHostedRunner() *schema.Resource {
...
+
+	runnerID, err := strconv.ParseInt(d.Id(), 10, 64)
+	if err != nil {
+		return diag.FromErr(err)
+	}
 
-	payload := make(map[string]any)
+	req := github.UpdateHostedRunnerRequest{}
 
 	if d.HasChange("name") {
```
stevehipwell: Why are we checking for changes here? I know that the GitHub API supports patching, but from the TF context I'm not sure of a use case where we'd conditionally send a value unless it's explicitly optional only and unset (checked with `d.Get`)?

github/resource_github_actions_hosted_runner_test.go
```diff
@@ -6,405 +6,303 @@ import (
...
+	"github.com/hashicorp/terraform-plugin-testing/statecheck"
+	"github.com/hashicorp/terraform-plugin-testing/tfjsonpath"
 )
 
 func TestAccGithubActionsHostedRunner(t *testing.T) {
 	t.Parallel()
+	skipUnlessHasPaidOrgs(t)
 
-	randomID := acctest.RandStringFromCharSet(5, acctest.CharSetAlphaNum)
+	runnerGroup := mustCreateTestOrganizationActionsRunnerGroup(t)
```
stevehipwell: Is there a benefit to doing this here instead of in each sub-test?

github/resource_github_actions_hosted_runner_test.go
```diff
@@ -6,405 +6,303 @@ import (
...
 
 		resource.Test(t, resource.TestCase{
 			PreCheck:          func() { skipUnlessHasPaidOrgs(t) },
 			ProviderFactories: providerFactories,
 			Steps: []resource.TestStep{
 				{
 					Config: config,
-					Check:  check,
+					ConfigStateChecks: []statecheck.StateCheck{
+						statecheck.ExpectKnownValue("github_actions_hosted_runner.test", tfjsonpath.New("name"), knownvalue.StringExact(hostedRunnerName)),
```
stevehipwell: We don't need to test values that have been configured as the test framework checks these at the end of each step.

github/resource_github_actions_hosted_runner_test.go
deiga: I was thinking about API rate limit reductions and saving a bit of time. The tests are already quite long since runners are async resources

stevehipwell: As it currently stands I'd make the tests isolated over optimising for rate limits as it simplifies the logic.

github/resource_github_actions_hosted_runner.go
```diff
@@ -30,6 +32,10 @@ func resourceGithubActionsHostedRunner() *schema.Resource {
 			Delete: schema.DefaultTimeout(10 * time.Minute),
 		},
 
+		Description: "This resource allows you to create and manage GitHub-hosted runners within your GitHub organization. You must have admin access to an organization to use this resource.",
```
stevehipwell:
```suggestion
		Description: "Resource to create and manage GitHub-hosted runners within an organization.",
```

I prefer the general/impersonal voice for the primary description.

RE Permissions and licence I think we ought to have a default block in each template where we add the required permissions and plan level. Something like below.

```markdown
| **Permission** |
| -------------- |
| `write`        |

| **Free** | **Team** | **Enterprise** | **Enterprise EMU** |
| -------- | -------- | -------------- | ------------------ |
| 🚫       | 🚫      | ✅             |                ✅ |
```

github/resource_github_actions_hosted_runner.go
```diff
@@ -183,366 +189,271 @@ func resourceGithubActionsHostedRunner() *schema.Resource {
...
-		result["id"] = fmt.Sprintf("%.0f", id)
-	}
-
-	if source, ok := image["source"].(string); ok {
-		result["source"] = source
-	}
-	if size, ok := image["size"].(float64); ok {
-		result["size_gb"] = int(size)
+func resourceGithubActionsHostedRunnerCreate(ctx context.Context, d *schema.ResourceData, m any) diag.Diagnostics {
+	err := checkOrganization(m)
```
stevehipwell: Could you switch over to `checkOrganizationOK`.

github/resource_github_actions_hosted_runner.go
```diff
@@ -69,35 +76,34 @@ func resourceGithubActionsHostedRunner() *schema.Resource {
...
 				Type:        schema.TypeString,
 				Required:    true,
-				Description: "Machine size (e.g., '4-core', '8-core'). Can be updated to scale the runner.",
+				Description: "Machine size for the hosted runner (e.g., `4-core`, `8-core`). Can be updated to scale the runner. To list available sizes, use the GitHub API: `GET /orgs/{org}/actions/hosted-runners/machine-sizes`.",
 			},
 			"runner_group_id": {
 				Type:        schema.TypeInt,
 				Required:    true,
-				Description: "The runner group ID.",
+				Description: "The ID of the runner group to assign this runner to.",
```
stevehipwell:
```suggestion
				Description: "ID of the runner group to assign this runner to.",
```

stevehipwell (changes requested):
A couple of minor change requests. I think the common permissions/plan docs should be a seperate PR but was just explaining why I don't think the top level description needs to mention this.

rilical (commented):
I found a few regressions in the move to the typed hosted-runner client that are worth addressing together. `last_active_on` can be null, but Create, Read, and Update currently format the zero value and leave `0001-01-01T00:00:00Z` in state. The `AcceptedError` path has a similar problem: `go-github` returns a nil runner with the error, then the code continues and copies empty or zero values into state. The typed response also no longer exposes `image_gen`, which the previous raw JSON Read preserved, so imports can lose that value. Could we keep the old null handling, avoid updating state from a nil runner, and either retain an `image_gen` fallback or document the limitation? It would also be good to restore the optional-parameter state checks that were removed from the acceptance test so these paths stay covered.

stevehipwell:
@deiga I'll wait for the changes requested by @rilical to be made before I review this again.

deiga:
This is now pending on https://github.com/integrations/terraform-provider-github/pull/3627, since the public IP checks are failing otherwise

PR #3527, opened 2026-07-07
fix: security_and_analysis.code_security bug

### LINKED ISSUE #3501: [BUG]: security_and_analysis.code_security causes a permanent diff (never read back / never written)
### Expected Behavior

When security_and_analysis.code_security.status is set to "enabled" (and the repository already has Code Security enabled on GitHub), terraform plan should show no changes. The provider should read the code_security status back from the GitHub API into state and send it on create/update, exactly as it already does for secret_scanning and secret_scanning_push_protection.

### Actual Behavior

terraform plan reports a permanent diff on every run, trying to add the code_security block even though the GitHub API already returns code_security.status = "enabled":

```tf
   ~ resource "github_repository" "this" {
       id = "<REPO>"
     ~ security_and_analysis {
         + code_security {
             + status = "enabled"
           }
           # (2 unchanged blocks hidden)   # secret_scanning / secret_scanning_push_protection
       }
   }
```

apply does not converge: the next plan shows the same diff again.

## Root cause

The code_security block is declared in the schema, but it is wired into neither the read nor the write path:

- flattenSecurityAndAnalysis() (READ) only sets advanced_security, secret_scanning, secret_scanning_push_protection — it never sets code_security, so the API value (enabled) is dropped and never stored in state.
- calculateSecurityAndAnalysis() (WRITE) only handles those same three fields — it never sends code_security to the API.

Because config always renders code_security {} but state never contains it, Terraform plans to add it forever, and apply never actually writes it.

##  Proposed fix

- https://github.com/integrations/terraform-provider-github/blob/f37a4b66a2832ca84af333459352d9dde6a2dbdf/github/resource_github_repository.go#L1190
- https://github.com/integrations/terraform-provider-github/blob/f37a4b66a2832ca84af333459352d9dde6a2dbdf/github/resource_github_repository.go#L553

```
   // calculateSecurityAndAnalysis() — WRITE
   if ok, status := tryGetSecurityAndAnalysisSettingStatus(lookup, "code_security"); ok {
       securityAndAnalysis.CodeSecurity = &github.CodeSecurity{
           Status: github.Ptr(status), // on main: new(status)
       }
   }
   
   // flattenSecurityAndAnalysis() — READ (conditional, like advanced_security)
   codeSecurity := securityAndAnalysis.GetCodeSecurity()
   if codeSecurity != nil {
       securityAndAnalysisMap["code_security"] = []any{map[string]any{
           "status": codeSecurity.GetStatus(),
       }}
   }
```

### Terraform Version

```
Terraform v1.15.2
on linux_amd64
```

### GitHub Installation Type

- [ ] I don't know

### Affected Resource(s)

- github_repository

```hcl
     name        = "code-security-repro"
     description = "repro for code_security perpetual diff"
     visibility  = "private"
   
     security_and_analysis {
       code_security {
         status = "enabled"
       }
       secret_scanning {
         status = "enabled"
       }
       secret_scanning_push_protection {
         status = "enabled"
       }
     }
   }


Precondition: the repository already has Code Security enabled on GitHub (e.g. enabled by an org-level security configuration). Confirm with:


gh api /repos/<OWNER>/<REPO> --jq '.security_and_analysis.code_security'
   {"status":"enabled"}
```

### Steps to Reproduce

1. Ensure the repo has code_security enabled on GitHub (verify with the gh api command above → "enabled").
2. Apply the configuration above (apply reports success).
3. Run `terraform plan` again.
   => Plan is NOT empty; it shows `+ code_security { status = "enabled" }`.
4. Run `terraform apply` then `terraform plan` repeatedly.
   => The same diff reappears every time; it never converges.

### Debug Output

```shell
# terraform plan (TF_LOG=DEBUG recommended; relevant excerpt)
   ~ resource "github_repository" "this" {
       id = "<REPO>"
     ~ security_and_analysis {
         + code_security {
             + status = "enabled"
           }
           # (2 unchanged blocks hidden)
       }
   }



   # Evidence the API already has it enabled (so this is a provider read/write gap, not API state):
   $ gh api /repos/<OWNER>/<REPO> --jq '.security_and_analysis'
   {
     "code_security": { "status": "enabled" },
     "secret_scanning": { "status": "enabled" },
     "secret_scanning_push_protection": { "status": "enabled" },
     ...
   }
```

### Code of Conduct

jtgorny:
Resolves #3501 

----

### Before the change?

- The `security_and_analysis.code_security` attribute was defined in the schema but never wired into the calculateSecurityAndAnalysis or flattenSecurityAndAnalysis functions. This meant that setting code_security had no effect on the API call, and the API response for code_security was never read back into state. This caused perpetual diffs. 

### After the change?

- calculateSecurityAndAnalysis now includes code_security when building the SecurityAndAnalysis request
- flattenSecurityAndAnalysis now reads CodeSecurity from the API response and writes it back into Terraform state.
- Added unit tests coverage

### Pull request checklist

- [ ] Yes
- [x] No

----


github/resource_github_repository_test.go
```diff
@@ -5,6 +5,7 @@ import (
 	"strings"
 	"testing"
 
+	goGithub "github.com/google/go-github/v88/github"
```
deiga:
```suggestion
	"github.com/google/go-github/v88/github"
```

deiga (changes requested):
Partial review. 


Please add an acceptance testcase as well

stevehipwell:
@deiga is this one of the parts of the repository resource that we planned to deprecate and replace with separate resources or data sources?

deiga:
@stevehipwell yep, though I haven't done a thorough analysis of feasibility yet

SamuZad:
Is there any chance this could get merged? I appreciate that this might get deprecated/replaced in the future, but in the interim it blocks enabling security features via terraform, which is not ideal

dekokun:
Not a maintainer, just helping with review triage, so this is a comment rather than a review.

The branch doesn't compile as it stands. Building the test package at 17566db:

```
github/resource_github_repository_test.go:8:2: no required module provides package github.com/google/go-github/v88/github
FAIL	github.com/integrations/terraform-provider-github/v6/github [setup failed]
```

`go.mod` here is already on `v89` (bumped in #3526) and there's no `v88` requirement left, and the rest of the package is on `v89` too (e.g. `acc_helpers_repository_test.go`). Switching that one import is enough: it compiles after that, and the three new unit tests pass for me.

Probably worth flagging that nothing has built this branch yet - the only checks showing up are `Setup`, `Label PRs` and `Check Acceptance Tests`, so the compile error isn't visible anywhere.

Also, FWIW, the acceptance test that was asked for looks like it already went in with the "addressing PR feedback" commit on Jul 8 (`manages code_security for a public repository`). I haven't run the acceptance tests myself.

PR #3530, opened 2026-07-09
feat: Check owner as part of provider configuration

stevehipwell:
Resolves #ISSUE_NUMBER

----

### Before the change?

- The owner was assumed to be valid with a check if it was an organization

### After the change?

- Owners are checked for validity (exists and can be accessed by the client) and if they're an organization

### Pull request checklist

- [ ] Yes
- [x] No

----

### DIFF
--- a/github/provider.go
+++ b/github/provider.go
@@ -2,7 +2,6 @@ package github
 
 import (
 	"context"
-	"errors"
 	"fmt"
 	"net/http"
 	"net/url"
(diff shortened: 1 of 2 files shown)


robert-crandall (approved):
Nice cleanup - using `Users.Get` to cover both users and orgs in one call is tidier than the old org-lookup-with-404-swallow, and validating the owner upfront is a genuinely helpful fail-fast. Tests look thorough too (I ran them locally and they pass). Thanks for this!

deiga (approved):
Based on the Users API description I would have guessed that this doesn't work, but I tested it with an Org and it works 🙃

PR #3535, opened 2026-07-13
feat: Add github_enterprise_app_installation resource

### LINKED ISSUE #3460: [FEAT]: Add support for installing GitHub Apps on all repositories in an organization
## Summary

Please add a new resource to support installing GitHub Apps on organizations with access to all repositories.

## Use Case

My goal is to install a GitHub App on an organization and grant it access to all repositories (`repository_selection = "all"`).

The existing `github_app_installation_repositories` resource does not provide a way to grant a GitHub App access to all repositories in an organization.

While it is possible to add repositories individually using the repository installation APIs, those APIs do not support an equivalent of `repository_selection = "all"`. As a result, Terraform users must enumerate and manage every repository individually.

For organizations with thousands of repositories, this approach is not practical and does not accurately model the desired state of "install this app on all repositories."

GitHub provides a [separate API ](https://docs.github.com/en/enterprise-cloud@latest/rest/enterprise-admin/organization-installations?apiVersion=2022-11-28#install-a-github-app-on-an-enterprise-owned-organization) that supports organization-wide installations with `repository_selection = "all"` within an enterprise, but this functionality is not currently represented by a Terraform resource.

## Proposed Solution

Introduce a new resource that manages GitHub App installations and supports granting access to all repositories in an organization within an enterprise.

## References

- https://docs.github.com/en/enterprise-cloud@latest/rest/apps/installations?apiVersion=2026-03-10#add-a-repository-to-an-app-installation
- https://docs.github.com/en/enterprise-cloud/rest/enterprise-admin/organization-installations#install-a-github-app-on-an-enterprise-owned-organization

### GitHub Installation Type

- [x] GitHub Enterprise Cloud with Managed Users/EMU (github.com)
- [ ] I don't know

### Relevant log output

```shell

```

### Code of Conduct

slang25:
Resolves #3460

----

### Before the change?

There is no way to install a GitHub App on an organization with access to **all** repositories via Terraform. `github_app_installation_repositories` only manages individual repositories on an existing installation, which is impractical for organizations with many repositories and doesn't model the "all repositories" desired state.

### After the change?

A new `github_enterprise_app_installation` resource manages GitHub App installations on enterprise-owned organizations via the [enterprise organization-installations API](https://docs.github.com/en/enterprise-cloud@latest/rest/enterprise-admin/organization-installations), available on GitHub Enterprise Cloud and GitHub Enterprise Server 3.19+ (verified against the [GHES 3.19/3.20 OpenAPI descriptions](https://github.com/github/rest-api-description/tree/main/descriptions/ghes-3.20)). The APIs are already present in go-github v89, so no new dependencies.

```hcl
resource "github_enterprise_app_installation" "example" {
  enterprise_slug      = "my-enterprise"
  organization         = "my-org"
  client_id            = "Iv1.abc123"
  repository_selection = "all" # or "selected" / "none"
}
```

Implementation notes:

- Resource ID is `<enterprise_slug>:<organization>:<client_id>`, so imports need no installation ID lookup; `installation_id` and `app_slug` are computed.
- `all ↔ selected` changes update in place via the toggle endpoint; repository set changes while staying `selected` use the grant/revoke endpoints (adding before removing so the selection is never empty). Transitions involving `none` force recreation, since the toggle endpoint only accepts `all`/`selected`.
- Repository lists are chunked to respect the API's 50-repositories-per-request limit.
- Plan-time validation requires `selected_repositories` exactly when `repository_selection = "selected"`.
- Acceptance tests are gated on a new `GH_TEST_ENTERPRISE_APP_CLIENT_ID` env var (the client ID of an installable app) alongside the existing enterprise test mode, so they skip cleanly where not configured.

I don't have access to an enterprise environment, so the CRUD paths have not been exercised against a live API — I'd appreciate a run of the enterprise acceptance tests. Happy to adjust naming/semantics if maintainers prefer a different shape; I noted #3460 is currently labeled blocked pending enterprise-scope work, but this follows the same `enterprise_slug` pattern as existing enterprise resources (e.g. `github_enterprise_organization`).

This PR was developed with AI assistance (Claude); I have reviewed the changes.

### Pull request checklist

- [ ] Schema migrations have been created if needed ([example](https://github.com/F-Secure-web/terraform-provider-github/blob/main/github/migrate_github_actions_organization_secret.go)) — not needed, new resource

- [ ] Yes
- [x] No

----

🤖 Generated with [Claude Code](https://claude.com/claude-code)


marcelosinhorini-jet (approved):

zry98:
Hi @slang25, I just found this PR after opening my https://github.com/integrations/terraform-provider-github/pull/3591 yesterday, and it seems better than my implementation in several places!

I noticed these are not covered currently:
1. It predates the [max per page to owner struct refactor](https://github.com/integrations/terraform-provider-github/pull/3555) merged on July 21, so it references the removed global.
2. data sources like `github_enterprise_app_installations`, `github_enterprise_app_installable_organizations` and `github_enterprise_app_accessible_organization_repositories`.

Would you update it?

zry98 (approved):

deiga:
- Please remove the docs template files and use the default template.
- Your PR description talks about installing a GitHub App to an Organization, yet you only created a resource for enterprise level installations. That seems out-of-place

deiga:
I still think there needs to be a distinction for "installation of an enterprise app to an Organisation" and "installation of an enterprise app to an Enterprise" and both of those should be covered in this PR.

zry98:
> I still think there needs to be a distinction for "installation of an enterprise app to an Organisation" and "installation of an enterprise app to an Enterprise" and both of those should be covered in this PR.

@deiga As far as I know, GitHub's current API doesn't have "installation of an enterprise app to an Enterprise"-related endpoints. Only installation to organizations owned by an enterprise.

deiga:
That is a fair point! 🤦‍♂️

slang25:
I've renamed the resources to avoid confusion, but happy to revert if we prefer terser names

github/resource_github_enterprise_organization_app_installation.go
```diff
@@ -0,0 +1,382 @@
...
+				Computed:    true,
+				Description: "The slug of the installed app.",
+			},
+		},
+
+		CustomizeDiff: customdiff.All(
+			// The API can only toggle an existing installation between 'all' and
+			// 'selected'; transitions involving 'none' require a reinstall.
+			customdiff.ForceNewIfChange("repository_selection", func(ctx context.Context, oldValue, newValue, meta any) bool {
+				return oldValue.(string) == "none" || newValue.(string) == "none"
```
mmizutani, line 91: When an existing `all` or `selected` installation receives a `repository_selection` that is unknown during planning and later resolves to `none`, this callback initially plans an in-place update but changes it to replacement during apply. Terraform then aborts with `Provider produced inconsistent final plan`. Please force replacement while the new selection is unknown so the initial and expanded plans agree.

github/resource_github_enterprise_organization_app_installation_test.go
```diff
@@ -0,0 +1,117 @@
...
+	"github.com/hashicorp/terraform-plugin-testing/statecheck"
+	"github.com/hashicorp/terraform-plugin-testing/tfjsonpath"
+)
+
+func TestAccGithubEnterpriseOrganizationAppInstallation(t *testing.T) {
+	appClientID := os.Getenv("GH_TEST_ENTERPRISE_APP_CLIENT_ID")
+	skipUnlessEnterpriseAppClientID := func(t *testing.T) {
+		t.Helper()
+		if appClientID == "" {
+			t.Skip("Skipping because GH_TEST_ENTERPRISE_APP_CLIENT_ID is not set")
```
mmizutani, line 20: `GH_TEST_ENTERPRISE_APP_CLIENT_ID` is not exported anywhere in the enterprise acceptance workflow, so this guard skips both new resource subtests on every CI run and leaves create/import/update/delete unexercised. Please add the fixture to the workflow environment so the new resource lifecycle actually runs in acceptance CI.

github/data_source_github_enterprise_app_installable_organizations.go
```diff
@@ -0,0 +1,88 @@
...
+
+	enterpriseSlug := d.Get("enterprise_slug").(string)
+
+	opts := &github.ListOptions{
+		PerPage: meta.maxPerPage,
+	}
+
+	results := make([]map[string]any, 0)
+	for {
+		organizations, resp, err := client.Enterprise.ListAppInstallableOrganizations(ctx, enterpriseSlug, opts)
```
mmizutani: This loop, the other two new data-source reads, and the resource Read/find helpers manually manage `NextPage` even though go-github v89 provides the corresponding generated `*Iter` methods (`ListAppInstallableOrganizationsIter`, `ListAppAccessibleOrganizationRepositoriesIter`, `ListAppInstallationsIter`, and `ListRepositoriesForOrgAppInstallationIter`). Please use those iterators for all five new loops so pagination and error handling stay centralized in the client, as required by the provider's pagination convention.

mmizutani (commented):
Thanks for working on this — I’m looking forward to using the feature. I found three issues that appear worth addressing before merge and left the details inline.

docs/data-sources/enterprise_organization_app_installations.md
deiga, line 1: issue: The docs generation plugin has an issue with nested Schemas and their descriptions. We have a few docs templates where the fix is to comment out the generated output and then copy-pasting the descriptions for those missing ones manually. Could you do that for this one as well? 🙏

examples/resources/github_enterprise_organization_app_installation/import.sh
deiga, line 1: Please also add a `import-by-string-id.tf` as well

github/resource_github_enterprise_organization_app_installation.go
```diff
@@ -0,0 +1,375 @@
...
+	return &schema.Resource{
+		Description: "Manage the installation of a GitHub App on an organization owned by an enterprise. " +
+			"The app is installed on the organization, not on the enterprise account itself; GitHub offers no API for the latter. " +
+			"This resource requires GitHub Enterprise Cloud or GitHub Enterprise Server 3.19+ and an authenticated user that is an enterprise owner.",
+		CreateContext: resourceGithubEnterpriseOrganizationAppInstallationCreate,
+		ReadContext:   resourceGithubEnterpriseOrganizationAppInstallationRead,
+		UpdateContext: resourceGithubEnterpriseOrganizationAppInstallationUpdate,
+		DeleteContext: resourceGithubEnterpriseOrganizationAppInstallationDelete,
+		Importer: &schema.ResourceImporter{
+			StateContext: schema.ImportStatePassthroughContext,
```
deiga, line 32: Please use a import function to ensure all necessary fields are set on import

github/resource_github_enterprise_organization_app_installation.go
```diff
@@ -0,0 +1,375 @@
...
+		},
+
```
deiga, line 34: Please add a `Timeouts` block

github/resource_github_enterprise_organization_app_installation.go
deiga: That seems to already been implemented in L87, right?

github/resource_github_enterprise_organization_app_installation.go
```diff
@@ -0,0 +1,375 @@
...
+				Description: "The ID of the installation.",
+			},
+			"app_slug": {
+				Type:        schema.TypeString,
+				Computed:    true,
+				Description: "The slug of the installed app.",
+			},
+		},
+
+		CustomizeDiff: customdiff.All(
```
deiga, line 80: Please create a separate function to include all of these

github/resource_github_enterprise_organization_app_installation.go
```diff
@@ -0,0 +1,375 @@
...
+				if selection != "selected" && repoCount > 0 {
+					return fmt.Errorf("'selected_repositories' can only be set when 'repository_selection' is 'selected'")
+				}
+				return nil
+			},
+		),
+	}
+}
+
+func resourceGithubEnterpriseOrganizationAppInstallationCreate(ctx context.Context, d *schema.ResourceData, meta any) diag.Diagnostics {
```
deiga, line 114:
```suggestion
func resourceGithubEnterpriseOrganizationAppInstallationCreate(ctx context.Context, d *schema.ResourceData, m any) diag.Diagnostics {
```

github/resource_github_enterprise_organization_app_installation.go
```diff
@@ -0,0 +1,375 @@
...
+		ClientID:            clientID,
+		RepositorySelection: d.Get("repository_selection").(string),
+		Repositories:        initial,
+	}
+
+	tflog.Debug(ctx, "Installing app on enterprise-owned organization", map[string]any{
+		"client_id":  clientID,
+		"org":        org,
+		"enterprise": enterpriseSlug,
+	})
```
deiga, line 142: Please compact these to a single line

github/resource_github_enterprise_organization_app_installation.go
```diff
@@ -0,0 +1,375 @@
...
+
+	enterpriseSlug, org, clientID, err := parseID3(d.Id())
+	if err != nil {
+		return diag.FromErr(err)
+	}
+
+	installation, err := findEnterpriseOrganizationAppInstallation(ctx, owner, enterpriseSlug, org, clientID)
+	if err != nil {
+		var ghErr *github.ErrorResponse
+		if errors.As(err, &ghErr) && ghErr.Response.StatusCode == http.StatusNotFound {
```
deiga, line 183: Please use the `errors.AsType` convention

deiga (changes requested):
Partial review

PR #3539, opened 2026-07-14
[BUG]: fix lowercase permadiff team members

### LINKED ISSUE #3533: [BUG]: v6.13.0 lower cases all team members
### Expected Behavior

Usernames shouldn't be lowercased, of if they are lowercased consistently lowercased.
v6.13.0 brought in some lowercasing on team members, v6.12.1 was good.

### Actual Behavior

```
  # github_team_members.misc_devs has changed
  ~ resource "github_team_members" "misc_devs" {
        id        = "123456"

      - members {
          - role     = "member" -> null
          - username = "ChrisSomethingSomething" -> null
        }
      + members {
          + role     = "member"
          + username = "chrissomethingsomething"
        }

        # (6 unchanged blocks hidden)
    }
```

```
  # github_team_members.misc_devs will be updated in-place
  ~ resource "github_team_members" "misc_devs" {
        id        = "123456"

      - members {
          - role     = "member" -> null
          - username = "chrissomethingsomething" -> null
        }
      + members {
          + role     = "member"
          + username = "ChrisSomethingSomething"
        }

        # (6 unchanged blocks hidden)
    }
```

### Terraform Version

OpenTofu v1.12.0
on linux_amd64
+ provider registry.opentofu.org/integrations/github v6.13.0

### GitHub Installation Type

- [ ] I don't know

### Affected Resource(s)

- github_team_members

```hcl

```

### Steps to Reproduce

_No response_

### Debug Output

```shell

```

### Code of Conduct

deiga:
Resolves #3533

----

### Before the change?

- username lower casing is causing permadiff

### After the change?

- fixes permadiff by storing username as lowercase in state

### Pull request checklist

- [ ] Yes
- [x] No

----

### DIFF
--- a/github/resource_github_team_members.go
+++ b/github/resource_github_team_members.go
@@ -406,11 +406,12 @@ func updateTeamMembers(ctx context.Context, meta *Owner, slug string, wantMember
 	}
 
 	for _, member := range currentMembers {
-		if _, ok := want[member.login]; !ok {
-			tflog.Debug(ctx, "Removing team member.", map[string]any{"team_slug": slug, "username": member.login})
+		login := strings.ToLower(member.login)
+		if _, ok := want[login]; !ok {
+			tflog.Debug(ctx, "Removing team member.", map[string]any{"team_slug": slug, "username": login})
 
-			if _, err := client.Teams.RemoveTeamMembershipBySlug(ctx, orgName, slug, member.login); err != nil {
-				return fmt.Errorf("could not remove existing team member %q: %w", member.login, err)
+			if _, err := client.Teams.RemoveTeamMembershipBySlug(ctx, orgName, slug, login); err != nil {
+				return fmt.Errorf("could not remove existing team member %q: %w", login, err)
 			}
 		}
 	}
--- a/github/resource_github_team_members.go
+++ b/github/resource_github_team_members.go
@@ -65,7 +65,8 @@ func resourceGithubTeamMembers() *schema.Resource {
 				Description: "List of users that should be members of the team.",
 				Set: func(v any) int {
 					username, _ := v.(map[string]any)["username"].(string)
-					return schema.HashString(strings.ToLower(username))
+					role, _ := v.(map[string]any)["role"].(string)
+					return schema.HashString(strings.ToLower(username) + ":" + strings.ToLower(role))
 				},
 				Elem: &schema.Resource{
 					Schema: map[string]*schema.Schema{
--- a/github/resource_github_team_members.go
+++ b/github/resource_github_team_members.go
@@ -64,16 +64,19 @@ func resourceGithubTeamMembers() *schema.Resource {
 				Required:    true,
 				Description: "List of users that should be members of the team.",
 				Set: func(v any) int {
-					username := v.(map[string]any)["username"].(string)
+					username, _ := v.(map[string]any)["username"].(string)
 					return schema.HashString(strings.ToLower(username))
 				},
 				Elem: &schema.Resource{
 					Schema: map[string]*schema.Schema{
 						"username": {
-							Type:             schema.TypeString,
-							Required:         true,
-							DiffSuppressFunc: caseInsensitive(),
-							Description:      "User to add to the team.",
+							Type:     schema.TypeString,
+							Required: true,
+							StateFunc: func(v any) string {
+								val, _ := v.(string)
+								return strings.ToLower(val)
+							},
+							Description: "User to add to the team.",
 						},
 						"role": {
 							Type:             schema.TypeString,
--- a/github/resource_github_team_members.go
+++ b/github/resource_github_team_members.go
@@ -63,11 +63,6 @@ func resourceGithubTeamMembers() *schema.Resource {
 				Type:        schema.TypeSet,
 				Required:    true,
 				Description: "List of users that should be members of the team.",
-				Set: func(v any) int {
-					username, _ := v.(map[string]any)["username"].(string)
-					role, _ := v.(map[string]any)["role"].(string)
-					return schema.HashString(strings.ToLower(username) + ":" + strings.ToLower(role))
-				},
 				Elem: &schema.Resource{
 					Schema: map[string]*schema.Schema{
 						"username": {
--- a/github/resource_github_team_members.go
+++ b/github/resource_github_team_members.go
@@ -69,11 +69,13 @@ func resourceGithubTeamMembers() *schema.Resource {
 							Type:             schema.TypeString,
 							Required:         true,
 							DiffSuppressFunc: caseInsensitive(),
+							Description:      "User to add to the team.",
+							// This seems to be the only way to ensure that the username is in lowercase.
+							// Without this the tests fail because the value is compared in a case-sensitive manner.
 							StateFunc: func(v any) string {
 								val, _ := v.(string)
 								return strings.ToLower(val)
 							},
-							Description: "User to add to the team.",
 						},
 						"role": {
 							Type:             schema.TypeString,
--- a/github/resource_github_team_members.go
+++ b/github/resource_github_team_members.go
@@ -70,8 +70,9 @@ func resourceGithubTeamMembers() *schema.Resource {
 				Elem: &schema.Resource{
 					Schema: map[string]*schema.Schema{
 						"username": {
-							Type:     schema.TypeString,
-							Required: true,
+							Type:             schema.TypeString,
+							Required:         true,
+							DiffSuppressFunc: caseInsensitive(),
 							StateFunc: func(v any) string {
 								val, _ := v.(string)
 								return strings.ToLower(val)
--- a/github/resource_github_team_members.go
+++ b/github/resource_github_team_members.go
@@ -63,6 +63,10 @@ func resourceGithubTeamMembers() *schema.Resource {
 				Type:        schema.TypeSet,
 				Required:    true,
 				Description: "List of users that should be members of the team.",
+				Set: func(v any) int {
+					username := v.(map[string]any)["username"].(string)
+					return schema.HashString(strings.ToLower(username))
+				},
 				Elem: &schema.Resource{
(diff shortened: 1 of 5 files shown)


stevehipwell (changes requested):
I think changing `Set` would technically be a breaking change, but I think we ought to do it anyway as any breaking issues would be catching incorrect usage.

RE the username casing, I think that we should keep the diff suppression pattern as that way we don't lose any of the input data.

stevehipwell:
@deiga FYI you should be able to add acceptance tests with alternative cases using the existing users.

github/resource_github_team_members.go
```diff
@@ -63,13 +63,21 @@ func resourceGithubTeamMembers() *schema.Resource {
...
+					return schema.HashString(strings.ToLower(username))
+				},
 				Elem: &schema.Resource{
 					Schema: map[string]*schema.Schema{
 						"username": {
 							Type:             schema.TypeString,
 							Required:         true,
 							DiffSuppressFunc: caseInsensitive(),
-							Description:      "User to add to the team.",
+							StateFunc: func(v any) string {
```
stevehipwell: I don't think we need this?

deiga: Without this, importing a mixed-case username will create a diff.

stevehipwell: Yeah, because the code is missing a check to reconcile the config members with the actual members as lower case. So we don't need this, but we do need a fix.

I need to revert the GraphQL change as `go-github` now supports filtering the members so I can take a look at this using your regression test (I think I've handled this in the repository collaborators resource).

stevehipwell (changes requested):
I don't think we should use `flippedCaseUsername` anywhere other than the targeted test. This will fix the imports test as it checks the state directly and does not use Terraform logic for equality.

deiga:
@stevehipwell that's definitely an option. I decided to add it to multiple tests as it highlighted problems with cases in multiple use-cases. But I can try to make the case sensitivity test cover all the same problems instead

github/resource_github_team_members.go
deiga: After going through multiple different iterations, it seems to me that `StateFunc` is the only proper solution here.
Yes, it's deprecated in Framework, but with suggestions to move the logic to Create.

I tried multiple different ways, but I couldn't get Create to store the username values in a way that would have allowed the case insensitive test to pass.

github/resource_github_team_members.go
```diff
@@ -70,6 +70,12 @@ func resourceGithubTeamMembers() *schema.Resource {
 							Required:         true,
 							DiffSuppressFunc: caseInsensitive(),
 							Description:      "User to add to the team.",
+							// This seems to be the only way to ensure that the username is in lowercase.
+							// Without this the tests fail because the value is compared in a case-sensitive manner.
+							StateFunc: func(v any) string {
+								val, _ := v.(string)
+								return strings.ToLower(val)
+							},
```
stevehipwell: Which tests? The import test is not valid when using non-lowercase usernames, but this is fine as it's an artificial pattern. You could add complexity and sore the initial case (either from config or import) and then use a lookup to retain the source case, but this would still fail the synthetic import test if the config case was different to the API case.

stevehipwell (changes requested):

github/resource_github_team_members.go
deiga: The `is_case_insensitive` test is failing without `StateFunc`. Both the ExpectKnowValue and Comparer Checks for username fail.

Since `username` is case insensitive in the GitHub API, I don't really understand why we'd need to store the initial case (in any other form than is already in the TF config)

stevehipwell: If we want to store values as lowercase we should handle it directly in our code, there is no need for `StateFunc` (which isn't present in the framework); this would fix the test and isn't a bad idea. However the real issue is the test, you're testing an implementation detail (state equality) instead of just the API surface (changing the case is a no-op). For me the following test steps would fully exercise the API surface.

- `is_case_insensitive`
  - Step 1
    - Config: Two members with the same username but different cases
    - Result: Error
  - Step 2
    - Config: Single member
    - Result: Success
  - Step 3
    - Config: Same member but change case of username
    - Result: No-op

github/resource_github_team_members.go
```diff
@@ -63,6 +63,12 @@ func resourceGithubTeamMembers() *schema.Resource {
 				Type:        schema.TypeSet,
 				Required:    true,
 				Description: "List of users that should be members of the team.",
+				// The Set hash function ensures that the same user cannot be added to the team multiple times with different case.
```
stevehipwell: This comment isn't correct, it would just block the same user plus role combo with different case. Given that it doesn't act as a block for duplicates and you need to use a diff function to actually block I don't think this code is necessary. I ended up at the same decision after copying your code form this PR in a different one.

github/resource_github_team_members.go
```diff
@@ -28,7 +28,7 @@ func resourceGithubTeamMembers() *schema.Resource {
 			StateContext: resourceGithubTeamMembersImport,
 		},
 
-		CustomizeDiff: customdiff.Sequence(diffLegacyTeamID, diffLegacyTeam),
+		CustomizeDiff: customdiff.Sequence(resourceGithubTeamMembersDiff, diffLegacyTeamID, diffLegacyTeam),
```
stevehipwell: This should just call `resourceGithubTeamMembersDiff` and it can call the other functions.

github/resource_github_team_members.go
```diff
@@ -85,6 +91,20 @@ func resourceGithubTeamMembers() *schema.Resource {
...
+
+	if err := diffNestedUsernameCheck(ctx, d, "members"); err != nil {
+		return fmt.Errorf("error diffing members config: %w", err)
+	}
+
+	if d.Id() == "" {
+		return nil
+	}
+
+	return nil
```
stevehipwell:
```suggestion
	return nil
```

github/util_diff.go
```diff
@@ -332,3 +333,33 @@ func suppressUnorderedListDiff(fieldKey string, f func(a, b any) int) schema.Sch
 		return reflect.DeepEqual(oldList, newList)
 	}
 }
+
+func diffNestedUsernameCheck(ctx context.Context, d *schema.ResourceDiff, fieldKey string) error {
```
stevehipwell:
```suggestion
func diffDuplicateUsernameCheck(ctx context.Context, d *schema.ResourceDiff, fieldKey string) error {
```
I think this would be clearer?

stevehipwell (changes requested):

github/resource_github_team_members.go
deiga: That seems like an unnecessary change as it increases verbosity without improving readability IMO, but I'll do it

github/resource_github_team_members.go
deiga: After removing the  `Set` function the case insensitive test starts failing as the default Set function considers different cased usernames as separate entries. This is IMO the wrong behaviour.

```
=== CONT  TestAccGithubTeamMembers/is_case_insensitive
    resource_github_team_members_test.go:585: Step 2/3 error: After applying this test step, the refresh plan was not empty.
        stdout
        
        
          ~ update in-place
        
          # github_team_members.test will be updated in-place
          ~ resource "github_team_members" "test" {
                id        = "19113114"
        
              - members {
                  - role     = "maintainer" -> null
                  - username = "gh-terraform-testing-test1" -> null
                }
              + members {
                  + role     = "maintainer"
                  + username = "Gh-terraform-testing-test1"
                }
            }
        
```

stevehipwell (approved):
LGTM

github/resource_github_team_members.go
```diff
@@ -400,11 +417,12 @@ func updateTeamMembers(ctx context.Context, meta *Owner, slug string, wantMember
 	}
 
 	for _, member := range currentMembers {
-		if _, ok := want[member.login]; !ok {
-			tflog.Debug(ctx, "Removing team member.", map[string]any{"team_slug": slug, "username": member.login})
+		login := strings.ToLower(member.login)
```
robert-crandall, line 420: I’m still seeing the perpetual diff for mixed-case usernames; can you add a test to verify this is case insensitive?

Here's an example test that's failing in this PR:

```go
package github

import (
	"testing"

	"github.com/hashicorp/go-cty/cty"
	"github.com/hashicorp/terraform-plugin-sdk/v2/terraform"
)

func TestGithubTeamMembersCaseOnlyDiff(t *testing.T) {
	for _, username := range []string{"mixedcaseuser", "MixedCaseUser"} {
		t.Run(username, func(t *testing.T) {
			r := resourceGithubTeamMembers()

			// Match the lowercase state produced by the read function.
			d := schema.TestResourceDataRaw(t, r.Schema, map[string]any{
				"team_slug": "example",
				"team_id":   "123",
				"members": []any{
					map[string]any{
						"username": "mixedcaseuser",
						"role":     "member",
					},
				},
			})
			d.SetId("123")

			config := cty.ObjectVal(map[string]cty.Value{
				"team_slug": cty.StringVal("example"),
				"team_id":   cty.NullVal(cty.String),
				"members": cty.SetVal([]cty.Value{
					cty.ObjectVal(map[string]cty.Value{
						"username": cty.StringVal(username),
						"role":     cty.StringVal("member"),
					}),
				}),
			})

			// CustomizeDiff reads RawConfig from the instance state.
			state := d.State()
			state.RawConfig = config

			diff, err := r.Diff(
				t.Context(),
				state,
				terraform.NewResourceConfigShimmed(config, r.CoreConfigSchema()),
				&Owner{},
			)
			if err != nil {
				t.Fatal(err)
			}

			if diff != nil && !diff.Empty() {
				for key, attr := range diff.Attributes {
					t.Logf("%s: %q -> %q", key, attr.Old, attr.New)
				}
				t.Fatal("username casing alone must not produce a diff")
			}
		})
	}
}
```

PR #3547, opened 2026-07-16
[BUG] fix rulesets `update_allows_fetch_and_merge` validation

### LINKED ISSUE #3543: `update_allows_fetch_and_merge` rejected as invalid rule type for branch-target `github_repository_ruleset` (regression in 6.11.0's CustomizeDiff validation)
### Expected Behavior
`update_allows_fetch_and_merge` should be accepted as a parameter of the `update` rule for a `branch`-target `github_repository_ruleset`, as documented, and as it worked prior to provider 6.11.0.

### Actual Behavior
`terraform plan` fails with:
```
Error: rule "update_allows_fetch_and_merge" is not valid for branch target; branch targets only support: [creation update deletion required_linear_history required_signatures pull_request required_status_checks non_fast_forward commit_message_pattern commit_author_email_pattern committer_email_pattern branch_name_pattern tag_name_pattern workflows code_scanning required_deployments merge_queue copilot_code_review]
```
This happens even though the same config previously worked and the setting ("Allow fork syncing" in the GitHub UI) is still active on the repository — GitHub's API itself accepts this parameter fine, so the current state already has `update_allows_fetch_and_merge = true` on that ruleset.

### Root cause
The `CustomizeDiff` validation added in v6.11.0 (#2958, `util_ruleset_validation.go`, `validateRules`/`branchTagOnlyRules`) iterates over every attribute set under `rules { }` and rejects any name that isn't in a hardcoded list of `github.RepositoryRuleType` values. `update_allows_fetch_and_merge` is not itself a rule type — it's a boolean *parameter* of the `update` rule type (mirrors `UpdateRuleParameters.UpdateAllowsFetchAndMerge` in go-github) — so it can never appear in that list and is always rejected once it's `true`.

### Terraform / Provider Version
- Terraform v1.14.8
- `integrations/github` v6.13.0 (reproduces on all versions ≥ 6.11.0; confirmed working on 6.6.0)

### Affected Resource
`github_repository_ruleset`

### Steps to Reproduce
```hcl
resource "github_repository_ruleset" "this" {
  name        = "main"
  repository  = "some-repo"
  target      = "branch"
  enforcement = "active"

  conditions {
    ref_name {
      include = ["refs/heads/main"]
      exclude = []
    }
  }

  rules {
    update                        = true
    update_allows_fetch_and_merge = true
  }
}
```
`terraform plan` (no apply needed — the error is raised by `CustomizeDiff`, purely client-side).

### Additional context
Confirmed this is not fixable via `lifecycle { ignore_changes = [...] }` on that attribute — Terraform Core reverts ignored attributes to their prior-state value before invoking the provider's plan, and since the attribute is already `true` in state, it's fed into the same buggy validation on every plan regardless of whether it's ignored, present, or absent.

deiga:
Resolves #3543

----

### Before the change?

- `rules.0.update_allows_fetch_and_merge` is being marked as invalid config for repository ruleset

### After the change?

- `rules.0.update_allows_fetch_and_merge` will be allowed as  config for repository rulesets
- Adds Create and Update validation to ensure only Forked repos can set `rules.0.update_allows_fetch_and_merge`
- Adds test helper for archiving a repository and for configuring repository visibility
- Refactors `github_repository_ruleset` tests to follow modern practices

### Pull request checklist

- [ ] Yes
- [x] No

----


github/acc_helpers_test.go
```diff
@@ -219,7 +222,38 @@ func mustAddTeamMember(t *testing.T, team *github.Team, username string) {
 	}
 }
 
+func mustArchiveTestRepository(t *testing.T, repo *github.Repository) {
```
stevehipwell: If you need to retry in this function, which I question for a test function, then could you use `retryUntilOK`. That said see below for what I think might be the actual solution.

github/acc_helpers_test.go
```diff
@@ -219,7 +222,38 @@ func mustAddTeamMember(t *testing.T, team *github.Team, username string) {
...
+
 func mustCreateTestRepository(t *testing.T) *github.Repository {
+	return mustCreateConfigurableTestRepository(t, "")
+}
+
+func mustCreateInternalTestRepository(t *testing.T) *github.Repository {
+	return mustCreateConfigurableTestRepository(t, "internal")
+}
+
+func mustCreateConfigurableTestRepository(t *testing.T, visibility string) *github.Repository {
```
stevehipwell: The pattern I'm working towards is below, with the expectation that we probably want to move them into separate files such as `acc_helpers_repository_test.go`.

```go
type createTestRepositoryOptionsFunc func(*github.Repository)

func mustCreateTestRepository(t *testing.T, f ...createTestRepositoryOptionsFunc) *github.Repository {
	t.Helper()

	randomID := acctest.RandString(testRandomIDLength)
	name := fmt.Sprintf("%s%s", testResourcePrefix, randomID)

	req := &github.Repository{
		Name:     &name,
		AutoInit: new(true),
	}

	for _, fn := range f {
		fn(req)
	}

	repo, _, err := testAccConf.meta.v3client.Repositories.Create(t.Context(), testAccConf.meta.name, req)
	if err != nil {
		t.Fatalf("failed to create test repository: %v", err)
	}

	t.Cleanup(func() {
		if _, err := testAccConf.meta.v3client.Repositories.Delete(context.Background(), testAccConf.meta.name, name); err != nil {
				return
			}
			t.Logf("failed to delete test repository %s: %v", name, err)
		}
	})

	return repo
}
```

github/resource_github_repository_ruleset_test.go
```diff
@@ -24,25 +24,17 @@ func TestAccGithubRepositoryRuleset(t *testing.T) {
...
 
 		config := fmt.Sprintf(`
-resource "github_repository" "test" {
-	name = "%s"
-	auto_init = true
-	vulnerability_alerts = true
-	visibility = "%s"
-}
-
 resource "github_repository_environment" "example" {
```
stevehipwell: You can create the environment outside of the TF cnfig too.

github/resource_github_repository_ruleset_test.go
```diff
@@ -124,31 +116,152 @@ resource "github_repository_ruleset" "test" {
...
-						resource.TestCheckResourceAttr("github_repository_ruleset.test", "rules.0.pull_request.0.allowed_merge_methods.#", "2"),
-						resource.TestCheckResourceAttr("github_repository_ruleset.test", "rules.0.required_code_scanning.0.required_code_scanning_tool.0.alerts_threshold", "errors"),
-						resource.TestCheckResourceAttr("github_repository_ruleset.test", "rules.0.required_code_scanning.0.required_code_scanning_tool.0.security_alerts_threshold", "high_or_higher"),
-						resource.TestCheckResourceAttr("github_repository_ruleset.test", "rules.0.required_code_scanning.0.required_code_scanning_tool.0.tool", "CodeQL"),
-						resource.TestCheckResourceAttr("github_repository_ruleset.test", "rules.0.copilot_code_review.0.review_on_push", "true"),
-						resource.TestCheckResourceAttr("github_repository_ruleset.test", "rules.0.copilot_code_review.0.review_draft_pull_requests", "false"),
-					),
+					ConfigStateChecks: []statecheck.StateCheck{
+						statecheck.ExpectKnownValue("github_repository_ruleset.test", tfjsonpath.New("name"), knownvalue.StringExact("test")),
```
stevehipwell: You don't need to check for value in the config, the framework does this for you.

github/acc_helpers_test.go
deiga: I'll look at retry until ok. 
Retry is my choice as a static sleep felt worse than a retry. And since repo creation is async and archiving right after creation usually returns a 422, something is necessary

stevehipwell: Given that it's for a test, wouldn't the flexible repo create pattern below be a better option as it's allow you to create a repo already archived?

deiga: It would be, but it's not possible to create repos as "archived" 🫠

github/acc_helpers_test.go
deiga: This is way better! I'll adopt this

github/acc_helpers_test.go
stevehipwell: I did wonder, so I think using `retryUntilOK` in the helper would be the best bet.

github/resource_github_repository_ruleset.go
```diff
@@ -32,6 +32,8 @@ func resourceGithubRepositoryRuleset() *schema.Resource {
 
 		CustomizeDiff: resourceGithubRepositoryRulesetDiff,
 
+		Description: "This resource allows you to create and manage rulesets on the repository level. When applied, a new ruleset will be created. When destroyed, that ruleset will be removed.",
```
stevehipwell:
```suggestion
		Description: "Resource to manage GitHub repository rulesets.",
```

github/resource_github_repository_ruleset.go
```diff
@@ -702,6 +705,10 @@ func resourceGithubRepositoryRulesetCreate(ctx context.Context, d *schema.Resour
 		return diag.Errorf("cannot create ruleset on archived repository %s/%s", owner, repoName)
 	}
 
+	if v, _ := d.Get("rules.0.update_allows_fetch_and_merge").(bool); v && !repo.GetFork() { //nolint:staticcheck // SA1019: d.GetOkExists is deprecated but necessary for bool fields
```
stevehipwell:
```suggestion
	if v, _ := d.Get("rules.0.update_allows_fetch_and_merge").(bool); v && !repo.GetFork() {
```

github/resource_github_repository_ruleset.go
```diff
@@ -814,6 +821,9 @@ func resourceGithubRepositoryRulesetUpdate(ctx context.Context, d *schema.Resour
 		tflog.Info(ctx, "Repository is archived, skipping ruleset update", map[string]any{"owner": owner, "repo_name": repoName})
 		return nil
 	}
+	if v, _ := d.Get("rules.0.update_allows_fetch_and_merge").(bool); v && !repo.GetFork() { //nolint:staticcheck // SA1019: d.GetOkExists is deprecated but necessary for bool fields
```
stevehipwell:
```suggestion
	if v, _ := d.Get("rules.0.update_allows_fetch_and_merge").(bool); v && !repo.GetFork() {
```

stevehipwell (changes requested):
Please add a _import-by-string-id.tf_ file to the examples.

rilical (commented):
The validation fix itself looks focused, but the surrounding cleanup dropped two useful pieces. `create_branch_ruleset` used to verify the resulting bypass actors and nested rules; it now only checks that Terraform plans a Create. The generated docs also lost the note explaining which rules apply to branch and tag targets versus push targets when the custom template was removed. Could we restore the state assertions alongside the new plan check and move that guidance into a schema description or small template section so it survives regeneration?

deiga:
> The validation fix itself looks focused, but the surrounding cleanup dropped two useful pieces. `create_branch_ruleset` used to verify the resulting bypass actors and nested rules; it now only checks that Terraform plans a Create. The generated docs also lost the note explaining which rules apply to branch and tag targets versus push targets when the custom template was removed. Could we restore the state assertions alongside the new plan check and move that guidance into a schema description or small template section so it survives regeneration?

@rilical 
The state assertions are automatic for things in the TF config, so adding them back doesn't bring benefits.
I'll re-add the information about rule applicability

rilical (approved):
looks good to me

stevehipwell (changes requested):
This PR looks good. But same as #3271 I'd like to see a template per resource/data-source.

PR #3549, opened 2026-07-17
feat: Refactor external groups data source

stevehipwell:
Related to #3537

----

### Before the change?

- `github_external_group` data source didn't use max per page resulting in a slow query time
- `github_external_group` didn't support filtering results
- `github_external_group` had no tests

### After the change?

- `github_external_group` has been refactored to the latest patterns
  - Use max per page
  - Add `display_name_filter` to filter returned groups
  - Updated documentation
  - Acceptance tests

### Pull request checklist

- [ ] Yes
- [x] No

----


github/data_source_github_external_groups_test.go
```diff
@@ -0,0 +1,82 @@
...
+	"github.com/hashicorp/terraform-plugin-testing/statecheck"
+	"github.com/hashicorp/terraform-plugin-testing/tfjsonpath"
+)
+
+func TestAccDataSourceGithubExternalGroups(t *testing.T) {
+	t.Parallel()
+
+	if testAccConf.authMode != enterprise && !testAccConf.enterpriseIsEMU {
+		t.Skip("Skipping as test mode is not enterprise using EMU")
+	}
```
deiga:
```suggestion
	skipUnlessEMUEnterprise(t)
```

deiga (dismissed):
LGTM

github/data_source_github_external_groups_test.go
stevehipwell: I'm happy to make this change if you're about to re-approve otherwise I'll pick it up in a separate PR.

stevehipwell: Actually, I'll make the change anyway; I want to add another minor tweak to the tests.

rilical (changes requested):
mistake.

rilical:
Could you take another look at the `updated_at` mapping and the new tests? `Timestamp.String()` changes the existing RFC3339 value and turns a nil timestamp into a zero timestamp. The tests also always skip because the external-group fixtures are never populated, and their assertions target `groups` rather than `external_groups`.

Once #3555 lands, this will also need a rebase to use `meta.maxPerPage`.

stevehipwell:
> Could you take another look at the updated_at mapping and the new tests? Timestamp.String() changes the existing RFC3339 value and turns a nil timestamp into a zero timestamp. 

@rilical the old code inconsistent and poorly implemented with the non standard `updated_at` date format undocumented; I consider this PR as a fix for this. RE the question of `nil`, the API schema has `updated_at` as a required field.

> The tests also always skip because the external-group fixtures are never populated, and their assertions target groups rather than external_groups.

The test are intended to be skipped as there is no EMU organization with external groups configured to test against. I've fixed the test assertions.

rilical:
@stevehipwell Fair enough on nil and the skipped EMU tests; the assertion fix resolves that part. One issue remains: this field currently returns RFC3339, and switching to Timestamp.String() breaks Terraform time functions such as formatdate. Could we keep it as RFC3339 with Format(time.RFC3339)?

stevehipwell:
@rilical you're absolutely right, timestamps in TF should always be RFC3339 (it's late and I'm getting things mixed up). On the back of this I think we need an issue for v7 to track fixing all timestamp fields as it was the secret & variable ones I kept as non-RFC3339.

rilical (approved):

stevehipwell:
@deiga could you please re-approve this?

deiga (approved):
LGTM

PR #3551, opened 2026-07-18
docs: update GitHub App env-var auth guidance for v6.13.0 blockless behavior

### LINKED ISSUE #3550: [DOCS]: Empty app_auth {} recommendation fails validate without credentials
### Describe the need

[`docs/index.md`, GitHub App Installation section](https://github.com/integrations/terraform-provider-github/blob/v6.13.0/docs/index.md?plain=1#L81) says:

> When using environment variables, an empty `app_auth` block is required to allow provider configurations from environment variables to be specified. See: <https://github.com/hashicorp/terraform-plugin-sdk/issues/142>

and the [`main.tf` example below it](https://github.com/integrations/terraform-provider-github/blob/v6.13.0/docs/index.md?plain=1#L93-L98) shows the empty block pattern. Following this recommendation produces a configuration that cannot pass `terraform validate` / `tofu validate` unless the `GITHUB_APP_*` environment variables happen to be set in the shell running validate:

```terraform
provider "github" {
  owner = "my-org"
  app_auth {} # documented pattern for env-var auth
}
```

**Why this happens:** the `app_auth` attributes are declared `Required: true` with `DefaultFunc: schema.EnvDefaultFunc(...)` ([`github/provider.go#L70-L90` at v6.13.0](https://github.com/integrations/terraform-provider-github/blob/v6.13.0/github/provider.go#L70-L90)). Validation is only satisfied when the DefaultFunc returns a value, i.e. when the environment variable exists at validate time. Apply-time environments (CI/CD runners with injected credentials) have them; validate-only contexts such as CI linting or local pre-commit hooks usually do not, so the documented pattern fails exactly where teams run `validate`.

**Why the recommendation is obsolete:** the note works around hashicorp/terraform-plugin-sdk#142 (still open), but since #3448 (first released in v6.13.0) the provider no longer depends on it: [`getAppAuth()`](https://github.com/integrations/terraform-provider-github/blob/v6.13.0/github/provider.go#L669-L677) reads `GITHUB_APP_ID`, `GITHUB_APP_INSTALLATION_ID` and `GITHUB_APP_PEM_FILE` directly, before consulting the block. Omitting `app_auth` entirely authenticates as the GitHub App and validates cleanly everywhere.

**Suggested fix** for the note and example in `docs/index.md` (no per-version guidance needed; registry docs are versioned, so older pins see the older docs):

```markdown
~> No `app_auth` block is needed when authenticating through the
`GITHUB_APP_XXX` environment variables; the provider reads them directly.
The previously documented empty `app_auth {}` block is no longer required
and should be removed: its arguments are schema-required, so it fails
`terraform validate` in environments where the variables are not set.
```

```terraform
provider "github" {
  owner = var.github_organization
  # Credentials come from the GITHUB_APP_XXX environment variables.
}
```

### Relevant log output

```shell
$ tofu validate   # no GITHUB_APP_* variables set
Error: Missing required argument
  The argument "id" is required, but no definition was found.
Error: Missing required argument
  The argument "installation_id" is required, but no definition was found.
Error: Missing required argument
  The argument "pem_file" is required, but no definition was found.
```

### Code of Conduct

SebastianBienert:
Resolves #3550

----

### Before the change?

- The GitHub App Installation docs state that an empty `app_auth {}` block is required when authenticating through the `GITHUB_APP_XXX` environment variables, and the `main.tf` example shows that pattern.
- Since #3448 this is obsolete: `getAppAuth()` reads the environment variables directly, before consulting the block.
- The documented pattern also fails `terraform validate` / `tofu validate` in any environment where the variables are not set (CI linting, local pre-commit hooks), because the `app_auth` arguments are schema-required with `EnvDefaultFunc`. Details and reproduction in #3550.

### After the change?

- The note states that no `app_auth` block is needed for environment-variable auth, and that the previously documented empty block should be removed because it fails `validate` without the variables set.
- The `main.tf` example no longer shows an empty `app_auth {}` block.
- No per-version guidance included: registry docs are versioned, users on older provider versions see the matching older docs.

### Pull request checklist

- [ ] Yes
- [x] No

----

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -67,6 +67,10 @@ To authenticate using a GitHub App installation, ensure that arguments in the `a
 
 Some API operations may not be available when using a GitHub App installation configuration. For more information, refer to the list of [supported endpoints](https://docs.github.com/en/rest/overview/endpoints-available-for-github-apps).
 
+~> The `app_auth` block is optional: the provider reads the `GITHUB_APP_XXX` environment variables directly. When you deliberately want GitHub App auth, keeping the block is recommended: its required arguments make `terraform validate` fail fast if a value is missing, and it lets you set non-secret values such as the App `id` directly. Drive auth from the environment without a block when the same configuration must also validate without the variables set, such as CI linting. The three patterns below cover each case.
+
+#### Block with values
+
 ```terraform
 provider "github" {
   owner = var.github_organization
@@ -78,9 +82,7 @@ provider "github" {
 }
 ```
 
-~> No `app_auth` block is needed when authenticating through the `GITHUB_APP_XXX` environment variables; the provider reads them directly. The previously documented empty `app_auth {}` block is no longer required and should be removed: its arguments are schema-required, so it fails `terraform validate` in environments where the variables are not set.
-
-#### .env
+#### Environment variables only
 
 ```shell
 export GITHUB_APP_ID="12332432" # Required: The GitHub App ID for authentication
@@ -88,12 +90,24 @@ export GITHUB_APP_INSTALLATION_ID="12435523" # Required: The GitHub App Installa
 export GITHUB_APP_PEM_FILE="..." # Required: Contents of the PEM file for the GitHub App, not the path to the PEM file
 ```
 
-#### main.tf
-
 ```terraform
 provider "github" {
   owner = var.github_organization
   # Credentials come from the `GITHUB_APP_XXX` environment variables.
+}
+```
+
+#### Non-secret values in configuration, secret from the environment
+
+```terraform
+provider "github" {
+  owner = var.github_organization
+  app_auth {
+    id              = "123456"   # the App ID and installation ID are not secret,
+    installation_id = "78901234" # so they can be set directly in configuration
+    # pem_file is omitted; it falls back to the GITHUB_APP_PEM_FILE variable,
+    # keeping the secret out of configuration
+  }
 }
 ```
 
--- a/docs/index.md
+++ b/docs/index.md
@@ -78,7 +78,7 @@ provider "github" {
 }
 ```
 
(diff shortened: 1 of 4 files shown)


deiga (changes requested):
Please note that docs files are generated. You need to update the template files as well

deiga (dismissed):

stevehipwell (changes requested):
I'm not sure I fully agree with these changes. The docs do need to be updated to show that all of the app auth config can be driven from environment variables and the `app_auth` block is not required. However setting the `app_auth` block if you explicitly want to use app auth is the recommended pattern as it errors if the env variables haven't been set, as is using the block for static configuration (e.g. app ID).

SebastianBienert:
Thanks @deiga, updated. Pushed the same change to the sources: `templates/index.md.tmpl` for the note, and `examples/provider/app_auth_env/main.tf` for the example, which the template inlines via `{{ tffile }}`. Ran `go generate ./...` locally to confirm `docs/index.md` regenerates identically, so generation stays idempotent.

SebastianBienert:
@stevehipwell reworked to address this. The note now states the `app_auth` block is optional (env vars are read directly) and recommends keeping the block when you want GitHub App auth deliberately, since its required arguments make `terraform validate` fail fast on a missing value. Three examples now cover: block with values, environment-variables-only (no block), and non-secret IDs in config with the PEM kept in the environment.

stevehipwell (approved):
Thanks for the quick response @SebastianBienert.

LGTM

deiga (approved):

stevehipwell:
@SebastianBienert could you please rebase this PR?

SebastianBienert:
@stevehipwell done

PR #3553, opened 2026-07-20
fix: remove vulnerability alerts from state for archived/deleted repos

### LINKED ISSUE #3369: [DOCS] Document how to handle Terraform config of vulnerability alerts when repo is archived
### Expected Behavior

1.  Vulnerability alerts are automatically disabled
2. Repository is archived

### Actual Behavior

```
│ Error: repository $REDATCTED is archived, please remove the resource from your configuration
│
│   with module.github_repository_vulnerability_alerts.vulnerability_alerts[0],
│   on redacted.tf line 40, in resource "github_repository_vulnerability_alerts" "vulnerability_alerts":
│   40: resource "github_repository_vulnerability_alerts" "vulnerability_alerts" {
│
```

### Terraform Version

Terraform v1.14.2
on darwin_arm64
+ provider registry.terraform.io/buildkite/buildkite v1.32.0
+ provider registry.terraform.io/hashicorp/aws v6.42.0
+ provider registry.terraform.io/hashicorp/github v6.11.1
+ provider registry.terraform.io/hashicorp/google v7.21.0
+ provider registry.terraform.io/integrations/github v6.12.0
+ provider registry.terraform.io/jianyuan/sentry v0.14.13
+ provider registry.terraform.io/newrelic/newrelic v3.85.0

### Affected Resource(s)

- github_repository
- github_repository_vulnerability_alerts

```hcl
resource "github_repository" "project" {
  name        = var.project_name
  description = var.project_description
  visibility  = var.public_repo ? "public" : "private"
  archived    = var.archived
}

resource "github_repository_vulnerability_alerts" "vulnerability_alerts" {
  repository = github_repository.project.name
  enabled    = true
}
```

### Steps to Reproduce

1. Set `archived = true` on the `github_repository` resource
2. Run `terraform apply`

### Workarounds

I tried a couple workarounds. This one didn't work:

```terraform
resource "github_repository_vulnerability_alerts" "vulnerability_alerts" {
  repository = github_repository.project.name
  enabled    = var.archived ? false : true
  depends_on = [github_repository.project]
}
```

But this did:

```terraform
resource "github_repository_vulnerability_alerts" "vulnerability_alerts" {
  repository = github_repository.project.name
  enabled    = true
  count      = var.archived ? 0 : 1
  depends_on = [github_repository.project]
}
```

Other resources, such as `github_branch_protection`, handled the archival of the repository just fine.

### Debug Output

```shell

```

### Panic Output

```shell

```

### Code of Conduct

DrFaust92:
Resolves #3369

----

### Before the change?

`resourceGithubRepositoryVulnerabilityAlertsRead` returned a hard error when the underlying repository was archived or no longer existed:

```
Error: repository <owner>/<name> is archived, please remove the resource from your configuration
```

Because this surfaces during **refresh**, every subsequent `plan`/`apply` fails until the resource is manually removed from both config *and* state — you can't even `terraform state rm` cleanly while the read errors. Archiving a repo auto-disables its vulnerability alerts on GitHub's side, so the resource is genuinely unmanageable at that point.

The existing 404 branch also set `d.SetId("")` and *then* returned `diag.Errorf(...)`; since the SDK discards state changes when Read returns an error, that removal never actually took effect.

### After the change?

Read now follows the provider's documented convention (`.github/instructions/schema-and-state.instructions.md`: *"In Read, a 404 from GitHub means 'remove from state' — call `d.SetId("")` and return nil"*):

- Alerts 404 (resource gone) → `d.SetId("")` + `return nil` (removed the stray `diag.Errorf`).
- Repository no longer exists → `d.SetId("")` + `return nil`.
- Repository archived → `d.SetId("")` + `return nil`, with a `tflog.Warn`.

Also guards against a nil `*http.Response` before dereferencing `StatusCode`.

Net effect: Terraform drops the resource from state gracefully. If the config still declares it on an archived repo, the follow-up plan is non-empty (recreate), and `Create` still returns the pre-existing `cannot enable vulnerability alerts on archived repository` error — so users get a clear signal to remove the block, instead of a wedged refresh.

### Pull request checklist

- [ ] Schema migrations have been created if needed
- [x] Tests for the changes have been added (reworked `errors_when_reading_archived_repository` → `removes_from_state_when_reading_archived_repository`, asserting graceful removal + non-empty plan instead of an error)
- [x] Docs have been reviewed and added / updated if needed (added a "Notes" section on archived-repo behavior; regenerated docs)

Read no longer returns an error for archived/deleted repositories; instead the resource is removed from state. This is a behavior change but not a schema breaking change.

- [ ] Yes
- [x] No

### DIFF
--- a/github/resource_github_repository_vulnerability_alerts.go
+++ b/github/resource_github_repository_vulnerability_alerts.go
@@ -97,20 +97,25 @@ func resourceGithubRepositoryVulnerabilityAlertsRead(ctx context.Context, d *sch
 	repoName := d.Get("repository").(string)
 	vulnerabilityAlertsEnabled, resp, err := client.Repositories.GetVulnerabilityAlerts(ctx, owner, repoName)
 	if err != nil {
-		if resp.StatusCode == http.StatusNotFound {
+		if resp != nil && resp.StatusCode == http.StatusNotFound {
+			tflog.Warn(ctx, "Vulnerability alerts don't exist, removing resource from state", map[string]any{"owner": owner, "repo_name": repoName})
 			d.SetId("")
-			return diag.Errorf("vulnerability alerts don't exist for repository %s/%s, removing resource from state", owner, repoName)
+			return nil
 		}
 		return diag.Errorf("error reading repository vulnerability alerts: %s", err.Error())
 	}
 	// If no error, but the response status code is 404, we need to check if the repository is accessible.
-	if resp.StatusCode == http.StatusNotFound {
+	if resp != nil && resp.StatusCode == http.StatusNotFound {
 		repo, _, err := client.Repositories.Get(ctx, owner, repoName)
 		if err != nil {
-			return diag.Errorf("repository doesn't exist anymore, please remove the resource from your configuration: %s", err.Error())
+			tflog.Warn(ctx, "Repository doesn't exist anymore, removing resource from state", map[string]any{"owner": owner, "repo_name": repoName})
+			d.SetId("")
+			return nil
 		}
 		if repo.GetArchived() {
-			return diag.Errorf("repository %s/%s is archived, please remove the resource from your configuration", owner, repoName)
+			tflog.Warn(ctx, "Repository is archived, removing vulnerability alerts resource from state", map[string]any{"owner": owner, "repo_name": repoName})
+			d.SetId("")
+			return nil
 		}
 	}
 	tflog.Debug(ctx, "Setting vulnerability alerts enabled state", map[string]any{"owner": owner, "repo_name": repoName, "vulnerability_alerts_enabled": vulnerabilityAlertsEnabled, "response_status": resp.StatusCode})
--- a/github/resource_github_repository_vulnerability_alerts.go
+++ b/github/resource_github_repository_vulnerability_alerts.go
@@ -98,7 +98,7 @@ func resourceGithubRepositoryVulnerabilityAlertsRead(ctx context.Context, d *sch
 	vulnerabilityAlertsEnabled, resp, err := client.Repositories.GetVulnerabilityAlerts(ctx, owner, repoName)
 	if err != nil {
 		if resp != nil && resp.StatusCode == http.StatusNotFound {
-			tflog.Warn(ctx, "Vulnerability alerts don't exist, removing resource from state", map[string]any{"owner": owner, "repo_name": repoName})
+			tflog.Warn(ctx, "vulnerability alerts don't exist, removing resource from state", map[string]any{"owner": owner, "repo_name": repoName})
 			d.SetId("")
 			return nil
 		}
--- a/github/resource_github_repository_vulnerability_alerts.go
+++ b/github/resource_github_repository_vulnerability_alerts.go
@@ -2,9 +2,11 @@ package github
 
 import (
 	"context"
+	"errors"
 	"net/http"
 	"strconv"
 
+	"github.com/google/go-github/v89/github"
 	"github.com/hashicorp/terraform-plugin-log/tflog"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
@@ -97,35 +99,29 @@ func resourceGithubRepositoryVulnerabilityAlertsRead(ctx context.Context, d *sch
 	repoName := d.Get("repository").(string)
 	vulnerabilityAlertsEnabled, resp, err := client.Repositories.GetVulnerabilityAlerts(ctx, owner, repoName)
 	if err != nil {
-		if resp != nil && resp.StatusCode == http.StatusNotFound {
-			tflog.Warn(ctx, "vulnerability alerts don't exist, removing resource from state", map[string]any{"owner": owner, "repo_name": repoName})
-			d.SetId("")
-			return nil
-		}
-		return diag.Errorf("error reading repository vulnerability alerts: %s", err.Error())
+		return diag.FromErr(err)
 	}
-	// If no error, but the response status code is 404, we need to check if the repository is accessible.
+	// GetVulnerabilityAlerts maps a 404 to enabled=false with no error, so a
+	// 404 here is ambiguous: alerts may simply be disabled, or the repository
+	// may be gone or archived (archiving disables alerts). Look the repository
+	// up to decide whether to drop the resource from state.
 	if resp != nil && resp.StatusCode == http.StatusNotFound {
-		repo, repoResp, err := client.Repositories.Get(ctx, owner, repoName)
+		repo, _, err := client.Repositories.Get(ctx, owner, repoName)
 		if err != nil {
-			if repoResp != nil && repoResp.StatusCode == http.StatusNotFound {
-				tflog.Warn(ctx, "Repository doesn't exist anymore, removing resource from state", map[string]any{"owner": owner, "repo_name": repoName})
+			if ghErr, ok := errors.AsType[*github.ErrorResponse](err); ok && ghErr.Response.StatusCode == http.StatusNotFound {
+				tflog.Warn(ctx, "repository no longer exists, removing vulnerability alerts from state", map[string]any{"owner": owner, "repo_name": repoName})
 				d.SetId("")
 				return nil
 			}
-			return diag.Errorf("error reading repository %s/%s: %s", owner, repoName, err.Error())
+			return diag.FromErr(err)
 		}
 		if repo.GetArchived() {
-			tflog.Warn(ctx, "Repository is archived, removing vulnerability alerts resource from state", map[string]any{"owner": owner, "repo_name": repoName})
+			tflog.Warn(ctx, "repository is archived, removing vulnerability alerts from state", map[string]any{"owner": owner, "repo_name": repoName})
 			d.SetId("")
 			return nil
 		}
 	}
-	responseStatus := 0
-	if resp != nil {
-		responseStatus = resp.StatusCode
-	}
-	tflog.Debug(ctx, "Setting vulnerability alerts enabled state", map[string]any{"owner": owner, "repo_name": repoName, "vulnerability_alerts_enabled": vulnerabilityAlertsEnabled, "response_status": responseStatus})
+	tflog.Debug(ctx, "Setting vulnerability alerts enabled state", map[string]any{"owner": owner, "repo_name": repoName, "vulnerability_alerts_enabled": vulnerabilityAlertsEnabled})
 	if err = d.Set("enabled", vulnerabilityAlertsEnabled); err != nil {
 		return diag.FromErr(err)
 	}
--- a/github/resource_github_repository_vulnerability_alerts.go
+++ b/github/resource_github_repository_vulnerability_alerts.go
@@ -106,11 +106,14 @@ func resourceGithubRepositoryVulnerabilityAlertsRead(ctx context.Context, d *sch
 	}
 	// If no error, but the response status code is 404, we need to check if the repository is accessible.
 	if resp != nil && resp.StatusCode == http.StatusNotFound {
-		repo, _, err := client.Repositories.Get(ctx, owner, repoName)
+		repo, repoResp, err := client.Repositories.Get(ctx, owner, repoName)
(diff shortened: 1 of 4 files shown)


github/resource_github_repository_vulnerability_alerts.go
```diff
@@ -97,20 +97,25 @@ func resourceGithubRepositoryVulnerabilityAlertsRead(ctx context.Context, d *sch
...
 	}
 	// If no error, but the response status code is 404, we need to check if the repository is accessible.
-	if resp.StatusCode == http.StatusNotFound {
+	if resp != nil && resp.StatusCode == http.StatusNotFound {
 		repo, _, err := client.Repositories.Get(ctx, owner, repoName)
 		if err != nil {
-			return diag.Errorf("repository doesn't exist anymore, please remove the resource from your configuration: %s", err.Error())
+			tflog.Warn(ctx, "Repository doesn't exist anymore, removing resource from state", map[string]any{"owner": owner, "repo_name": repoName})
+			d.SetId("")
+			return nil
```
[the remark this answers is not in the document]
DrFaust92: Good catch — fixed in 51c87ad7. The out-of-band `Repositories.Get` now captures its response and only clears state when it returns 404; permission, rate-limit, network, and 5xx errors are propagated via `diag.Errorf` instead of silently dropping valid state and forcing a recreate.

github/resource_github_repository_vulnerability_alerts_test.go
```diff
@@ -233,9 +233,13 @@ func TestAccGithubRepositoryVulnerabilityAlerts(t *testing.T) {
 					},
 				},
 				{
+					// Archiving the repository disables vulnerability alerts on
+					// GitHub's side. Read must remove the resource from state
+					// instead of erroring. Because the config still declares the
+					// resource, the follow-up plan is non-empty (Terraform wants
+					// to recreate it) rather than failing the whole run.
 					Config:             fmt.Sprintf(withAlertsConfig, repoName, true),
 					ExpectNonEmptyPlan: true,
```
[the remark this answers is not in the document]
DrFaust92: Added in 51c87ad7 — new `removes_from_state_when_repository_deleted` acceptance step deletes the repository out of band, then `RefreshState: true` + `ExpectNonEmptyPlan: true` asserts the deleted-repo Read path removes the resource from state. Used `RefreshState` (matching `resource_github_branch_default_test.go`) rather than a re-apply, since re-applying would recreate the repo and yield an empty plan.

github/resource_github_repository_vulnerability_alerts.go
```diff
@@ -97,23 +97,35 @@ func resourceGithubRepositoryVulnerabilityAlertsRead(ctx context.Context, d *sch
 	repoName := d.Get("repository").(string)
 	vulnerabilityAlertsEnabled, resp, err := client.Repositories.GetVulnerabilityAlerts(ctx, owner, repoName)
 	if err != nil {
-		if resp.StatusCode == http.StatusNotFound {
+		if resp != nil && resp.StatusCode == http.StatusNotFound {
```
deiga: Please use the established pattern for error handling: https://github.com/integrations/terraform-provider-github/blob/main/ARCHITECTURE.md#error-handling

github/resource_github_repository_vulnerability_alerts.go
```diff
@@ -97,23 +97,35 @@ func resourceGithubRepositoryVulnerabilityAlertsRead(ctx context.Context, d *sch
...
+			tflog.Warn(ctx, "Vulnerability alerts don't exist, removing resource from state", map[string]any{"owner": owner, "repo_name": repoName})
```
deiga: Please lowercase log messages
```suggestion
			tflog.Warn(ctx, "vulnerability alerts don't exist, removing resource from state", map[string]any{"owner": owner, "repo_name": repoName})
```

github/resource_github_repository_vulnerability_alerts.go
```diff
@@ -97,23 +97,35 @@ func resourceGithubRepositoryVulnerabilityAlertsRead(ctx context.Context, d *sch
...
+			tflog.Warn(ctx, "Repository is archived, removing vulnerability alerts resource from state", map[string]any{"owner": owner, "repo_name": repoName})
+			d.SetId("")
+			return nil
 		}
 	}
-	tflog.Debug(ctx, "Setting vulnerability alerts enabled state", map[string]any{"owner": owner, "repo_name": repoName, "vulnerability_alerts_enabled": vulnerabilityAlertsEnabled, "response_status": resp.StatusCode})
+	responseStatus := 0
+	if resp != nil {
+		responseStatus = resp.StatusCode
+	}
```
deiga: question: What's the purpose of this?

github/resource_github_repository_vulnerability_alerts.go
```diff
@@ -97,23 +97,35 @@ func resourceGithubRepositoryVulnerabilityAlertsRead(ctx context.Context, d *sch
...
 			d.SetId("")
-			return diag.Errorf("vulnerability alerts don't exist for repository %s/%s, removing resource from state", owner, repoName)
+			return nil
 		}
 		return diag.Errorf("error reading repository vulnerability alerts: %s", err.Error())
 	}
 	// If no error, but the response status code is 404, we need to check if the repository is accessible.
-	if resp.StatusCode == http.StatusNotFound {
-		repo, _, err := client.Repositories.Get(ctx, owner, repoName)
+	if resp != nil && resp.StatusCode == http.StatusNotFound {
```
deiga, line 108: question: Is there a case when `resp` could be nil here?

github/resource_github_repository_vulnerability_alerts_test.go
```diff
@@ -233,9 +233,59 @@ func TestAccGithubRepositoryVulnerabilityAlerts(t *testing.T) {
 					},
 				},
 				{
+					// Archiving the repository disables vulnerability alerts on
+					// GitHub's side. Read must remove the resource from state
+					// instead of erroring. Because the config still declares the
+					// resource, the follow-up plan is non-empty (Terraform wants
+					// to recreate it) rather than failing the whole run.
```
deiga, line 241: suggestion: Let's add a ConfigPlanCheck to verify that it tries to create a new resource

github/resource_github_repository_vulnerability_alerts_test.go
```diff
@@ -233,9 +233,59 @@ func TestAccGithubRepositoryVulnerabilityAlerts(t *testing.T) {
...
+
+		randomID := acctest.RandString(5)
+		repoName := fmt.Sprintf("%svuln-alerts-%s", testResourcePrefix, randomID)
+
+		withAlertsConfig := `
+			resource "github_repository" "test" {
+				name       = "%s"
+				visibility = "private"
+				auto_init  = true
+			}
```
deiga: Please use the `mustCreateTestRepository` function instead of declaring the repo in Terraform code

github/resource_github_repository_vulnerability_alerts_test.go
```diff
@@ -233,9 +233,59 @@ func TestAccGithubRepositoryVulnerabilityAlerts(t *testing.T) {
...
+		`
+
+		resource.Test(t, resource.TestCase{
+			PreCheck:          func() { skipUnauthenticated(t) },
+			ProviderFactories: providerFactories,
+			Steps: []resource.TestStep{
+				{
+					Config: fmt.Sprintf(withAlertsConfig, repoName),
+					ConfigStateChecks: []statecheck.StateCheck{
+						statecheck.ExpectKnownValue("github_repository_vulnerability_alerts.test", tfjsonpath.New("enabled"), knownvalue.Bool(true)),
```
deiga: No need to check for same value as config
```suggestion
```

github/resource_github_repository_vulnerability_alerts_test.go
```diff
@@ -233,9 +233,59 @@ func TestAccGithubRepositoryVulnerabilityAlerts(t *testing.T) {
...
+					},
+				},
+				{
+					// Delete the repository out of band. On refresh, the
+					// vulnerability alerts Read hits a deleted repository and must
+					// remove the resource from state instead of erroring, leaving a
+					// non-empty plan (Terraform wants to recreate it).
```
deiga, line 278: Please add a plan check here as well for the Create Action

deiga (changes requested):

github/resource_github_repository_vulnerability_alerts.go
DrFaust92: Done in d882da7 — the deleted-repo case now uses `errors.AsType[*github.ErrorResponse](err)` + `diag.FromErr`, matching ARCHITECTURE.md#error-handling. I also dropped the 404 branch on `GetVulnerabilityAlerts`: go-github's `parseBoolResponse` maps a 404 to `(false, nil)`, so that call never returns a 404 *error* — the real disambiguation (disabled vs. archived/deleted) is on the follow-up `Repositories.Get`, where the AsType pattern now applies.

github/resource_github_repository_vulnerability_alerts.go
DrFaust92: When `err == nil`, `GetVulnerabilityAlerts` always returns a non-nil `resp`, so the `resp != nil` guard is defensive only — kept per the schema-and-state rule "Nil-check API response fields before dereferencing" rather than dereferencing `resp.StatusCode` bare. d882da7.

github/resource_github_repository_vulnerability_alerts.go
DrFaust92: It only existed to log the status code in the debug line. Removed the field and the nil-guard scaffolding with it in d882da7.

github/resource_github_repository_vulnerability_alerts_test.go
DrFaust92: Added in d882da7 via `ConfigPlanChecks.PostApplyPostRefresh` asserting `ResourceActionCreate` — the archive disables alerts server-side during apply, so the removal→recreate only surfaces on the post-apply refresh, not PreApply.

github/resource_github_repository_vulnerability_alerts_test.go
DrFaust92: Done in d882da7 — the repo is now provisioned out of band with `mustCreateTestRepository(t)` and referenced by name; the Terraform config only declares the alerts resource.

github/resource_github_repository_vulnerability_alerts_test.go
DrFaust92: Added in d882da7. Since this is a `RefreshState` step, the framework honors `RefreshPlanChecks.PostRefresh` (not `ConfigPlanChecks`) — used that with `ExpectResourceAction(..., ResourceActionCreate)`.

github/resource_github_repository_vulnerability_alerts.go
```diff
@@ -97,23 +99,29 @@ func resourceGithubRepositoryVulnerabilityAlertsRead(ctx context.Context, d *sch
 	repoName := d.Get("repository").(string)
 	vulnerabilityAlertsEnabled, resp, err := client.Repositories.GetVulnerabilityAlerts(ctx, owner, repoName)
 	if err != nil {
-		if resp.StatusCode == http.StatusNotFound {
-			d.SetId("")
-			return diag.Errorf("vulnerability alerts don't exist for repository %s/%s, removing resource from state", owner, repoName)
-		}
-		return diag.Errorf("error reading repository vulnerability alerts: %s", err.Error())
+		return diag.FromErr(err)
```
deiga: No need to worsen the error message here
```suggestion
		return diag.Errorf("error reading repository vulnerability alerts: %s", err.Error())
```

deiga (changes requested):
Minor change request, otherwise this is looking quite good!

github/resource_github_repository_vulnerability_alerts.go
DrFaust92: Done — restored the descriptive error in 9dfe90a. Also rebased on main.

deiga (approved):
LGTM

robert-crandall (approved):
Thanks for addressing the archived and deleted repository cases so thoughtfully! The state-removal behavior and added coverage look good to me.

PR #3555, opened 2026-07-20
chore: Refactor max per page to owner struct

stevehipwell:
Resolves #ISSUE_NUMBER

----

### Before the change?

- The provider scoped input for `max_per_page` is set to a global package variable

### After the change?

- The provider metadata holds `maxPerPage` variable

### Pull request checklist

- [x] Schema migrations have been created if needed ([example](https://github.com/F-Secure-web/terraform-provider-github/blob/main/github/migrate_github_actions_organization_secret.go))
- [x] Tests for the changes have been added (for bug fixes / features)
- [x] Docs have been reviewed and added / updated if needed (for bug fixes / features)

### Does this introduce a breaking change?

Please see our docs on [breaking changes](https://github.com/octokit/.github/blob/master/community/breaking_changes.md) to help!

- [ ] Yes
- [x] No

----

PR #3557, opened 2026-07-20
feat: Add data sources to replace summary deprecations

stevehipwell:
Resolves #ISSUE_NUMBER

----

### Before the change?

- Data source fields set when `summary_only = false` (the default) not deprecated

### After the change?

- Deprecated fields set when `summary_only = false` ready for removal in next major version
- Added new data sources to replace the deprecated fields
  - `github_organization_members`
  - `github_organization_repositories`
  - `github_team_members`
  - `github_team_repositories`

### Pull request checklist

- [ ] Yes
- [x] No

----


rilical (changes requested):
mistake.

rilical:
Could we make the replacement data sources preserve the fields and behavior they are deprecating? `github_organization_members` drops `role` and `email`, `github_team_members` cannot distinguish direct from inherited members, and `github_team_repositories` drops `role_name`.

Once #3555 lands, this will also need a rebase to use `meta.maxPerPage` in the four new data sources.

stevehipwell:
> Could we make the replacement data sources preserve the fields and behavior they are deprecating?

Not with the current REST API surface area, but we're not trying to directly replicate the existing patterns some of which were fundamentally flawed in the first place.

> github_organization_members drops role and email

Email and role are missing from the response (@robert-crandall could we get role added given that it's in the filter?).

> github_team_members cannot distinguish direct from inherited members,

I'll fix this, it's not in the example so I missed it. I'll also add emil and role as these are available.

> github_team_repositories drops role_name.

Role name is not returned from the list endpoint, this would need fixing to make this possible.

github/data_source_github_organization_members.go
```diff
@@ -0,0 +1,74 @@
...
+}
+
+func dataSourceGithubOrganizationMembersRead(ctx context.Context, d *schema.ResourceData, m any) diag.Diagnostics {
+	meta, _ := m.(*Owner)
+
+	if ok, diags := checkOrganizationOK(meta); !ok {
+		return diags
+	}
+
+	users := make([]map[string]any, 0)
```
deiga: Nit: could we call this members?

deiga (dismissed):
LGTM

rilical:
The team-member update covers that part, and I’m fine dropping the organization email/role points. I checked `role_name`: the list-team-repositories response does include it—the REST schema lists it, and the existing `queries_team_details` acceptance test gets `"read"` from the same `ListTeamReposBySlugIter` call. Could we add `role_name` to `github_team_repositories`?

stevehipwell:
> The team-member update covers that part, and I’m fine dropping the organization email/role points. I checked role_name: the list-team-repositories response does include it—the REST schema lists it, and the existing queries_team_details acceptance test gets "read" from the same ListTeamReposBySlugIter call. Could we add role_name to github_team_repositories?

Done. This is an area where the example should show the `role_name` and I think that the schema should have it as required?

rilical (approved):

deiga (approved):

PR #3563, opened 2026-07-20
Add Team Review Router label-routing workflow

robert-crandall:
## What this does

Adds a single workflow that requests the GitHub-employee review **team** as a reviewer when a maintainer applies the `needs-github-review` label to a PR. This unblocks contributions that would otherwise stall waiting on one designated maintainer.

This workflow authenticates as a **GitHub App** (which does have org read), mints a short-lived installation token, and makes the request.

## How it's wired

Reads from the `team-review-router` environment on this repo:

- Secrets: `TF_TEAM_REVIEW_APP_ID`, `TF_TEAM_REVIEW_APP_PRIVATE_KEY`
- Variables: `TF_TEAM_REVIEW_TEAM_SLUG`, `TF_TEAM_REVIEW_DRY_RUN`

Flow on each `labeled` event:
1. Ignore bot actors.
2. Re-verify the labeler has `write`+ on the repo (run-time check, not just trusting who could label).
3. Skip if the team is already a requested reviewer (idempotency).
4. Request the team - unless `TF_TEAM_REVIEW_DRY_RUN` is `true`, in which case it only logs the intended action.

## Safety

- `pull_request_target` runs with secrets in scope, so this workflow **does not check out or execute any PR code** - it only calls the GitHub API. No untrusted-code path.
- `GITHUB_TOKEN` is down-scoped to `contents: read`; every write goes through the App token.
- The `team-review-router` environment should be restricted to the default branch, with no required reviewers and a zero wait timer.

## Before this is merged / goes live

- [x] GitHub App installed on this repo with: Pull requests (write), Administration (read), Members (read), Metadata (read).
- [x] `needs-github-review` label exists; review team has read/triage+ access.
- [x] Keep `TF_TEAM_REVIEW_DRY_RUN=true` and confirm a labeled PR logs `[dry-run] would request team ...`.
- [x] Pin `actions/create-github-app-token@v2` to a full commit SHA.
- [ ] Flip `TF_TEAM_REVIEW_DRY_RUN=false` only after the dry-run check passes.

## Verification status

YAML and the embedded shell script were syntax-checked locally. **Not yet run against a live PR** - that needs the App installed and a dry-run, which is why this is a draft.

### DIFF
--- a/.github/workflows/team-review-router.yml
+++ b/.github/workflows/team-review-router.yml
@@ -35,10 +35,9 @@ jobs:
     # timer (either would pause this "immediate" path).
     environment: team-review-router
     steps:
-      # Pin this to a full commit SHA before production use (repo convention).
       - name: Mint GitHub App token
         id: app-token
-        uses: actions/create-github-app-token@v2
+        uses: actions/create-github-app-token@fee1f7d63c2ff003460e3d139729b119787bc349 # v2.2.2
         with:
           app-id: ${{ secrets.TF_TEAM_REVIEW_APP_ID }}
           private-key: ${{ secrets.TF_TEAM_REVIEW_APP_PRIVATE_KEY }}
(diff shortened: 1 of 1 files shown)


stevehipwell (approved):
LGTM

deiga (approved):
Nice!

Issue #2305, opened 2024-07-02
[BUG]: github_organization_settings resource fails with HTTP 422


giuliov-unum:
### Expected Behavior

We want to use the GitHub Terraform provider to manage our organizations but the `github_organization_settings` resource invariably fails with HTTP 422.

Tried many different combinations with no luck:
- an existing Org manually created referred via data
- an Org created via `github_enterprise_organization` in the same module
- as the single resource in the module and import

The PAT belongs to an Enterprise and Org owner and all permissions selected.
This is an EMU Enterprise account.

### Actual Behavior

Here is the message

[ERROR] provider.terraform-provider-github_v6.2.2.exe: Response contains error diagnostic: @caller=github.com/hashicorp/terraform-plugin-go@v0.23.0/tfprotov5/internal/diag/diagnostics.go:58 @module=sdk.proto diagnostic_severity=ERROR diagnostic_summary="PATCH https://api.github.com/orgs/OMISSIS: 422 []" tf_proto_version=5.6 tf_provider_addr=provider tf_req_id=491343e0-ef75-6d3c-e12a-cfba58c7c4ca tf_rpc=ApplyResourceChange diagnostic_detail="" tf_resource_type=github_organization_settings timestamp="2024-07-01T15:13:06.315+0100"

### Terraform Version

Terraform version: 1.8.5 on Win64 using integrations/github 6.2.2 

### Affected Resource(s)

github_organization_settings

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

giuliov-unum:
I tried the same code on a personal account and it works fine, so it seems related to EMU.

stuartpurgavie:
Had this issue as well when attempting to provide a parameter to an organization when that parameter was being enforced by the parent Enterprise. Removing the offending parameter resolved the issue.

The API response was much more enlightening:
```json
{
  "message": "Validation Failed",
  "errors": "Forking policy has been set by enterprise administrators.",
  "documentation_url": "https://docs.github.com/rest/orgs/orgs#update-an-organization",
  "status": "422"
}
```

It would be great if the Terraform Provider passed that error message through!

giuliov-unum:
> Had this issue as well when attempting to provide a parameter to an organization when that parameter was being enforced by the parent Enterprise. Removing the offending parameter resolved the issue.

That's my problem: how do you find that? The payload is not in the provider log, not even at the most detailed level.
[...]

Issue #3429, opened 2026-05-14
[BUG]: github_organization_settings silently drifts repo-creation booleans due to partial PATCH × deprecated legacy field


naokit-dev:
### Expected Behavior

When `terraform apply` updates any of the four repo-creation boolean fields on `github_organization_settings`:

- `members_can_create_repositories`
- `members_can_create_public_repositories`
- `members_can_create_private_repositories`
- `members_can_create_internal_repositories`

…the resulting server-side state should match the Terraform config exactly, regardless of which subset of those four actually changed in the diff.

### Actual Behavior

Provider silently leaves the org in a state that contradicts the Terraform config, with no warning at apply time. The next `terraform plan` may report no drift because state is refreshed from the (now-wrong) server values.

Real-world example (Enterprise Cloud org, 2026-05-14):

- Config: `create=true, public=false, private=true, internal=false`
- State before apply: `create=false, private=false` (drift from UI), `public=false, internal=false` (matched)
- `terraform apply` sent a PATCH containing only the two `HasChange` fields (`members_can_create_repositories=true`, `members_can_create_private_repositories=true`) per [resource_github_organization_settings.go L183](https://github.com/integrations/terraform-provider-github/blob/v6.11.1/github/resource_github_organization_settings.go#L183)
- Server response: `create=true, public=true, private=true, internal=true`, `members_allowed_repository_creation_type=all`

The GitHub Org API, when receiving a PATCH that sets `members_can_create_repositories=true` without also specifying the granular booleans, resets the deprecated `members_allowed_repository_creation_type` to `all`, which then silently overrides the unspecified granular booleans server-side. We had to manually reconcile via `gh api -X PATCH` after every apply that touches these fields.

### Terraform Version

```
Terraform v1.14.8
+ provider registry.terraform.io/integrations/github v6.11.1
```

### Affected Resource(s)

- `github_organization_settings`

### Steps to Reproduce

1. Real org with Enterprise plan and config:

```hcl
resource "github_organization_settings" "this" {
  billing_email                            = "x@example.com"
  members_can_create_repositories          = true
  members_can_create_public_repositories   = false
  members_can_create_private_repositories  = true
  members_can_create_internal_repositories = false
}
```

2. Have server state diverge such that *some but not all* of the four booleans match config (easiest: toggle "members can create repositories" off in the UI, leaving the others at their existing values).
3. `terraform apply`.
[...]

Issue #2689, opened 2025-06-22
[BUG]: `github_organization_setting` fails with 422 when a fork policy is defined on enterprise level


MariusStorhaug:
The reconfiguration of a organization setting resource fails with the following changes when we have a policy defining a forking policy on enterprise level:

Policy:

![Image](https://github.com/user-attachments/assets/2dcbe241-2b21-4c38-9523-05d410ef7563)

Deployment output:
```plaintext
╷
│ Error: PATCH https://mydomain.ghe.com/api/v3/orgs/dnb-acme: 422  []
│ 
│   with module.organization-settings.github_organization_settings.this,
│   on .terraform/modules/organization-settings/main.settings.tf line 1, in resource "github_organization_settings" "this":
│    1: resource "github_organization_settings" "this" {
│ 
╵
```

Upon disabling the forking setting on the organization, the deployment succeeds:

![Image](https://github.com/user-attachments/assets/d7c89103-4435-454e-bb55-80dbf5afa74f)

Logs from deployment are clean and green.

After first deployment i set the fork settings on organization same as the enterprise policy and rerun tf apply and no issues arrise on the second run.

brgsstm:
any update on this one?

deiga:
@brgsstm @MariusStorhaug If you are encountering this with the latest provider version, could you please share a DEBUG level output of `terraform apply` and a sample Terraform config to reproduce this?

deiga:
There also doesn't seem to be a REST endpoint which controls that field 🙃

brgsstm:
redacted debug log attached (plan + apply).

providers.tf

```hcp
terraform {
  required_providers {
    github = {
      source  = "integrations/github"
      version = "~> 6.11.1"
    }
  }
}
```

main.tf

```hcp
resource "github_organization_settings" "xxx" {
  billing_email                            = "xxx@xxx.com"
  members_can_fork_private_repositories    = false
}
```

Same enterprise setting as @MariusStorhaug 

[terraform_debug.redacted.log](https://github.com/user-attachments/files/26267395/terraform_debug.redacted.log)

deiga:
Fudge. We don't have modernized logging for `github_organization_setting`

Could you manually run [this](https://docs.github.com/en/rest/orgs/orgs?apiVersion=2026-03-10#get-an-organization) GET against the GH API and share the redacted output?

And I'm beginning to realize the situation now. 
[...]

PR #2807, opened 2025-10-23
Fix HTTP 422 error in github_organization_settings resource

netflash:
# Fix HTTP 422 error in github_organization_settings resource

Fixes #2305

## Problem Description

The `github_organization_settings` resource was failing with HTTP 422 errors, particularly when used with Enterprise Managed User (EMU) organizations. Users reported that the resource "invariably fails with HTTP 422" regardless of configuration attempts.

### Root Cause Analysis

The issue was caused by three main problems:

1. **Unnecessary API Call**: The code was making an `Organizations.Edit(ctx, org, nil)` call with nil parameters, which could trigger API validation errors.

2. **Empty String Pollution**: The resource was unconditionally sending all fields to the GitHub API, including empty strings for optional fields that weren't configured, causing API rejections.

3. **Boolean False Handling**: The code used `d.GetOk()` for boolean fields, which returns `false` for boolean `false` values, causing these fields to be omitted from the API payload even when explicitly set to `false`.

### Core Fixes

- **Replaced unnecessary API call**: Changed `Organizations.Edit(ctx, org, nil)` to `Organizations.Get()` to safely retrieve organization plan information
- **Created conditional payload builder**: Implemented `buildOrganizationSettings()` helper function that only includes explicitly configured fields
- **Fixed boolean handling**: Used `d.GetOkExists()` for boolean fields to properly detect when `false` values are explicitly configured
- **Optimized API payloads**: Only send fields that are actually configured, avoiding empty string issues

### Enterprise Support

- **Enterprise detection**: Properly detect enterprise organizations using plan name
- **Conditional enterprise fields**: Only include enterprise-specific fields like `members_can_create_internal_repositories` for enterprise organizations
- **EMU compatibility**: Full support for Enterprise Managed User organizations

### Files Modified

- `github/resource_github_organization_settings.go` - Core implementation fixes
- `github/resource_github_organization_settings_test.go` - Comprehensive test coverage

### Key Implementation Details

1. **`buildOrganizationSettings()` Function**:
   - Conditionally builds `github.Organization` struct
   - Only includes fields that are explicitly configured
   - Handles enterprise-specific fields appropriately
   - Uses `d.GetOkExists()` for proper boolean false detection

2. **API Call Optimization**:
   - Replaced problematic `Edit(nil)` call with safe `Get()` call
   - Ensures organization plan information is retrieved without side effects

3. **Enterprise Field Handling**:
   - Detects enterprise organizations via `orgInfo.GetPlan().GetName() == "enterprise"`
[...]

PR #3570, opened 2026-07-23
fix: include configured false booleans on organization_settings create

### LINKED ISSUE #3493: [BUG]: `github_organization_settings` silently ignores boolean attributes set to `false` on create (GetOk zero-value bug in shouldInclude)
### Expected Behavior

On the **initial create** of a `github_organization_settings` resource, every boolean attribute explicitly set in the configuration should be sent to the GitHub Org API, including attributes set to `false`. After a single `terraform apply`, the server-side org settings should match the configuration exactly, and a subsequent `terraform plan` should show no changes.

### Actual Behavior

On **create**, any boolean attribute the user explicitly sets to `false` is **omitted from the API PATCH payload**. The GitHub API therefore never receives those values and applies its own **server-side defaults** instead, leaving the org in a state that contradicts the Terraform configuration — with no error or warning at apply time.

A **second** `terraform apply` (which goes through the *update* code path) then corrects the values, because the update path uses `d.HasChange()` instead of `d.GetOk()`. So the symptom is: the first apply silently produces the wrong state, and the diff "reappears" until a second apply.

Concrete real-world example (Enterprise Cloud org, fresh resource create):

- Config: `members_can_create_repositories = true`, `members_can_create_public_repositories = false`
- After the first `apply`: server ends with public repo creation **ENABLED** (and `members_can_create_public_pages` enabled) — the opposite of the `false` in config.
- After a second `apply` (update path): the values are correctly set to `false`.

This generalizes to **any** boolean attribute on this resource that the user intends to be `false` on create, e.g. `web_commit_signoff_required`, the various fork-related flags, and the security/Dependabot toggles.

### Terraform Version

```
Terraform v1.14.x
+ provider registry.terraform.io/integrations/github v6.12.1  (constraint ~> 6.0)
```

Confirmed still present on the current `main` tip (the `shouldInclude` helper still calls `d.GetOk` on the create path — see Root Cause).

### GitHub Installation Type

GitHub Enterprise Cloud with Personal Accounts (github.com)

(Note: this is a provider-side logic bug on the resource **create** path and is not specific to the installation type — it reproduces on any org where a boolean is configured `false` on the initial create.)

### Affected Resource(s)

- `github_organization_settings`

```hcl
resource "github_organization_settings" "this" {
  billing_email                          = "x@example.com"
  members_can_create_repositories        = true
  members_can_create_public_repositories = false
}
```

### Steps to Reproduce

1. Ensure no `github_organization_settings` resource currently exists in state (this bug is on the **create** path), and that the org's server-side `members_can_create_public_repositories` is currently `true` (e.g. GitHub default).
2. Apply the configuration above for the **first time**:
   ```
   terraform apply
   ```
3. Inspect the server state:
   ```
   gh api orgs/<org> --jq '{members_can_create_repositories, members_can_create_public_repositories, members_can_create_public_pages}'
   ```
   Observe `members_can_create_public_repositories = true` (and `members_can_create_public_pages = true`) — i.e. the `false` in config was ignored.
4. Run `terraform plan` again — it shows `members_can_create_public_repositories: true -> false`.
5. Run `terraform apply` a **second** time. This time the update path runs and the value is correctly set to `false`. `terraform plan` is now clean.

### Root Cause

`buildOrganizationSettings` in
[`github/resource_github_organization_settings.go`](https://github.com/integrations/terraform-provider-github/blob/main/github/resource_github_organization_settings.go#L185-L193)
gates each field on a `shouldInclude` helper that branches on create vs. update:

```go
shouldInclude := func(fieldName string) bool {
    if !isUpdate {
        // For creates, include if explicitly configured
        _, ok := d.GetOk(fieldName)
        return ok
    }
    // For updates, only include if the field has changed
    return d.HasChange(fieldName)
}
```

The fields are then conditionally added as pointers, e.g.:

```go
if shouldInclude("members_can_create_public_repositories") {
    settings.MembersCanCreatePublicRepos = github.Ptr(d.Get("members_can_create_public_repositories").(bool))
}
```

On the **create** path (`!isUpdate`) it uses `d.GetOk(fieldName)`. For a `TypeBool`, `d.GetOk` returns `ok = false` whenever the resolved value is the zero value (`false`) — it **cannot distinguish an explicitly configured `false` from an unset attribute**. This is the well-known Terraform SDK `GetOk` + zero-value footgun.

As a result, `shouldInclude("...")` returns `false` for every boolean the user set to `false`, the corresponding `*bool` field on the go-github `Organization` struct is left `nil`, and because those fields are tagged `omitempty`, they are dropped from the PATCH body entirely. GitHub then applies its server-side default.

The **update** path is unaffected because it uses `d.HasChange()`, which correctly reports a change from `true -> false`. Hence the "second apply fixes it" behavior.

This `shouldInclude`/`GetOk`-on-create logic was introduced in PR **#2807** ("fix: HTTP 422 error in `github_organization_settings` resource"). Note: the PR description claims it "Used `d.GetOkExists()` for boolean fields to properly detect when `false` values are explicitly configured," but the merged code (and current `main`) actually uses `d.GetOk()` in `shouldInclude`, so the intended `false`-detection was not achieved on the create path.

### Why `Default: true` does not rescue this

Several of these booleans declare `Default: true` in the schema (e.g. `members_can_create_repositories`, `members_can_create_public_repositories`, `members_can_create_public_pages`). The schema default does **not** help: when the user explicitly sets the attribute to `false`, the *resolved* value is `false`, which is still the `TypeBool` zero value. `d.GetOk` keys off the zero value, not off whether the value differs from the schema default, so `ok` is still `false` and the field is still omitted. (`web_commit_signoff_required` has `Default: false` and is affected identically.)

### Generalization

This is not limited to the repo-creation booleans in the reproduction. It affects **every** `TypeBool` attribute on `github_organization_settings` that a user intends to set to `false` on the initial create, including (non-exhaustive):

- `members_can_create_public_repositories`, `members_can_create_private_repositories`, `members_can_create_internal_repositories`
- `members_can_create_pages`, `members_can_create_public_pages`, `members_can_create_private_pages`
- `members_can_fork_private_repositories`
- `web_commit_signoff_required`
- the `advanced_security_*` / `dependabot_*` / `secret_scanning_*` security toggles

Any of these set to `false` in the config will be silently dropped from the create PATCH and left at GitHub's server-side default.

### Suggested Fix

Make the create-path inclusion test detect *explicit configuration* rather than relying on the zero-value semantics of `GetOk`. Two viable directions:

1. **Use raw config to detect explicit set-ness.** On create, include the field when it is present (non-null) in the raw configuration, e.g. via `d.GetRawConfig()` and checking the attribute is not null. This correctly distinguishes a configured `false` from an unset attribute for `TypeBool`.

2. **Always include booleans on create.** Since the schema supplies a `Default` for these attributes, the resolved value (`d.Get`) is always well-defined on create. The create branch of `shouldInclude` can simply return `true` for the boolean fields (or the build function can unconditionally set the `*bool` fields on create via `github.Ptr(d.Get(name).(bool))`). This sends the user's intended value (including `false`) every time.

A minimal patch in the spirit of option (2), mirroring the fix pattern already proposed for the analogous `GetOk`-zero-value bug in #3223:

```go
shouldInclude := func(fieldName string) bool {
    if !isUpdate {
        // On create, include any attribute that is explicitly present in config.
        // GetOk() is wrong for TypeBool: it returns ok=false for an explicitly
        // configured `false`, silently dropping it from the PATCH.
        return !d.GetRawConfig().GetAttr(fieldName).IsNull()
    }
    return d.HasChange(fieldName)
}
```

### Relation to existing issues

- **#3429** (`github_organization_settings silently drifts repo-creation booleans due to partial PATCH × deprecated legacy field`) — same resource and a similar *observable symptom* (repo-creation booleans not matching config), but a **different root cause**. #3429 is about the **update** path: a partial `HasChange`-driven PATCH that resets the deprecated `members_allowed_repository_creation_type` server-side, entangling the four repo-creation booleans. This issue is about the **create** path, where `d.GetOk` omits *any* `false` boolean before the API is even called. The two are independent and would need separate fixes.
- **#3223** (`sha_pinning_required = false is silently ignored due to d.GetOk() zero-value bug`) — **same root cause** (`d.GetOk` zero-value footgun on a `TypeBool`), but in **different resources** (`github_actions_organization_permissions` / `github_actions_repository_permissions`). This issue reports the identical class of bug in `github_organization_settings`, where it additionally manifests on every `false`-intended boolean across the whole resource.

### Code of Conduct

somaz94:
Resolves #3493

----

### Before the change?

On create, a boolean explicitly set to `false` was dropped from the payload. Inclusion was decided with `d.GetOk`, which reports `false` the same as unset, so GitHub kept its own value and only a second `apply` (through the `HasChange` update path) converged.

### After the change?

Create sends an attribute when it is written in the configuration (read from the raw config, so an explicit `false` counts) or when its schema default is non-zero, which the create payload already contained before. Unset attributes that default to `false`/`""` still stay out of the request, so the narrow payload from #2807 (#2305) is preserved. Update is unchanged: only changed attributes are sent.

Following review, create and update are separate CRUD functions with the logic inline, neither calls Read, logging uses `tflog`, and the raw-config check lives in `util.go` as `isConfigured`.

### Pull request checklist

- [ ] Schema migrations have been created if needed → N/A (no schema change).
- [x] Tests for the changes have been added → acceptance case `sends booleans configured as false on create` (explicit `false` on five default-`true` booleans, then a plan-only step that must be empty). `TestAccGithubOrganizationSettings` is still skipped by the existing `t.Skip("TODO: Make this test cleanup correctly")`, so it does not run in CI, and I have not run it against a live org.
- [x] Docs have been reviewed and added / updated if needed → N/A (no schema or docs change).

- [ ] Yes
- [x] No

----

> AI assistance disclosure: this PR was largely written with an AI coding assistant (Claude Code). I reviewed it and validated it locally with `gofmt`, `go build ./...`, `go vet ./github/`, `go test ./github/`, and `golangci-lint` with `.golangci.new.yml` (0 new issues). The acceptance test has not been run against a live organization.


github/resource_github_organization_settings_unit_test.go
deiga, line 1: Please don't create a separate unit test file. Add the unit tests to the existing test file instead ☺️

github/resource_github_organization_settings_unit_test.go
```diff
@@ -0,0 +1,125 @@
...
+	"testing"
+
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+)
+
+// TestBuildOrganizationSettingsCreateIncludesFalseBool verifies that a boolean
+// attribute explicitly configured as false is included in the create payload.
+// Regression test for the create path dropping false booleans (only fixed by a
+// second apply through the update/HasChange path).
+func TestBuildOrganizationSettingsCreateIncludesFalseBool(t *testing.T) {
```
deiga: Please use an underscore to separate function under test and the testcase in the name

deiga (changes requested):
Nice work!

Could you refactor the unit tests to use a single testing function with a table for test cases?

somaz94:
@deiga Thanks for the review — all three points addressed in `8cead3a`:

- **No separate file** — deleted `resource_github_organization_settings_unit_test.go`; the tests now live at the bottom of the existing `resource_github_organization_settings_test.go`.
- **Single table-driven function** — the five separate test functions are now one `Test_buildOrganizationSettings` with a 6-case table, following the same inline-anonymous-slice idiom as `Test_configureProviderMeta` in `provider_test.go`. Splitting the old enterprise test into two rows (`enterprise_…` / `non_enterprise_…`) made that case clearer too.
- **Underscore naming** — `Test_buildOrganizationSettings` matches the repo's `Test_<functionUnderTest>` convention, and the case names use underscores (`create_includes_explicitly_false_booleans`, `update_omits_unchanged_booleans`, …).

I also verified the table still has teeth: temporarily reverting the `shouldInclude` bool branch makes exactly `create_includes_explicitly_false_booleans` and `enterprise_create_includes_internal_repositories_boolean` fail, and restoring it turns them green.

Rebased on latest `main` as well, so the branch is no longer behind. `go vet` / `gofmt` / the unit tests are all clean.

github/resource_github_organization_settings_unit_test.go
somaz94: Done in `8cead3a` — this file is deleted. The unit tests now live at the bottom of the existing `github/resource_github_organization_settings_test.go`.

github/resource_github_organization_settings_unit_test.go
somaz94: Done in `8cead3a` — the five separate funcs are now a single table-driven `Test_buildOrganizationSettings`, following the repo's existing `Test_getBaseURL` / `Test_configureProviderMeta` convention. Testcase names use underscores (e.g. `create_includes_explicitly_false_booleans`, `enterprise_create_includes_internal_repositories_boolean`).

github/resource_github_organization_settings.go
deiga, line 1: suggestion: There is little benefit on using `shouldInclude` for boolean fields, if it will just return true. What id you instead remove the call to `shouldInclude` from the boolean fields?

github/resource_github_organization_settings_test.go
```diff
@@ -613,3 +614,149 @@ func TestAccGithubOrganizationSettings(t *testing.T) {
...
+				d.SetId(tt.id)
+			}
+
+			got := buildOrganizationSettings(d, tt.isEnterprise)
+
+			assertStringPtr(t, "billing_email", got.BillingEmail, tt.wantBillingEmail)
+			assertStringPtr(t, "company", got.Company, tt.wantCompany)
+			assertBoolPtr(t, "has_organization_projects", got.HasOrganizationProjects, tt.wantHasOrganizationProjects)
+			assertBoolPtr(t, "web_commit_signoff_required", got.WebCommitSignoffRequired, tt.wantWebCommitSignoffRequired)
+			assertBoolPtr(t, "members_can_create_internal_repositories", got.MembersCanCreateInternalRepos, tt.wantMembersCanCreateInternalRepos)
```
deiga: issue: this seems superfluous, wouldn't the following cover the same cases without extra custom code?
```
if diff := cmp.Diff(got, tt.want); diff != "" {
	t.Fatalf("got %+v, want %+v: %s", got, tt.want, diff)
}
```

deiga (changes requested):
Thanks for the work!

github/resource_github_organization_settings.go
somaz94: Done in `cbe668a8`, with one wrinkle worth flagging.

Removing the call outright would have changed the update path: `shouldInclude` returns `true` for booleans only on create — on update it still falls through to `d.HasChange(fieldName)`, which is what keeps unchanged fields out of the payload and avoids the API validation errors the function comment mentions.

So I split it in two instead. The schema lookup and the `TypeBool` special case are gone, and the boolean fields no longer call `shouldInclude`:

```go
// Non-boolean fields are included only when explicitly configured on create,
// and only when changed on update.
shouldInclude := func(fieldName string) bool {
	if isUpdate {
		return d.HasChange(fieldName)
	}
	_, ok := d.GetOk(fieldName)
	return ok
}

// Boolean attributes always carry a definite value from their schema
// default, and d.GetOk cannot tell an explicit false from an unset field,
// so on create they are included unconditionally.
includeBool := func(fieldName string) bool {
	return !isUpdate || d.HasChange(fieldName)
}
```

Happy to collapse it further if you'd rather see something else.

github/resource_github_organization_settings_test.go
somaz94: Agreed — done in `ecd0631f`. `assertBoolPtr`, `assertStringPtr`, the `omitted` map and the `wantOmitted` field are all gone; each case now declares a single `want *github.Organization` and the assertion is just:

```go
if diff := cmp.Diff(tt.want, got); diff != "" {
	t.Errorf("buildOrganizationSettings() mismatch (-want +got):\n%s", diff)
}
```

Two notes on what changed as a result:

1. Comparing the whole payload means the create cases have to spell out every attribute, since on create each boolean is sent with its schema default. That makes the table longer, but it documents the exact create payload — which is the thing this PR is fixing — and it will now fail if a future attribute silently changes what create sends.

2. It caught something the old per-field assertions were blind to. `update_omits_unchanged_booleans` was also emitting eight attributes I had not been asserting. That turned out not to be a bug in the resource but an artifact of `schema.TestResourceDataRaw`: an attribute left out of `raw` reads as the zero value from state but as its schema default from `d.Get`, so anything defaulting to a non-zero value (`true`, `"read"`) looks changed. I encoded the real output and left a comment explaining it. The case's actual point — that the explicitly configured `has_organization_projects: false` is dropped on update — still holds.

`gofmt` and `go vet` are clean and all six subtests pass.

github/resource_github_organization_settings.go
deiga: @stevehipwell do I remember correctly that we want to remove all unnecessary partial updates? 
In that case the whole shouldInclude function wpuld be unnecessary

somaz94:
@deiga @stevehipwell — happy to go either way on this, just let me know which you prefer so I do not build the wrong thing.

**Option A (this PR as it stands):** keep `shouldInclude`, but stop it from dropping explicitly-configured `false` booleans on create. Narrow, low-risk, fixes #3429.

**Option B (what @deiga raised):** remove `shouldInclude` entirely and always send the full set of fields on create. That is a superset of A — it fixes the same false-boolean bug as a side effect — but it is a broader behavioural change and touches every field, not just the booleans.

I am glad to do B in this PR if that is the agreed direction. The only thing blocking me is knowing which one you want.

For what it is worth, the three earlier review threads are already addressed (`8cead3a`, `cbe668a8`, `ecd0631f`) — they are just still showing unresolved.

github/resource_github_organization_settings.go
somaz94: @deiga — one data point that might make this cheaper to decide than it looks: `shouldInclude` is used in exactly one place in the repo. It's defined in this file and called 27 times, all inside the same function. So removing it entirely would be a single-resource change, not a cross-provider refactor.

I also went looking for the broader rule you're recalling and couldn't find it written down anywhere — no issue, PR or CONTRIBUTING note about removing partial updates. The closest thing is #3429, which is the bug this PR fixes.

stevehipwell hasn't been active since Jul 27, so rather than hold this on his answer: if you'd like the full-payload version, say the word and I'll push it in this PR. If you'd rather keep the current `includeBool` split and treat removing `shouldInclude` as its own change, that works too. Happy either way — I just don't want to guess and build the wrong one.

Separately, and regardless of the above: `CI` and `CodeQL` are sitting at `action_required` on `ecd0631f` and have never run on this branch, so the green checkmarks are only the `pull_request_target` workflows. The Go tests haven't actually executed here yet. If you approve the run we'd both be looking at real results.

Happy to rebase too — it's one commit behind.

lloydmcl:
I hit this on a first apply against a new organisation, so a data point in favour of fixing it, plus one thing worth weighing before `shouldInclude` comes out entirely.

The partial payload on create is load-bearing. It arrived in #2807 to fix #2305, where the resource returned a 422 on every apply for EMU organisations because unconfigured fields were being sent. Dropping `shouldInclude` would send the full field set on create again, which is the shape that caused #2305 in the first place.

There is also a live conflict with #3360. That PR fixes #2689, a 422 when an enterprise policy locks forking, by removing `Default: false` from `members_can_fork_private_repositories` so the attribute is only sent when it is explicitly configured. `includeBool` sends it on every create regardless, so whichever of the two lands second undoes the other. The rationale that booleans always carry a definite value from their schema default also stops holding for that attribute once #3360 removes its default.

Option 1 in #3493 sidesteps both. Testing the raw configuration for a null attribute distinguishes a configured `false` from an unset one without widening the create payload at all, so it fixes this bug and stays compatible with #3360.

somaz94:
Thanks for digging into this — the #3360 point is a fair catch and I'd missed it. `includeBool` does send `members_can_fork_private_repositories` on every create, which is the shape that produces the 422 in #2689, and my comment's premise about booleans always carrying a definite schema default stops holding for that attribute once #3360 makes it Computed.

On which way to fix it, I don't think I should pick on my own. @deiga suggested the opposite direction on 28 Jul — dropping the `shouldInclude` call from the boolean fields — and then asked @stevehipwell on 29 Jul whether the project wants to remove unnecessary partial updates entirely, which would remove `shouldInclude` altogether. That question hasn't been answered, and I think it's what this PR is actually waiting on.

Either answer is small work from here. If partial updates stay, I'll switch the create check to a raw-config null test (the sample patch in #3493), which fixes the explicit `false` without widening the create payload and stays compatible with #3360. If they go, this branch is already most of the way there and I'll drop `shouldInclude` from the non-boolean fields too.

@stevehipwell @deiga — could I get a steer? Happy to re-cut either way.

stevehipwell (changes requested):
For me before any other changes are made the helper needs removing and the code needs to be put back into create/update. Then the API surface needs to be correctly designed to support the limitations of Terraform (SDK v2), as once a field is set there is no way to check if was done so by the user or a read.

FYI for boolean schema values `Get` should be used as `getOk` always returns `ok` as `true`.

somaz94:
Thanks @stevehipwell — that's the steer I was after. One thing I want to pin down before re-cutting: with `shouldInclude` gone, should create send the full field set? That's the shape #2807 narrowed to fix #2305, and @lloydmcl flagged it would also collide with #3360 on `members_can_fork_private_repositories`. If you'd rather keep the payload narrow I'll use the raw-config null check from #3493 instead, and either way I'll split `resourceGithubOrganizationSettingsCreateOrUpdate` so create and update stop sharing a helper that re-detects the operation from `d.Id()`.

stevehipwell:
@somaz94 if a field is optional and can be overridden at a higher layer we shouldn't set/read it if the user hasn't explicitly set it.

somaz94:
Done. The helper is gone and create and update are separate handlers again.

Inclusion is now decided from the raw config rather than `GetOk`, so an attribute is only sent when the user actually wrote it. Unconfigured fields stay out of the create payload the way #2305 needs, and a boolean explicitly set to `false` now reaches the API instead of being silently dropped (#3493).

That also means `members_can_fork_private_repositories` is only sent when it's configured, so it no longer collides with #3360.

Boolean values are read with `d.Get`.

I split it into a refactor commit and the fix so the two read separately.

somaz94:
Copilot's first point is right, fixed in bb3b9c16. Create now also sends an attribute when its schema default is a non-zero value, which is what the GetOk path already did before this PR, so the create payload is unchanged apart from explicitly configured false values now reaching it. Attributes defaulting to false or an empty string are still only sent when configured, so nothing new is sent to EMU orgs. The unit case for an unset default-true attribute now expects it to be sent, and the test helper fills schema defaults into the attributes the way the planned state does.

github/resource_github_organization_settings.go
```diff
@@ -173,153 +175,288 @@ func resourceGithubOrganizationSettings() *schema.Resource {
...
+	return v
+}
+
+// boolAttr returns the boolean value of the named attribute, with the same
+// type guarantee as stringAttr.
+func boolAttr(d *schema.ResourceData, name string) bool {
+	v, _ := d.Get(name).(bool)
+
+	return v
+}
```
stevehipwell: These are unnecessary and make the code harder to read, please remove them and keep the logic inline.

github/resource_github_organization_settings.go
```diff
@@ -173,153 +175,288 @@ func resourceGithubOrganizationSettings() *schema.Resource {
...
+
+	v, diags := d.GetRawConfigAt(cty.GetAttrPath(name))
+	if diags.HasError() {
+		_, ok := d.GetOk(name)
+
+		return ok
+	}
+
+	return !v.IsNull()
+}
```
stevehipwell: If this is a reusable helper then it should be in one of the utils files.

stevehipwell (changes requested):
@somaz94 there are still too many abstractions in this logic, please refactor back to the CRUD functions and remove all but the obvious abstractions. For example if the logic for getting the payload for create and update are different then there shouldn't be a common function, it should be inline.

And also RE tests, please add acceptance tests and not unit tests.

github/resource_github_organization_settings.go
```diff
@@ -173,153 +175,265 @@ func resourceGithubOrganizationSettings() *schema.Resource {
...
-	}
-
-	// Build settings using helper function
-	isEnterprise := orgInfo.GetPlan().GetName() == "enterprise"
-	settings := buildOrganizationSettings(d, isEnterprise)
-
-	// Debug: Log the settings being sent to the API with detailed field information
+// logOrganizationSettings records the payload about to be sent to the API.
+func logOrganizationSettings(org string, isEnterprise bool, settings *github.Organization) {
+	log.Printf("[DEBUG] Built settings for org %s (enterprise: %v)", org, isEnterprise)
```
deiga: Please use tflog instead of log

github/resource_github_organization_settings.go
```diff
@@ -412,129 +540,177 @@ func resourceGithubOrganizationSettingsCreateOrUpdate(d *schema.ResourceData, me
...
+
+		return diag.FromErr(err)
 	}
-	id := strconv.FormatInt(orgSettings.GetID(), 10)
-	d.SetId(id)
 
-	return resourceGithubOrganizationSettingsRead(d, meta)
+	d.SetId(strconv.FormatInt(orgSettings.GetID(), 10))
+
+	return resourceGithubOrganizationSettingsRead(ctx, d, owner)
```
deiga: Never call a CRUD function

deiga (changes requested):

somaz94:
Done: the logic is back inline in create/update, the helpers are gone apart from `isConfigured`, which moved to `util.go` since it's generic, logging uses `tflog`, and create/update no longer call Read. The unit tests are replaced by an acceptance test (explicit `false` on default-`true` booleans, then an empty plan).

One question: the whole `TestAccGithubOrganizationSettings` suite is still behind the existing `t.Skip("TODO: Make this test cleanup correctly")`, so the new case won't run in CI. Should I move it into its own enabled test, or leave the suite as is?

stevehipwell:
@somaz94 ideally `TestAccGithubOrganizationSettings` needs fixing so it cleans up correctly, all resources tests should be grouped under this function.

somaz94:
Done — added a `mustSnapshotOrganizationSettings` helper that captures the org settings and puts them back in `t.Cleanup`, following the same shape as `mustAssignOrganizationRoleToUser`, and dropped the `t.Skip`. All 26 attributes the resource manages are captured, with the enterprise-only `members_can_create_internal_repositories` guarded the way delete guards it. I can't run the acceptance suite myself, so it needs the `acctest` label to actually be verified.

github/resource_github_organization_settings.go
```diff
@@ -173,368 +174,448 @@ func resourceGithubOrganizationSettings() *schema.Resource {
...
+	if !ok {
+		return diag.Errorf("unexpected provider meta type %T", meta)
 	}
 
-	// Required field - always include if configured (API requires it even if unchanged)
-	if billingEmail, ok := d.GetOk("billing_email"); ok {
-		settings.BillingEmail = new(billingEmail.(string))
+	ctx = context.WithValue(ctx, ctxId, d.Id())
+
+	orgInfo, _, err := owner.v3client.Organizations.Get(ctx, owner.name)
```
stevehipwell:
```suggestion
func resourceGithubOrganizationSettingsCreate(ctx context.Context, d *schema.ResourceData, m any) diag.Diagnostics {
	meta, _ := m.(*Owner)

	if ok, diags := checkOrganizationOK(meta); !ok {
		return diags
	}

	orgInfo, _, err := meta.v3client.Organizations.Get(ctx, owner.name)
```

This is the current pattern.

github/resource_github_organization_settings.go
```diff
@@ -173,368 +174,448 @@ func resourceGithubOrganizationSettings() *schema.Resource {
...
-			settings.MembersCanCreateInternalRepos = new(d.Get("members_can_create_internal_repositories").(bool))
+		if _, ok := d.GetOk("members_can_create_internal_repositories"); ok || isConfigured(d, "members_can_create_internal_repositories") {
+			v, _ := d.Get("members_can_create_internal_repositories").(bool)
+			settings.MembersCanCreateInternalRepos = new(v)
 		}
 	}
 
-	return settings
-}
+	tflog.Debug(ctx, "Creating organization settings", map[string]any{
```
stevehipwell: Please keep `tflog` on a single line.

github/resource_github_organization_settings.go
```diff
@@ -173,368 +174,448 @@ func resourceGithubOrganizationSettings() *schema.Resource {
...
+	orgSettings, _, err := owner.v3client.Organizations.Edit(ctx, owner.name, settings)
 	if err != nil {
-		return err
+		if ghErr, ok := errors.AsType[*github.ErrorResponse](err); ok {
+			tflog.Debug(ctx, "GitHub API error", map[string]any{
+				"status":  ghErr.Response.StatusCode,
+				"message": ghErr.Message,
+				"errors":  ghErr.Errors,
+			})
+		}
```
stevehipwell: This is unnecessary.

github/resource_github_organization_settings.go
```diff
@@ -173,368 +174,448 @@ func resourceGithubOrganizationSettings() *schema.Resource {
...
 
-	orgInfo, _, err := client.Organizations.Get(ctx, org)
-	if err != nil {
-		return err
+	d.SetId(strconv.FormatInt(orgSettings.GetID(), 10))
+
+	return nil
+}
+
+func resourceGithubOrganizationSettingsUpdate(ctx context.Context, d *schema.ResourceData, meta any) diag.Diagnostics {
```
stevehipwell: We order the functions as CRUD(I) (see `ARCHITECTURE.md`).

github/resource_github_organization_settings.go
somaz94: Done — the split-out helpers are gone and the logic is inline in Create and Update now.

github/resource_github_organization_settings.go
somaz94: Moved `isConfigured` into `github/util.go`.

github/resource_github_organization_settings.go
somaz94: Done — this file uses `tflog` only now, no `log`.

github/resource_github_organization_settings.go
somaz94: Done — Create and Update no longer call Read.

github/resource_github_organization_settings.go
somaz94: Applied, and did the same for Read, Update and Delete so all four match.

github/resource_github_organization_settings.go
somaz94: Done.

github/resource_github_organization_settings.go
somaz94: Removed.

github/resource_github_organization_settings.go
somaz94: Reordered to Create, Read, Update, Delete.

PR #3572, opened 2026-07-23
fix(resource_github_organization_custom_properties): support bool default values

### LINKED ISSUE #3580: [BUG]: github_organization_custom_properties of type `true_false` detects drift on `default_value` indefinitely

secustor:
This is a minimal fix for users of default values and is intentionally tries not to reach the scope of https://github.com/integrations/terraform-provider-github/pull/3234

Closes: #3580

Resolves infinite drift if when using `resource_github_organization_custom_properties` of `value_type` `true_false` and `default_value`.

Infinite terraform plan: 
```
  ~ update in-place

  # github_organization_custom_properties.stedisentry_enabled will be updated in-place
  ~ resource "github_organization_custom_properties" "foo" {
      + default_value      = "true"
        id                 = "foo"
    }

```
----

### Before the change?

`customProperty.DefaultValueString()` will return `"", nill`  because of https://github.com/google/go-github/blob/a3951f514384b06cc428f394672020c14f7efd84/github/orgs_properties.go#L59

```hcl
resource "github_organization_custom_properties" "foo" {
  property_name = "foo"
  value_type    = "true_false"
  required      = true

  // currently there is an infinite drift setting creating an addition on each plan 
  default_value = "true" // `true` has the same behavior 
}
``` 

This forces following workaround 

```hcl
resource "github_organization_custom_properties" "foo" {
  property_name = "foo"
  value_type    = "true_false"
  required      = true

  default_value = "true"

  lifecycle {
    ignore_changes = [default_value]
  }
}
``` 

### After the change?

With this instead `customProperty.DefaultValueBool()` is called and with that the return value is `true, nill` which will then correctly set in the state

### Pull request checklist

- [ ] Yes
- [x] No

----

### DIFF
--- a/github/resource_github_organization_custom_properties.go
+++ b/github/resource_github_organization_custom_properties.go
@@ -2,6 +2,7 @@ package github
 
 import (
 	"context"
+	"strconv"
 
 	"github.com/google/go-github/v89/github"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/customdiff"
@@ -121,8 +122,19 @@ func resourceGithubCustomPropertiesRead(d *schema.ResourceData, meta any) error 
 		return err
 	}
 
-	// TODO: Add support for other types of default values
-	defaultValue, _ := customProperty.DefaultValueString()
+	// multi_select is not supported: its default value is a []string, which
+	// cannot round-trip through the TypeString default_value attribute.
+	var defaultValue string
+	switch customProperty.ValueType {
+	case github.PropertyValueTypeTrueFalse:
+		if b, ok := customProperty.DefaultValueBool(); ok {
+			defaultValue = strconv.FormatBool(b)
+		}
+	default:
+		if s, ok := customProperty.DefaultValueString(); ok {
+			defaultValue = s
+		}
+	}
 
 	d.SetId(*customProperty.PropertyName)
 	_ = d.Set("allowed_values", customProperty.AllowedValues)
--- a/github/resource_github_organization_custom_properties.go
+++ b/github/resource_github_organization_custom_properties.go
@@ -45,7 +45,7 @@ func resourceGithubOrganizationCustomProperties() *schema.Resource {
 			},
 			"default_value": {
 				Type:        schema.TypeString,
-				Description: "The default value of the custom property",
+				Description: "The default value of the custom property. Not supported for multi_select properties.",
 				Optional:    true,
 				Computed:    true,
 			},
--- a/github/resource_github_organization_custom_properties.go
+++ b/github/resource_github_organization_custom_properties.go
@@ -5,6 +5,7 @@ import (
 	"strconv"
 
 	"github.com/google/go-github/v89/github"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/customdiff"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/validation"
@@ -12,12 +13,12 @@ import (
 
 func resourceGithubOrganizationCustomProperties() *schema.Resource {
 	return &schema.Resource{
-		Create: resourceGithubCustomPropertiesCreate,
-		Read:   resourceGithubCustomPropertiesRead,
-		Update: resourceGithubCustomPropertiesUpdate,
-		Delete: resourceGithubCustomPropertiesDelete,
+		CreateContext: resourceGithubCustomPropertiesCreate,
+		ReadContext:   resourceGithubCustomPropertiesRead,
+		UpdateContext: resourceGithubCustomPropertiesUpdate,
+		DeleteContext: resourceGithubCustomPropertiesDelete,
 		Importer: &schema.ResourceImporter{
-			State: resourceGithubCustomPropertiesImport,
+			StateContext: resourceGithubCustomPropertiesImport,
 		},
 
 		CustomizeDiff: customdiff.Sequence(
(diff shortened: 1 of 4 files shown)


deiga:
Please mark clearly which issue this is going to resolve!

secustor:
> Resolves infinite drift if when using resource_github_organization_custom_properties of value_type true_false and default_value.

I have been under the impression that the explanation above is clear enough, now there is also an example of the Terraform plan which will be generated on each plan and is not reconciled.

deiga:
@secustor I meant which issue number/URL.
If no issue exists yet, please create one :)

github/resource_github_organization_custom_properties_test.go
```diff
@@ -265,6 +265,42 @@ resource "github_organization_custom_properties" "test" {
...
+}
+`, name)
+
+		resource.Test(t, resource.TestCase{
+			PreCheck:          func() { skipUnlessHasOrgs(t) },
+			ProviderFactories: providerFactories,
+			Steps: []resource.TestStep{
+				{
+					Config: config,
+					Check: resource.ComposeTestCheckFunc(
```
deiga: issue: Please use `ConfigStateChecks` instead

templates/resources/organization_custom_properties.md.tmpl
```diff
@@ -40,7 +40,7 @@ The following arguments are supported:
 
 - `description` - (Optional) The description of the custom property.
 
-- `default_value` - (Optional) The default value of the custom property.
+- `default_value` - (Optional) The default value of the custom property. Not supported for `multi_select` properties.
```
deiga, line 43: Please update the `Description` in the resource as well

deiga (changes requested):
Thanks for the effort so far!

github/resource_github_organization_custom_properties.go
```diff
@@ -44,7 +45,7 @@ func resourceGithubOrganizationCustomProperties() *schema.Resource {
 			},
 			"default_value": {
 				Type:        schema.TypeString,
-				Description: "The default value of the custom property",
+				Description: "The default value of the custom property. Not supported for multi_select properties.",
```
deiga, line 50: Would it make sense to add a validation to not use default with multi-select?

github/resource_github_organization_custom_properties_test.go
```diff
@@ -265,6 +268,42 @@ resource "github_organization_custom_properties" "test" {
...
+`, name)
+
+		resource.Test(t, resource.TestCase{
+			PreCheck:          func() { skipUnlessHasOrgs(t) },
+			ProviderFactories: providerFactories,
+			Steps: []resource.TestStep{
+				{
+					Config: config,
+					ConfigStateChecks: []statecheck.StateCheck{
+						statecheck.ExpectKnownValue("github_organization_custom_properties.test", tfjsonpath.New("default_value"), knownvalue.StringExact("true")),
```
deiga: There is no value in checkong config values, testing Framework already does that.

github/resource_github_organization_custom_properties_test.go
```diff
@@ -265,6 +268,42 @@ resource "github_organization_custom_properties" "test" {
...
+					},
+				},
+				{
+					// Read must round-trip the true_false default_value so the
+					// second plan is empty (regression guard for perpetual drift).
+					Config:             config,
+					PlanOnly:           true,
+					ExpectNonEmptyPlan: false,
+				},
```
deiga: Could this be replaced with a postapplypostrefresh configplancheck in the previous step?

deiga (changes requested):

github/resource_github_organization_custom_properties.go
secustor: My understanding is this would be breaking the setup for existing users which use the `ignore_changes` workaround 🤔

deiga: I'm not sure that breaking the wrong way would be a problem. But if we want to be careful we should at least introduce a warning for this case

secustor: What kind of warning do you want me to add? 

I see different patterns here in the repo. 
- Logging with `log.Printf("[WARN] ...")` 
- Using diagnostics, though that would need more changes in that PR
- or simply in the docs

deiga: diagnostics would be the best, let's go with that 🙏

stevehipwell:
I think #3234 should fix this, but if not this change would need to be made on top of the other PR.

secustor:
> I think #3234 should fix this, but if not this change would need to be made on top of the other PR.

Potentially, I have referred this PR in the PR body. 
The intention is to deliver a minimal fix until #3234 lands.

stevehipwell:
@secustor I don't think it's worth fixing this resource given that it's going to be replaced, I'd rather incentivise people to `remove` -> `import` sooner than later.

PR #3578, opened 2026-07-24
feat: Add explicit auth mode support to provider

### LINKED ISSUE #3576: [BUG]: GITHUB_APP_* env vars silently override PAT-configured provider alias in v6.13.0
Follow up to the discussioin in #3531

### Expected Behavior

When multiple `github` provider configurations use different authentication methods, authentication should be selected per provider instance.

An instance with an explicit `app_auth {}` block should use GitHub App authentication. An alias with no `app_auth` block should continue to use its configured token / `GITHUB_TOKEN`, even when `GITHUB_APP_*` environment variables are present for the other provider instance.

This is the behavior in v6.12.1 and matches the v6.13.0 documentation:

> When using environment variables, an empty `app_auth` block is required to allow provider configurations from environment variables to be specified.

### Actual Behavior

After upgrading to v6.13.0, `GITHUB_APP_ID`, `GITHUB_APP_INSTALLATION_ID`, and `GITHUB_APP_PEM_FILE` activate App authentication for **every** GitHub provider instance in the process, including an alias that has no `app_auth` block and is intended to use `GITHUB_TOKEN`.

The PAT alias therefore ignores `GITHUB_TOKEN`. Resources assigned to that alias fail when they call user-token endpoints:

```text
GET https://api.github.com/user/installations/<redacted>/repositories?per_page=100:
403 Resource not accessible by integration []
```

All four `github_app_installation_repositories` resources assigned to the PAT alias fail this way. The same configuration worked with v6.12.1.

### Terraform Version

```text
Terraform v1.15.8
provider registry.terraform.io/integrations/github:
  works: v6.12.1
  fails: v6.13.0
```

### GitHub Installation Type

- [x] GitHub Enterprise Cloud with Managed Users/EMU (github.com)
- [ ] I don't know

### Affected Resource(s)

- `github_app_installation_repositories`
- Potentially any resource using a token-authenticated provider instance while all `GITHUB_APP_*` variables are present

This is related to the PAT-auth regression described in [#3531 comment](https://github.com/integrations/terraform-provider-github/issues/3531#issuecomment-4973755418), but differs from the original #3531 report, where App authentication was intentional. A maintainer requested a separate issue for the classic PAT case.

```hcl
terraform {
  required_providers {
    github = {
      version = "~> 6.8"
    }
  }
}

provider "github" {
  owner = "my-org"

  app_auth {
    # Populated from GITHUB_APP_* environment variables.
  }
}

provider "github" {
  owner = "my-org"
  alias = "pat"
  # No app_auth block. This instance should use GITHUB_TOKEN.
}

resource "github_app_installation_repositories" "example" {
  installation_id       = "<target-installation-id>"
  selected_repositories = ["example-repository"]
  provider              = github.pat
}
```

The CI environment contains:

```text
GITHUB_APP_ID=<set>
GITHUB_APP_INSTALLATION_ID=<set>
GITHUB_APP_PEM_FILE=<set>
GITHUB_TOKEN=<classic PAT>
```

### Steps to Reproduce

1. Configure two provider instances as above.
2. Set all three `GITHUB_APP_*` variables for the provider with `app_auth {}`, and set `GITHUB_TOKEN` for the alias without `app_auth`.
3. With provider v6.12.1, run `terraform plan`: the alias uses the PAT and refresh succeeds.
4. With provider v6.13.0, run the same plan: the alias uses App authentication and `github_app_installation_repositories` refresh fails with HTTP 403.

### Debug Output

Relevant sanitized excerpts only:

```text
- Installing integrations/github v6.13.0...

# Two GitHub provider configurations exist. Both are incorrectly configured as App auth:
provider.terraform-provider-github_v6.13.0: Using GitHub App authentication.
provider.terraform-provider-github_v6.13.0: Using GitHub App authentication.

# There is no "Using token from provider configuration" Configure message.

github_app_installation_repositories.sync_infra_app:
GET https://api.github.com/user/installations/<redacted>/repositories?per_page=100:
403 Resource not accessible by integration []

github_app_installation_repositories.local_sync_infra_app: same 403
github_app_installation_repositories.fluxcd_yaml_updater_app: same 403
github_app_installation_repositories.fluxcd_yaml_app: same 403
```

The Terraform graph in the same debug log assigns all four resources to `provider["registry.terraform.io/integrations/github"].pat`.

### Root Cause

PR [#3448](https://github.com/integrations/terraform-provider-github/pull/3448) introduced `getAppAuth`. In v6.13.0 it reads `GITHUB_APP_*` unconditionally and returns valid App credentials even when `app_auth` is absent:

```go
func getAppAuth(d *schema.ResourceData) (*string, *string, []byte, bool) {
    appID := os.Getenv("GITHUB_APP_ID")
    appInstallationID := os.Getenv("GITHUB_APP_INSTALLATION_ID")
    appPEM := os.Getenv("GITHUB_APP_PEM_FILE")

    v, ok := d.GetOk("app_auth")
    if !ok {
        return validateAppAuth(appID, appInstallationID, appPEM)
    }
    // ...
}
```

`configureProvider` then reads the token only when App auth did not activate:

```go
if appID, appInstallationID, appPEM, ok := getAppAuth(d); ok {
    config.AppID = appID
    // ...
}
if config.AppID == nil {
    // token / GITHUB_TOKEN is read only here
}
```

In v6.12.1, App auth was gated by the presence of the block:

```go
if appAuth, ok := d.Get("app_auth").([]any); ok && len(appAuth) > 0 && appAuth[0] != nil {
    // generate and select App token
}
```

Suggested behavior: return no App auth when the current provider instance has no `app_auth` block. Environment variables should populate fields inside an explicitly present block, preserving v6.12.1 behavior and the documented requirement for an empty block.

References:

- [PR #3448](https://github.com/integrations/terraform-provider-github/pull/3448)
- [v6.13.0 `getAppAuth`](https://github.com/integrations/terraform-provider-github/blob/v6.13.0/github/provider.go#L668-L687)
- [v6.13.0 auth selection](https://github.com/integrations/terraform-provider-github/blob/v6.13.0/github/provider.go#L398-L416)
- [v6.12.1 block-gated behavior](https://github.com/integrations/terraform-provider-github/blob/v6.12.1/github/provider.go#L352-L426)
- [v6.13.0 documentation](https://github.com/integrations/terraform-provider-github/blob/v6.13.0/docs/index.md#L64-L97)

### Code of Conduct

stevehipwell:
Resolves #3576

----

### Before the change?

- The provider auto detects which authentication mode to use
- Env variables can't be used when configuring multiple providers

### After the change?

- The provider authentication mode can be explicitly set
- App auth env variables can be targeted to a single provider instance

### Pull request checklist

- [ ] Yes
- [x] No

----


github/provider.go
```diff
@@ -397,30 +410,38 @@ func configureProvider(version, commit string) func(context.Context, *schema.Res
 			}
 		}
 
-		if appID, appInstallationID, appPEM, ok := getAppAuth(d); ok {
-			tflog.Debug(ctx, "Using GitHub App authentication.", map[string]any{"app_id": appID, "app_installation_id": appInstallationID})
-			config.AppID = appID
-			config.AppInstallationID = appInstallationID
-			config.AppPEM = appPEM
+		if authMode == "app" || authMode == "auto" {
+			if appID, appInstallationID, appPEM, ok := getAppAuth(d, appAuthEnvPrefix); ok {
```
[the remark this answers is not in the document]
stevehipwell: App auth has a higher priority than token auth, this is the expected behaviour.

github/provider.go
```diff
@@ -53,13 +53,26 @@ func NewProvider(version, commit string) func() *schema.Provider {
...
 					DefaultFunc: schema.EnvDefaultFunc("GITHUB_TOKEN", nil),
 					Description: "GitHub OAuth or Personal Access Token (PAT) to use for authentication. This can also be set by the `GITHUB_TOKEN` environment variable.",
 					// ConflictsWith: []string{"app_auth"}, // TODO: Enable as part of v7.
 				},
+				"app_auth_env_prefix": {
+					Type:        schema.TypeString,
+					Optional:    true,
+					Default:     "GITHUB_APP_",
+					Description: "The environment variable prefix for the GitHub App authentication used to determine the environment variable names for the GitHub App's ID, installation ID, and PEM file content. This defaults to `GITHUB_APP_`.",
+				},
```
deiga, line 79: suggestion: Since token auth could be used in multiple providers with different tokens, would it make sense to change this to `auth_env_prefix`?  Or have separate attribute for `token_auth_env_prefix`?

deiga (commented):
Nice!

github/provider.go
stevehipwell: Yeah, I was thinking the exact same thing over the weekend. Incoming.

deiga (approved):
Sweet!

robert-crandall (commented):
Thanks for tackling this, Steve, the explicit `auth_mode` is a really nice shape for the multi-provider case. Can you verify that the config from #3576 works unchanged on this branch? When I tried it, the PAT alias (no `app_auth` block, `GITHUB_APP_*` and `GITHUB_TOKEN` both in the environment) still authenticated as the App rather than with the PAT. My read is that `auto` mode calls `getAppAuth`, which picks up `GITHUB_APP_*` even for an instance with no `app_auth` block, so the reporter would still need to add `auth_mode = "token"` to get v6.12.1 behaviour back. Would gating the env-only App lookup on the block being present in `auto` mode (and keeping env-only for explicit `auth_mode = "app"`) work here?

stevehipwell:
> Thanks for tackling this, Steve, the explicit auth_mode is a really nice shape for the multi-provider case. Can you verify that the config from https://github.com/integrations/terraform-provider-github/issues/3576 works unchanged on this branch? When I tried it, the PAT alias (no app_auth block, GITHUB_APP_* and GITHUB_TOKEN both in the environment) still authenticated as the App rather than with the PAT. My read is that auto mode calls getAppAuth, which picks up GITHUB_APP_* even for an instance with no app_auth block, so the reporter would still need to add auth_mode = "token" to get v6.12.1 behaviour back. Would gating the env-only App lookup on the block being present in auto mode (and keeping env-only for explicit auth_mode = "app") work here?

@robert-crandall this is the expected behaviour, reverting the behaviour in `v6.13.0` would now also be "breaking" and it'd make the both the behaviour and code much more complex. This has been communicated in https://github.com/integrations/terraform-provider-github/issues/3576#issuecomment-5067773914.

robert-crandall (approved):

PR #3582, opened 2026-07-28
feat: support use_immutable_subject and sub_claim_prefix on repository OIDC subject claim template

### LINKED ISSUE #3548: Support use_immutable_subject on github_actions_repository_oidc_subject_claim_customization_template
## Problem

As of July 15, 2026, GitHub automatically uses immutable OIDC subject claims for newly created repositories (and repos that are renamed/transferred). The `sub` claim now includes both the owner ID and repository ID:

```
repo:my-org@123456/my-repo@789012:ref:refs/heads/main
```

GitHub also provides a **per-repository opt-in** via the REST API field `use_immutable_subject` on `PUT /repos/{owner}/{repo}/actions/oidc/customization/sub`:

```json
{
  "use_default": false,
  "include_claim_keys": ["repo"],
  "use_immutable_subject": true
}
```

This allows repositories created **before** July 15, 2026 to adopt the immutable format without an org-wide change.

The underlying Go SDK (`google/go-github` v89) already carries both `UseImmutableSubject` and `SubClaimPrefix` on the `OIDCSubjectClaimCustomTemplate` struct:

```go
type OIDCSubjectClaimCustomTemplate struct {
    UseDefault          *bool    `json:"use_default,omitempty"`
    IncludeClaimKeys    []string `json:"include_claim_keys,omitempty"`
    UseImmutableSubject *bool    `json:"use_immutable_subject,omitempty"`
    SubClaimPrefix      *string  `json:"sub_claim_prefix,omitempty"`
}
```

However, the Terraform resource `github_actions_repository_oidc_subject_claim_customization_template` only exposes `repository`, `use_default`, and `include_claim_keys` — the `use_immutable_subject` and `sub_claim_prefix` fields are not wired through.

## Impact

Organizations that:
- Use Azure (or other cloud) federated identity credentials matching the OIDC `sub` claim
- Have a **mixed fleet** of repositories (some created before July 15, some after)
- Cannot enable immutable subjects org-wide (shared org with other teams)

...have no way to **per-repo opt in** to immutable subjects via Terraform. This forces them to either:
- Use out-of-band REST API calls (losing drift detection and state management)
- Switch to non-human-readable ID-only claims (`repository_id`) to get a uniform subject format

Both are workarounds for a missing Terraform capability.

## Proposed Solution

Add two optional attributes to the `github_actions_repository_oidc_subject_claim_customization_template` resource:

```hcl
resource "github_actions_repository_oidc_subject_claim_customization_template" "example" {
  repository   = "my-repo"
  use_default  = false

  include_claim_keys = ["repo"]

  # NEW: opt this repo into immutable subject claims
  use_immutable_subject = true

  # NEW: optional prefix for the sub claim
  sub_claim_prefix = "custom-prefix"
}
```

### Implementation notes

The resource's `CreateOrUpdate` function already builds an `OIDCSubjectClaimCustomTemplate` struct and calls `SetRepoOIDCSubjectClaimCustomTemplate`. The change is:
1. Add `use_immutable_subject` (`TypeBool`, `Optional`, `Computed`) to the schema
2. Add `sub_claim_prefix` (`TypeString`, `Optional`, `Computed`) to the schema  
3. Set both fields on the struct in `CreateOrUpdate` when provided
4. Read both fields back in `Read`

This should be ~12-15 lines of code change in the resource file.

## References

- [GitHub OIDC docs — Immutable subject claims](https://docs.github.com/en/actions/reference/security/oidc#immutable-subject-claims)
- [REST API — Set customization template for a repository](https://docs.github.com/en/rest/actions/oidc#set-the-customization-template-for-an-oidc-subject-claim-for-a-repository)
- Related SDK: `google/go-github` v89 `OIDCSubjectClaimCustomTemplate` struct in `github/actions_oidc.go`

## Willingness to contribute

I'm happy to submit a PR for this if the maintainers agree with the approach.

madriaanse377:
## Summary

Wires `use_immutable_subject` and `sub_claim_prefix` — already present on go-github v89's `OIDCSubjectClaimCustomTemplate` struct — through to `github_actions_repository_oidc_subject_claim_customization_template`.

This lets a repository created before GitHub's July 15, 2026 immutable-subject-claims rollout opt in individually via Terraform, instead of requiring an org-wide toggle or an out-of-band REST call that Terraform can't track.

## Changes

- `use_immutable_subject` (bool, optional, computed) and `sub_claim_prefix` (string, optional, computed) added to the resource schema
- Both fields wired into `CreateOrUpdate` and `Read`, following the existing `GetOk` + `new(...)` pattern already used elsewhere in this resource (`include_claim_keys`) and in sibling resources (e.g. `resource_github_enterprise_actions_workflow_permissions.go`)
- Docs (`docs/resources/...md` and its `.tmpl` source) and the example under `examples/resources/...` updated to cover both new arguments
- New acceptance test subtest asserting both fields round-trip correctly

Scoped to the repository-level resource only, matching the issue — the organization-level resource and this resource's `Delete` behavior are unchanged.

Closes #3548

## Test plan

- [ ] Maintainer-run acceptance suite (`TestAccGithubActionsRepositoryOIDCSubjectClaimCustomizationTemplate`), since these require a live GitHub token this environment doesn't have
- [x] Manual diff review against the existing `use_default`/`include_claim_keys` and `can_approve_pull_request_reviews` patterns for consistency

### DIFF
--- a/docs/resources/actions_repository_oidc_subject_claim_customization_template.md
+++ b/docs/resources/actions_repository_oidc_subject_claim_customization_template.md
@@ -29,6 +29,11 @@ resource "github_actions_repository_oidc_subject_claim_customization_template" "
   repository         = github_repository.example.name
   use_default        = false
   include_claim_keys = ["actor", "context", "repository_owner"]
+
+  # Opt this repository into immutable subject claims (owner and repository IDs)
+  # ahead of an org-wide rollout.
+  use_immutable_subject = true
+  sub_claim_prefix      = "custom-prefix"
 }
 ```
 
@@ -38,6 +43,8 @@ The following arguments are supported:
 
 - `use_default` - (Required) Whether to use the default template or not. If `true`, `include_claim_keys` must not be set.
 - `include_claim_keys` - (Optional) A list of OpenID Connect claims.
+- `use_immutable_subject` - (Optional) Whether to use immutable subject claims, which include the immutable repository and owner IDs, for this repository.
+- `sub_claim_prefix` - (Optional) An optional prefix to add to the repository's subject claim.
 
 ## Import
 
--- a/examples/resources/actions_repository_oidc_subject_claim_customization_template/example_1.tf
+++ b/examples/resources/actions_repository_oidc_subject_claim_customization_template/example_1.tf
@@ -6,4 +6,9 @@ resource "github_actions_repository_oidc_subject_claim_customization_template" "
   repository         = github_repository.example.name
   use_default        = false
   include_claim_keys = ["actor", "context", "repository_owner"]
+
+  # Opt this repository into immutable subject claims (owner and repository IDs)
+  # ahead of an org-wide rollout.
+  use_immutable_subject = true
+  sub_claim_prefix      = "custom-prefix"
 }
--- a/github/resource_github_actions_repository_oidc_subject_claim_customization_template.go
+++ b/github/resource_github_actions_repository_oidc_subject_claim_customization_template.go
@@ -39,6 +39,18 @@ func resourceGithubActionsRepositoryOIDCSubjectClaimCustomizationTemplate() *sch
 					Type: schema.TypeString,
 				},
 			},
+			"use_immutable_subject": {
+				Type:        schema.TypeBool,
+				Optional:    true,
+				Computed:    true,
+				Description: "Whether to use immutable subject claims, which include the immutable repository and owner IDs, for this repository.",
+			},
+			"sub_claim_prefix": {
+				Type:        schema.TypeString,
+				Optional:    true,
+				Computed:    true,
+				Description: "An optional prefix to add to the repository's subject claim.",
+			},
 		},
 	}
 }
@@ -73,6 +85,14 @@ func resourceGithubActionsRepositoryOIDCSubjectClaimCustomizationTemplateCreateO
 		customOIDCSubjectClaimTemplate.IncludeClaimKeys = claimsStr
 	}
 
+	if v, ok := d.GetOk("use_immutable_subject"); ok {
+		customOIDCSubjectClaimTemplate.UseImmutableSubject = new(v.(bool))
+	}
+
+	if v, ok := d.GetOk("sub_claim_prefix"); ok {
+		customOIDCSubjectClaimTemplate.SubClaimPrefix = new(v.(string))
+	}
+
 	ctx := context.Background()
 	_, err := client.Actions.SetRepoOIDCSubjectClaimCustomTemplate(ctx, owner, repository, customOIDCSubjectClaimTemplate)
 	if err != nil {
@@ -102,6 +122,12 @@ func resourceGithubActionsRepositoryOIDCSubjectClaimCustomizationTemplateRead(d 
 		return err
 	}
 	if err = d.Set("include_claim_keys", template.IncludeClaimKeys); err != nil {
+		return err
+	}
+	if err = d.Set("use_immutable_subject", template.UseImmutableSubject); err != nil {
+		return err
+	}
+	if err = d.Set("sub_claim_prefix", template.SubClaimPrefix); err != nil {
 		return err
 	}
 
--- a/github/resource_github_actions_repository_oidc_subject_claim_customization_template_test.go
+++ b/github/resource_github_actions_repository_oidc_subject_claim_customization_template_test.go
@@ -61,6 +61,45 @@ func TestAccGithubActionsRepositoryOIDCSubjectClaimCustomizationTemplate(t *test
 		})
 	})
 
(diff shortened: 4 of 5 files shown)


RulerOf:
Just an FYI, I got bit by this today trying to resolve an issue with a new microservice's OIDC config. The API publishes `use_immutable_subject` but that value seems to lie. 

In this older repo, the UI and API agree:

Google Chrome 2026-08-04 at 17 03 53

```json
╰─❯ gh api /repos/BuyerQuest/redacted-old-repo/actions/oidc/customization/sub
{
  "use_default": true,
  "use_immutable_subject": false,
  "sub_claim_prefix": "repo:BuyerQuest/redacted-old-repo"
}
```

But this newer repo created after the compulsive switchover they disagree entirely:

Google Chrome 2026-08-04 at 16 57 22

```json
gh api /repos/BuyerQuest/redacted-new-repo/actions/oidc/customization/sub
{
  "use_default": true,
  "use_immutable_subject": false,
  "sub_claim_prefix": "repo:BuyerQuest@123445689/redacted-new-repo@987654321"
}
```

I bring this up because what I _wanted_ to do was inspect the true/false of `use_immutable_subject` from the data source to construct the sub, but it doesn't appear to be reliable via the API.

I'm using this as a workaround:
```hcl
data "github_repository" "this" {
  full_name = "${local.github_org}/${var.github_repository}"
}

data "github_rest_api" "repo_oidc" {
  endpoint = "repos/${data.github_repository.this.full_name}/actions/oidc/customization/sub"
}

module "github_actions_role" {
  source  = "philips-labs/github-oidc/aws"
  version = "0.8.1"

  openid_connect_provider_arn = "arn:aws:iam::${data.aws_caller_identity.current.account_id}:oidc-provider/token.actions.githubusercontent.com"
  repo                        = data.github_repository.this.full_name
  role_name                   = "${local.name_prefix}-deploy"
  role_path                   = "/github-actions/"

  default_conditions = [
    "deny_pull_request",
  ]

  // GitHub reports the exact prefix it uses for this repository. This handles both the
  // name-based and immutable name@id formats without duplicating GitHub's selection logic.
  conditions = [
    {
      test     = "StringEquals"
      variable = "token.actions.githubusercontent.com:sub"
      values = [
        "${jsondecode(data.github_rest_api.repo_oidc.body).sub_claim_prefix}:environment:${github_repository_environment.this.environment}",
      ]
    },
  ]
  ```
  
Unless I'm mistaken, for the data source it may be worth computing it based on the qualities of `sub_claim_prefix` (e.g. an `@` symbol in it anywhere means it should be `true`), or possibly disabling it until github fixes whatever the problem is.

**Comment by dekokun** [comment]
Not a maintainer, just helping with review triage, so this is a comment rather than a review.

Building on what @RulerOf found above: I checked the write side too, and `sub_claim_prefix` isn't just unreliable to read, it can't be set at all. It isn't a documented body parameter for `PUT /repos/{owner}/{repo}/actions/oidc/customization/sub` (it only appears in the GET response), and sending it is silently ignored. On a throwaway repo:

```
$ gh api -X PUT repos/OWNER/REPO/actions/oidc/customization/sub \
    --input - <<< '{"use_default": false, "include_claim_keys": ["repo"], "sub_claim_prefix": "test-prefix"}'
{}

$ gh api repos/OWNER/REPO/actions/oidc/customization/sub
{"use_default": false, "use_immutable_subject": false, "include_claim_keys": ["repo"],
 "sub_claim_prefix": "repo:OWNER@108519/REPO@1356554838"}
```

Same derived value before and after, exactly as in @RulerOf's examples.

That has two consequences for this PR specifically. The attribute is `Optional` + `Computed`, so a config value of `"custom-prefix"` against a read-back of the derived value should give "Provider produced inconsistent result after apply" - which means the example added to the docs wouldn't work as written. And the new acceptance subtest asserts `sub_claim_prefix` equals `"custom-prefix"`, so I'd expect it to fail when someone runs it. Computed-only looks like the right shape for this field.

Separately, `use_immutable_subject = false` never reaches the API, because `d.GetOk` can't tell an explicit `false` from an unset bool. With a scratch test against a mocked API and `use_immutable_subject = false` in the config, the PUT body is:

```
{"use_default":false,"include_claim_keys":["repo"]}
```

The field is dropped. This matters because opting back out does work at the API level: on the same throwaway repo I set it to `true` and then back to `false`, and both took effect. Since the attribute is also `Computed`, the Read after apply writes `true` back into state, so someone trying to opt out via Terraform gets a silent no-op and a permanent diff.

There's an existing idiom for this in the repo: `d.GetOkExists` with a nolint comment, e.g. `resource_github_repository_pages.go:193` ("necessary for bool fields") and `resource_github_repository.go:636/645/1001`.

I haven't run the acceptance tests.

Issue #1967, opened 2023-10-17
[FEAT]: Add `prevent_self_review` boolean to the `github_repository_environment` resource


nnellanspdl:
### Describe the need

Requesting an update to the [github_repository_environment](https://registry.terraform.io/providers/integrations/github/latest/docs/resources/repository_environment) resource to support the `prevent_self_review` boolean as defined in the [API here](https://docs.github.com/en/rest/deployments/environments?apiVersion=2022-11-28#create-or-update-an-environment)

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

Nmishin:
going to work on that

PR #3583, opened 2026-07-28
fix: prevent inconsistent final plan for `prevent_self_review` in `github_repository_environment`

kishaningithub:
Fixes an issue where Terraform reports "Provider produced inconsistent final plan" for the `prevent_self_review` attribute on `github_repository_environment` resources that have no reviewers configured.

Resolves a missed edge case in  https://github.com/integrations/terraform-provider-github/issues/1967

### Problem

When an environment has no reviewers, the GitHub API does not return a `required_reviewers` protection rule. The read function only set `prevent_self_review` inside the `case "required_reviewers"` branch, so the attribute was never written to state—leaving it as `null`. Terraform's plan expected `false` (the schema default), causing the following mismatch:

Error: Provider produced inconsistent final plan .prevent_self_review: was cty.False, but now null.

### Root Cause

1. `prevent_self_review` was only set in state when a `required_reviewers` protection rule existed in the API response.
2. `pr.PreventSelfReview` (a `*bool`) was passed directly to `d.Set()` without nil-checking, which could also store `null` in state.

### Fix

- Set `prevent_self_review` to `false` before iterating protection rules, ensuring it always has a value in state.
- Safely dereference `pr.PreventSelfReview` with a nil guard, defaulting to `false`.

### Testing

Environments without reviewers (the failing case) are already exercised by the `create_with_id_separator_in_name` and `update_to_add_reviewers` acceptance tests which create environments with no initial reviewers.

### Pull request checklist

- [ ] Yes
- [x] No

----

### DIFF
--- a/github/resource_github_repository_environment.go
+++ b/github/resource_github_repository_environment.go
@@ -270,11 +270,7 @@ func resourceGithubRepositoryEnvironmentRead(ctx context.Context, d *schema.Reso
 				return diag.FromErr(err)
 			}
 
-			preventSelfReview := false
-			if pr.PreventSelfReview != nil {
-				preventSelfReview = *pr.PreventSelfReview
-			}
-			if err = d.Set("prevent_self_review", preventSelfReview); err != nil {
+			if err = d.Set("prevent_self_review", pr.GetPreventSelfReview()); err != nil {
 				return diag.FromErr(err)
 			}
 		}
--- a/github/resource_github_repository_environment.go
+++ b/github/resource_github_repository_environment.go
@@ -232,6 +232,12 @@ func resourceGithubRepositoryEnvironmentRead(ctx context.Context, d *schema.Reso
 		return diag.FromErr(err)
 	}
 
+	// Default prevent_self_review to false; it will be overwritten if the
+	// required_reviewers protection rule is present in the API response.
+	if err := d.Set("prevent_self_review", false); err != nil {
+		return diag.FromErr(err)
+	}
+
 	for _, pr := range env.ProtectionRules {
 		switch pr.GetType() {
 		case "wait_timer":
@@ -264,7 +270,11 @@ func resourceGithubRepositoryEnvironmentRead(ctx context.Context, d *schema.Reso
 				return diag.FromErr(err)
 			}
 
-			if err = d.Set("prevent_self_review", pr.PreventSelfReview); err != nil {
+			preventSelfReview := false
+			if pr.PreventSelfReview != nil {
+				preventSelfReview = *pr.PreventSelfReview
+			}
+			if err = d.Set("prevent_self_review", preventSelfReview); err != nil {
 				return diag.FromErr(err)
 			}
 		}
--- a/github/resource_github_repository_environment.go
+++ b/github/resource_github_repository_environment.go
@@ -269,11 +269,7 @@ func resourceGithubRepositoryEnvironmentRead(ctx context.Context, d *schema.Reso
 				return diag.FromErr(err)
 			}
 
-			preventSelfReview := false
-			if pr.PreventSelfReview != nil {
-				preventSelfReview = *pr.PreventSelfReview
-			}
-			if err = d.Set("prevent_self_review", preventSelfReview); err != nil {
+			if err = d.Set("prevent_self_review", pr.GetPreventSelfReview()); err != nil {
 				return diag.FromErr(err)
 			}
 		}
--- a/github/resource_github_repository_environment.go
+++ b/github/resource_github_repository_environment.go
@@ -231,6 +231,12 @@ func resourceGithubRepositoryEnvironmentRead(ctx context.Context, d *schema.Reso
 		return diag.FromErr(err)
 	}
 
+	// Default prevent_self_review to false; it will be overwritten if the
+	// required_reviewers protection rule is present in the API response.
+	if err := d.Set("prevent_self_review", false); err != nil {
+		return diag.FromErr(err)
+	}
+
 	for _, pr := range env.ProtectionRules {
 		switch pr.GetType() {
 		case "wait_timer":
@@ -263,7 +269,11 @@ func resourceGithubRepositoryEnvironmentRead(ctx context.Context, d *schema.Reso
 				return diag.FromErr(err)
 			}
 
-			if err = d.Set("prevent_self_review", pr.PreventSelfReview); err != nil {
+			preventSelfReview := false
+			if pr.PreventSelfReview != nil {
+				preventSelfReview = *pr.PreventSelfReview
+			}
+			if err = d.Set("prevent_self_review", preventSelfReview); err != nil {
 				return diag.FromErr(err)
 			}
 		}
--- a/github/resource_github_repository_environment_test.go
+++ b/github/resource_github_repository_environment_test.go
@@ -284,23 +284,18 @@ resource "github_repository_environment" "test" {
 	t.Run("prevent_self_review_defaults_false_without_reviewers", func(t *testing.T) {
 		t.Parallel()
 
-		randomID := acctest.RandStringFromCharSet(5, acctest.CharSetAlphaNum)
-		repoName := fmt.Sprintf("%s%s", testResourcePrefix, randomID)
+		skipUnlessHasOrgs(t)
+
+		repo := mustCreateTestRepository(t)
 
 		config := fmt.Sprintf(`
-resource "github_repository" "test" {
-	name       = "%s"
-	visibility = "public"
-}
-
 resource "github_repository_environment" "test" {
-	repository  = github_repository.test.name
+	repository  = "%s"
 	environment = "test"
 }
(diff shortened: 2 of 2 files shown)


deiga:
Nice, thanks! Could you add a regression acceptance test case as well?

kishaningithub:
@deiga Have added the test. Can you kindly review?

github/resource_github_repository_environment.go
```diff
@@ -264,7 +270,11 @@ func resourceGithubRepositoryEnvironmentRead(ctx context.Context, d *schema.Reso
 				return diag.FromErr(err)
 			}
 
-			if err = d.Set("prevent_self_review", pr.PreventSelfReview); err != nil {
+			preventSelfReview := false
+			if pr.PreventSelfReview != nil {
+				preventSelfReview = *pr.PreventSelfReview
+			}
```
deiga: I think this might be replaceable by using pr.GetPreventSelfReview()

github/resource_github_repository_environment_test.go
```diff
@@ -281,6 +281,38 @@ resource "github_repository_environment" "test" {
...
+		t.Parallel()
+
+		randomID := acctest.RandStringFromCharSet(5, acctest.CharSetAlphaNum)
+		repoName := fmt.Sprintf("%s%s", testResourcePrefix, randomID)
+
+		config := fmt.Sprintf(`
+resource "github_repository" "test" {
+	name       = "%s"
+	visibility = "public"
+}
```
deiga: Please use mustCreateTestRepo instead

deiga (changes requested):

github/resource_github_repository_environment.go
kishaningithub: Have addressed this feedback @deiga

github/resource_github_repository_environment_test.go
kishaningithub: Have addressed this feedback @deiga

kishaningithub:
@deiga Have resolved your PR review feedback. Can you kindly check once?

deiga:
@kishaningithub I ran your test without any other code changes and the test doesn't fail. Can you please provide DEBUG level logs of the issue you are trying to resolve here?

kishaningithub:
@deiga The issue was with the test

The original test did not reproduce the issue because it asserted the value immediately after creation. At that point, the schema default had already populated `prevent_self_review = false`, so the test passed even without the read-path fix.

The failing scenario is a refresh/read from imported or legacy state where `prevent_self_review` is absent and the GitHub API returns no `required_reviewers` protection rule.

I replaced the test with a read-path regression test that starts with the attribute absent from state and mocks that API response. 

The corrected test fails without the production change and passes with it. I also removed `prevent_self_review` from `ImportStateVerifyIgnore` so the import test verifies this state value.

github/resource_github_repository_environment_test.go
```diff
@@ -15,6 +18,93 @@ import (
...
+				responseBody:  `{"name":"test","protection_rules":[{"type":"required_reviewers","prevent_self_review":true,"reviewers":[]}]}`,
+				expectedValue: "true",
+			},
+		}
+
+		for _, test := range tests {
+			t.Run(test.name, func(t *testing.T) {
+				t.Parallel()
+
+				ts := githubApiMock([]*mockResponse{
```
deiga: issue: no mocked tests inside TestAcc functions. 

We need an acceptance test against the real API to verify the behaviour you mention.

deiga (changes requested):

github/resource_github_repository_environment_test.go
kishaningithub: @deiga The code now much simpler. The acceptance test "import_without_reviewers" fails when the fix is not present

github/resource_github_repository_environment.go
```diff
@@ -232,6 +232,12 @@ func resourceGithubRepositoryEnvironmentRead(ctx context.Context, d *schema.Reso
 		return diag.FromErr(err)
 	}
 
+	// Default prevent_self_review to false; it will be overwritten if the
+	// required_reviewers protection rule is present in the API response.
+	if err := d.Set("prevent_self_review", false); err != nil {
+		return diag.FromErr(err)
+	}
+
```
deiga, line 239: issue: this is the wrong way of handling this. If prevent_self_review gets into the wrong state at import, then the import function needs to set it and not a double set in Read

deiga (changes requested):

github/resource_github_repository_environment.go
kishaningithub: Understood.. will figure out the exact steps which led to that state and ideate a fix..

kishaningithub:
@deiga  This is the exact scenario.. I have a state which is like this

## Result of `terraform show` (redacted)

```terraform
resource "github_repository_environment" "repo_environment" {
    can_admins_bypass = true
    environment       = "dev_bot"
    id                = "example-repo:dev_bot"
    repository        = "example-repo"
    repository_id     = 123456789
    wait_timer        = 0

    reviewers {}
}
```

## Result of `terraform plan` (redacted)

Then we have a plan run where `prevent_self_review` is being set explicitly

```terraform
~ resource "github_repository_environment" "repo_environment" {
      id                  = "example-repo:dev_bot"
    + prevent_self_review = false
      # (5 unchanged attributes hidden)
    ~ reviewers {
        + teams = (known after apply)
      }
  }
```

## Result of `terraform apply (redacted)

The apply fails with the below error

```bash
Error: Provider produced inconsistent final plan

When expanding the plan for
github_repository_environment.repo_environment["dev_bot"]
to include new values learned so far during apply, provider
"registry.terraform.io/integrations/github" produced an invalid new value for
.prevent_self_review: was cty.False, but now null.

issue tracker.
```

kishaningithub:
@deiga Given the above context. Which route do you think i must take for the fix? Any suggestions?

deiga:
@kishaningithub Please provide a testcase that reproduces your error-case and/or DEBUG level logs if the apply run

deiga:
I think this PR would/should resolve #3609 as well

deiga:
@kishaningithub I tried creating a test case which follows your example, but I couldn't get it to fail like that

kishaningithub:
@deiga Yes I tried it too.. But could not make it fail.. We need to simulate it like this somehow in the test case (which I am trying)

Use provider version 6.8.3, Create the github_repository_environment without reviewers and prevent_self_review 

Upgrade to latest provider version setting the legacy_client provider attribute to false,Then set prevent_self_review setting it to false for this resource and run a terraform plan and apply

That's my exact production scenario

deiga:
@kishaningithub Ah! This is about a migration between provider versions? 
That makes this clearer. 
Writing a test for that is slightly harder, we don't have a working example of that currently

kishaningithub:
Yes @deiga We hit this issue when upgrading the provider version from 6.8.3 to latest version disabling the legacy client

deiga:
@kishaningithub Have you tried pinpointing which version upgrade causes this?

kishaningithub:
@deiga Thats the challenge. Putting here the exact state file entries (redacted) causing this issue for reference

```json
 {
  "index_key": "example_environment",
  "schema_version": 1,
  "attributes": {
    "can_admins_bypass": true,
    "deployment_branch_policy": [],
    "environment": "example_environment",
    "id": "example-repo:example_environment",
    "prevent_self_review": null,
    "repository": "example-repo",
    "repository_id": 123456789,
    "reviewers": [
      {
        "teams": null,
        "users": null
      }
    ],
    "wait_timer": 0
  },
  "sensitive_attributes": [],
}
```

deiga:
@kishaningithub Thanks!

I wonder how that state was even achieved, since the default has been "false" for years. 

But a plan/apply with 6.8.3 does not have those issue? Can you test if it already happens on 6.9 or 6.10?

kishaningithub:
[For issue followers] In case you are wondering how to work around this issue without directly editing the statefile here is a possible solution (this is what i did)

- Download the statefile json. Find out the affected repos by searching for `"prevent_self_review": null` in the file the line above that contains the `id` from which you can extract the repository name
- Delete the affected environements either via UI or via curl
```bash
curl -L \
  -X DELETE \
  -H "Authorization: Bearer YOUR_GITHUB_TOKEN" \
  https://api.github.com/repos/OWNER/REPO/environments/ENVIRONMENT_NAME
```
- Run a targetted `terraform plan` and `terraform apply` on the affected resources

kishaningithub:
@deiga IMO this i think predates much longer.. the affected resources are the ones that have been around since 2023 and over time multiple version bumps have happened.

The one where apply failed was this version (`6.13.0`) which is very wierd

kishaningithub:
As per the git history these are the versions.. one of the upgrades caused this.. will try to simulate via acceptance test

```bash
5.0
5.32.0
6.6.0
6.8.1
6.8.3
6.9.0
6.13.0
```

PR #3585, opened 2026-07-29
fix(branch_protection_v3): send each required status check only once

### LINKED ISSUE #3420: [BUG]: github_branch_protection_v3 imports both checks and contexts
### Expected Behavior

Users should be able to import a github_branch_protection_v3 resource and then modify the required_status_checks. 

### Actual Behavior

When you import a github_branch_protection_v3 resource and then try to modify the required_status_checks, you get the following error:
```
PUT https://api.github.com/repos/{orgname}/branch-protection-test/branches/main/protection: 422 Validation Failed [{Resource: Field: Code: Message:Context must be unique per branch protection.}]
```

I did some investigation. It appears that both checks and contexts get imported into the state. This is because the  [endpoint](https://docs.github.com/en/rest/branches/branch-protection?apiVersion=2026-03-10#get-branch-protection) returns both checks and contexts, but they contain the same checks. Both checks and contexts get [combined into one list when doing an update](https://github.com/integrations/terraform-provider-github/blob/main/github/resource_github_branch_protection_v3_utils.go#L267-L308). This means that each existing check gets sent twice to the [endpoint](https://docs.github.com/en/rest/branches/branch-protection?apiVersion=2026-03-10#update-branch-protection) and that causes the 422 response.

### Terraform Version

Terraform v1.15.2
on darwin_arm64

### Affected Resource(s)

- github_branch_protection_v3

```hcl
resource "github_branch_protection_v3" "main" {
  repository  = "branch-protection-test"
  branch      = "main"
  required_status_checks {
    checks = [
      "check_types:15368",
      "lint:15368",
      "test:15368",
    ]
  }
}
```

### Steps to Reproduce

```
$ terragrunt import github_branch_protection_v3.main branch-protection-test:main
$ terraform apply
```

### Debug Output

```shell
❯ terraform import github_branch_protection_v3.main branch-protection-test:main
github_branch_protection_v3.main: Importing from ID "branch-protection-test:main"...
github_branch_protection_v3.main: Import prepared!
  Prepared github_branch_protection_v3 for import
github_branch_protection_v3.main: Refreshing state... [id=branch-protection-test:main]

Import successful!

The resources that were imported are shown above. These resources are now in
your Terraform state and will henceforth be managed by Terraform.

❯ terraform state show github_branch_protection_v3.main
# github_branch_protection_v3.main:
resource "github_branch_protection_v3" "main" {
    branch                          = "main"
    enforce_admins                  = false
    id                              = "branch-protection-test:main"
    repository                      = "branch-protection-test"
    require_conversation_resolution = false
    require_signed_commits          = false

    required_status_checks {
        checks         = [
            "check_types:15368",
            "lint:15368",
        ]
        contexts       = [
            "check_types",
            "lint",
        ]
        include_admins = false
        strict         = false
    }
}

❯ terraform apply
github_branch_protection_v3.main: Refreshing state... [id=branch-protection-test:main]

  ~ update in-place

  # github_branch_protection_v3.main will be updated in-place
  ~ resource "github_branch_protection_v3" "main" {
        id                              = "branch-protection-test:main"

      ~ required_status_checks {
          ~ checks         = [
              + "test:15368",
                # (2 unchanged elements hidden)
            ]
        }
    }

Do you want to perform these actions?
  Terraform will perform the actions described above.
  Only 'yes' will be accepted to approve.

  Enter a value: yes

github_branch_protection_v3.main: Modifying... [id=branch-protection-test:main]
╷
│ Error: PUT https://api.github.com/repos/{orgname}/branch-protection-test/branches/main/protection: 422 Validation Failed [{Resource: Field: Code: Message:Context must be unique per branch protection.}]
│
│   with github_branch_protection_v3.main,
│   on main.tf line 1, in resource "github_branch_protection_v3" "main":
│    1: resource "github_branch_protection_v3" "main" {
│
╵
```

### Panic Output

```shell

```

### Code of Conduct

dekokun:
Resolves #3420

----

### Before the change?

Updating a `github_branch_protection_v3` resource that has non-empty `required_status_checks` fails:

```
Error: PUT https://api.github.com/repos/<org>/<repo>/branches/main/protection: 422 Validation Failed
[{Resource: Field: Code: Message:Context must be unique per branch protection.}]
```

The branch protection API returns every required check under both `contexts` and `checks`, so `flattenAndSetRequiredStatusChecks` puts the same set into both state fields. Both are `Optional + Computed`, so the field that is not configured keeps its state value forever, and `expandRequiredStatusChecks` merges both fields into one `checks` array — sending each context twice.

Configuration cannot hit this on its own (`checks` conflicts with `contexts`); the duplicate always comes from state. `import` is one way to get there (as in #3420), but so is a plain `apply`: `Create` only sees the configured field and succeeds, and the read that follows it populates both. From then on **any** update to that resource fails, even when the change has nothing to do with status checks — I hit this in production on a change that only flipped `require_code_owner_reviews`, on resources this provider had created itself.

Note on #2232, which proposed a similar fix and was closed as unnecessary because both fields would be made `Computed: true`: both fields already are `Optional + Computed` in v6.13.0 and the bug is still there. The schema is not what produces the duplicate — the read path writing both fields plus the write path merging them is.

### After the change?

Requests carry each context once. When a context is present in both fields, an `app_id` that came from `checks` wins, since `contexts` cannot express one.

A check read back as `app_id: null` allows any app, which is `-1` on write, so the deduplicated check keeps `app_id: -1` instead of omitting the field — [omitting it](https://docs.github.com/en/rest/branches/branch-protection#update-branch-protection) makes GitHub select the app that recently provided the check, which would silently pin an app that was not pinned before.

Tests: a unit test on `expandRequiredStatusChecks` for the state shapes above, and an acceptance case that updates an unrelated setting after the first apply (the existing no-churn tests only re-plan the same config, so they never perform an update).

I ran the acceptance case against a real organization in `organization` mode. With only the fix reverted it fails at step 2 with the 422 above, for both the `contexts` and the `checks` variant; with the fix in place both pass. `make test`, `golangci-lint run ./...` and `make lintcheck-new` are clean.

Per the AI use policy: the patch and this description were drafted with an AI coding assistant. I reviewed the change and ran every test and lint run described above myself.

### Pull request checklist

- [ ] Yes
- [x] No

----


github/resource_github_branch_protection_v3_test.go
```diff
@@ -360,6 +360,66 @@ func TestAccGithubBranchProtectionV3_computed_status_contexts_no_churn(t *testin
...
+			t.Parallel()
+
+			randomID := acctest.RandStringFromCharSet(5, acctest.CharSetAlphaNum)
+			testRepoName := fmt.Sprintf("%sbranch-protection-%s", testResourcePrefix, randomID)
+			config := func(enforceAdmins bool) string {
+				return fmt.Sprintf(`
+				resource "github_repository" "test" {
+					name      = "%s"
+					auto_init = true
+				}
```
deiga: Please use mustCreateTestRepository instead

github/resource_github_branch_protection_v3_test.go
```diff
@@ -360,6 +360,66 @@ func TestAccGithubBranchProtectionV3_computed_status_contexts_no_churn(t *testin
...
+							"ci/test",
+							"ci/build"
+						]
+					}
+				}
+			`, testRepoName, enforceAdmins, statusChecksField)
+			}
+
+			resource.Test(t, resource.TestCase{
+				PreCheck:          func() { skipUnlessHasOrgs(t) },
```
deiga: Does branch protection rules need an Org?

deiga (changes requested):

github/resource_github_branch_protection_v3_test.go
dekokun: Done in dd7463c — switched to `mustCreateTestRepository(t)` and dropped the inline `github_repository` resource from the config.

github/resource_github_branch_protection_v3_test.go
dekokun: No — branch protection works on user-owned repos too, so I copied `skipUnlessHasOrgs` from the neighbouring test without thinking it through. Relaxed it to `skipUnauthenticated(t)` in dd7463c, matching what the `github_branch_protection` (v4) tests do.

github/resource_github_branch_protection_v3_test.go
```diff
@@ -360,6 +360,66 @@ func TestAccGithubBranchProtectionV3_computed_status_contexts_no_churn(t *testin
...
+						Config: config(true),
+						Check: resource.ComposeAggregateTestCheckFunc(
+							resource.TestCheckResourceAttr(
+								"github_branch_protection_v3.test", "enforce_admins", "true",
+							),
+							resource.TestCheckResourceAttr(
+								"github_branch_protection_v3.test",
+								fmt.Sprintf("required_status_checks.0.%s.#", statusChecksField), "2",
+							),
+						),
```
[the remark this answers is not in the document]
dekokun: Converted in dd7463c: the step now uses `ConfigStateChecks` with `statecheck.ExpectKnownValue` (`knownvalue.Bool(true)` for `enforce_admins` and `knownvalue.SetSizeExact(2)` for the status-check set), with the `knownvalue`/`statecheck`/`tfjsonpath` imports added.

dekokun:
Thanks for the review! Pushed dd7463c addressing all three points:

- use `mustCreateTestRepository(t)` instead of an inline `github_repository` resource
- `skipUnlessHasOrgs` → `skipUnauthenticated`, since branch protection does not require an org
- assertions converted from `Check`/`TestCheckResourceAttr` to `ConfigStateChecks` + `statecheck.ExpectKnownValue`

@deiga ready for another look when you have a moment.

dekokun:
@deiga gentle ping on this one — the three review comments were addressed in dd7463c a week ago and CI is green. Happy to make further changes if anything still looks off.

For context on why this might be worth prioritising: with `required_status_checks` set, `github_branch_protection_v3` currently cannot be updated in place *at all* — a read populates both `contexts` and `checks`, so any subsequent update sends every check twice and GitHub rejects it with `422 Context must be unique per branch protection` (#3420). The only workaround today is reconciling the branch protection by hand via `gh api`, which needs admin rights on the repo.

Also happy to rebase onto `main` if you would prefer the branch up to date before merging.

deiga (dismissed):
LGTM

dekokun:
@stevehipwell   friendly ping — this has been sitting on your review since @deiga approved it and requested you on Aug 13. CI is green and there are no conflicts with `main`.

The change is small and narrowly scoped: on update, the provider was sending each required status check twice (once from the deprecated `contexts` field, once from `checks`, both of which a read populates from the same API response), so GitHub rejected every update with `422 Context must be unique per branch protection`. The practical impact is that a `github_branch_protection_v3` resource with a non-empty `required_status_checks` block cannot be updated in place at all — even for unrelated fields — and the only workaround is reconciling branch protection by hand with `gh api`, which needs admin on the repo (#3420).

No rush if you are busy — happy to leave it in the queue, just flagging that it is unblocked from my side. I can also update the branch onto `main` whenever that helps.

deiga:
@dekokun To help get your PR reviewed, could you please help us with reviewing other open PRs?
That can help get this and other features implemented faster 🙏

For review guidance, familiarize yourself with our Contributing and Architecture guidelines ☺️

dekokun:
@deiga Sure, happy to help. I have left a review on #3621 and I will keep picking up a couple of open PRs as I go.

github/resource_github_branch_protection_v3_test.go
```diff
@@ -360,6 +363,64 @@ func TestAccGithubBranchProtectionV3_computed_status_contexts_no_churn(t *testin
...
+						Config: config(false),
+					},
+					{
+						Config: config(true),
+						ConfigStateChecks: []statecheck.StateCheck{
+							statecheck.ExpectKnownValue(
+								"github_branch_protection_v3.test",
+								tfjsonpath.New("enforce_admins"),
+								knownvalue.Bool(true),
+							),
```
stevehipwell: Please could you make this (and the one below) a single line to match the current coding standards.

stevehipwell (changes requested):
This PR looks good, just a minor nit.

github/resource_github_branch_protection_v3_test.go
dekokun: Done in f3fdde0, both are single lines now. Thanks for taking a look.

deiga (approved):
LGTM

dekokun:
Thanks @deiga! I've resolved the four review threads and the branch is up to date, so CI (including the required Continuous Integration check) is green.

The only thing left blocking the merge is @stevehipwell's change request from Sep 7, which was just the single-line formatting nit — fixed in f3fdde0 and replied to on the thread. I've re-requested their review. If it's easier for you to dismiss it, that works too, but no rush either way.

stevehipwell (approved):
LGTM

robert-crandall (approved):
Thanks for fixing this and adding regression coverage. The fix preserves app bindings while removing duplicate checks. Looks good to me!

stevehipwell:
@dekokun if you rebase this PR we should be able to get it merged.

dekokun:
Updated onto main, @stevehipwell — should be good to go now. Thanks!

PR #3587, opened 2026-07-29
fix: remove repository file from state when its commit no longer exists

### LINKED ISSUE #3586: github_repository_file: read errors instead of removing the resource from state when its commit no longer exists
### Summary

`github_repository_file`'s read returns a hard error instead of removing the resource from state when the commit it recorded is no longer reachable — for example because the repository, the branch, or the relevant history is gone.

### Current behaviour

`resourceGithubRepositoryFileRead` looks the file's commit up via `client.Repositories.GetCommit` (when `commit_sha` is in state) or `getFileCommit` (when it is not). Both error paths fall through to a bare `return diag.FromErr(err)`:

- a GitHub **404** from `GetCommit` is surfaced as a read failure
- `getFileCommit` returns `cannot find file %s in repo %s/%s` when no commit reachable from the ref contains the file, which is likewise surfaced as a read failure

Other resources in the provider already handle this correctly. `resource_github_repository_custom_property.go`, `resource_github_repository_ruleset.go` and `resource_github_branch_default.go` all detect a 404 on read and call `d.SetId("")` so the resource is removed from state.

### Why this matters

Terraform's contract for read is that a missing remote object means "clear the ID", not "fail". Because this resource errors instead, once the underlying commit is unreachable the resource can neither be refreshed nor cleanly removed — every subsequent plan/apply fails on the read, and the entry cannot be reconciled away without manual state surgery.

The effect is more pronounced for consumers that drive this provider through a reconciliation loop rather than a one-shot `terraform apply`, since a read that always errors is retried indefinitely.

### Expected behaviour

A read that establishes the file's commit no longer exists should remove the resource from state (`d.SetId("")`, no error), consistent with the other repository resources listed above. Errors that are not "not found" should keep propagating unchanged.

I have a patch for this and will open a PR shortly.

🤖

erikmiller-gusto:
Resolves #3586

----

### Before the change?

- `resourceGithubRepositoryFileRead` returns a hard error when the file's commit can no longer be found. Both lookup paths fall through to a bare `return diag.FromErr(err)`: a GitHub 404 from `GetCommit` (used when `commit_sha` is in state), and `getFileCommit`'s `cannot find file ...` case (used when it is not).
- Because read fails rather than clearing the ID, once the commit is unreachable the resource can neither be refreshed nor removed — every later plan/apply fails on the same read, and the entry needs manual state surgery to clear.
- This is inconsistent with the sibling repository resources: `resource_github_repository_custom_property.go`, `resource_github_repository_ruleset.go` and `resource_github_branch_default.go` already detect a 404 on read and call `d.SetId("")`.

### After the change?

- A read that establishes the commit is gone removes the resource from state, matching Terraform's contract and the sibling resources.
- `getFileCommit`'s exhaustion case is wrapped in an `errFileCommitNotFound` sentinel so the read can tell it apart from a genuine failure without string matching. That case is reachable independently of any HTTP status, which is why a sentinel is used rather than only checking for a 404.
- Every error that is not "not found" still propagates unchanged.

### Pull request checklist

- [ ] ~Schema migrations have been created if needed~ — no schema change; this only affects read behaviour.
- [ ] ~Docs have been reviewed and added / updated if needed~ — no user-facing schema or option change, so I did not find anything in `website/` that describes this behaviour. Happy to document it if you would like it called out.

- [ ] Yes
- [x] No

A configuration whose file still exists is unaffected. The change only alters what happens when the commit is already gone, where the previous outcome was a permanent read error.

----

### Testing

`Test_getFileCommit_NoCommitContainsFile` drives `getFileCommit` against an `httptest` server that returns an empty commit list, and asserts `errors.Is(err, errFileCommitNotFound)`. It uses the existing `mustCreateTestGitHubClient` helper and needs no credentials. I verified it discriminates: removing the `%w` wrap makes it fail with `errors.Is(...) = false`, which is the regression that would silently restore the old behaviour.

I could not exercise the resource-level 404 path, since the acceptance tests need a live org and credentials I do not have. Note that `go test ./github/...` does not pass cleanly on a fresh checkout in my environment either — `TestAccGithubActionsOrganizationPermissions` panics without credentials — and I confirmed that failure is identical with and without this change. `go build ./...`, `go vet ./github/...` and `gofmt -l` are clean.

### AI use

Per the AI Use Policy: this change was written with AI assistance. I have reviewed it, and the verification described above is what I actually ran rather than a summary of intent. The reasoning for the sentinel over a status-code-only check, and the scope limits above, are stated deliberately — please push back if you would rather solve it differently, for example by handling the exhaustion case at the call site instead.

🤖

### DIFF
--- a/github/util_repo_test.go
+++ b/github/util_repo_test.go
@@ -1,0 +1,34 @@
+package github
+
+import (
+	"context"
+	"errors"
+	"net/http"
+	"net/http/httptest"
+	"testing"
+)
+
+// Test_getFileCommit_NoCommitContainsFile locks in the errFileCommitNotFound
+// sentinel. resourceGithubRepositoryFileRead relies on it to remove the resource
+// from state instead of returning a hard error, so unwrapping this error would
+// silently reintroduce that failure.
+func Test_getFileCommit_NoCommitContainsFile(t *testing.T) {
+	ts := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, _ *http.Request) {
+		w.Header().Set("Content-Type", "application/json; charset=utf-8")
+		if _, err := w.Write([]byte("[]")); err != nil {
+			t.Errorf("failed to write response: %s", err)
+		}
+	}))
+	defer ts.Close()
+
+	client := mustCreateTestGitHubClient(t, ts.URL+"/")
+
+	_, err := getFileCommit(context.Background(), client, "owner", "repo", "some/file.txt", "main")
+	if err == nil {
+		t.Fatal("expected an error when no commit contains the file, got nil")
+	}
+
+	if !errors.Is(err, errFileCommitNotFound) {
+		t.Errorf("errors.Is(err, errFileCommitNotFound) = false, want true; got error: %s", err)
+	}
+}
--- a/github/resource_github_repository_file.go
+++ b/github/resource_github_repository_file.go
@@ -2,6 +2,7 @@ package github
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"net/http"
 	"net/url"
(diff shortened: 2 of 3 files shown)


github/util_repo_test.go
```diff
@@ -0,0 +1,34 @@
...
+		w.Header().Set("Content-Type", "application/json; charset=utf-8")
+		if _, err := w.Write([]byte("[]")); err != nil {
+			t.Errorf("failed to write response: %s", err)
+		}
+	}))
+	defer ts.Close()
+
+	client := mustCreateTestGitHubClient(t, ts.URL+"/")
+
+	_, err := getFileCommit(context.Background(), client, "owner", "repo", "some/file.txt", "main")
```
deiga, line 26: Please use t.Context() instead

deiga (changes requested):

PR #3594, opened 2026-08-04
fix: populate database_id on enterprise org create; fix enterprise permissions test config

### LINKED ISSUE #3593: Enterprise acceptance tests use enterprise_slug, but github_enterprise_organization only accepts enterprise_id
### Expected Behavior

`TestAccGithubActionsEnterprisePermissions` should be able to run.

### Actual Behavior

It fails at plan time, before touching the API. The test config declares `github_enterprise_organization` with an `enterprise_slug` argument, but the resource schema only has `enterprise_id` (`github/resource_github_enterprise_organization.go`, `Required: true`). There is no `enterprise_slug` attribute in the schema.

```
Error: Missing required argument

  on terraform_plugin_test.tf line 16, in resource "github_enterprise_organization" "org":
  16:             resource "github_enterprise_organization" "org" {

The argument "enterprise_id" is required, but no definition was found.

Error: Unsupported argument

  on terraform_plugin_test.tf line 17, in resource "github_enterprise_organization" "org":
  17:                 enterprise_slug = "github-terraform-testing"

An argument named "enterprise_slug" is not expected here.
```

Both subtests fail this way, so the test cannot pass in its current form.

The same `enterprise_slug = ` pattern appears in four other enterprise test files, so they may be affected too - I only confirmed this one directly:

- `github/resource_github_enterprise_actions_permissions_test.go` (7)
- `github/resource_github_enterprise_security_analysis_settings_test.go` (8)
- `github/resource_github_enterprise_actions_workflow_permissions_test.go` (4)
- `github/resource_github_enterprise_ip_allow_list_entry_test.go` (2)
- `github/resource_github_enterprise_actions_runner_group_test.go` (1)

I have not looked into which side is wrong - whether the tests should pass `enterprise_id` or the resource should accept a slug.

### Terraform Version

Terraform v1.8.1

### GitHub Installation Type

GitHub Enterprise Cloud with Personal Accounts (github.com)

### Affected Resource(s)

- `github_enterprise_organization`

### Steps to Reproduce

On `main` (observed at `f3c63cb2`), with an enterprise-mode acceptance test environment:

```
TF_ACC=1 go test ./github -run '^TestAccGithubActionsEnterprisePermissions$' -v
```

robert-crandall:
Fixes #3593.

## What I did

Two things, and the second one is the interesting one.

**1. The test config was wrong.** `TestAccGithubActionsEnterprisePermissions` declared `github_enterprise_organization` blocks with `enterprise_slug`, but that resource requires `enterprise_id` (the enterprise node ID) and has no slug attribute. Both subtests failed at plan time before touching the API. Fixed by adding a `data "github_enterprise"` block and passing `enterprise_id = data.github_enterprise.enterprise.id`, matching the pattern already used in `resource_github_enterprise_organization_test.go` and in `examples/resources/enterprise_actions_runner_group/example_1.tf`.

The same configs also passed the org's GraphQL node ID to `enabled_organizations_config.organization_ids`. That attribute is a set of `TypeInt` and the read path populates it from REST `Organization.ID` (database IDs), so a node ID string can never satisfy it. Switched to `.database_id`.

**2. Fixing that first issue surfaced a real provider bug.** This provider deliberately doesn't read after write, and `resourceGithubEnterpriseOrganizationCreate` never set the computed `database_id` - it only set the resource ID. So `database_id` stayed `0` in state until the next refresh, and any same-apply reference to it was broken. That includes the example we ship at `examples/resources/enterprise_actions_runner_group/example_1.tf:17`, which does exactly that. Create now selects `databaseId` from the `createEnterpriseOrganization` mutation payload and writes it to state.

### DIFF
--- a/github/resource_github_enterprise_organization_test.go
+++ b/github/resource_github_enterprise_organization_test.go
@@ -9,8 +9,10 @@ import (
 	"testing"
 
 	"github.com/google/go-github/v89/github"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
 	"github.com/hashicorp/terraform-plugin-testing/helper/acctest"
 	"github.com/hashicorp/terraform-plugin-testing/helper/resource"
+	"github.com/shurcooL/githubv4"
 )
 
 func TestIsSAMLEnforcementError(t *testing.T) {
@@ -110,6 +112,9 @@ func TestAccGithubEnterpriseOrganization(t *testing.T) {
 			"before": resource.ComposeTestCheckFunc(
 				resource.TestCheckResourceAttrSet(
 					"github_enterprise_organization.org", "enterprise_id",
+				),
+				resource.TestCheckResourceAttrSet(
+					"github_enterprise_organization.org", "database_id",
 				),
 				resource.TestCheckResourceAttr(
 					"github_enterprise_organization.org", "name",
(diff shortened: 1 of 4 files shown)


github/resource_github_enterprise_organization_test.go
```diff
@@ -565,3 +570,60 @@ func TestAccGithubEnterpriseOrganization(t *testing.T) {
 		})
 	})
 }
+
+func TestResourceGithubEnterpriseOrganizationCreateSetsDatabaseID(t *testing.T) {
+	// IMPORTANT: This test is not parallelized because it uses a shared HTTP handler.
```
deiga: Is this true? I might misunderstand something, but it seems like the HTTP handler is fully internal to this test

github/resource_github_enterprise_organization_test.go
```diff
@@ -565,3 +570,60 @@ func TestAccGithubEnterpriseOrganization(t *testing.T) {
...
+		if !strings.Contains(body, "databaseId") {
+			t.Errorf("mutation does not select databaseId: %s", body)
+		}
+		w.Header().Set("Content-Type", "application/json")
+		mustWrite(w, createResponse)
+	})
+
+	meta := &Owner{
+		v4client: githubv4.NewClient(&http.Client{Transport: localRoundTripper{handler: mux}}),
+	}
```
deiga, line 602: Some parts of this should maybe be refactored into a `mustCreateTestGitHubGraphQLClient` function, but not necessary in this PR

deiga (dismissed):
Looks good! I don't have enough permissions to run this locally 😬

github/resource_github_enterprise_organization_test.go
robert-crandall: You understand it correctly - I was wrong. The mux, the client and the ResourceData are all local to the test, so there is nothing shared. I copied that note from `util_v4_repository_test.go` without checking whether it applied. Swapped it for `t.Parallel()` in f14a63d2, and it passes under `-race`.

github/resource_github_enterprise_organization_test.go
robert-crandall: Done in f14a63d2 - added `newTestGraphQLClient(handler)` next to `localRoundTripper` and used it here and in `TestGetRepositoryIDPositiveMatches`.

One deviation: I dropped the `must` prefix you suggested. In this package `must` means "panics on error" (`mustRead`, `mustWrite`), and `githubv4.NewClient` cannot fail, so `mustCreateTestGitHubGraphQLClient` would be misleading. Happy to rename if you would rather keep it consistent with your suggestion.

deiga: Hmm, I proposed "must" prefix, since the mock rest API func is called that way and we've named all new test setup funcs that way 🤔

I personally don't have a strong preference either way, but appreciate consistency

blkdm0n (approved):

PR #3598, opened 2026-08-06
feat(organization_ruleset): support the repository target

### LINKED ISSUE #2777: [FEAT]: Add repository target to github_organization_ruleset
### Describe the need

The provider currently only supports `branch` and `tag` as values for the `target` argument of the `github_organization_ruleset` resource. `repository` should be added.

The API [already supports this](https://docs.github.com/en/rest/orgs/rules?apiVersion=2022-11-28#update-an-organization-repository-ruleset). Relates to #2714 (adding the `push` target).

### SDK Version

_No response_

### API Version

_No response_

### Relevant log output

```shell

```

### Code of Conduct

yordis:
Resolves #2777

----

### Before the change?

- Organization rulesets accept a `repository` target that governs repository creation, deletion, naming, transfer and visibility. The provider rejected it, so those org policies could not be managed in Terraform at all. Importing one was worse than unsupported: `target` is `Required` but not `ForceNew`, so an import succeeded and the next apply silently rewrote the ruleset as a branch ruleset.
- #2777 was closed as completed, but the change referenced there (#2958) shipped the `push` target, not `repository`. #3110 implements these rules only for a new `github_enterprise_ruleset` and treats the repository target as enterprise-only; the REST docs list it as a plain ruleset target ("Can be one of: `branch`, `tag`, `push`, `repository`"), and all of the testing below was done against a regular org on the Team plan.

### After the change?

- `github_organization_ruleset` accepts `target = "repository"`, so repository creation, deletion, naming, transfer and visibility policies are declarable and importable like every other ruleset.
- Five rules are added for that target: `repository_create`, `repository_delete`, `repository_transfer`, `repository_name` (`pattern`, `negate`) and `repository_visibility`.
- Rules and conditions are validated per target at plan time rather than failing at the API, and `target` is now checked against the four values the API documents instead of accepting any string.

### Testing

Verified end to end against a live organization using a local build of this branch via `dev_overrides`. Create, read, update and delete all exercised against the API.

**Create.** A ruleset using all five rules, applied clean, and the server stored exactly the intended payload:

```json
"rules": [
  {"type": "repository_create"},
  {"type": "repository_delete"},
  {"type": "repository_name", "parameters": {"negate": false, "pattern": "^tmp-writepath-"}},
  {"type": "repository_transfer"},
  {"type": "repository_visibility", "parameters": {"internal": true, "private": true, "public": false}}
]
```

This confirms the parameter shapes: `repository_create`/`repository_delete`/`repository_transfer` marshal as bare rules with no parameters, and `repository_name` as `{negate, pattern}`. A follow-up plan was empty, so all five round-trip through read.

**Update.** Removing a bare rule, flipping `negate`, changing `pattern` and flipping `internal` applied as a single in-place change, the server reflected each one, and the follow-up plan was empty.

**Import.** An existing repository-target ruleset created through the UI imported cleanly, both directly and through an `import` block (`1 to import, 0 to add, 0 to change, 0 to destroy`), with an empty plan afterwards.

**Delete.** Destroy removed it; the API returns 404.

**Validation**, rejected at plan time:

| Config | Error |
| --- | --- |
| `repository` target with `ref_name` | `ref_name must not be set for repository target` |
| `repository` target with `creation` | `rule "creation" is not valid for repository target; repository targets only support: [...]` |
| `branch` target with `repository_delete` | `rule "repository_delete" is not valid for branch target; ...` |
| `target = "bogus"` | `expected target to be one of ["branch" "tag" "push" "repository"], got bogus` |

Unit tests, `gofmt`, `go vet` and `golangci-lint` are clean. New unit coverage: round-trip expand/flatten for all five rules, a check that the rules are omitted for `github_repository_ruleset` (whose schema lacks them), a pairwise-disjointness test across the three target rule lists, and conditions validation for the new target. Acceptance tests are included but I cannot run the mutating suite in CI myself.

### Pull request checklist

The new rules are optional attributes on an existing resource, so no state migration is needed.

- [ ] Yes
- [x] No

Existing `branch`, `tag` and `push` configurations are untouched. The one behaviour change to call out is that `target` is now validated against those four values at plan time, where it previously accepted any string and failed at the API instead.

----

I used an AI assistant while writing this. I have read every line, verified the API behaviour against the REST docs and against live calls rather than taking the model's word for it, and ran the resources end to end against a real organization before opening this, as tabulated above.


deiga:
Please show a screenshot of an Org level ruleset with `repository` target. AFAIK the docs lie and the `repository` target is only available on Enterprise level

yordis:
@deiga unless I do not understand my own GitHub org, my apologies if I did a mistake here, I honestly thought this is the normal flow

https://github.com/user-attachments/assets/36f18d7b-091b-437b-8faf-05724c9fb9a6

deiga:
@yordis you have understood totally fine. I wasn't aware that it's under a different UI menu! 🙏

templates/resources/organization_ruleset.md.tmpl
deiga, line 1: question: would you have the courage to refactor this template to use the auto-generated schema instead of manually doing the schema changes?

examples/resources/organization_ruleset/example_1.tf
deiga, line 1: suggestion: Please split examples into multiple files and move them to `examples/resources/github_organization_ruleset/resource_<name|number>.tf`

deiga (changes requested):
Nice changes! Would you be open for improving the docs bit?

yordis:
@deiga check again and let me know 🙏🏻

yordis:
@deiga should I address Copilot until it is happy with things?

github/util_rules.go
```diff
@@ -655,6 +655,37 @@ func expandRules(input []any, org bool) *github.RepositoryRulesetRules {
...
+				Pattern: repositoryNameMap["pattern"].(string),
+			}
+		}
+
+		if v, ok := rulesMap["repository_visibility"].([]any); ok && len(v) != 0 {
+			repositoryVisibilityMap := v[0].(map[string]any)
+			rulesetRules.RepositoryVisibility = &github.RepositoryVisibilityRuleParameters{
+				Internal: repositoryVisibilityMap["internal"].(bool),
+				Private:  repositoryVisibilityMap["private"].(bool),
+			}
```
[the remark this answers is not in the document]
yordis: There is no `public` flag on this rule to carry through.

- go-github [v89](https://github.com/google/go-github/blob/v89.0.0/github/rules.go#L602-L605), [v90](https://github.com/google/go-github/blob/v90.0.0/github/rules.go) and `master`: `RepositoryVisibilityRuleParameters` is `Internal` + `Private` only, unchanged since https://github.com/google/go-github/pull/3850.
- [github/rest-api-description](https://github.com/github/rest-api-description): `repository_visibility` does not appear in `api.github.com.json` or `ghec.json`.

`expandRules` and `flattenRules` are symmetric over those two fields, so there is nothing to drop on update.

yordis:
Heads up on a related gap I hit while working on this, and filed upstream as google/go-github#4455.

The `repository_visibility` rule has a third toggle, `public`, alongside `internal` and `private`. `RepositoryVisibilityRuleParameters` in go-github does not model it, so `encoding/json` drops it on decode. Because updating a ruleset is a full replace, a read-modify-write through the SDK omits the key and the API resolves it back to `false`.

I verified this against a live `target: repository` ruleset: the server sends `{"public":false,"internal":true,"private":true}`, the SDK decodes it as `{"internal":true,"private":true}`, and setting `public` through the typed API is not expressible at all today.

That means this PR ships `repository_visibility` with two of the three toggles, and Terraform managing such a ruleset will reset `public` to `false`. Nothing to change here, since the field cannot be plumbed through until the SDK supports it. Once #4455 lands and is released, the follow-up is a dependency bump plus a `public` attribute on the `repository_visibility` block.

Filing this so the limitation is on the record rather than discovered later.

github/resource_github_organization_ruleset_test.go
```diff
@@ -566,6 +566,72 @@ resource "github_organization_ruleset" "test" {
...
+						statecheck.ExpectKnownValue("github_organization_ruleset.test", tfjsonpath.New("rules").AtSliceIndex(0).AtMapKey("repository_delete"), knownvalue.Bool(true)),
+						statecheck.ExpectKnownValue("github_organization_ruleset.test", tfjsonpath.New("rules").AtSliceIndex(0).AtMapKey("repository_transfer"), knownvalue.Bool(true)),
+						statecheck.ExpectKnownValue("github_organization_ruleset.test", tfjsonpath.New("rules").AtSliceIndex(0).AtMapKey("repository_name").AtSliceIndex(0).AtMapKey("pattern"), knownvalue.StringExact("^tf-acc-")),
+						statecheck.ExpectKnownValue("github_organization_ruleset.test", tfjsonpath.New("rules").AtSliceIndex(0).AtMapKey("repository_name").AtSliceIndex(0).AtMapKey("negate"), knownvalue.Bool(false)),
+						statecheck.ExpectKnownValue("github_organization_ruleset.test", tfjsonpath.New("rules").AtSliceIndex(0).AtMapKey("repository_visibility").AtSliceIndex(0).AtMapKey("internal"), knownvalue.Bool(true)),
+						statecheck.ExpectKnownValue("github_organization_ruleset.test", tfjsonpath.New("rules").AtSliceIndex(0).AtMapKey("repository_visibility").AtSliceIndex(0).AtMapKey("private"), knownvalue.Bool(true)),
```
deiga: No need to check values in the config, those are checked automatically

deiga (changes requested):
Minor changes still ☺️

yordis:
@deiga will you take the PR as is, or should we wait for the upstream pkg to be released first? the `public` field will be dropped, that is the downside

github/util_rules.go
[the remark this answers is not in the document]
yordis: Fixed by https://github.com/google/go-github/pull/4455 so it requires it to be released, or I can use the commit sha to pin it here

github/resource_github_organization_ruleset_test.go
yordis: Dropped in 3ad49c16 — the step just carries the config now, and the follow-on import step covers the round-trip.

github/util_rules.go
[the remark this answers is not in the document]
deiga: We don't know when a new release is going to show up, and commit pinning isn't something we want to do either.

Could you add a TODO comment to the lines/sections that need changing when v91 come sout?

yordis:
@deiga :shipit:

yordis:
@deiga run the tools again! 🙏🏻

github/util_rules.go
```diff
@@ -655,6 +655,46 @@ func expandRules(input []any, org bool) *github.RepositoryRulesetRules {
...
+		}
+
+		if v, ok := rulesMap["repository_transfer"].(bool); ok && v {
+			rulesetRules.RepositoryTransfer = &github.EmptyRuleParameters{}
+		}
+
+		if v, ok := rulesMap["repository_name"].([]any); ok && len(v) != 0 {
+			if repositoryNameMap, ok := v[0].(map[string]any); ok {
+				negate, _ := repositoryNameMap["negate"].(bool)
+				pattern, _ := repositoryNameMap["pattern"].(string)
```
deiga, line 675: I'm not sure, but it might be possible for pattern to be nil, right? Does that pose silent-failure with the error swallowing?

yordis: `pattern` is Required with `StringIsNotEmpty` validation, and SDKv2 always materializes every attribute of a nested block, so the value is a non-empty string whenever the block is present.

The assertion can't fail; the comma-ok form is only there to satisfy the `forcetypeassert` gate on new code, since `expandRules` has no error return. And if it ever were empty, the API rejects it with a 422 rather than accepting silently. 

Happy to restructure if you'd prefer something more explicit.

yordis:
Waiting for #3640

yordis:
@danielpops no more TODO comments, everything now would work, since the `public` is supported now

yordis:
@deiga any opportunity to get this one review now? 🙏🏻

PR #3602, opened 2026-08-10
Allow anonymous acceptance tests without credentials

rilical:
I found that `getTestMeta` still attempted token or GitHub App credential setup when `GH_TEST_AUTH_MODE=anonymous`, so credential-free anonymous acceptance tests exited during setup. This change skips credential setup only for anonymous mode, and the authenticated paths are unchanged.

I added `Test_getTestMeta_anonymousDoesNotRequireCredentials` to cover the regression. I ran `make test`, `go vet ./...`, and `CGO_ENABLED=0 go build ./...`; `gofmt -l .` returned no files.

### DIFF
--- a/github/provider_test.go
+++ b/github/provider_test.go
@@ -19,6 +19,24 @@ func TestProvider(t *testing.T) {
 	})
 }
 
+func Test_getTestMeta_anonymousDoesNotRequireCredentials(t *testing.T) {
+	baseURL, _, err := getBaseURL(DotComAPIURL)
+	if err != nil {
+		t.Fatalf("parsing base URL: %v", err)
+	}
+
+	meta, err := getTestMeta(&testAccConfig{
+		authMode: anonymous,
+		baseURL:  baseURL,
+	})
+	if err != nil {
+		t.Fatalf("configuring anonymous provider meta: %v", err)
+	}
+	if meta == nil {
+		t.Fatal("anonymous provider meta is nil")
+	}
+}
+
 func Test_configureProviderMeta(t *testing.T) {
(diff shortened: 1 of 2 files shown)


stevehipwell:
@rilical I'm not sure I agree with the design here or even the problem description, at first glance this change looks to add significant complexity which goes against to Go idiom of making code as readable as possible. The acceptance test design is to gate the tests, the helpers are expected to fail if they're used by a test that doesn't have permission as this is a coding error.

FYI for a large change such as this it'd be best to open an issue to track the actual problem being solved rather than come with an implementation.

rilical:
@stevehipwell Would you mind giving this another look?

robert-crandall (dismissed):
Thank you for simplifying this. This PR fixes a real problem of anonymous tests failing during setup.

github/provider_test.go
```diff
@@ -19,6 +19,24 @@ func TestProvider(t *testing.T) {
...
+		baseURL:  baseURL,
+	})
+	if err != nil {
+		t.Fatalf("configuring anonymous provider meta: %v", err)
+	}
+	if meta == nil {
+		t.Fatal("anonymous provider meta is nil")
+	}
+}
+
```
stevehipwell:
```suggestion
```
I don't think we want or need to test a test function.

stevehipwell (changes requested):
The code change looks good to me, but we don't need the test.

stevehipwell (approved):
LGTM

stevehipwell:
@rilical if you can rebase your branch we can get this merged in.

PR #3605, opened 2026-08-12
fix: Simplify new client throttle logic to avoid leaks

### LINKED ISSUE #3603: [BUG]: New client leaks concurrency permits on non-2xx REST responses
### Expected Behavior

With `legacy_client = false`, every concurrency permit acquired for a GitHub REST request should be released after the response has been consumed, including when GitHub returns a non-2xx response.

Expected responses such as `404 Not Found` from the vulnerability-alerts endpoint should not progressively exhaust the provider's concurrency limit. A Terraform plan managing more than 100 repositories should complete without requests becoming permanently blocked inside the provider.

### Actual Behavior

The new client introduced in v6.13.0 leaks one concurrency permit whenever `go-github` processes a non-2xx REST response.

After 100 such responses, all permits in the provider's shared semaphore are exhausted. Subsequent requests block in `semaphore.Acquire` before reaching the network. They eventually fail after the provider's five-minute client timeout with:

```text
Get "https://api.github.com/repos/example-org/example-repository": (http.RoundTripper).RoundTrip failed: context deadline exceeded (Client.Timeout exceeded while awaiting headers)
```

This was reproduced on both:

- Terraform v1.14.4 on `linux_amd64`, running on a self-hosted GitHub Actions runner
- Terraform v1.14.4 on `darwin_arm64`

The affected configuration manages 331 existing `github_repository` resources using GitHub App authentication.

The repeated trigger in this configuration is:

```http
GET /repos/{owner}/{repository}/vulnerability-alerts
```

GitHub returns `404 Not Found` when vulnerability alerts are disabled. `go-github` intentionally converts that response to `false` without returning an error, but the concurrency permit has already been leaked.

#### Root cause

The v6.13.0 [`throttler`](https://github.com/integrations/terraform-provider-github/blob/v6.13.0/internal/ghclient/throttle.go#L34-L50):

1. Acquires one semaphore permit.
2. Wraps the original response body in `throttlerReadCloser`.
3. Releases the permit only when that wrapper's `Close` method is called.

For non-2xx responses, [`go-github.CheckResponse`](https://github.com/google/go-github/blob/v88.0.0/github/github.go#L1591-L1610) reads the wrapped body and replaces `r.Body` with a new `io.NopCloser`, without closing the original body.

[`bareDo`](https://github.com/google/go-github/blob/v88.0.0/github/github.go#L1124-L1127) subsequently closes only the replacement body. The original `throttlerReadCloser` is no longer referenced, so its `Close` method is never called and the permit is never released.

The provider's concurrency limit is exactly 100, and its client timeout is five minutes: [`internal/ghclient/const.go`](https://github.com/integrations/terraform-provider-github/blob/v6.13.0/internal/ghclient/const.go#L21-L41).

For the observed trigger, `GetVulnerabilityAlerts` calls `client.Do` and then [`parseBoolResponse`](https://github.com/google/go-github/blob/v88.0.0/github/github.go#L1658-L1674), which hides the `404` and returns `false, nil`. The plan therefore continues while silently losing one permit for each repository with vulnerability alerts disabled.

Control tests established that this is not a network-connectivity or retry problem:

- Direct authenticated `curl` requests from the same runner completed successfully over HTTP/2.
- The provider process had no outbound TCP socket for the requests waiting on the exhausted semaphore.
- `max_retries = 0` produced the same behavior.
- `legacy_client = true` with `parallel_requests = true` completed the same plan successfully.
- The issue occurs with `legacy_client = false`, where `parallel_requests` is documented as ignored.

### Terraform Version

```text
Terraform v1.14.4
on linux_amd64
```

Also reproduced with Terraform v1.14.4 on `darwin_arm64`.

### GitHub Installation Type

- [ ] I don't know

### Affected Resource(s)

- `github_repository`
- Potentially every resource or data source using the new REST client, because the defect is in the shared `internal/ghclient` transport

```hcl
terraform {
  required_providers {
    github = {
      version = "6.13.0"
    }
  }
}

provider "github" {
  owner         = var.github_owner
  legacy_client = false
  max_retries   = 0

  app_auth {
    id              = var.github_app_id
    installation_id = var.github_app_installation_id
    pem_file        = var.github_app_pem
  }
}

resource "github_repository" "existing" {
  for_each = toset(var.existing_repository_names)

  name = each.value
}
```

The production configuration uses modules, but the failure occurs in the provider's shared HTTP transport rather than module evaluation.

### Steps to Reproduce

A deterministic unit test can reproduce the leak without contacting GitHub.

Add the following test under `internal/ghclient` at tag `v6.13.0`:

```go
package ghclient

import (
	"io"
	"net/http"
	"strings"
	"testing"

	"github.com/google/go-github/v88/github"
	"golang.org/x/sync/semaphore"
)

type notFoundTransport struct{}

func (t *notFoundTransport) RoundTrip(req *http.Request) (*http.Response, error) {
	return &http.Response{
		Status:     "404 Not Found",
		StatusCode: http.StatusNotFound,
		Header:     make(http.Header),
		Body:       io.NopCloser(strings.NewReader(`{"message":"Not Found"}`)),
		Request:    req,
	}, nil
}

func TestThrottlerReleasesPermitAfterErrorResponse(t *testing.T) {
	sema := semaphore.NewWeighted(1)

	client, err := github.NewClient(
		github.WithTransport(&throttler{
			sema:  sema,
			inner: &notFoundTransport{},
		}),
		github.WithDisableRateLimitCheck(),
	)
	if err != nil {
		t.Fatal(err)
	}

	req, err := client.NewRequest(
		t.Context(),
		http.MethodGet,
		"repos/example/missing",
		nil,
	)
	if err != nil {
		t.Fatal(err)
	}

	if _, err := client.Do(req, nil); err == nil {
		t.Fatal("expected GitHub API error")
	}

	if !sema.TryAcquire(1) {
		t.Fatal("semaphore permit leaked after non-2xx response")
	}
	sema.Release(1)
}
```

Run:

```shell
go test ./internal/ghclient \
  -run TestThrottlerReleasesPermitAfterErrorResponse \
  -count=1
```

The test fails on v6.13.0 with:

```text
semaphore permit leaked after non-2xx response
```

A Terraform-level reproduction is:

1. Configure `integrations/github` v6.13.0 with `legacy_client = false`.
2. Manage or refresh more than 100 repositories for which the vulnerability-alerts check returns `404`.
3. Run `TF_LOG=DEBUG terraform plan`.
4. Observe that the provider initially refreshes resources successfully.
5. After the semaphore permits are exhausted, subsequent API operations stop reaching the network.
6. After five minutes, blocked requests begin returning `Client.Timeout exceeded while awaiting headers`.
7. Repeat with `legacy_client = true` and `parallel_requests = true`; the plan completes.

### Debug Output

```shell
[ERROR] vertex "github_repository.example" error:
Get "https://api.github.com/repos/example-org/example-repository":
(http.RoundTripper).RoundTrip failed:
context deadline exceeded
(Client.Timeout exceeded while awaiting headers)
```

At the time of the timeout, the provider process remained alive but had no TCP connection corresponding to the blocked request. Direct API requests from the same runner continued to succeed.

#### Additional Context

The new client and explicit concurrency control were introduced in v6.13.0. The previous release, v6.12.1, did not contain `internal/ghclient`, the 100-permit semaphore, or the `legacy_client` option.

One possible correction would be for `throttlerReadCloser` to release its permit when reading reaches `io.EOF` as well as on `Close`, using the existing `sync.Once`. Another option is to ensure that the original wrapped response body is explicitly closed before `go-github` replaces it. The key requirement is that all response paths, including non-2xx responses, return the acquired permit exactly once.

### Code of Conduct

stevehipwell:
Resolves #3603

----

### Before the change?

- New client would leak concurrency permits
- New client concurrency constraint based on reading the response body

### After the change?

- New client concurrency permits can't leak
- New client concurrency constraint based on remote processing

### Pull request checklist

- [ ] Yes
- [x] No

----

### DIFF
--- a/internal/ghclient/helpers_test.go
+++ b/internal/ghclient/helpers_test.go
@@ -6,16 +6,23 @@ import (
 	"path/filepath"
 	"sync/atomic"
 	"testing"
+	"time"
 )
 
 type testRoundTripper struct {
+	delay  time.Duration
 	called atomic.Int32
 	resp   *http.Response
 	err    error
 }
 
 func (r *testRoundTripper) RoundTrip(_ *http.Request) (*http.Response, error) {
+	if r.delay > 0 {
+		time.Sleep(r.delay)
+	}
+
 	r.called.Add(1)
+
 	if r.err != nil {
 		return nil, r.err
 	}
@@ -43,4 +50,15 @@ func mustReadAppPrivateKey(t *testing.T) []byte {
 	}
 
 	return privateKeyData
+}
+
+func mustCreateRequest(t *testing.T, method, url string) *http.Request {
+	t.Helper()
+
+	req, err := http.NewRequestWithContext(t.Context(), method, url, nil)
+	if err != nil {
+		t.Fatalf("failed to create request: %v", err)
+	}
+
+	return req
 }
--- a/internal/ghclient/helpers_test.go
+++ b/internal/ghclient/helpers_test.go
@@ -16,18 +16,22 @@ type testRoundTripper struct {
 	err    error
 }
 
-func (r *testRoundTripper) RoundTrip(_ *http.Request) (*http.Response, error) {
-	if r.delay > 0 {
-		time.Sleep(r.delay)
+func (tr *testRoundTripper) RoundTrip(req *http.Request) (*http.Response, error) {
+	if tr.delay > 0 {
+		ctx := req.Context()
+		select {
+		case <-time.After(tr.delay):
+		case <-ctx.Done():
+		}
 	}
 
-	r.called.Add(1)
+	tr.called.Add(1)
 
-	if r.err != nil {
-		return nil, r.err
+	if tr.err != nil {
+		return nil, tr.err
 	}
 
-	return r.resp, nil
+	return tr.resp, nil
 }
 
 func mustMkdirTemp(t *testing.T, dir, pattern string) string {
--- a/internal/ghclient/throttle.go
+++ b/internal/ghclient/throttle.go
@@ -1,28 +1,10 @@
 package ghclient
 
 import (
-	"io"
 	"net/http"
-	"sync"
 
 	"golang.org/x/sync/semaphore"
 )
-
-// throttlerReadCloser is a wrapper around an io.ReadCloser that releases a semaphore weight when the ReadCloser is closed. This is used to ensure that the semaphore controlling concurrent requests is properly released after the response body has been fully read and closed, preventing resource leaks and allowing other requests to proceed.
-type throttlerReadCloser struct {
-	io.ReadCloser
-	sema *semaphore.Weighted
-	once sync.Once
-}
-
-// Close releases the semaphore weight when the ReadCloser is closed. It ensures that the release operation is only performed once, even if Close is called multiple times, preventing potential double-release issues that could lead to incorrect semaphore state.
-func (c *throttlerReadCloser) Close() error {
-	err := c.ReadCloser.Close()
-	c.once.Do(func() {
-		c.sema.Release(1)
-	})
-	return err
-}
 
 // throttler is an HTTP RoundTripper that limits the number of concurrent requests to a specified maximum. It uses a weighted semaphore to control access to the underlying RoundTripper, ensuring that no more than the allowed number of requests are in flight at any given time. This is useful for preventing overwhelming a server or API with too many simultaneous requests.
 type throttler struct {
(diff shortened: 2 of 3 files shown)


internal/ghclient/helpers_test.go
```diff
@@ -6,16 +6,23 @@ import (
...
 type testRoundTripper struct {
+	delay  time.Duration
 	called atomic.Int32
 	resp   *http.Response
 	err    error
 }
 
 func (r *testRoundTripper) RoundTrip(_ *http.Request) (*http.Response, error) {
+	if r.delay > 0 {
+		time.Sleep(r.delay)
```
deiga: question: Should this use the Context-aware `sleep`?

deiga (dismissed):
Looks good, but I can't fully vouch for the functionality. This isn't my strongest area 😬

internal/ghclient/helpers_test.go
stevehipwell: It's probably best to do so.

deiga (approved):

robert-crandall (approved):
Nice cleanup — dropping the `throttlerReadCloser` in favor of a plain `defer Release` makes the leak impossible by construction, and moving the constraint to remote processing time feels like the right semantics here. Thanks for tracking this down!

shxdxwraze:
@robert-crandall @deiga hi, when you plan to release this?

stevehipwell:
@shxdxwraze there is currently no planned release date for `6.14.0`.

shxdxwraze:
i'm suffering

PR #3608, opened 2026-08-16
feat: Add downgrade_to option to github_membership resource

### LINKED ISSUE #3607: [FEAT]: Support turning organization members into outside collaborators
### Describe the need

I'm a maintainer of tooling for [managing GitHub configuration as code](https://github.com/ipdxco/github-as-code). One of the use cases I'd like to support is the ability to turn existing organisation members into [outside collaborators](https://docs.github.com/en/rest/orgs/outside-collaborators?apiVersion=2026-03-10). This is especially important for organisations on non-free plans, as they often want to free up organisation membership seats without removing access to public repositories for their collaborators.

Currently, when destroying [github_membership](https://registry.terraform.io/providers/integrations/github/latest/docs/resources/membership), one has to choose between deleting the membership or downgrading it to `member` (if said membership is higher).

I'd like to propose adding a new option for controlling what a membership can be downgraded to, with a choice between `member` and `outside_collaborator`. I have started working on this already and, hopefully, will be able to put it up for review shortly.

### GitHub Installation Type

- [ ] I don't know

### Relevant log output

```shell

```

### Code of Conduct

galargh:
Resolves #3607

----

### Before the change?

- [github_membership](https://registry.terraform.io/providers/integrations/github/latest/docs/resources/membership) accepted optional `downgrade_on_destroy` parameter
- when enabled, on destroy, if a user was an admin, their permissions were downgraded to a regular member
- if they were already a member, the resource was destroyed without modifying the user's status

### After the change?

- `github_membership` gets a new optional `downgrade_to` parameter
- it defaults to `member`, which preserves the previous behaviour in a backwards-compatible manner
- it also accepts `outside_collaborator`
- when the new value is passed, an existing member is turned into an outside collaborator on destroy (i.e. their repository collaborator access is preserved where GitHub allows it)

### Pull request checklist

- [x] ~~Schema migrations have been created if needed ([example](https://github.com/F-Secure-web/terraform-provider-github/blob/main/github/migrate_github_actions_organization_secret.go))~~ (n/a: the proposal is backwards-compatible)

- [ ] Yes
- [x] No, this is a backwards-compatible change. We're not modifying the defaults.

----

### Testing

> [!WARNING]
> `TestAccGithubMembership/downgrades_organization_membership_to_outside_collaborator` requires `GH_TEST_EXTERNAL_USER1_TOKEN` to be a **Classic PAT** with `write:org` permissions. This is because only active organisation members can be turned into outside collaborators, so the invite it issues must be accepted. Unfortunately, the fine-grained tokens do not cover organisation invitation acceptance. The newly added behaviour is also covered by unit tests, so we could remove this acceptance test case if we don't want to use a classic token here.

#### ✅ Acceptance Tests

`TestAccGithubMembership/downgrades_organization_membership_to_outside_collaborator`

1. Creates member-level membership for the external user
2. Accepts the org invitation
3. Adds the user as a direct collaborator with write access to a test repo.
4. Destroys the `github_membership` resource
5. Asserts the user is now an outside collaborator

#### ✅ Unit Tests

- `Test_resourceGithubMembershipDelete/removes membership` - asserts that destroy with `downgrade_on_destroy = false` is successful (uses mocks);
- `Test_resourceGithubMembershipDelete/member default`; asserts that destroy with `downgrade_on_destroy = true` is successful (uses mocks);
- `Test_resourceGithubMembershipDelete/outside collaborator`; asserts that destroy with `downgrade_on_destroy = true` and `downgrade_to = outside_collaborator` is successful (uses mocks);
- `Test_resourceGithubMembershipDelete/outside collaborator pending invitation` - asserts that destroy with `downgrade_on_destroy = true` and `downgrade_to = outside_collaborator` when the org invitation has not been accepted is NOT successful (uses mocks);
- `Test_resourceGithubMembershipDelete/membership not found` - asserts that destroy with `downgrade_on_destroy = true` and `downgrade_to = outside_collaborator` exits cleanly when the user is not an organisation member (uses mocks);
- `Test_resourceGithubMembershipDowngradeToValidation` - verifies valid `downgrade_to` values:
    - `member`
    - `outside_collaborator`

### DIFF
--- a/github/resource_github_membership.go
+++ b/github/resource_github_membership.go
@@ -12,6 +12,12 @@ import (
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
 )
 
+const (
+	membershipDowngradeToMember              = "member"
+	membershipDowngradeToOutsideCollaborator = "outside_collaborator"
+	membershipStateActive                    = "active"
+)
+
 func resourceGithubMembership() *schema.Resource {
 	return &schema.Resource{
 		CreateContext: resourceGithubMembershipCreateOrUpdate,
@@ -45,7 +51,14 @@ func resourceGithubMembership() *schema.Resource {
 				Type:        schema.TypeBool,
 				Optional:    true,
 				Default:     false,
-				Description: "Instead of removing the member from the org, you can choose to downgrade their membership to 'member' when this resource is destroyed. This is useful when wanting to downgrade admins while keeping them in the organization",
+				Description: "Instead of removing the member from the org, you can choose to downgrade their membership when this resource is destroyed. This is useful when wanting to downgrade admins while keeping them in the organization, or to downgrade members while keeping their access to public repositories intact.",
+			},
+			"downgrade_to": {
+				Type:             schema.TypeString,
+				Optional:         true,
+				ValidateDiagFunc: validateValueFunc([]string{membershipDowngradeToMember, membershipDowngradeToOutsideCollaborator}),
+				Default:          membershipDowngradeToMember,
+				Description:      "The target membership state when downgrade_on_destroy is true. Must be one of 'member' or 'outside_collaborator'.",
 			},
 		},
 	}
@@ -64,6 +77,9 @@ func resourceGithubMembershipCreateOrUpdate(ctx context.Context, d *schema.Resou
 	roleName := d.Get("role").(string)
 	if !d.IsNewResource() {
 		ctx = context.WithValue(ctx, ctxId, d.Id())
+		if !d.HasChange("role") {
+			return resourceGithubMembershipRead(ctx, d, meta)
+		}
 	}
 
 	_, resp, err := client.Organizations.EditOrgMembership(ctx,
@@ -148,7 +164,10 @@ func resourceGithubMembershipDelete(ctx context.Context, d *schema.ResourceData,
 
 	username := d.Get("username").(string)
 	downgradeOnDestroy := d.Get("downgrade_on_destroy").(bool)
-	downgradeTo := "member"
+	downgradeTo, ok := d.Get("downgrade_to").(string)
+	if !ok || downgradeTo == "" {
+		downgradeTo = membershipDowngradeToMember
+	}
 
 	if downgradeOnDestroy {
 		tflog.Info(ctx, fmt.Sprintf("Downgrading '%s' membership for '%s' to '%s'", orgName, username, downgradeTo), map[string]any{
@@ -176,7 +195,7 @@ func resourceGithubMembershipDelete(ctx context.Context, d *schema.ResourceData,
 			return diag.FromErr(err)
 		}
 
-		if *membership.Role == downgradeTo {
+		if downgradeTo == membershipDowngradeToMember && membership.GetRole() == downgradeTo {
 			tflog.Info(ctx, fmt.Sprintf("Not downgrading '%s' membership for '%s' because they are already '%s'", orgName, username, downgradeTo), map[string]any{
 				"org_name": orgName,
 				"username": username,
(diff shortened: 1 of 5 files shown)


BigLep:
Project maintainers: this would be really useful for the github organization management situations I have right now. When adjusting membership to my open source organization, I'd like the way to programatically still keep various contributors as collaborators even if I'm freeing up their organization membership seat. This change makes sense to me. What would be the next steps for reviewing and hopefully merging this?

deiga:
@BigLep while we can't promise timelines or capacity for getting this merged, if ypu want you can support us on this. 

By looking through our Contributing and Architecture guides and based on those do a review of this PR (or any other in the repo) 🙏

github/resource_github_membership.go
```diff
@@ -64,6 +77,9 @@ func resourceGithubMembershipCreateOrUpdate(ctx context.Context, d *schema.Resou
 	roleName := d.Get("role").(string)
 	if !d.IsNewResource() {
 		ctx = context.WithValue(ctx, ctxId, d.Id())
+		if !d.HasChange("role") {
+			return resourceGithubMembershipRead(ctx, d, meta)
```
BigLep: main just merged #3637 ("Correct legacy etag handling logic"), which added a `d.Set("etag", nil)` call right before this block in the same function. This branch is behind main so that line isn't in your diff yet, but I merged main into a local copy of this branch to see how the two changes combine, and the etag ends up getting cleared before the `HasChange("role")` check runs. That means an apply that only touches `downgrade_on_destroy` or `downgrade_to` correctly skips the PUT, but it also throws away the etag, so the following read does a full GET instead of a conditional one, losing the caching benefit this resource was just set up for.

`resource_github_branch.go`'s update function does the equivalent `HasChange` check before touching any state, so it might be worth moving this check above the etag reset once you rebase, to keep that same ordering. Also curious what motivated adding this skip, since it's not mentioned in the PR description and doesn't look covered by a test yet.

github/resource_github_membership.go
```diff
@@ -148,7 +164,10 @@ func resourceGithubMembershipDelete(ctx context.Context, d *schema.ResourceData,
 
 	username := d.Get("username").(string)
 	downgradeOnDestroy := d.Get("downgrade_on_destroy").(bool)
-	downgradeTo := "member"
+	downgradeTo, ok := d.Get("downgrade_to").(string)
+	if !ok || downgradeTo == "" {
```
BigLep: `downgradeTo, ok := d.Get("downgrade_to").(string)` followed by the `!ok || downgradeTo == ""` fallback: since the schema already sets `Default: membershipDowngradeToMember`, can this fallback actually be hit? If not, might be worth dropping it, or adding a short note on why it's there.

github/resource_github_membership.go
BigLep, line 81: [ARCHITECTURE.md](https://github.com/integrations/terraform-provider-github/blob/main/ARCHITECTURE.md) is explicit about this exact function: `return nil // Never call Read at end of Update, set any Computed fields in Update`. This new branch calls `resourceGithubMembershipRead` at the end of Update, which also cuts against "Minimize API Calls" in the same doc, since it replaces the PUT it's trying to avoid with a full GET instead. Nothing computed actually changes here when `role` is unchanged, so `return nil` (or skipping the block entirely) looks like a better fit for the guide.

On top of that: main just merged #3637 ("Correct legacy etag handling logic"), which added a `d.Set("etag", nil)` call right before this block in the same function. This branch is behind main so that line isn't in your diff yet, but I merged main into a local copy of this branch to see how the two changes combine, and the etag ends up getting cleared before the `HasChange("role")` check runs. So even with the Read call kept, an apply that only touches `downgrade_on_destroy`/`downgrade_to` throws away the etag first, meaning the read does a full GET instead of a conditional one, losing the caching benefit this resource was just set up for.

`resource_github_branch.go`'s update function has the same Read-at-end-of-Update pattern, so this isn't unprecedented in the codebase, just worth squaring against the doc since it's called out so specifically there. Also curious what motivated adding this skip in the first place, it's not mentioned in the PR description and doesn't look covered by a test yet.

github/resource_github_membership.go
BigLep, line 168: (nit) `downgradeTo, ok := d.Get("downgrade_to").(string)` followed by the `!ok || downgradeTo == ""` fallback: since the schema already sets `Default: membershipDowngradeToMember`, can this fallback actually be hit? If not, might be worth dropping it, or adding a short note on why it's there.

BigLep (changes requested):

BigLep (changes requested):
Note: I am no veteran on this codebase or `go` in general. I spent time going through it with agent help to better understand and ask questions, and see any callouts missed from https://github.com/integrations/terraform-provider-github/blob/main/CONTRIBUTING.md and https://github.com/integrations/terraform-provider-github/blob/main/ARCHITECTURE.md

The outside_collaborator conversion path is covered well: the mock-based delete tests, the pending-invitation error path, and the acceptance test all check out, and I ran the new unit tests locally, they pass. Two comments below: one on how the new update-skip logic here squares with ARCHITECTURE.md and the etag fix that just landed on main in #3637, and a small question about a defensive fallback in the delete function.

PR #3610, opened 2026-08-17
fix: tolerate a repository license the API cannot serve

### LINKED ISSUE #2092: [BUG]: github_repository data source crashes with license not found
### Expected Behavior

no errors

### Actual Behavior

error 
```
 Error: GET https://api.github.com/repos/org/repo/license: 404 Not Found []
│ 
│   with data.github_repository.repository_data["repo"],
```

### Terraform Version

Terraform v1.6.6
on linux_amd64
+ provider registry.terraform.io/hashicorp/google v5.11.0
+ provider registry.terraform.io/hashicorp/google-beta v5.11.0
+ provider registry.terraform.io/integrations/github v5.44.0

### Affected Resource(s)

- data.github_repository

_No response_

### Steps to Reproduce

_No response_

### Debug Output

_No response_

### Panic Output

_No response_

### Code of Conduct

dimoschi:
Resolves #2092

----

### Before the change?

- Adding a `LICENSE` file to a repository and later removing it leaves it classified as `other` with a null license URL, while `GET /repos/{owner}/{repo}/license` returns 404.
- `data.github_repository` calls the license endpoint whenever `license` is non-null and returns any error as-is, so the whole data source fails and you can't read any other attribute of the repository. Reported in #2092 on versions 5.44.0 through 6.6, usually on repositories people can't easily change.

### After the change?

- A 404 from the license endpoint is treated as no license, the same way a missing repository already is, and the rest of the repository is returned. Other errors still fail the read, so real API problems aren't hidden.
- Kept deliberately narrow. It doesn't work around the GitHub behaviour itself, and it doesn't split the license lookup into its own data source, which came up in the issue but is a bigger change.

### Pull request checklist

No schema change, so no migration is needed.

An acceptance test creates a repository from a license template, deletes `LICENSE` and reads it back. I confirmed it fails without the fix, with the same error as the issue. A unit test covers the 404, 200 and 500 branches, including that a 500 still fails the read.

- [ ] Yes
- [x] No

A repository in this state failed the read before, so nothing could have depended on the old behaviour. `repository_license` is now empty for it, the same as a repository with no license.

----

Co-authored with an AI assistant, as described in the AI use policy in CONTRIBUTING.md. I've reviewed and tested the change myself, including running the acceptance test against a real account and confirming it fails without the fix.

### DIFF
--- a/github/acc_helpers_repository_test.go
+++ b/github/acc_helpers_repository_test.go
@@ -10,8 +10,6 @@ import (
 	"github.com/hashicorp/terraform-plugin-testing/helper/acctest"
 )
 
-const licenseFilePath = "LICENSE"
-
 type createTestRepositoryOptionsFunc func(*github.Repository)
 
 func mustCreateTestRepository(t *testing.T, f ...createTestRepositoryOptionsFunc) *github.Repository {
@@ -62,19 +60,19 @@ func mustRenameTestRepository(t *testing.T, repo *github.Repository, newName str
 	}
 }
 
-func mustDeleteTestRepositoryLicense(t *testing.T, repo *github.Repository) {
+func mustDeleteRepositoryFile(t *testing.T, repo *github.Repository, path string) {
 	t.Helper()
 
-	license, _, _, err := testAccConf.meta.v3client.Repositories.GetContents(t.Context(), testAccConf.meta.name, repo.GetName(), licenseFilePath, nil)
+	file, _, _, err := testAccConf.meta.v3client.Repositories.GetContents(t.Context(), testAccConf.meta.name, repo.GetName(), path, nil)
 	if err != nil {
-		t.Fatalf("failed to read %s of test repository %s: %v", licenseFilePath, repo.GetName(), err)
+		t.Fatalf("failed to read %s of test repository %s: %v", path, repo.GetName(), err)
 	}
 
-	_, _, err = testAccConf.meta.v3client.Repositories.DeleteFile(t.Context(), testAccConf.meta.name, repo.GetName(), licenseFilePath, &github.RepositoryContentFileOptions{
-		Message: new(fmt.Sprintf("Remove %s", licenseFilePath)),
-		SHA:     license.SHA,
+	_, _, err = testAccConf.meta.v3client.Repositories.DeleteFile(t.Context(), testAccConf.meta.name, repo.GetName(), path, &github.RepositoryContentFileOptions{
+		Message: new(fmt.Sprintf("Remove %s", path)),
+		SHA:     file.SHA,
 	})
 	if err != nil {
-		t.Fatalf("failed to delete %s of test repository %s: %v", licenseFilePath, repo.GetName(), err)
+		t.Fatalf("failed to delete %s of test repository %s: %v", path, repo.GetName(), err)
 	}
 }
--- a/github/acc_helpers_repository_test.go
+++ b/github/acc_helpers_repository_test.go
@@ -10,6 +10,8 @@ import (
 	"github.com/hashicorp/terraform-plugin-testing/helper/acctest"
 )
 
+const licenseFilePath = "LICENSE"
+
 type createTestRepositoryOptionsFunc func(*github.Repository)
 
 func mustCreateTestRepository(t *testing.T, f ...createTestRepositoryOptionsFunc) *github.Repository {
@@ -57,5 +59,22 @@ func mustRenameTestRepository(t *testing.T, repo *github.Repository, newName str
 	_, _, err := testAccConf.meta.v3client.Repositories.Edit(t.Context(), testAccConf.meta.name, repo.GetName(), &github.Repository{Name: &newName})
 	if err != nil {
 		t.Fatalf("failed to rename test repository %s to %s: %v", repo.GetName(), newName, err)
+	}
+}
+
+func mustDeleteTestRepositoryLicense(t *testing.T, repo *github.Repository) {
+	t.Helper()
+
+	license, _, _, err := testAccConf.meta.v3client.Repositories.GetContents(t.Context(), testAccConf.meta.name, repo.GetName(), licenseFilePath, nil)
+	if err != nil {
+		t.Fatalf("failed to read %s of test repository %s: %v", licenseFilePath, repo.GetName(), err)
+	}
+
+	_, _, err = testAccConf.meta.v3client.Repositories.DeleteFile(t.Context(), testAccConf.meta.name, repo.GetName(), licenseFilePath, &github.RepositoryContentFileOptions{
+		Message: new(fmt.Sprintf("Remove %s", licenseFilePath)),
+		SHA:     license.SHA,
+	})
+	if err != nil {
+		t.Fatalf("failed to delete %s of test repository %s: %v", licenseFilePath, repo.GetName(), err)
 	}
 }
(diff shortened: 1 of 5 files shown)


github/data_source_github_repository.go
```diff
@@ -432,11 +432,20 @@ func dataSourceGithubRepositoryRead(ctx context.Context, d *schema.ResourceData,
...
+		repositoryLicense, _, err := client.Repositories.License(ctx, owner, repoName)
 		if err != nil {
-			return diag.FromErr(err)
+			// A repository can report a license while the license endpoint has nothing to serve,
+			// for example after a LICENSE file is added and later removed, which leaves the
+			// repository classified as `other` with a null license URL. Treat that like a
+			// repository without a license instead of failing the whole read, matching how a
+			// missing repository is handled above.
+			var ghErr *github.ErrorResponse
+			if !errors.As(err, &ghErr) || ghErr.Response.StatusCode != http.StatusNotFound {
```
deiga: issue: Please use the errors.AsType pattern as instructed in CONTRIBUTORS.md

github/data_source_github_repository_test.go
```diff
@@ -376,3 +381,92 @@ data "github_repository" "test" {
 		})
 	})
 }
+
+func Test_dataSourceGithubRepositoryReadLicense(t *testing.T) {
```
deiga, line 417: issue: Please use an acceptance test which sets up a repo with license and deletes it, to actually reproduce the behaviour

deiga (changes requested):

github/data_source_github_repository.go
dimoschi: Done, thanks for pointing to the right direction.

github/data_source_github_repository_test.go
dimoschi: Added a test, check my comment below.

dimoschi:
Both done, thanks @deiga

Added the acceptance test case as "tolerates a license the API cannot serve", following your suggestion.  The repo lands in the bad state right after the file is deleted, so no polling or sleeps are needed. Full `TestAccDataSourceGithubRepository` run is 9 pass, 1 skip (the skip is the existing TODO: Fix this test.).

Also, I kept the unit test instead of replacing it because it also covers the 500 case, which is what stops the fix from swallowing real API errors. An acceptance test can't make GitHub return a 500 on demand. Happy to drop it if you'd rather only have the acceptance test.

templates/data-sources/repository.md.tmpl
deiga, line 1: Suggestion: would you be willing to refactor this template to use the generated fields? It might even suffice to delete the file and allow the default template to render the docs. 

Then you we wouldn't need to modify the template and data source

deiga (commented):
Looks good! Left a discussion to resolve before continuing

templates/data-sources/repository.md.tmpl
dimoschi: I tried deleting it, but then the default template loses the example and renders all attributes empty.

So doing it without losing docs means adding **Description** to the attributes, including the nested ones, moving the example, then deleting the template. That's fine by me, it's just a lot bigger than this fix, and it would leave `github_repository` as the only data source without a template.

I'd prefer to keep this PR as the 404 fix and open a separate one for the docs refactor right after, which also matches the note in CONTRIBUTING about keeping changes focused.

deiga: Sure, a follow-up PR sounds good to me!

deiga (dismissed):
LGMT

dimoschi:
Hello @stevehipwell any chance we can have this reviewed and if nothing comes up merged?

deiga:
@dimoschi To help get your PR reviewed, could you please help us with reviewing other open PRs?
That can help get this and other features implemented faster 🙏

For review guidance, familiarize yourself with our Contributing and Architecture guidelines ☺️

github/acc_helpers_repository_test.go
```diff
@@ -10,6 +10,8 @@ import (
 	"github.com/hashicorp/terraform-plugin-testing/helper/acctest"
 )
 
+const licenseFilePath = "LICENSE"
```
stevehipwell: Why do we need this?

github/acc_helpers_repository_test.go
```diff
@@ -59,3 +61,20 @@ func mustRenameTestRepository(t *testing.T, repo *github.Repository, newName str
...
+	}
+
+	_, _, err = testAccConf.meta.v3client.Repositories.DeleteFile(t.Context(), testAccConf.meta.name, repo.GetName(), licenseFilePath, &github.RepositoryContentFileOptions{
+		Message: new(fmt.Sprintf("Remove %s", licenseFilePath)),
+		SHA:     license.SHA,
+	})
+	if err != nil {
+		t.Fatalf("failed to delete %s of test repository %s: %v", licenseFilePath, repo.GetName(), err)
+	}
+}
```
stevehipwell: Why do we need a whole function to delete a single repo file path, please can this be converted to `mustDeleteRepositoryFile`.

stevehipwell (changes requested):
Is this behaviour documented by GitHub anywhere?

CC @robert-crandall

github/acc_helpers_repository_test.go
dimoschi: We don't. I pulled it out because LICENSE appeared a few times in the one function, which isn't worth a package level const. Removed in https://github.com/integrations/terraform-provider-github/pull/3610/commits/34deed21b638b9da2ee8085867b30aa992bb8005.

github/acc_helpers_repository_test.go
dimoschi: No good reason, I wrote it for the single call site and named it after that. I will convert to `mustDeleteRepositoryFile(t, repo, path)`, with the test passing LICENSE in. Fixed in https://github.com/integrations/terraform-provider-github/pull/3610/commits/34deed21b638b9da2ee8085867b30aa992bb8005

stevehipwell:
@dimoschi you might want to rebase this PR while you're at it?

stevehipwell (dismissed):
LGTM

github/data_source_github_repository_test.go
```diff
@@ -376,4 +381,133 @@ data "github_repository" "test" {
...
+					w.WriteHeader(http.StatusOK)
+					_, _ = w.Write([]byte(repoResponse))
+					return
+				}
+
+				w.WriteHeader(http.StatusNotFound)
+			}))
+			t.Cleanup(ts.Close)
+
+			client, err := github.NewClient(github.WithURLs(new(ts.URL+"/"), nil))
```
deiga: question: could this be replaced with `githubApiMock` and `mustCreateTestGitHubClient`?

deiga (changes requested):
Looks good, but the unit test might have room for improvement

github/data_source_github_repository_test.go
dimoschi: I think it does, let me take a hit on it. It should tighten the test up as well, since the mock asserts the exact URIs and the call order, and my handler doesn't.

dimoschi: Thanks for pointing out, I've totally missed it. Fixed in https://github.com/integrations/terraform-provider-github/pull/3610/changes/36e99ccbf8def04b8aa54f508824aef6d9e3becb

deiga (approved):
LGTM!

stevehipwell (approved):
LGTM

robert-crandall (approved):
Thanks for the focused fix! I’m happy with treating a missing license as empty while preserving other API errors.

PR #3464, opened 2026-06-01
feat: add github_organization_invitation resource

nitinjain999:
Closes #2197

Picking up the work from #3229 (original author inactive) and addressing all maintainer review feedback.

----

### Before the change?

`github_membership` only works when a user already has a GitHub account and is identified by their username. There was no way to invite someone to an org by email address, which is a common need when onboarding new users who may not yet have GitHub accounts.

### After the change?

New resource `github_organization_invitation`:

```hcl
# Invite by email
resource "github_organization_invitation" "example" {
  email = "newmember@example.com"
  role  = "direct_member"
}

# Invite by GitHub user ID
data "github_user" "invitee" {
  username = "someuser"
}

resource "github_organization_invitation" "example" {
  invitee_id = data.github_user.invitee.id
}
```

### Changes vs #3229

All three issues flagged by @deiga are addressed, plus one additional bug found during testing:

1. **Redundant `invitation_id` computed field removed** — the resource ID already is the invitation ID
2. **`ctx = context.WithValue(ctx, ctxId, ...)` removed** from Read and Delete — not needed
3. **Idempotency fixed** — when an invitation is no longer in the pending list, Read now checks whether the invitee accepted and became an org member via `IsMember`. If they are a member, the resource stays in state. Only if the invitee is neither pending nor a member (expired/cancelled) is the resource removed from state
4. **Email drift bug fixed** — the GitHub API always returns the invitee's email in the invitation response, even for `invitee_id`-based invitations. The original PR set `email` from the API response unconditionally, which caused a perpetual destroy/recreate diff when `invitee_id` was the configured field. Now `email` is only written to state when the resource was created with `email`

### Testing

`invite_by_invitee_id` **PASSES** verified against a real org (`nitin-sv`) with `deiga` as the external user.

`invite_by_email` and `invite_with_admin_role` skip when `GH_TEST_INVITATION_EMAIL` is not set — they require an email address not already associated with an org member.

### Pull request checklist

- [x] Schema migrations have been created if needed — not applicable
- [x] Tests for the changes have been added
- [x] Docs have been reviewed and added / updated if needed

### Does this introduce a breaking change?

- [ ] Yes
- [x] No

----

PR #3462, opened 2026-05-30
feat: add github_copilot_organization_seat_assignment and github_copilot_team_seat_assignment resources

nitinjain999:
Closes #1629

----

### Before the change?

There was no way to manage or inspect GitHub Copilot configuration via this provider. Organisations had to assign seats and check settings manually through the GitHub UI or call the API directly.

### After the change?

**Two new resources** (require seat management set to `assign_selected`):

**`github_copilot_organization_seat_assignment`** — assigns a Copilot seat to a specific org member:
```hcl
resource "github_copilot_organization_seat_assignment" "example" {
  username = "someuser"
}
```
Computed attributes: `created_at`, `pending_cancellation_date`, `last_activity_at`, `last_activity_editor`, `plan_type`.

**`github_copilot_team_seat_assignment`** — assigns Copilot seats to all members of a team:
```hcl
resource "github_copilot_team_seat_assignment" "example" {
  team = github_team.example.slug
}
```

**One new data source** (works for any org with Copilot enabled, no `assign_selected` required):

**`github_copilot_organization_settings`** — reads org-level Copilot billing settings:
```hcl
data "github_copilot_organization_settings" "org" {}

output "seat_management" {
  value = data.github_copilot_organization_settings.org.seat_management_setting
}
```
Attributes: `seat_management_setting`, `public_code_suggestions`, `seat_breakdown` (total, active, pending, etc.).

The data source exists because the GitHub API has no write endpoint for org Copilot settings — they can only be changed via the GitHub UI.

**Known limitation — team Read performance:** The GitHub API has no single-team Copilot seat lookup endpoint. `github_copilot_team_seat_assignment` Read paginates through all org seats to find the team entry. For large orgs with many Copilot seats this adds latency on every `plan`/`refresh`. This is a GitHub API constraint — if a dedicated endpoint is added in future it can be adopted without a breaking change.

### Testing

- `github_copilot_organization_settings` data source: **acceptance test passes** against a real Copilot-enabled org (`nitin-sv`)
- `github_copilot_organization_seat_assignment` and `github_copilot_team_seat_assignment`: acceptance tests skip automatically when seat management is not set to `assign_selected`, with a clear message

### Pull request checklist

- [x] Schema migrations have been created if needed — not applicable
- [x] Tests for the changes have been added (for bug fixes / features)
- [x] Docs have been reviewed and added / updated if needed

### Does this introduce a breaking change?

- [ ] Yes
- [x] No

----

PR #3616, opened 2026-08-22
fix: detect out-of-band removal of environment reviewers on refresh

### LINKED ISSUE #3609: [BUG]: Out-of-band removal of environment reviewers is not detected on refresh
### Expected Behavior

When a required reviewer is removed from an environment outside Terraform, the next `terraform plan` should report a change, because the configuration still declares the reviewer and GitHub no longer has it.

### Actual Behavior

`terraform plan` reports `No changes`, and the reviewer stays removed. The drift is invisible for as long as the configuration is left alone.

I checked whether the API is at fault and it is not. `GET /repos/{owner}/{repo}/environments/{name}` reports the removal faithfully: once the last reviewer is gone, the response's `protection_rules` array simply has no `required_reviewers` entry in it. The information the provider needs is in the response.

The cause is in `resourceGithubRepositoryEnvironmentRead`, in `github/resource_github_repository_environment.go`. Five fields are populated by three different patterns, and only two of them survive an entry disappearing from `protection_rules`:

| Field | How it is populated | Refreshes on removal? |
|---|---|---|
| `wait_timer` | Reset to `nil` before the loop | Yes |
| `can_admins_bypass` | Read from the top-level `env` object | Yes |
| `deployment_branch_policy` | `if env.DeploymentBranchPolicy != nil { … } else { d.Set(…, []any{}) }` | Yes |
| `reviewers` | Set **only** inside `case "required_reviewers":` | **No** |
| `prevent_self_review` | Set **only** inside the same case | **No** |

`wait_timer` gets it right, and it is worth reading as the fix already present in the same function:

```go
if err := d.Set("wait_timer", nil); err != nil {
	return diag.FromErr(err)
}
```

That pre-loop reset means a vanished `wait_timer` entry leaves `nil` behind. `reviewers` and `prevent_self_review` have no equivalent, so when the `required_reviewers` case never executes, `d.Set` is never called for either and the prior state value survives the refresh untouched. `terraform plan` then compares the configuration against stale state and correctly reports no changes about a state that is wrong.

`prevent_self_review` is caught by the same defect for the same reason, since it is set inside that case as well.

This looks like a two-line fix: reset both fields before the loop, the way `wait_timer` already is.

Note this is not #3156, which was about removing `reviewers` through configuration and is fixed. This is the reverse direction: a removal made outside Terraform is never detected.

### Terraform Version

```
Terraform v1.15.8
on darwin_arm64
```

Also read at `main`, where `github/resource_github_repository_environment.go` is byte-identical to `v6.13.0` apart from a `go-github` v88 to v89 bump and a type-assertion style change. Neither touches this logic.

### GitHub Installation Type

GitHub.com (Free, Pro, or Team)

### Affected Resource(s)

- `github_repository_environment`

```hcl
resource "github_repository_environment" "example" {
  repository  = "example-repo"
  environment = "production"

  reviewers {
    users = [1234567]
  }
}
```

### Steps to Reproduce

1. `terraform apply` the configuration above, on a public repository or a plan tier that allows required reviewers.
2. In the GitHub UI, go to Settings > Environments > production and remove the reviewer, so no reviewers remain.
3. `terraform plan`.

Expected: a change is reported, restoring the reviewer. Actual: `No changes`.

Comparing the API response before and after step 2 shows the `required_reviewers` entry present, then absent, with the rest of the object unchanged.

### Code of Conduct

nitinjain999:
## Summary

Fixes #3609.

`reviewers` and `prevent_self_review` in `resourceGithubRepositoryEnvironmentRead` were only ever written to state inside `case "required_reviewers":` of the `protection_rules` loop. Once the last required reviewer is removed outside Terraform, the GitHub API stops returning a `required_reviewers` entry in `protection_rules` at all, so that case never runs, `d.Set` is never called for either field, and the prior state value silently survives the refresh. `terraform plan` then reports `No changes` even though the reviewer is gone.

`wait_timer` already avoids this exact problem by being reset to `nil` before the loop runs. This PR applies the same pattern to `reviewers` and `prevent_self_review`.

## Changes

- Reset `reviewers` to `[]any{}` and `prevent_self_review` to `false` before the `protection_rules` loop in `github/resource_github_repository_environment.go`, so a rule missing from the API response is correctly treated as absent.
- Added a unit test (`github/resource_github_repository_environment_unit_test.go`) that mocks a `GetEnvironment` response with `protection_rules: []` (the exact shape GitHub returns once the last reviewer is removed, per the debug logs in the issue) against a prior state that has reviewers set, and asserts both fields refresh to empty/false.

## Test plan

- [x] Added a unit test that fails against the pre-fix code and passes after the fix (no live GitHub credentials required).
- [x] `go build ./...`, `go vet ./github/...`, `gofmt -l` all clean.
- [x] Full non-acceptance test suite passes (`go test ./github/... -run '^Test[^A]|^TestA[^c]|^TestAc[^c]'`).
- [ ] Maintainer to confirm against a real environment (steps to reproduce are in the issue) if desired.


nitinjain999:
This PR is functionally ready for review — it's still marked as a draft only because this account is currently at the repo's open-PR limit (#3464, #3462 also open). Happy to flip it to "Ready for review" as soon as a slot frees up, or a maintainer is welcome to do so directly. @deiga @stevehipwell

deiga:
~I think this is a duplicate of https://github.com/integrations/terraform-provider-github/pull/3583~

~Please coordinate there instead of opening a new PR~

I'll undraft this and draft the others

github/resource_github_repository_environment_unit_test.go
deiga, line 1: issue: don't use unit tests to prove resource functionality, create a acceptance test instead

github/resource_github_repository_environment.go
```diff
@@ -227,6 +227,12 @@ func resourceGithubRepositoryEnvironmentRead(ctx context.Context, d *schema.Reso
 	if err := d.Set("wait_timer", nil); err != nil {
 		return diag.FromErr(err)
 	}
+	if err := d.Set("reviewers", []any{}); err != nil {
+		return diag.FromErr(err)
+	}
+	if err := d.Set("prevent_self_review", false); err != nil {
+		return diag.FromErr(err)
+	}
```
deiga: issue: If the current problem is conditionally setting values, then the conditions need to be improved instead of setting these values to empty

deiga (changes requested):

github/resource_github_repository_environment.go
nitinjain999: Restructured `resourceGithubRepositoryEnvironmentRead` in `github/resource_github_repository_environment.go` so nothing is pre-set to empty. `wait_timer`, `reviewers` and `prevent_self_review` are now collected in a single pass over `env.ProtectionRules` and set unconditionally afterwards, which is the same shape as the existing `deployment_branch_policy` if/else. That also removes the pre-existing `d.Set("wait_timer", nil)` reset. ✅ Fixed

github/resource_github_repository_environment_unit_test.go
nitinjain999: Dropped `github/resource_github_repository_environment_unit_test.go` entirely and added a `detects_out_of_band_reviewer_removal` subtest to `TestAccGithubRepositoryEnvironment` in `github/resource_github_repository_environment_test.go`. It creates the environment with a team reviewer and `prevent_self_review = true`, then `PreConfig` clears the reviewers through `Repositories.CreateUpdateEnvironment` so GitHub drops the `required_reviewers` rule, and a `RefreshState` step asserts the post-refresh plan is a non-empty in-place update via `plancheck.ExpectResourceAction`. A final apply step confirms the drift is corrected. Same `PreConfig` + `RefreshState` + `ExpectNonEmptyPlan` shape as the existing out-of-band tests in `resource_github_repository_vulnerability_alerts_test.go`. ✅ Fixed

github/resource_github_repository_environment.go
```diff
@@ -224,19 +224,20 @@ func resourceGithubRepositoryEnvironmentRead(ctx context.Context, d *schema.Reso
...
+	)
+	reviewers := make([]any, 0)
 
 	for _, pr := range env.ProtectionRules {
 		switch pr.GetType() {
 		case "wait_timer":
-			if err = d.Set("wait_timer", pr.WaitTimer); err != nil {
-				return diag.FromErr(err)
-			}
+			waitTimer = pr.WaitTimer
```
deiga, line 240: If there is a GetWaitTimer funcrion, please use that

github/resource_github_repository_environment.go
```diff
@@ -254,21 +255,30 @@ func resourceGithubRepositoryEnvironmentRead(ctx context.Context, d *schema.Reso
...
 					"users": users,
 				},
-			}); err != nil {
-				return diag.FromErr(err)
-			}
-
-			if err = d.Set("prevent_self_review", pr.PreventSelfReview); err != nil {
-				return diag.FromErr(err)
 			}
+			preventSelfReview = pr.PreventSelfReview
```
deiga, line 265: Use a getter function if available

github/resource_github_repository_environment.go
```diff
@@ -254,21 +255,30 @@ func resourceGithubRepositoryEnvironmentRead(ctx context.Context, d *schema.Reso
...
+	if err := d.Set("wait_timer", waitTimer); err != nil {
+		return diag.FromErr(err)
+	}
+	if err := d.Set("reviewers", reviewers); err != nil {
+		return diag.FromErr(err)
+	}
+	if err := d.Set("prevent_self_review", preventSelfReview); err != nil {
+		return diag.FromErr(err)
+	}
+	if err := d.Set("can_admins_bypass", env.CanAdminsBypass); err != nil {
```
deiga, line 278: Use a getter function if available

github/resource_github_repository_environment_test.go
```diff
@@ -243,6 +245,87 @@ resource "github_repository_environment" "test" {
...
+resource "github_repository" "test" {
+	name       = "%[1]s"
+	visibility = "public"
+}
+
+resource "github_team_repository" "test" {
+	team_id    = github_team.test.id
+	repository = github_repository.test.name
+	permission = "pull"
+}
```
deiga, line 271: Use the mustCreateTest functions to setup test resources

github/resource_github_repository_environment_test.go
```diff
@@ -243,6 +245,87 @@ resource "github_repository_environment" "test" {
...
+					// state in place, leaving a non-empty plan that restores them.
+					PreConfig: func() {
+						if _, _, err := testAccConf.meta.v3client.Repositories.CreateUpdateEnvironment(t.Context(), testAccConf.meta.name, repoName, envName, &github.CreateUpdateEnvironment{
+							Reviewers:       []*github.EnvReviewers{},
+							CanAdminsBypass: new(true),
+						}); err != nil {
+							t.Errorf("failed to remove environment reviewers out-of-band: %s", err)
+						}
+					},
+					RefreshState:       true,
```
deiga, line 309: This step seems to be missing the Config attribute

deiga (changes requested):

stevehipwell:
@nitinjain999 please comment on the threads as you address review comments and when you're ready for a re-review please add a comment summarising the changes you've made.

PR #3619, opened 2026-08-27
fix: read every default_value type for github_organization_custom_properties

### LINKED ISSUE #3618: [BUG]: github_organization_custom_properties: default_value of true_false properties reads back empty, causing a permanent plan diff
### Expected Behavior

A `github_organization_custom_properties` resource with `value_type = "true_false"` and `default_value = "false"` (or `"true"`) should plan clean after apply: GitHub stores the default, so Terraform should read it back unchanged.

### Actual Behavior

Every plan shows the default being (re-)added, forever:

```
  # github_organization_custom_properties.this["fossa"] will be updated in-place
  ~ resource "github_organization_custom_properties" "this" {
      + default_value      = "false"
        id                 = "fossa"
    }
```

GitHub has the value: `GET /orgs/{org}/properties/schema/fossa` returns `"default_value": "false"` and repositories report `"false"` for the property. The provider's Read uses `customProperty.DefaultValueString()`, which in go-github only returns a value for `string`, `single_select` and `url` types (it is documented as such); for `true_false` (and `multi_select`) it returns `("", false)`, so `default_value` is always written to state as `""`. The code carries a `// TODO: Add support for other types of default values` at that line.

A secondary issue in the same resource: Create/Update always sends `default_value` in the request body, as `""` when unset. GitHub does not accept an empty default for non-required properties, so `required = false` without a default cannot be applied reliably.

### Terraform Version

Terraform v1.14.9, provider integrations/github v6.13.0 (also present on `main`).

### GitHub Installation Type

GitHub.com (organization on the Team plan)

### Affected Resource(s)

- `github_organization_custom_properties`

```hcl
resource "github_organization_custom_properties" "fossa" {
  property_name = "fossa"
  value_type    = "true_false"
  required      = true
  default_value = "false"
  description   = "Require the fossa status check"
}
```

### Steps to Reproduce

1. `terraform apply` the configuration above (succeeds; the property is created with default `"false"`).
2. `terraform plan` → `+ default_value = "false"` is planned again. Repeat indefinitely.

### Debug Output

n/a — reproducible from the plan output above; root cause is visible in `resourceGithubCustomPropertiesRead` (`DefaultValueString()` for a `true_false` property).

### Code of Conduct

AhmadMasry:
Resolves #3618

----

### Before the change?

- `resourceGithubCustomPropertiesRead` obtains the default via go-github's `DefaultValueString()`, which only returns a value for `string`, `single_select` and `url` property types. For `true_false` (and `multi_select`) properties it returns `("", false)`, so `default_value` is always stored as `""` and every plan shows `+ default_value = "false"` even though GitHub has the default (`GET /orgs/{org}/properties/schema/{name}` returns it, repositories report it). The code carried a `// TODO: Add support for other types of default values`.
- Create/Update always sent `default_value` in the request body — as `""` when unset — which GitHub rejects for non-required properties.

### After the change?

- New `customPropertyDefaultValueString()` renders the default for every value type using go-github's typed accessors (`DefaultValueBool` → `"true"`/`"false"`, `DefaultValueStrings` → comma-joined, `DefaultValueString` otherwise), with a raw-string fallback. `true_false` properties with a default now plan clean after apply.
- `default_value` is only included in the Create/Update body when configured, so `required = false` without a default applies.
- Unit tests for the helper (`go test ./github -run TestCustomPropertyDefaultValueString`), and an acceptance test that creates a `true_false` property with `default_value = "false"` and asserts an empty follow-up plan.

### Pull request checklist

- [ ] Schema migrations have been created if needed ([example](https://github.com/F-Secure-web/terraform-provider-github/blob/main/github/migrate_github_actions_organization_secret.go)) — not needed, no schema change
- [x] Docs have been reviewed and added / updated if needed (for bug fixes / features) — no doc change needed; `default_value` docs already describe the intended behaviour

- [ ] Yes
- [x] No

Note: `multi_select` defaults are now read back comma-joined into the string attribute (previously `""`). Users who never set `default_value` on a `multi_select` property are unaffected; users who did will see a one-time correction instead of a permanent diff.

----

### DIFF
--- a/github/resource_github_organization_custom_properties.go
+++ b/github/resource_github_organization_custom_properties.go
@@ -2,6 +2,8 @@ package github
 
 import (
 	"context"
+	"strconv"
+	"strings"
 
 	"github.com/google/go-github/v89/github"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/customdiff"
@@ -92,9 +94,14 @@ func resourceGithubCustomPropertiesCreate(d *schema.ResourceData, meta any) erro
 		PropertyName:  &propertyName,
 		ValueType:     valueType,
 		Required:      &required,
-		DefaultValue:  &defaultValue,
 		Description:   &description,
 		AllowedValues: allowedValuesString,
+	}
+	// Only send a default when one is configured; an empty string is not a
+	// valid default for any value type and is rejected for non-required
+	// properties.
+	if defaultValue != "" {
+		customProperty.DefaultValue = &defaultValue
 	}
 
 	if val, ok := d.GetOk("values_editable_by"); ok {
@@ -121,8 +128,7 @@ func resourceGithubCustomPropertiesRead(d *schema.ResourceData, meta any) error 
 		return err
 	}
 
-	// TODO: Add support for other types of default values
-	defaultValue, _ := customProperty.DefaultValueString()
+	defaultValue := customPropertyDefaultValueString(customProperty)
 
 	d.SetId(*customProperty.PropertyName)
 	_ = d.Set("allowed_values", customProperty.AllowedValues)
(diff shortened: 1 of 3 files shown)


deiga:
This PR is duplicating changes in https://github.com/integrations/terraform-provider-github/pull/3234

Please contribute there instead of opening a new one

PR #3627, opened 2026-09-02
fix: wait for hosted runner public IP provisioning

### LINKED ISSUE #3626: [BUG]: `github_actions_hosted_runner` does not wait for public IP provisioning

krukowskid:
Resolves #3626

----

### Before the change?

`github_actions_hosted_runner` returned after a single read following create or update. GitHub provisions hosted runners asynchronously, so Terraform could finish while the runner was still provisioning and persist an empty `public_ips` value.

### After the change?

Create and update wait until the runner is ready. When static public IPs are enabled, they also wait until GitHub returns the allocated IPs, so dependent outputs are populated by the completed operation.

The implementation was prepared with Copilot assistance. Tests were run locally

### Pull request checklist

- [x] Schema migrations have been created if needed (not needed; the state schema shape is unchanged)

- [ ] Yes
- [x] No

----

### DIFF
--- a/github/resource_github_actions_hosted_runner.go
+++ b/github/resource_github_actions_hosted_runner.go
@@ -336,7 +336,7 @@ func resourceGithubActionsHostedRunnerCreate(d *schema.ResourceData, meta any) e
 		return fmt.Errorf("failed to get runner ID from response: %+v", runner)
 	}
 
-	if err := waitForRunnerReady(ctx, client, orgName, d.Id(), d.Get("public_ip_enabled").(bool), d.Timeout(schema.TimeoutCreate)); err != nil {
+	if err := waitForRunnerReady(ctx, client, orgName, d.Id(), nil, d.Get("public_ip_enabled").(bool), d.Timeout(schema.TimeoutCreate)); err != nil {
 		return err
 	}
 
@@ -495,14 +495,14 @@ func resourceGithubActionsHostedRunnerUpdate(d *schema.ResourceData, meta any) e
 		}
 	}
 
-	if err := waitForRunnerReady(ctx, client, orgName, runnerID, d.Get("public_ip_enabled").(bool), d.Timeout(schema.TimeoutUpdate)); err != nil {
+	if err := waitForRunnerReady(ctx, client, orgName, runnerID, payload, d.Get("public_ip_enabled").(bool), d.Timeout(schema.TimeoutUpdate)); err != nil {
 		return err
 	}
 
 	return resourceGithubActionsHostedRunnerRead(d, meta)
 }
 
-func waitForRunnerReady(ctx context.Context, client *github.Client, orgName, runnerID string, requirePublicIPs bool, timeout time.Duration) error {
+func waitForRunnerReady(ctx context.Context, client *github.Client, orgName, runnerID string, expectedUpdate map[string]any, requirePublicIPs bool, timeout time.Duration) error {
 	conf := &retry.StateChangeConf{
 		Pending: []string{"pending"},
 		Target:  []string{"ready"},
@@ -525,7 +525,7 @@ func waitForRunnerReady(ctx context.Context, client *github.Client, orgName, run
 				return nil, "", fmt.Errorf("no runner data returned from API")
 			}
 
-			state, err := hostedRunnerProvisioningState(runner, requirePublicIPs)
+			state, err := hostedRunnerProvisioningState(runner, expectedUpdate, requirePublicIPs)
 			return runner, state, err
 		},
 		Timeout:    timeout,
@@ -537,7 +537,7 @@ func waitForRunnerReady(ctx context.Context, client *github.Client, orgName, run
 	return err
 }
 
-func hostedRunnerProvisioningState(runner map[string]any, requirePublicIPs bool) (string, error) {
+func hostedRunnerProvisioningState(runner, expectedUpdate map[string]any, requirePublicIPs bool) (string, error) {
 	status, ok := runner["status"].(string)
 	if !ok {
 		return "", fmt.Errorf("failed to get hosted runner status from response: %+v", runner)
@@ -550,6 +550,10 @@ func hostedRunnerProvisioningState(runner map[string]any, requirePublicIPs bool)
 		return "pending", nil
 	}
 
+	if !hostedRunnerUpdateApplied(runner, expectedUpdate) {
+		return "pending", nil
+	}
+
 	if requirePublicIPs {
 		publicIPs, ok := runner["public_ips"].([]any)
 		if !ok || len(publicIPs) == 0 {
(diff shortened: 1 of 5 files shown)


github/resource_github_actions_hosted_runner.go
```diff
@@ -489,9 +495,71 @@ func resourceGithubActionsHostedRunnerUpdate(d *schema.ResourceData, meta any) e
...
+
 	return resourceGithubActionsHostedRunnerRead(d, meta)
 }
 
+func waitForRunnerReady(ctx context.Context, client *github.Client, orgName, runnerID string, requirePublicIPs bool, timeout time.Duration) error {
+	conf := &retry.StateChangeConf{
+		Pending: []string{"pending"},
+		Target:  []string{"ready"},
+		Refresh: func() (any, string, error) {
+			req, err := client.NewRequest(ctx, "GET", fmt.Sprintf("orgs/%s/actions/hosted-runners/%s", orgName, runnerID), nil)
```
deiga, line 510: Is there a reason this uses bare requests and not client methods?

github/resource_github_actions_hosted_runner.go
```diff
@@ -489,9 +495,71 @@ func resourceGithubActionsHostedRunnerUpdate(d *schema.ResourceData, meta any) e
 		}
 	}
 
+	if err := waitForRunnerReady(ctx, client, orgName, runnerID, d.Get("public_ip_enabled").(bool), d.Timeout(schema.TimeoutUpdate)); err != nil {
```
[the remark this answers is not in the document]
krukowskid: Fixed in df9fd2df. The update waiter now requires the GET response to match every submitted PATCH value (including the mapped size, static-IP, and image-version fields) before accepting `Ready`. The acceptance test now enables `public_ip_enabled` from false to true and verifies both `Ready` and `public_ips.0.prefix`.

deiga (changes requested):
The changes look good, barring the open discussion.

@stevehipwell Since I have the https://github.com/integrations/terraform-provider-github/pull/3521 PR which does all the necessary refactors, this PR IMO doesn't need to bother with them.

deiga:
@krukowskid please rebase, address the failing CI checks and look into the open discussion

github/resource_github_actions_hosted_runner.go
```diff
@@ -489,9 +495,131 @@ func resourceGithubActionsHostedRunnerUpdate(d *schema.ResourceData, meta any) e
...
 	}
 
+	if err := waitForRunnerReady(ctx, client, orgName, runnerID, payload, d.Get("public_ip_enabled").(bool), d.Timeout(schema.TimeoutUpdate)); err != nil {
+		return err
+	}
+
 	return resourceGithubActionsHostedRunnerRead(d, meta)
 }
 
+func waitForRunnerReady(ctx context.Context, client *github.Client, orgName, runnerID string, expectedUpdate map[string]any, requirePublicIPs bool, timeout time.Duration) error {
```
stevehipwell, line 505: This is unnecessary, we have the generic `retryUntilResourceFound` function to support this pattern.

github/resource_github_actions_hosted_runner.go
```diff
@@ -27,6 +27,8 @@ func resourceGithubActionsHostedRunner() *schema.Resource {
 		},
 
 		Timeouts: &schema.ResourceTimeout{
+			Create: schema.DefaultTimeout(30 * time.Minute),
+			Update: schema.DefaultTimeout(30 * time.Minute),
```
stevehipwell, line 31: Do you have documentation as to why you're picking 30 mins here? IMHO a better solution would be to switch to the context aware functions which have a default 20 mins timeout. The delete timeout should be removed in the next major version a it's not consistent with the rest of the provider.

stevehipwell (changes requested):
This is just a first pass review.

github/resource_github_actions_hosted_runner.go
deiga: That function doesn't support waiting for a specific state of a resource AFAIK

krukowskid: as deiga noted, `retryUntilResourceFound` can't wait for specific status

github/resource_github_actions_hosted_runner.go
krukowskid: consistency with other calls in the file. Happy to switch if you'd prefer it regardless.

github/resource_github_actions_hosted_runner.go
krukowskid: There's no documented provisioning time from GitHub. In my experience it takes from 5 to 15 minutes (static IP allocation being the slow part), so a 30-minute default should cover most deployments while letting users adjust it via timeouts if needed. Happy to change it to 20 minutes to match the SDK default if you'd prefer.

github/resource_github_actions_hosted_runner.go
deiga: Yes, since there are SDK methods available those should be used

krukowskid: changed to client method

github/resource_github_actions_hosted_runner.go
stevehipwell: Ok, but `retryUntilOK` which `retryUntilResourceFound` wraps does.

github/resource_github_actions_hosted_runner.go
stevehipwell: If it's not documented then you need to switch to the context based pattern and drop the custom timeouts as the first implementation. From an API design perspective I'd suggest that if the value can't be looked up that we store the zero value and allow a re-run to pick up the actual value.

stevehipwell (changes requested):

github/resource_github_actions_hosted_runner.go
krukowskid: I'm happy to change the default, but storing the zero value and letting a re-run pick it up is the current behaviour.

github/resource_github_actions_hosted_runner.go
krukowskid: I tested it and `retryUntilOK` with a 5-minute timeout and a never-true condition fails after ~3 min with couldn't find resource (21 retries). A false `ok` returns a nil result, which `StateChangeConf` counts toward `NotFoundChecks` (default 20, and retryOptions.maxRetries isn't wired to it). Since runners provision for 5 - 15 min, it would fail every create as-is.

The fix is small, and I can do it here, but @deiga preferred keeping other changes out of this PR, so deferring to you two on scope.

stevehipwell: A fix for `retryUntilOK`? If so you can add it here and we can decide if it needs a seperate PR.

github/resource_github_actions_hosted_runner.go
stevehipwell: The current behaviour doesn't attempt to wait for it to be created, using context and `retryUntilOK` changes when the zero value would be stored.

PR #3635, opened 2026-09-07
fix: Set correct transport order for new client cache

### LINKED ISSUE #3612: [BUG]: Transport ordering makes the response cache a no-op when legacy_client = false
### Expected Behavior

With `legacy_client = false`, the provider builds a REST client that includes a conditional-request cache ([`ghct.NewTransport`](https://github.com/integrations/terraform-provider-github/blob/67f3fd10bda01461da6241543c1cd07e8e553f86/internal/ghclient/transport.go#L61-L68), `github.com/bored-engineer/github-conditional-http-transport`). `Cache: true` is [set unconditionally](https://github.com/integrations/terraform-provider-github/blob/67f3fd10bda01461da6241543c1cd07e8e553f86/github/provider.go#L609-L618) for this client.

Repeated `GET`s of an unchanged resource should revalidate via `If-None-Match` and receive `304 Not Modified`, which GitHub does not charge against the primary rate limit.

### Actual Behavior

The cache never produces a hit. Every request is a full `200`.

`newTransport` composes the chain so the cache is **outermost** and `oauth2.Transport` is **innermost** ([`internal/ghclient/transport.go`](https://github.com/integrations/terraform-provider-github/blob/67f3fd10bda01461da6241543c1cd07e8e553f86/internal/ghclient/transport.go#L32-L71)):

```
cloneTransport -> oauth2 -> logging -> retryablehttp -> throttler -> ratelimit -> ghct
                  ^ innermost                                                     ^ outermost
```

[`oauth2` is applied first](https://github.com/integrations/terraform-provider-github/blob/67f3fd10bda01461da6241543c1cd07e8e553f86/internal/ghclient/transport.go#L35-L40) so it ends up deepest; [ghct is applied last](https://github.com/integrations/terraform-provider-github/blob/67f3fd10bda01461da6241543c1cd07e8e553f86/internal/ghclient/transport.go#L61-L68) so it ends up on top. RoundTripper chains execute outermost-first, so **ghct sees every request before the `Authorization` header exists**.

GitHub derives ETags from `Authorization` and advertises `Vary: Accept, Authorization`. ghct handles this — its [`VaryHeaders`](https://github.com/bored-engineer/github-conditional-http-transport/blob/94778cbb26ea34bb63c577dfffaa99ebfb59e1cc/hash.go#L13-L18) are exactly `Accept`, `Authorization`, `Cookie`. With the credential invisible, three paths fail:

| # | Path | Source | What goes wrong |
|---|---|---|---|
| 1 | store | [`transport.go:165-174`](https://github.com/bored-engineer/github-conditional-http-transport/blob/94778cbb26ea34bb63c577dfffaa99ebfb59e1cc/transport.go#L165-L174) | `req.Header.Values("Authorization")` is empty, so the `X-Varied-Authorization` marker is never written |
| 2 | vary/reuse | [`vary.go:21-25`](https://github.com/bored-engineer/github-conditional-http-transport/blob/94778cbb26ea34bb63c577dfffaa99ebfb59e1cc/vary.go#L21-L25) | compares `HashToken("")` against a missing stored value — always false |
| 3 | recompute | [`conditional.go:45-50`](https://github.com/bored-engineer/github-conditional-http-transport/blob/94778cbb26ea34bb63c577dfffaa99ebfb59e1cc/conditional.go#L45-L50) | recomputes the ETag *without* `Authorization` while GitHub hashed *with* it |

**Two things this is not.** Authentication still works — `oauth2` authenticates below the cache, so requests reach GitHub fully credentialed; "the cache broke auth" is the wrong reading. And credential rotation is not implicated: [`addConditionalHeaders`](https://github.com/bored-engineer/github-conditional-http-transport/blob/94778cbb26ea34bb63c577dfffaa99ebfb59e1cc/conditional.go#L13-L53) has a third branch that recomputes from the cached body, which works correctly once `Authorization` is visible.

**Impact** is wasted rate limit and latency only — no correctness or state issue. Users on the default `legacy_client = true` are unaffected, which likely explains why this went unnoticed. The cache is REST-only; [`getGraphQLClientOptions` never sets `Cache`](https://github.com/integrations/terraform-provider-github/blob/67f3fd10bda01461da6241543c1cd07e8e553f86/internal/ghclient/options.go#L42-L54).

**Fix:** apply `oauth2.Transport` so it wraps ghct rather than sitting beneath it.

Distinct from #3519, which concerns the legacy etag path caching too aggressively into stale state; this is the new client not caching at all.

### Terraform Version

```
Terraform v1.15.8
on linux_amd64
```

### GitHub Installation Type

### Affected Resource(s)

Not resource-specific — every REST-backed resource and data source when `legacy_client = false`. The defect is in the shared client transport.

```hcl
provider "github" {
  owner         = "your-org"
  legacy_client = false # required to reach the affected client
}
```

### Steps to Reproduce

A standalone repro is at [`TEMPORARY-repro/cache-ordering`](https://github.com/garnertb/terraform-provider-github/tree/garnertb/cache-ordering-repro/TEMPORARY-repro/cache-ordering). It builds both orderings **directly from the pinned libraries without importing `internal/ghclient`**, so the core comparison can be audited without trusting this fork.

```sh
git clone -b garnertb/cache-ordering-repro https://github.com/garnertb/terraform-provider-github
cd terraform-provider-github
GOTOOLCHAIN=auto go run ./TEMPORARY-repro/cache-ordering

# optional, against the real API
GITHUB_TOKEN=$(gh auth token) GOTOOLCHAIN=auto go run ./TEMPORARY-repro/cache-ordering -live
```

Hermetic mode, against an `httptest` origin that computes ETags with GitHub's rule and returns `401` when unauthenticated:

```
ORDERING    ORIGIN STATUSES     CLIENT STATUSES   X-Cache:HIT   STORED X-Varied-Authorization
upstream    200 x4              200 x4            0             (marker absent)
patched     200 x1, 304 x3      200 x4            3             present, len=44
provider    200 x4              200 x4            0             (marker absent)
```

Live mode against `api.github.com`: `upstream=0`, `patched=3` of 4.

Note that **statuses must be counted at the origin**. ghct [rewrites a wire `304` into a `200`](https://github.com/bored-engineer/github-conditional-http-transport/blob/94778cbb26ea34bb63c577dfffaa99ebfb59e1cc/transport.go#L135-L137) before returning, so client-side codes are `200` in every row; the client-side signal is `X-Cache`/`Cache-Status`.

### Code of Conduct

stevehipwell:
Resolves #3612

----

### Before the change?

- New client transport cache was ineffective due to incorrect transport ordering

### After the change?

- New client caches correctly based on authorization header

### Pull request checklist

- [ ] Yes
- [x] No

----

### DIFF
--- a/internal/ghclient/helpers_test.go
+++ b/internal/ghclient/helpers_test.go
@@ -7,7 +7,26 @@ import (
 	"sync/atomic"
 	"testing"
 	"time"
+
+	"golang.org/x/oauth2"
 )
+
+// sequentialTokenSource is an oauth2.TokenSource that returns a different token on each call,
+// following the order of the provided tokens. Once exhausted, it keeps returning the last token.
+// This is used to simulate multiple distinct callers/tokens sharing the same underlying transport.
+type sequentialTokenSource struct {
+	tokens []string
+	idx    atomic.Int32
+}
+
+func (s *sequentialTokenSource) Token() (*oauth2.Token, error) {
+	i := int(s.idx.Add(1)) - 1
+	if i >= len(s.tokens) {
+		i = len(s.tokens) - 1
+	}
+
+	return &oauth2.Token{AccessToken: s.tokens[i]}, nil
+}
 
 type testRoundTripper struct {
 	delay  time.Duration
--- a/internal/ghclient/transport.go
+++ b/internal/ghclient/transport.go
@@ -32,6 +32,23 @@ func cloneTransport(tr http.RoundTripper, opts ClientOptions) http.RoundTripper 
 func newTransport(tokenSource oauth2.TokenSource, opts ClientOptions) (http.RoundTripper, error) {
 	tr := cloneTransport(http.DefaultTransport, opts)
 
+	// The cache transport must be wrapped directly around the base transport, before the
+	// OAuth2 transport is applied. The OAuth2 transport injects the Authorization header on a
+	// cloned request immediately before invoking its Base RoundTripper, so any transport wrapping
+	// it from the outside (i.e. added to tr afterwards) would only ever see requests without the
+	// Authorization header. Since the cache transport partitions/validates its cache entries based
+	// on the request's Authorization header (see the Vary handling in
+	// github.com/bored-engineer/github-conditional-http-transport), placing it outside of the
+	// OAuth2 transport silently breaks per-token cache validation for authenticated requests.
+	if opts.Cache {
+		store, err := createCacheStore(opts.CachePath)
+		if err != nil {
+			return nil, fmt.Errorf("failed to create cache store: %w", err)
+		}
+
+		tr = ghct.NewTransport(store, tr)
+	}
+
 	if tokenSource != nil {
 		tr = &oauth2.Transport{
 			Base:   tr,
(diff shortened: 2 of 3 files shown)


deiga (approved):
LGTM

robert-crandall (approved):
Thanks for fixing the transport ordering and adding regression coverage! This looks good to me.

PR #3640, opened 2026-09-14
chore: Update go-github to 92.0.0

stevehipwell:
Resolves #ISSUE_NUMBER

----

### Before the change?

- Provider used old `go-github` version

### After the change?

- Provider uses latest `go-github` version (`92.0.0`)

### Pull request checklist

- [ ] Yes
- [x] No

----


github/resource_github_actions_environment_secret.go
```diff
@@ -7,12 +7,14 @@ import (
...
 
-	"github.com/google/go-github/v89/github"
+	"github.com/google/go-github/v91/github"
 	"github.com/hashicorp/terraform-plugin-log/tflog"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/customdiff"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/validation"
+
+	"github.com/integrations/terraform-provider-github/v6/internal/tfschemautil"
```
deiga, line 17: Sweet <3

github/resource_github_issue.go
```diff
@@ -86,43 +88,51 @@ func resourceGithubIssueCreateOrUpdate(d *schema.ResourceData, meta any) error {
...
 			log.Printf("[DEBUG] Response from creating issue: %#v", *resp)
 		}
 	} else {
-		number := d.Get("number").(int)
+		req := github.UpdateIssueRequest{
+			Title:     new(title),
+			Body:      new(body),
+			Labels:    labels,
+			Assignees: asignees,
+		}
```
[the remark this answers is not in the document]
deiga: I don't think this should be an issue, right?

github/resource_github_organization_repository_role.go
```diff
@@ -94,109 +113,126 @@ func resourceGithubOrganizationRepositoryRoleCreate(ctx context.Context, d *sche
...
-			}
+		if ghErr, ok := errors.AsType[*github.ErrorResponse](err); ok && ghErr.Response.StatusCode == http.StatusNotFound {
+			tflog.Warn(ctx, "GitHub organization repository role not found, removing from state", map[string]any{"orgName": orgName, "roleId": roleID})
+			d.SetId("")
+			return nil
 		}
 		return diag.FromErr(err)
 	}
 
-	if err = d.Set("role_id", role.GetID()); err != nil {
```
deiga, line 126: question: I assume you removed `d.Set("role_id", role.GetID())` on purpose and didn't omit it by accident?

github/resource_github_organization_repository_role.go
```diff
@@ -94,109 +113,126 @@ func resourceGithubOrganizationRepositoryRoleCreate(ctx context.Context, d *sche
...
+		BaseRole:    new(baseRole),
+		Permissions: permissions,
 	}
 
-	permissions := d.Get("permissions").(*schema.Set).List()
-	permissionsStr := make([]string, len(permissions))
-	for i, v := range permissions {
-		permissionsStr[i] = v.(string)
+	if _, _, err := client.Organizations.UpdateCustomRepoRole(ctx, orgName, roleID, req); err != nil {
+		return diag.FromErr(fmt.Errorf("error updating GitHub organization repository role (%s/%s): %w", orgName, name, err))
```
deiga, line 179: question: Is this always an error that needs to be surfaced? I guess so 🤔

github/resource_github_organization_repository_role.go
```diff
@@ -94,109 +113,126 @@ func resourceGithubOrganizationRepositoryRoleCreate(ctx context.Context, d *sche
...
 	}
 
-	_, _, err = client.Organizations.UpdateCustomRepoRole(ctx, orgName, roleId, update)
-	if err != nil {
-		return diag.FromErr(fmt.Errorf("error updating GitHub organization repository role (%s/%s): %w", orgName, d.Get("name").(string), err))
+	roleIDInt, _ := d.Get("role_id").(int)
+	roleID := int64(roleIDInt)
+
+	if _, err := client.Organizations.DeleteCustomRepoRole(ctx, orgName, roleID); err != nil {
+		return diag.FromErr(fmt.Errorf("Error deleting organization repository role %d: %w", roleID, err))
```
deiga, line 203: issue: This needs the 404 handling at least

github/resource_github_team.go
```diff
@@ -168,13 +163,21 @@ func resourceGithubTeamCreate(ctx context.Context, d *schema.ResourceData, m any
...
-	if newTeam.ParentTeamID != nil && team.Parent == nil {
-		_, resp, err = client.Teams.EditTeamBySlug(ctx, ownerName, slug, newTeam, false)
+	if req.ParentTeamID != nil && team.Parent == nil {
+		_, resp, err = client.Teams.UpdateTeamBySlug(ctx, ownerName, slug, github.UpdateTeamRequest{ParentTeamID: req.ParentTeamID})
 		if err != nil {
 			return diag.FromErr(err)
 		}
 	}
 
+	// TODO: Put this back into the request once it's been reinstated.
```
deiga, line 173: question: What does `once it's been reinstated.` refer to?

github/resource_github_team.go
```diff
@@ -168,13 +163,21 @@ func resourceGithubTeamCreate(ctx context.Context, d *schema.ResourceData, m any
...
 		on the parent team, the operation might still fail to set the parent team.
 	*/
-	if newTeam.ParentTeamID != nil && team.Parent == nil {
-		_, resp, err = client.Teams.EditTeamBySlug(ctx, ownerName, slug, newTeam, false)
+	if req.ParentTeamID != nil && team.Parent == nil {
+		_, resp, err = client.Teams.UpdateTeamBySlug(ctx, ownerName, slug, github.UpdateTeamRequest{ParentTeamID: req.ParentTeamID})
 		if err != nil {
 			return diag.FromErr(err)
 		}
 	}
```
deiga, line 171: question: The comment says "best effort" and "might fail to set", but we treat every error as "will fail", is the comment wrong or is the code?

deiga (changes requested):
Nice changes! I like the tfschemautil package <3

adamcrews:
v92 of go-github was just released a few hours ago, and includes support for configuring agents secrets https://github.com/google/go-github/pull/4533.  It also introduces a fix for a regression introduced between v89 and v91 on ghes servers https://github.com/google/go-github/pull/4553
Would it be possible to pick v92 for this update instead of v91?
Here is a PR to this branch to get the v92 work. (https://github.com/integrations/terraform-provider-github/pull/3643)

github/resource_github_issue.go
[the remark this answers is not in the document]
stevehipwell: This is the correct behaviour, the resource should operate authoratativly.

github/resource_github_organization_repository_role.go
stevehipwell: Yeah, I normalised it to use the value in `role_id` to lookup and to follow the pattern of not setting an immutable value on read.

github/resource_github_organization_repository_role.go
stevehipwell: Yeah, if it can't be updated then it fails.

github/resource_github_organization_repository_role.go
stevehipwell: Yeah, I am doing some fixup on this resource so should also fix this.

github/resource_github_team.go
stevehipwell: https://github.com/google/go-github/pull/4553

github/resource_github_team.go
stevehipwell: Both are correct (assuming the API still behaves in this manner); we need to use the fallback to correctly set the parent, but if the token in use doesn't have the permission to do this it needs to error.

stevehipwell:
@adamcrews I'll be bumping this PR to target go-github v92 once I've addressed the other review comments.

stevehipwell:
@deiga what do you think about `tfschemautil` -> `tfpluginutil` showing that the focus of the package is to support TF Plugin v2 usage but make it wider than just the schema?

deiga:
> @deiga what do you think about `tfschemautil` -> `tfpluginutil` showing that the focus of the package is to support TF Plugin v2 usage but make it wider than just the schema?

@stevehipwell with `tfschemautil` I was dreaming of of using that as a basis for some future custom linter implementation, but both names seem fine to me.

stevehipwell:
@deiga I was wanting to keep it just for the legacy SDK v2 code and was thinking that some of the other helpers such as the ID helpers should live in there too. Basically anything generic and simple that makes using SDK v2 easier to use.

adamcrews:
If you are interested, this commit https://github.com/adamcrews/terraform-provider-github/commit/eee32434da0bbc5cf8c9cf7dc3cebcbfed98c598 adds a test to check for the double escaping issue that was fixed.  And this commit https://github.com/adamcrews/terraform-provider-github/commit/ad52515d1b29a180106ea22dcbfd1135b5c20cdd adds a test for the label rename issue.

deiga (dismissed):

stevehipwell:
@adamcrews thank you but AFAIK there isn't a double escaping bug in the code and I don't see the value in proving a negative against a dependency (`go-github`) and the GitHub API. RE the label rename, feel free to add a PR once this has been merged if you think it has value.

stevehipwell:
@deiga I renamed the package, please could you re-approve.

deiga (dismissed):
LGTM!

github/resource_github_team.go
```diff
@@ -168,11 +168,12 @@ func resourceGithubTeamCreate(ctx context.Context, d *schema.ResourceData, m any
 		Note that this is best-effort: when running this with a PAT that does not have admin permissions
 		on the parent team, the operation might still fail to set the parent team.
 	*/
-	if newTeam.ParentTeamID != nil && team.Parent == nil {
-		_, resp, err = client.Teams.EditTeamBySlug(ctx, ownerName, slug, newTeam, false)
+	if (req.ParentTeamID != nil || req.ParentTeamSlug != nil) && team.Parent == nil {
+		t, _, err := client.Teams.UpdateTeamBySlug(ctx, ownerName, slug, github.UpdateTeamRequest{ParentTeamID: req.ParentTeamID, ParentTeamSlug: req.ParentTeamSlug})
```
robert-crandall: [UpdateTeamRequest.Name](https://github.com/google/go-github/blob/5149b4d74590b63154fcc43c4dac05e881f9aea3/github/teams.go#L234-L237) lacks `omitempty`, so leaving it unset sends `"name": null`, which GitHub rejects.

Could you include `Name: new(name)` here?

robert-crandall (commented):
Thanks for handling this upgrade! Only 1 comment, the rest of this looks good!

github/resource_github_team.go
stevehipwell: Would it work correctly if `name` was tagged with `omitempty`? It looks like it's a bug given that the field is a pointer.

I've fixed it.

deiga (approved):

robert-crandall (approved):
Thanks for updating the client! I’ve reviewed the latest changes, and this looks good to me.

PR #3642, opened 2026-09-14
fix: treat 404 as already-deleted in organization_custom_properties Read/Delete

### LINKED ISSUE #3641: github_organization_custom_properties Read doesn't handle 404, breaking recovery after deletion

gilfthde:
## Description

`resourceGithubCustomPropertiesRead` propagated any error from `GetCustomProperty` verbatim, including a 404 for a property that no longer exists on GitHub. This breaks the standard Terraform "resource already gone" recovery convention: `Read` should call `d.SetId("")` and return `nil` when the remote resource is confirmed absent, not surface a hard error.

In practice this means: once the underlying custom property is deleted (whether by `terraform destroy`, an external actor, or any controller built on this provider that re-observes state after a delete), the next `Read` never converges - it keeps reporting a 404 as a reconcile error instead of confirming deletion.

`resourceGithubCustomPropertiesDelete` has the same gap: removing an already-removed property currently errors, instead of being treated as a no-op.

## Fix

Mirrors the handling already used by the repo-scoped sibling, `resource_github_repository_custom_property.go`'s `Read`/`Delete` (`errors.AsType[*github.ErrorResponse](err)` with `StatusCode == 404`).

## Testing

- `go build ./github/...` and `go vet ./github/...` pass.
- No existing acceptance test in this repo exercises the disappears/404 case for either the org- or repo-scoped custom property resource, so none was added here, consistent with the existing test coverage for this resource; happy to add one if maintainers want it and can point me at credentials/CI conventions for it.

Fixes #3641


deiga:
Thanks for opening a PR on this! This is already being addressed in https://github.com/integrations/terraform-provider-github/pull/3234/changes#diff-759376803bb14871b2470ba914f08b5604e1cc642ab3ce6a6d748ab8df5fc407R253-R257

gilfthde:
Hello @deiga, the mentioned PR is open since February. I see that there is recent activity, but is there any chance to get this simple fix merged in advance? It is affecting us in production.

deiga:
@gilfthde there are so many bugs in custom properties currently, that I doubt there will be a release without #3234, so merging this would just slow that PR down due to merge conflicts

